PR #5, opened 2025-12-11
feat: implement AI companion response handler

kristofferremback:
## Summary

- Add async companion response generation via outbox pattern
- When a user sends a message to a stream with `companion_mode='on'`, Ariadne (or configured persona) responds
- Update companion_mode values: `off` | `mentions` | `on` (default: `mentions`)

## Changes

**New files:**
- `PersonaRepository` - queries for personas from DB
- `AIService` - wraps OpenRouter client for LLM calls  
- `CompanionService` - orchestrates the response flow

**Modified files:**
- `OutboxListener` - triggers companion after message:created events
- `stream-repository` - updated CompanionMode type
- `stream-handlers` - updated validation
- `server.ts` - wires up new services
- `002_core_schema.sql` - updated companion_mode comment/default

## Flow

```
User sends message
    ↓
OutboxListener picks up 'message:created' event
    ↓
Check: stream.companion_mode === 'on'?
    ↓ yes
Load persona → Get context → Call LLM → Post response
```

## Test plan

- [x] All 36 E2E tests pass
- [x] Companion triggers correctly (verified via logs)
- [x] Gracefully handles missing API key (warns, doesn't crash)
- [ ] Manual test with actual OPENROUTER_API_KEY

### DIFF
--- a/apps/backend/src/db/migrations/002_core_schema.sql
+++ b/apps/backend/src/db/migrations/002_core_schema.sql
@@ -288,7 +288,7 @@ VALUES (
     'You are Ariadne, an AI thinking companion in Threa. You help users explore ideas, think through problems, and make decisions. You have access to their previous conversations and knowledge base through the GAM (General Agentic Memory) system.
 
 Be concise but thoughtful. Ask clarifying questions when needed. When referencing previous knowledge, cite your sources.',
-    'anthropic:claude-sonnet-4-20250514',
+    'anthropic/claude-3.5-sonnet',
     'system',
     'active'
 );


apps/backend/src/db/migrations/002_core_schema.sql
```diff
@@ -288,7 +288,7 @@ VALUES (
     'You are Ariadne, an AI thinking companion in Threa. You help users explore ideas, think through problems, and make decisions. You have access to their previous conversations and knowledge base through the GAM (General Agentic Memory) system.
 
 Be concise but thoughtful. Ask clarifying questions when needed. When referencing previous knowledge, cite your sources.',
-    'anthropic:claude-sonnet-4-20250514',
+    'anthropic/claude-3.5-sonnet',
```
kristofferremback: Hmm, this is not the right one, here's the one we should use `anthropic/claude-haiku-4.5`

kristofferremback (commented):
## Code Review: AI Companion Response Handler

### Overview
This PR adds AI companion response functionality that automatically generates and posts AI responses when users send messages in streams with companion mode enabled.

### Key Issues

#### 🔴 `mentions` mode is defined but not implemented
**File:** `companion-service.ts:96`

The schema and validation accept `'mentions'` as a valid companion mode, but `shouldRespond()` only triggers on `'on'`:
```typescript
return stream.companionMode === "on"
```
This means `mentions` behaves identically to `off`, which will confuse users. Either implement mention detection or remove the mode until it's ready.

---

#### 🔴 Inconsistent model format in seed data
**File:** `002_core_schema.sql:291`

The seed data uses `anthropic/claude-3.5-sonnet` (OpenRouter format), but `AIService.mapModelToOpenRouter()` expects and converts `provider:model` format. This inconsistency means:
- The mapping function won't transform it (no `:` found)
- Other personas might use the colon format, creating inconsistency

Pick one format and use it consistently. Since the codebase documents `provider:model` as the convention, the seed should use `anthropic:claude-3.5-sonnet`.

---

#### 🟡 Redundant database queries
**File:** `companion-service.ts:88-116`

`shouldRespond()` and `getPersonaForStream()` both fetch the stream separately. This means 2 DB round-trips per message. Consolidate into a single method:

```typescript
private async getStreamAndPersona(streamId: string) {
  return withClient(this.pool, async (client) => {
    const stream = await StreamRepository.findById(client, streamId)
    if (\!stream || stream.companionMode \!== "on") return null
    
    const persona = stream.companionPersonaId
      ? await PersonaRepository.findById(client, stream.companionPersonaId)
      : await PersonaRepository.findSystemDefault(client)
    
    return { stream, persona }
  })
}
```

---

#### 🟡 Unsafe type assertion
**File:** `outbox-listener.ts:193`

```typescript
authorType: payload.message.authorType as "user" | "persona"
```

If `authorType` is an unexpected value, this silently passes through. Add validation:
```typescript
const authorType = payload.message.authorType
if (authorType \!== "user" && authorType \!== "persona") {
  logger.warn({ authorType }, "Unknown authorType in message payload")
  return
}
```

---

#### 🟡 Unused import
**File:** `companion-service.ts:4`

`MessageRepository` is imported but never used.

---

### Minor Suggestions

1. **No test coverage** - Consider adding tests for `AIService.mapModelToOpenRouter()` and `CompanionService.shouldRespond()` at minimum.

2. **No rate limiting** - Users could spam messages and trigger many concurrent AI calls. Consider debouncing per-stream.

3. **No typing indicator** - Users won't know the AI is generating a response. Consider emitting a typing event before the AI call.

### What's Good

- Clean separation of concerns (AIService, CompanionService, PersonaRepository)
- Fire-and-forget pattern is appropriate here
- Good logging throughout
- Follows existing repository patterns
- Proper error handling that doesn't block message delivery

apps/backend/src/lib/outbox-listener.ts
```diff
@@ -34,6 +36,10 @@ export class OutboxListener {
     this.fallbackPollMs = options.fallbackPollMs ?? 500
   }
 
+  setCompanionService(service: CompanionService): void {
```
kristofferremback: Is there a reason for a setting instead of params? I don't think we would want the companion service to be optional either, so why not just pass it in directly?

If you're thinking of the e2e tests, we should use a stub similar to auth if we don't actually want the AI there (which we don't).

apps/backend/src/lib/outbox-listener.ts
```diff
@@ -159,8 +167,37 @@ export class OutboxListener {
         events.map((e) => e.id),
       )
 
+      processedEvents = events
       logger.debug({ count: events.length }, "Processed outbox events")
     })
+
+    // Trigger companion responses after transaction commits (non-blocking)
```
kristofferremback: ...What? No no no! The message isn't handled until all its side effects are dispatched, such as enqueuing a durable session for the companion to run in.

Outbox events should have minimal handling and fan out to more complex stuff, but we should **ensure everything happens**, so when we trigger the response outside the transaction, we have no guarantees the agent will respond. That's not OK.

I wonder if the outbox is designed incorrectly right now where we assume a single worker? We have two options as I see it:
- A single worker that fans out to other jobs (what we kinda have). This is simplest, but has the drawback of possibly re-triggering previous steps on failure of later stuff (e.g., the agent dispatch works but the notify fails, a retry will do the agent dispatch twice and the notify twice, which requires a more complicated setup overall IMO)
- Multiple workers, one per listener which lets us minimize duplicate actions as each action is basically atomic.

I want us to take the second approach of multiple workers, they are the most durable and durability is important.

apps/backend/src/services/companion-service.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export class CompanionService {
+  constructor(
+    private pool: Pool,
+    private aiService: AIService,
+    private eventService: EventService,
+  ) {}
+
+  async handleMessageCreated(params: HandleMessageCreatedParams): Promise<void> {
```
kristofferremback: This is entirely too simplistic. We're not building an MVP or a PoC, we already did that in the backup repo we based the legacy exploration on. There are several issues I see with this setup:
- This isn't durable. If the server is restarted, we hit an intermittent error or similar, we don't get a response, we might log an error, but that's it. And that's bad. We should expect failure and build for it.
- This is just plain chat that doesn't follow the core principle of GAM whatsoever. This **must** be an agentic chat, otherwise what's the point?

Together with my other comment on the outbox handling (with per-listener workers), let's build an actually retriable agentic pipeline for this, and let's use langgraph. I want something like:
- In the outbox event listener, dispatch a job (using `pg-boss` perhaps?)
- Pick the job up, check if we should handle the message. If not, complete the job with a noop.
- Build up context for the agent and spin up an agentic loop which follows the GAM pattern
- Once the agent considers itself done, then send the message. Please ensure we don't shoot ourselves in the foot with streaming.

I suggest reading the section "Architectural Decision: Streaming & Recovery" in the legacy exploration.

apps/backend/src/services/ai-service.ts
```diff
@@ -0,0 +1,71 @@
...
+    const [provider, modelName] = model.split(":")
+    if (!modelName) {
+      // Already in OpenRouter format or just a model name
+      return model
+    }
+    return `${provider}/${modelName}`
+  }
+
+  private getDefaultPrompt(): string {
+    return `You are a helpful AI assistant. Be concise but thoughtful in your responses.`
```
kristofferremback: Let's try a little bit harder, ok?

PR #12, opened 2025-12-18
feat: implement fullstack chat frontend with draft scratchpads

kristofferremback:
## Problem

The frontend needed a complete implementation to support real-time chat functionality with scratchpads (personal AI-assisted notes), channels, and a polished UI foundation. Additionally, scratchpads needed draft behavior similar to Claude.ai/ChatGPT where:
- Users can start typing without immediately creating a server-side resource
- Draft names and message content persist across browser refreshes
- The UX for renaming scratchpads should be intuitive (hover-to-edit)

## Solution

This PR implements the full frontend chat foundation with React 19, Vite, and Shadcn UI, plus draft scratchpad improvements.

### Key design decisions

**1. IndexedDB for drafts instead of localStorage**

IndexedDB provides structured storage with reactive queries via `useLiveQuery` from dexie-react-hooks. This enables automatic UI updates when draft data changes.

**2. Direct IndexedDB reads in mutations**

When creating a stream from a draft, we read directly from IndexedDB instead of using hook cache to avoid stale data issues.

**3. Shadcn UI with Golden Thread theme**

Accessible, customizable components copied into the codebase. The Golden Thread theme uses warm neutrals with gold accents.

**4. Separate draft message persistence**

Draft messages are stored separately from draft scratchpads, keyed by stream or thread ID. This enables draft persistence for ALL stream types.

## Test plan

- [x] Create draft scratchpad - appears in sidebar with "(draft)" badge
- [x] Rename draft - name persists across refresh
- [x] Type message in draft - content persists across refresh
- [x] Send first message - creates server stream with displayName
- [x] Hover scratchpad title - shows outline and pen icon
- [x] Rename real scratchpad - updates in sidebar and header immediately
- [ ] WebSocket propagation for cross-tab sync (follow-up task)

## Follow-up tasks

1. **WebSocket stream updates** - Propagate stream rename/archive across browser tabs
2. **Naming service refactor** - Move naming logic from event service to outbox/listener pattern


apps/frontend/src/types/domain.ts
```diff
@@ -0,0 +1,145 @@
+// Domain types that mirror the backend exactly
+// Keep in sync with apps/backend/src/repositories/*
```
kristofferremback: Can we share the types? One of things we both thought we could improve from the PoC was to share types between the frontend and the backend. Let's do that instead of copying and leaving comments to keep things up to date!

apps/frontend/src/api/streams.ts
```diff
@@ -0,0 +1,98 @@
...
+  Stream,
+  StreamMember,
+  StreamEvent,
+  StreamType,
+  Visibility,
+  CompanionMode,
+} from "@/types/domain"
+
+// Bootstrap response - everything needed to render a stream
+export interface StreamBootstrap {
```
kristofferremback: Can these be the same as are used in the backend?

apps/backend/src/handlers/workspace-handlers.ts
```diff
@@ -55,5 +57,34 @@ export function createWorkspaceHandlers({ workspaceService }: Dependencies) {
...
+        workspaceService.getMembers(workspaceId),
+        streamService.list(workspaceId, userId),
+      ])
+
+      if (!workspace) {
+        return res.status(404).json({ error: "Workspace not found" })
+      }
+
+      // Get stream memberships for all streams the user has access to
+      const streamMemberships = await Promise.all(
```
kristofferremback: Similar to the review that Claude made, perhaps we can introduce a batch endpoint for stream membership?

apps/frontend/src/contexts/query-client.tsx
```diff
@@ -0,0 +1,49 @@
...
+  })
+}
+
+let browserQueryClient: QueryClient | undefined = undefined
+
+function getQueryClient() {
+  // Server: always make a new query client
+  if (typeof window === "undefined") {
+    return makeQueryClient()
+  }
```
kristofferremback: I don't think we need server side handling? The app should never be server side rendered. We'll do an Electron app for that instead.

apps/frontend/src/components/timeline/timeline-view.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const {
+    events,
+    isLoading,
+    error,
+    fetchOlderEvents,
+    hasOlderEvents,
+    isFetchingOlder,
+  } = useEvents(workspaceId!, streamId!, { enabled: !isDraft })
+
+  // Auto-scroll to bottom when new events arrive
```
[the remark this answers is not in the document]
kristofferremback: Created follow-up task in docs/requests/scroll-behavior.md with detailed requirements for unread positioning and conditional auto-scroll.

PR #14, opened 2025-12-18
refactor: extract naming service to outbox/listener pattern

kristofferremback:
## Problem

The naming service was coupled with the EventService - `createMessage()` directly called `streamNamingService.attemptAutoNaming()` (fire-and-forget). This violated:
- Single responsibility principle (EventService handling naming concerns)
- INV-13 (Construct, Don't Assemble) via the `setStreamNamingService()` setter pattern
- Consistency with other async features that use the outbox/listener pattern

## Solution

Extract naming into the same outbox/listener pattern used by companion and broadcast features.

### How it works

```
message:created (outbox)
       │
       ▼
NamingListener
  - Filters for scratchpad/thread streams needing names
  - Dispatches job to pg-boss
       │
       ▼
pg-boss (naming.generate queue)
  - Durable storage
  - Automatic retries
       │
       ▼
NamingWorker
  - Thin handler
  - Delegates to StreamNamingService.attemptAutoNaming()
```

### Key design decisions

**1. Listener dispatches to pg-boss, not direct service call**

LLM calls can fail transiently. pg-boss provides automatic retry with backoff, which the previous fire-and-forget approach lacked.

**2. Reuse `message:created` event**

No new outbox event type needed - the naming listener simply filters existing events, same as companion listener.

**3. Keep business logic in StreamNamingService**

Worker is a thin handler. All naming logic (locking, LLM calls, updates) stays in the service for testability and reusability.

## New files

| File | Purpose |
|------|---------|
| `apps/backend/src/lib/naming-listener.ts` | Listens for `message:created`, dispatches naming jobs for streams that need auto-naming |
| `apps/backend/src/workers/naming-worker.ts` | Thin pg-boss handler that delegates to `StreamNamingService` |

## Modified files

| File | Change |
|------|--------|
| `apps/backend/src/lib/job-queue.ts` | Added `NAMING_GENERATE` queue and `NamingJobData` type |
| `apps/backend/src/server.ts` | Wire up naming worker/listener, removed `setStreamNamingService()` call |
| `apps/backend/src/services/event-service.ts` | Removed naming coupling (property, setter, fire-and-forget call) |

## Test plan

- [x] Unit tests pass (63 tests)
- [ ] Manual verification: create scratchpad, send message, verify auto-naming still works

---

Resolves: `docs/requests/naming-service-outbox.md`

### DIFF
--- a/apps/backend/src/lib/display-name.ts
+++ b/apps/backend/src/lib/display-name.ts
@@ -110,11 +110,11 @@ export function formatParticipantNames(participants: { id: string; name: string 
  * Checks whether a stream needs auto-naming.
  * Returns true if:
  * - Stream is a scratchpad or thread
- * - Display name has not been generated yet
+ * - Display name is not set (neither manually nor generated)
  */
 export function needsAutoNaming(stream: Stream): boolean {
   if (stream.type !== "scratchpad" && stream.type !== "thread") {
     return false
   }
-  return stream.displayNameGeneratedAt === null
+  return stream.displayName === null
 }


kristofferremback (commented):
## Code Review

Good refactor - fixes INV-13 violation and uses consistent outbox/listener pattern. The worker is properly thin.

### Concerns

**1. Double database lookup** (`naming-listener.ts:32-39`)
```typescript
const stream = await StreamRepository.findById(client, streamId)
if (!stream) { ... }
if (!needsAutoNaming(stream)) { return }
```
The message payload already has stream context from the event. Consider including stream type in the `message:created` payload to avoid the extra query.

**2. Interface location** (`naming-worker.ts:5-7`)
```typescript
export interface StreamNamingServiceLike {
  attemptAutoNaming(streamId: string): Promise<boolean>
}
```
Good for testing, but the interface could live with the actual service for better discoverability.

**3. Manual test not verified**
From the PR description:
```markdown
- [ ] Manual verification: create scratchpad, send message, verify auto-naming still works
```
The checkbox is unchecked. This should be verified before merge.

### Suggestions
- Verify manual test scenario works before merging
- Consider enriching `message:created` payload to avoid extra DB lookup
- Move interface to `stream-naming-service.ts`

PR #19, opened 2025-12-18
refactor: unify draft and stream handling with useStreamOrDraft hook

kristofferremback:
## Problem

Draft vs non-draft handling is scattered throughout the codebase. Components need to check `isDraft` and handle both cases separately, which is error-prone and creates code duplication.

Example from the old `sidebar.tsx`:
```typescript
if (isDraft) {
  onRename(trimmed)
} else {
  await updateStream.mutateAsync({ displayName: trimmed })
}
```

This pattern repeated in:
- `sidebar.tsx` - rename, archive operations
- `message-input.tsx` - create stream from draft on first message
- `stream.tsx` - header display, title editing

## Solution

Create a unified `useStreamOrDraft` hook that provides a transparent interface for both drafts and real streams. Components call the same methods regardless of whether they're working with a draft or a persisted stream.

### How it works

```typescript
const { stream, isDraft, rename, archive, sendMessage } = useStreamOrDraft(workspaceId, streamId)

// These work identically for drafts and real streams:
await rename("New name")
await archive()
await sendMessage({ content: "Hello", contentFormat: "markdown" })
```

For drafts, `sendMessage()` handles the entire promotion flow automatically:
1. Creates the real stream on the server
2. Sends the message to the new stream
3. Deletes the draft from IndexedDB
4. Returns navigation URL for the new stream

### Key design decisions

**1. Hook name: `useStreamOrDraft`**

Chose explicit naming over `useStream` (which already exists) or `useVirtualStream`. The name makes it clear this handles both cases.

**2. `sendMessage()` includes promotion logic**

The draft-to-real promotion flow (create → send → delete → navigate) is tightly coupled. Including it in the hook centralizes all draft→real transitions in one place, making the abstraction seamless for callers.

**3. Timeline keeps `isDraft` prop**

The timeline's behavior is fundamentally different for drafts (no events exist on server). This is a display concern, not an operation concern, so the prop remains rather than adding complexity to the hook.

## New files

| File | Purpose |
|------|---------|
| `apps/frontend/src/hooks/use-stream-or-draft.ts` | Unified hook abstracting draft/stream operations |
| `apps/frontend/src/components/ui/empty.tsx` | Shadcn Empty component for draft timeline state |
| `docs/plans/draft-abstraction/work_notes.md` | Implementation notes and design decisions |

## Modified files

| File | Change |
|------|--------|
| `apps/frontend/src/hooks/index.ts` | Export new hook and types |
| `apps/frontend/src/pages/stream.tsx` | Use `useStreamOrDraft` instead of 6 separate hooks |
| `apps/frontend/src/components/layout/sidebar.tsx` | `ScratchpadItem` now self-contained with hook |
| `apps/frontend/src/components/timeline/message-input.tsx` | Use `sendMessage()` for seamless draft promotion |
| `apps/frontend/src/components/timeline/timeline-view.tsx` | Use shadcn Empty component for draft state |

## Test plan

- [x] Build passes (`bun vite build`)
- [ ] Create new scratchpad (draft) - verify empty state shows
- [ ] Rename draft from sidebar - verify name updates
- [ ] Rename draft from header - verify name updates
- [ ] Send first message to draft - verify promotion and navigation
- [ ] Rename real scratchpad - verify API call works
- [ ] Archive real scratchpad - verify removal from sidebar

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,24 @@ on:
     branches: [main]
 
 jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - name: Install dependencies
+        run: bun install
+
+      - name: Lint frontend
+        run: bun run --cwd apps/frontend lint
+
   test:
     name: Tests
     runs-on: ubuntu-latest
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,6 +166,7 @@ Invariants are constraints that must hold across the entire codebase. Reference 
 | **INV-12** | Pass Dependencies, Not Configuration | Pass constructed objects (`pool`, `registry`), not raw config (`connectionString`, `apiKey`). Config only goes to factories/constructors that create dependencies.                                                                                                                                          |
 | **INV-13** | Construct, Don't Assemble            | Never `doThing(deps, params)` where caller assembles deps. Instead, construct objects with their deps at startup (`new Thing(deps)`), then callers just call `thing.doThing(params)`. Callers should know interfaces, not implementation dependencies.                                                      |
 | **INV-14** | Shadcn UI Components                 | Always use Shadcn UI for frontend components. Never build custom buttons, inputs, dialogs, etc. from scratch. Install missing components via `bunx shadcn@latest add <component>`. Components live in `apps/frontend/src/components/ui/`. See "Shadcn UI Reference" section below for available components. |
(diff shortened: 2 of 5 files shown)


apps/frontend/src/hooks/use-stream-or-draft.ts
```diff
@@ -0,0 +1,183 @@
...
+  const draft = useLiveQuery(() => (isDraft ? db.draftScratchpads.get(streamId) : undefined), [isDraft, streamId])
+
+  // For real streams: fetch from server
+  const { data: bootstrap, isLoading: isRealLoading } = useStreamBootstrap(workspaceId, streamId, {
+    enabled: !isDraft,
+  })
+
+  // Normalize to VirtualStream
+  const stream: VirtualStream | undefined = isDraft
+    ? draft
```
kristofferremback: We don't do nested ternaries.

PR #22, opened 2025-12-19
feat(THR-19): implement agent loop with explicit send_message and new message checks

kristofferremback:
**Linear:** [THR-19](https://linear.app/threa/issue/THR-19)

## Problem

Ariadne's agent loop terminated implicitly when generating a response, making it impossible to:
- Send multiple messages in one session
- Check for new user messages during tool execution
- Adapt to context changes mid-response

## Solution

Refactored the agent loop to use `send_message` as an explicit tool and added new message checks at key boundaries.

### How it works

```
Agent activation
    → Fetch thread context
    → LangGraph loop:
        → If no tool call → Check for new messages → End
        → If tool call → Check for new messages first
            → If new messages → Inject into context, re-run LLM
            → If no new messages → Execute tool, continue loop
```

### Key design decisions

**1. Two-point new message checks**

Check before tool execution AND before END to prevent message gaps. A user message arriving between the last tool call and termination would otherwise be missed.

**2. Concurrency control via `FOR UPDATE SKIP LOCKED`**

Only one agent can run per stream. If a second job arrives while one is running, it skips instead of queuing (the running agent will pick up new messages via its polling).

**3. `last_seen_sequence` tracking**

Prevents re-triggering for messages the agent chose not to respond to. If an agent sees a message but decides no response is needed, we don't want subsequent messages to re-trigger for the old one.

**4. Safeguards**

- Max 20 iterations per session
- Max 5 messages per session
- 60s timeout

## New files

| File | Purpose |
|------|---------
| `src/agents/tools/send-message-tool.ts` | Explicit send_message tool with max messages enforcement |
| `src/db/migrations/009_agent_session_last_seen.sql` | Adds `last_seen_sequence` column |
| `tests/integration/agent-session.test.ts` | Integration tests for new repository methods |

## Modified files

| File | Change |
|------|--------|
| `src/agents/companion-graph.ts` | Major refactor with real tools, state tracking, new message checks |
| `src/agents/companion-runner.ts` | Updated with tool execution callbacks |
| `src/agents/companion-agent.ts` | Added concurrency check and session lifecycle callbacks |
| `src/lib/companion-listener.ts` | Added seen-message check before job dispatch |
| `src/repositories/agent-session-repository.ts` | Added `findRunningByStream`, `findLatestByStream`, `updateLastSeenSequence` |
| `src/repositories/message-repository.ts` | Added `listSince` for checking new messages |

## Test plan

- [x] 99 unit tests passing
- [x] 99 integration tests passing (including 14 new tests for this feature)
- [x] Manual testing: agent responds to messages in scratchpad

---

### DIFF
--- a/apps/backend/src/agents/companion-agent.ts
+++ b/apps/backend/src/agents/companion-agent.ts
@@ -95,6 +95,7 @@ export async function withSession(
 
       await AgentSessionRepository.updateStatus(client, session.id, SessionStatuses.COMPLETED, {
         responseMessageId: sentMessageIds[0] ?? null,
+        sentMessageIds,
       })
 
       logger.info({ sessionId: session.id, messagesSent, sentMessageIds }, "Session completed")
--- a/apps/backend/src/db/migrations/010_agent_session_sent_messages.sql
+++ b/apps/backend/src/db/migrations/010_agent_session_sent_messages.sql
@@ -1,0 +1,8 @@
+-- =============================================================================
+-- Agent Sessions: Track sent message IDs
+-- =============================================================================
+-- Stores all message IDs sent by the agent during a session.
+-- Provides a strong link from session -> messages for crash recovery and auditing.
+
+ALTER TABLE agent_sessions
+    ADD COLUMN sent_message_ids TEXT[] DEFAULT '{}';
--- a/apps/backend/src/repositories/agent-session-repository.ts
+++ b/apps/backend/src/repositories/agent-session-repository.ts
@@ -34,6 +34,7 @@ interface SessionRow {
   response_message_id: string | null
   error: string | null
   last_seen_sequence: string | null
+  sent_message_ids: string[] | null
   created_at: Date
   completed_at: Date | null
 }
@@ -62,6 +63,7 @@ export interface AgentSession {
   responseMessageId: string | null
   error: string | null
   lastSeenSequence: bigint | null
+  sentMessageIds: string[]
   createdAt: Date
   completedAt: Date | null
 }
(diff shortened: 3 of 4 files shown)


apps/backend/src/agents/companion-runner.ts
```diff
@@ -44,52 +79,99 @@ export class LangGraphResponseGenerator implements ResponseGenerator {
...
         messageCount: messages.length,
+        streamId,
+        sessionId,
       },
       "Running companion graph"
     )
 
+    // Track messages sent for the tool
+    const sentMessageIds: string[] = []
+    let messagesSentCount = 0
```
kristofferremback: What happens if the agent process crashes?  I think we should persist exactly which messages the agent sent in a given session

kristofferremback:
## Response to Review Comments

### Thread 1: Persisting sent message IDs (FIXED)

Good question\! **Fixed in this update.**

I have added:
- New `sent_message_ids TEXT[]` column on `agent_sessions` (migration 010)
- Updated `updateStatus()` to accept and persist `sentMessageIds` array  
- Updated `withSession()` to pass all `sentMessageIds` on completion

Now the session has a strong link to all messages it sent. On crash recovery, we can query the session's `sentMessageIds` to see what was already sent.

Added tests: "should persist sent message IDs on session completion", "should default to empty array when no messages sent"

### Thread 2: Concurrency control (ALREADY HANDLED)

Yes, this is already handled\! `findRunningByStream()` uses `FOR UPDATE SKIP LOCKED`:

```sql
SELECT ... FROM agent_sessions
WHERE stream_id = $1 AND status = 'running'
FOR UPDATE SKIP LOCKED
LIMIT 1
```

When two concurrent events are picked up:
1. First worker acquires lock on the row
2. Second worker's query skips the locked row (returns null)
3. Second worker proceeds to check `findByTriggerMessage()` which prevents duplicate sessions

### Thread 3: Concurrency tests (ADDED)

Added true concurrency tests using promise-based coordination (no sleeps):

1. **FOR UPDATE SKIP LOCKED prevents concurrent access** - First transaction holds lock, second gets null
2. **Concurrent session creation for same trigger message** - Tests race condition handling
3. **Multiple streams can have concurrent running sessions** - Different streams don't block each other

All 104 integration tests now pass (up from 14).

PR #26, opened 2025-12-19
feat(THR-11, THR-22): implement optimistic message sending with visual feedback

kristofferremback:
**Linear:** [THR-11](https://linear.app/threa/issue/THR-11), [THR-22](https://linear.app/threa/issue/THR-22)

## Problem

When sending messages, users had no visual feedback until the server confirmed receipt. On slow or unreliable connections this felt broken - the message would just disappear from the input with no indication it was being sent. Failed messages were lost entirely.

## Solution

Implement optimistic message sending with visual pending/failed states and automatic retry on reconnection.

### How it works

```
User sends message
       │
       ▼
┌──────────────────┐
│ Add optimistic   │ ← Immediate UI feedback
│ event to cache   │
└────────┬─────────┘
         │
         ▼
┌──────────────────┐
│ Persist to       │ ← For recovery/retry
│ IndexedDB        │
└────────┬─────────┘
         │
         ▼
┌──────────────────┐     ┌─────────────────┐
│ Send to server   │────►│ On success:     │
│                  │     │ - Mark sent     │
└────────┬─────────┘     │ - Cleanup       │
         │               └─────────────────┘
         │ (failure)
         ▼
┌──────────────────┐
│ Mark as failed   │ ← Red border, retry button
│ in timeline      │
└──────────────────┘
```

When the real WebSocket event arrives, it atomically replaces the optimistic event - preventing duplicate message flashes even on very slow connections (tested with GPRS throttling).

### Key design decisions

**1. Separate pending state from wire types**

Client-only state (pending/failed) is tracked in a React context, not added to `@threa/types`. This keeps wire types clean and makes the client state easy to reason about independently.

**2. 500ms delay before showing "Sending..."**

Fast sends (the common case) show no loading indicator at all. Only slow connections see "Sending..." - implemented with CSS `animation-delay` to avoid JS timers.

**3. All failures show in timeline with retry button**

Rather than restoring failed message content to the input box, all failures appear inline in the timeline. This is clearer (user sees exactly what failed) and simpler (one failure UI, not two).

**4. Atomic deduplication on WebSocket events**

When the real event arrives via WebSocket, we find any matching optimistic event (by content + actorId) and remove it in the same cache update that adds the real event. This eliminates the race condition where both could briefly appear.

## New files

| File | Purpose |
|------|---------|
| `apps/frontend/src/contexts/pending-messages-context.tsx` | Tracks pending/failed message IDs, provides retry function |
| `apps/frontend/src/hooks/use-pending-message-retry.ts` | Auto-retries pending messages on socket reconnect (max 3 attempts) |

## Modified files

| File | Change |
|------|--------|
| `apps/frontend/src/hooks/use-stream-or-draft.ts` | Optimistic updates with proper sequencing, status tracking |
| `apps/frontend/src/hooks/use-stream-socket.ts` | Atomic deduplication when WebSocket events arrive |
| `apps/frontend/src/components/timeline/message-event.tsx` | Pending/failed visual states with retry button |
| `apps/frontend/src/components/timeline/message-input.tsx` | Simplified error handling (failures show in timeline) |
| `apps/frontend/src/App.tsx` | Added PendingMessagesProvider and retry handler |
| `apps/frontend/src/contexts/index.ts` | Export new context |
| `apps/frontend/src/hooks/index.ts` | Export new hook |
| `apps/frontend/tailwind.config.js` | Added `fade-in-delayed` animation for 500ms delay |

## Test plan

- [x] Happy path - message sends, appears immediately, shows normally after confirm
- [x] Slow network (GPRS throttle) - "Sending..." appears after 500ms, no duplicates
- [x] Network failure - message shows "Failed" with retry button
- [x] Auto-retry on reconnect - pending messages retry when socket reconnects
- [x] Race condition - WebSocket event during `await` doesn't cause duplicates

---

### DIFF
--- a/apps/frontend/src/components/timeline/message-event.tsx
+++ b/apps/frontend/src/components/timeline/message-event.tsx
@@ -1,3 +1,4 @@
+import type { ReactNode } from "react"
 import type { StreamEvent } from "@threa/types"
 import { Avatar, AvatarFallback } from "@/components/ui/avatar"
 import { Button } from "@/components/ui/button"
@@ -17,24 +18,19 @@ interface MessageEventProps {
   streamId: string
 }
 
-export function MessageEvent({ event }: MessageEventProps) {
-  const payload = event.payload as MessagePayload
-  const isPersona = event.actorType === "persona"
-  const { getStatus, retryMessage } = usePendingMessages()
+interface MessageLayoutProps {
+  event: StreamEvent
+  payload: MessagePayload
+  statusIndicator: ReactNode
+  actions?: ReactNode
+  containerClassName?: string
+}
 
-  const status = getStatus(event.id)
-  const isPending = status === "pending"
-  const isFailed = status === "failed"
+function MessageLayout({ event, payload, statusIndicator, actions, containerClassName }: MessageLayoutProps) {
+  const isPersona = event.actorType === "persona"
 
   return (
-    <div
-      className={cn(
-        "group flex gap-3 py-2",
-        isPersona && "bg-muted/30 -mx-4 px-4 rounded-lg",
-        isPending && "opacity-60",
-        isFailed && "border-l-2 border-destructive pl-2"
-      )}
-    >
+    <div className={cn("group flex gap-3 py-2", isPersona && "bg-muted/30 -mx-4 px-4 rounded-lg", containerClassName)}>
       <Avatar className="h-8 w-8 shrink-0">
         <AvatarFallback className={cn(isPersona && "bg-primary text-primary-foreground")}>
           {isPersona ? "AI" : getInitials(event.actorId)}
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/timeline/message-event.tsx
```diff
@@ -18,9 +20,21 @@ interface MessageEventProps {
 export function MessageEvent({ event }: MessageEventProps) {
```
kristofferremback: I feel like it might be time to split this component into sub components now? I foresee more variance here, and it's already kinda messy, so let's try out a FailedMessageEvent, PendingMessageEvent, and the normal message event, all wrapped in this one MessageEvent if that makes sense?

kristofferremback: Done! Split into:

- `MessageLayout` - shared structure (avatar, content wrapper, status slot, actions slot)
- `SentMessageEvent` - normal sent messages with timestamp
- `PendingMessageEvent` - reduced opacity, delayed "Sending..." indicator
- `FailedMessageEvent` - red border, "Failed to send", retry button

The main `MessageEvent` now just dispatches based on status via a switch statement. Much cleaner for future variance!

PR #32, opened 2025-12-23
feat(THR-36): implement hybrid search with filter operators

kristofferremback:
**Linear:** [THR-36](https://linear.app/threa/issue/THR-36)

## Problem

Users need to find past messages across conversations. The ticket specifies hybrid search combining keyword and semantic capabilities with filter operators for precise queries.

## Solution

Implements full hybrid search with:
- **Keyword search**: PostgreSQL tsvector with GIN indexes
- **Semantic search**: pgvector with HNSW index using OpenRouter embeddings
- **RRF combination**: Reciprocal Rank Fusion (60% keyword, 40% semantic)
- **Filter operators**: `from:@user`, `with:@user`, `is:scratchpad`, `is:channel`
- **Permission filtering**: Users only see messages from accessible streams

### How it works

```
Query: "project deadline from:@alice is:channel"
                     │
                     ▼
            ┌─────────────────┐
            │  Filter Parser  │
            └────────┬────────┘
                     │ terms: "project deadline"
                     │ filters: { from: ["alice"], is: ["channel"] }
                     ▼
        ┌────────────┴────────────┐
        ▼                         ▼
┌───────────────┐       ┌──────────────────┐
│  Keyword      │       │    Semantic      │
│  Search       │       │    Search        │
│  (tsvector)   │       │   (pgvector)     │
└───────┬───────┘       └────────┬─────────┘
        │                        │
        └──────────┬─────────────┘
                   ▼
          ┌────────────────┐
          │  RRF Combine   │
          │  (60/40 weight)│
          └────────┬───────┘
                   ▼
            Ranked Results
```

### Key design decisions

**1. OpenRouter for embeddings**

Using `openai/text-embedding-3-small` via OpenRouter API. This avoids running Ollama infrastructure while keeping costs low (~$0.02/1M tokens).

**2. Async embedding generation**

Embeddings are generated via a pg-boss background worker. Messages are immediately keyword-searchable; semantic search works once embedding is generated (typically < 1s).

**3. RRF over linear combination**

Reciprocal Rank Fusion handles the different score scales between keyword (ts_rank) and semantic (cosine similarity) naturally. The k=60 constant provides good stability.

**4. Permission filtering in search service**

Stream access is checked before search, not after. This ensures we never load messages the user shouldn't see and allows the database to optimize with stream_id filters.

## New files

| File | Purpose |
|------|---------|
| `src/db/migrations/014_message_embeddings.sql` | Adds embedding column with HNSW index |
| `src/lib/search/filter-parser.ts` | Parses queries into terms + filter operators |
| `src/lib/search/filter-parser.test.ts` | 19 unit tests for filter parser |
| `src/lib/search/rrf.ts` | Reciprocal Rank Fusion algorithm |
| `src/lib/search/rrf.test.ts` | 10 unit tests for RRF |
| `src/repositories/search-repository.ts` | Full-text and vector search SQL |
| `src/services/embedding-service.ts` | OpenRouter embedding API client |
| `src/services/search-service.ts` | Orchestrates hybrid search |
| `src/handlers/search-handlers.ts` | HTTP handler for `/api/workspaces/:id/search` |
| `src/workers/embedding-worker.ts` | Background job for embedding generation |
| `src/lib/embedding-listener.ts` | Dispatches embedding jobs for new messages |
| `tests/e2e/search.test.ts` | 14 e2e tests for search API |

## Modified files

| File | Change |
|------|--------|
| `src/lib/job-queue.ts` | Added `EMBEDDING_GENERATE` job type |
| `src/repositories/message-repository.ts` | Added `updateEmbedding` method |
| `src/repositories/user-repository.ts` | Added `findByEmailOrDisplayName` for `from:@user` lookup |
| `src/routes.ts` | Added search route |
| `src/server.ts` | Wired up embedding and search services |
| `tests/client.ts` | Added `search` helper function |

## Test plan

- [x] 19 unit tests for filter parser
- [x] 10 unit tests for RRF algorithm
- [x] 14 e2e tests covering:
  - Basic keyword search
  - Empty results handling
  - Limit parameter
  - Multi-stream search
  - Permission filtering (private vs public)
  - Filter operators (`is:`, `from:@`, `with:@`)
  - Combined filters
  - Edge cases (special characters, filter-only queries)
- [x] All 128 unit tests pass
- [x] All 86 e2e tests pass

---

### DIFF
--- a/apps/backend/src/services/embedding-service.ts
+++ b/apps/backend/src/services/embedding-service.ts
@@ -11,20 +11,32 @@ export interface EmbeddingServiceConfig {
 /**
  * Service for generating embeddings using the configured provider.
  * Wraps ProviderRegistry with a configured default model.
+ *
+ * Uses lazy initialization to avoid failing at startup when API keys
+ * aren't configured (e.g., in test environments).
  */
 export class EmbeddingService {
-  private embeddingModel: EmbeddingModel<string>
+  private providerRegistry: ProviderRegistry
+  private modelId: string
+  private embeddingModel: EmbeddingModel<string> | null = null
 
   constructor(config: EmbeddingServiceConfig) {
-    const model = config.model ?? DEFAULT_MODEL
-    this.embeddingModel = config.providerRegistry.getEmbeddingModel(model)
+    this.providerRegistry = config.providerRegistry
+    this.modelId = config.model ?? DEFAULT_MODEL
+  }
+
+  private getModel(): EmbeddingModel<string> {
+    if (!this.embeddingModel) {
+      this.embeddingModel = this.providerRegistry.getEmbeddingModel(this.modelId)
+    }
+    return this.embeddingModel
   }
 
   /**
    * Generate embedding for a single text.
    */
   async embed(text: string): Promise<number[]> {
-    const result = await embed({ model: this.embeddingModel, value: text })
+    const result = await embed({ model: this.getModel(), value: text })
     return result.embedding
   }
 
@@ -35,7 +47,7 @@ export class EmbeddingService {
     if (texts.length === 0) {
       return []
     }
-    const result = await embedMany({ model: this.embeddingModel, values: texts })
+    const result = await embedMany({ model: this.getModel(), values: texts })
     return result.embeddings
   }
 }


apps/backend/src/lib/search/filter-parser.test.ts
```diff
@@ -0,0 +1,140 @@
+import { describe, test, expect } from "bun:test"
+import { parseQuery } from "./filter-parser"
+
+describe("parseQuery", () => {
+  test("extracts plain search terms", () => {
```
kristofferremback: Please convert this to a table test us with names to keep the "should ..." pattern going, and please use object matching instead of continuous asserts!

PR #41, opened 2025-12-26
fix(THR-63): exclude threads from auto-naming

kristofferremback:
## Summary

- Excludes threads from the naming worker - they don't need auto-generated titles
- Threads will use their default display ("Thread in #channel" or "New thread")
- Reduces unnecessary AI calls

## Changes

- `needsAutoNaming()` now only returns true for scratchpads, not threads
- Updated test to reflect new behavior

## Test plan

- [x] Ran stream-naming tests (23 pass)
- [ ] Manual: Create a thread and verify naming worker doesn't run for it

Closes THR-63

### DIFF
--- a/apps/backend/src/lib/display-name.ts
+++ b/apps/backend/src/lib/display-name.ts
@@ -109,8 +109,12 @@ export function formatParticipantNames(participants: { id: string; name: string 
 /**
  * Checks whether a stream needs auto-naming.
  * Returns true if:
- * - Stream is a scratchpad (threads excluded for now - see THR-63)
+ * - Stream is a scratchpad
  * - Display name is not set (neither manually nor generated)
+ *
+ * Threads are excluded because they already have contextual display names
+ * derived from their parent stream (e.g., "Thread in #general"). Generating
+ * AI titles for threads adds cost without clear value.
  */
 export function needsAutoNaming(stream: Stream): boolean {
   if (stream.type !== "scratchpad") {


kristofferremback (commented):
## Code Review

### Overview

This PR excludes threads from the auto-naming worker. Previously, both scratchpads and threads would trigger AI-powered name generation. Now only scratchpads get auto-named, while threads rely on their default display names ("Thread in #channel" or "New thread").

### Analysis

**Code Correctness** ✅
- The logic change is correct
- Test updated to match new behavior

**Following Project Conventions** ✅
- Comment references ticket ID (THR-63) - good traceability
- Test name is descriptive and includes rationale

**Performance** ✅
- Reduces AI calls for threads - positive impact

**Security** ✅
- No security implications

### Questions/Concerns

1. **Silent behavior change for existing threads**: If there are threads that already had `displayName === null`, they'll now skip naming forever. Is that intentional? Should there be a migration or cleanup task?

2. **Comment accuracy**: The code comment says "threads excluded for now" which implies this is temporary. Is there a follow-up ticket for thread naming? If not, "for now" suggests unfinished work that may never happen.

### Recommendations

1. **Minor**: Clarify the comment. Either:
   - Remove "for now" if this is the permanent design
   - Add a TODO with ticket reference if there's a follow-up planned

2. **Consider**: Add a brief code comment in `needsAutoNaming` explaining *why* threads are excluded (e.g., "Threads use contextual display names like 'Thread in #channel'")

### Verdict

**Approve with minor suggestions**. The change is small, focused, and well-tested. The suggestions above are optional improvements.

PR #45, opened 2025-12-27
feat(THR-102): implement dark mode with theme picker

kristofferremback:
**Linear:** [THR-102](https://linear.app/threa/issue/THR-102)

## Problem

Users have no way to switch between light and dark mode. The app only displays in light mode, even though the CSS dark theme variables were already defined in `index.css`.

## Solution

Implement a complete dark mode system with a theme picker dropdown in the sidebar header.

### How it works

```
User clicks theme button → Dropdown shows 3 options → Selection stored in localStorage
                                                    → Theme class applied to <html>
                                                    → CSS variables switch automatically
```

### Key design decisions

**1. Three-way toggle (System/Light/Dark) instead of binary toggle**

Respects user preference while allowing override. "System" follows OS `prefers-color-scheme` and updates automatically if OS preference changes.

**2. Theme state at outermost provider level**

`ThemeProvider` wraps the entire app (before `AuthProvider`) so theme is available everywhere, even on unauthenticated pages.

**3. localStorage persistence with `threa-theme` key**

Simple, synchronous storage. No need for server-side persistence since theme is a device-level preference.

**4. ThemeDropdown as standalone component (INV-18)**

Extracted to its own file per new invariant—sidebar.tsx should contain sidebar logic, not theme picker logic.

**5. Flash prevention via inline script**

Inline script in `index.html` applies theme class before React hydrates, preventing flash of wrong theme on initial load.

## New files

| File | Purpose |
|------|---------|
| `apps/frontend/src/contexts/theme-context.tsx` | Theme state management with localStorage, system preference detection, and media query listener. Exports `Theme` type. |
| `apps/frontend/src/components/theme-dropdown.tsx` | Theme picker dropdown UI component |

## Modified files

| File | Change |
|------|--------|
| `CLAUDE.md` | Added INV-18: No Inline Components |
| `apps/frontend/index.html` | Added inline script for flash prevention |
| `apps/frontend/src/App.tsx` | Wrap with `ThemeProvider` as outermost provider |
| `apps/frontend/src/components/layout/sidebar.tsx` | Add `<ThemeDropdown />` to header |
| `apps/frontend/src/contexts/index.ts` | Export `ThemeProvider`, `useTheme`, and `Theme` type |
| `apps/frontend/src/index.css` | Add 150ms transition for smooth theme switching |

## Test plan

- [x] ESLint passes
- [ ] Toggle between System/Light/Dark modes
- [ ] Verify localStorage persistence across page refresh
- [ ] Change OS preference while in System mode (should update automatically)
- [ ] Verify no flash of wrong theme on initial load
- [ ] Verify all UI components render correctly in dark mode

---


kristofferremback (commented):
## Code Review: Dark Mode Implementation

### Overview
This PR implements a dark mode feature with a three-way theme picker (System/Light/Dark) in the sidebar. The implementation is clean, follows React patterns well, and integrates properly with the existing codebase.

### ✅ What's Good

**Architecture**
- Clean separation: `ThemeProvider` context handles state, sidebar only handles UI
- Correct provider ordering: `ThemeProvider` wraps everything so theme is available even on unauthenticated pages
- localStorage persistence is appropriate for device-level preferences

**Implementation Quality**
- Proper cleanup of media query event listener
- `useCallback` for stable function references
- SSR-safe guards (`typeof window === "undefined"`)
- Type-safe theme values (`Theme` type restricts to valid options)
- Smooth 150ms transition for polished UX

**UX Considerations**
- System preference detection with real-time updates when OS changes
- Icon changes to reflect current state (Monitor for system, Sun/Moon for explicit)
- Check mark shows current selection in dropdown

### 🔸 Minor Suggestions

**1. Flash of wrong theme on initial load**
The theme is applied in a `useEffect`, which means there's a brief moment where the default (light) theme shows before the stored preference is applied. Consider adding an inline script in `index.html` to apply the theme class before React hydrates:

```html
<script>
  (function() {
    const stored = localStorage.getItem('threa-theme');
    const theme = stored === 'dark' || (stored !== 'light' && window.matchMedia('(prefers-color-scheme: dark)').matches);
    if (theme) document.documentElement.classList.add('dark');
  })();
</script>
```

This prevents the flash completely but is optional - the 150ms transition makes it less jarring anyway.

**2. Consider exporting the Theme type**
If other components need to work with theme values (e.g., for conditional styling logic), exporting the `Theme` type would be helpful:

```tsx
export type Theme = "light" | "dark" | "system"
```

**3. `toggleTheme` could be removed if unused**
The `toggleTheme` function is defined but I don't see it used in the sidebar (which uses explicit `setTheme` calls). If it's not planned for keyboard shortcuts or other UI, consider removing it to avoid dead code. If it IS planned for later use, that's fine to keep.

### 📋 Test Coverage
No automated tests were added for the theme context. While the functionality is straightforward, consider adding tests for:
- Initial theme loading from localStorage
- Theme persistence on change
- System preference change handling

### 🔒 Security
No concerns - localStorage is appropriate for this use case and there's no user input handling.

### Summary
**Approve** - This is a well-implemented feature. The suggestions above are minor polish items, not blockers. The code follows project conventions and the three-way theme picker with system preference support is the right design choice.

PR #47, opened 2025-12-28
feat(THR-28): implement agent invocation via @mentions

kristofferremback:
**Linear:** [THR-28](https://linear.app/threa/issue/THR-28)

## Problem

Users cannot invoke AI personas by @mentioning them. The only way to get AI assistance is through companion mode, which must be explicitly enabled on a stream. This limits the natural conversational flow where users might want to ask a specific persona for help on-demand.

## Solution

Implement a mention-based invocation system that triggers personas when @mentioned in any stream type, with context-appropriate response locations.

### How it works

```
1. User sends: "@ariadne help me with X"
2. mention-invoke-listener extracts @mentions from message content
3. For each slug, looks up if it's a persona (workspace-scoped → system)
4. If persona found:
   - For channels: creates thread on the message first
   - Dispatches PERSONA_INVOKE job with targetStreamId
5. mention-invoke-worker runs MentionInvokeAgent
6. Agent generates response using same LangGraph infrastructure as companion
```

### Response location by stream type

| Stream Type | Response Location |
|-------------|-------------------|
| **Channel** | Creates a thread on the message, persona responds there |
| **Thread** | Responds directly in the thread |
| **Scratchpad** | Responds directly in the scratchpad |
| **DM** | Responds directly in the DM |

### Key design decisions

**1. Separate agent rather than extending CompanionAgent**

While there's code duplication with CompanionAgent, keeping them separate allows:
- Different trigger logic (mentions vs companionMode)
- Different response targeting (targetStreamId can differ from source)
- Future divergence without affecting companion mode
- Clearer debugging and logging

**2. Thread creation for channels**

When mentioned in a channel, the persona responds in a thread to keep the main channel clean. This mirrors how humans might "take it to a thread" for focused discussions.

**3. Reusing LangGraph infrastructure**

The MentionInvokeAgent reuses the same `ResponseGenerator` interface and LangGraph setup as CompanionAgent, ensuring consistent AI behavior and tool availability.

## New files

| File | Purpose |
|------|---------|
| `apps/backend/src/lib/mention-extractor.ts` | Parses `@slug` patterns from message content |
| `apps/backend/src/lib/mention-invoke-listener.ts` | Outbox listener that detects persona mentions and dispatches jobs |
| `apps/backend/src/agents/mention-invoke-agent.ts` | Agent that responds when @mentioned |
| `apps/backend/src/workers/mention-invoke-worker.ts` | Job handler for `PERSONA_INVOKE` jobs |

## Modified files

| File | Change |
|------|--------|
| `apps/backend/src/lib/job-queue.ts` | Added `PERSONA_INVOKE` job type and `PersonaInvokeJobData` interface |
| `apps/backend/src/server.ts` | Wired up mention invoke agent, worker, and listener |

## Test plan

- [x] Server starts with no errors
- [x] `persona.invoke` job queue handler registered
- [x] `mention-invoke` outbox listener started
- [ ] End-to-end: mention @ariadne in a channel → thread created with response
- [ ] End-to-end: mention @ariadne in a thread → response in same thread
- [ ] End-to-end: mention @ariadne in a scratchpad → response in scratchpad

---

### DIFF
--- a/apps/backend/src/services/stream-service.ts
+++ b/apps/backend/src/services/stream-service.ts
@@ -395,15 +395,18 @@ export class StreamService {
     return withClient(this.pool, (client) => StreamMemberRepository.findByStreamsAndUser(client, streamIds, userId))
   }
 
+  // TODO: This is a permission check masquerading as a membership check. "isMember" is
+  // misleading because for threads we actually check root stream membership, not direct
+  // membership. Should be broken out into a proper authz module (e.g., canParticipate,
+  // canRead, canWrite) that encapsulates the permission model cleanly.
   async isMember(streamId: string, userId: string): Promise<boolean> {
     return withClient(this.pool, async (client) => {
-      // Check direct membership first
       const directMember = await StreamMemberRepository.isMember(client, streamId, userId)
       if (directMember) {
         return true
       }
 
-      // For threads, also check membership in the root stream
+      // Threads inherit participation rights from root stream
       const stream = await StreamRepository.findById(client, streamId)
       if (stream?.rootStreamId) {
         return StreamMemberRepository.isMember(client, stream.rootStreamId, userId)


kristofferremback:
## 🔒 Security Review: THR-28 Agent Invocation via @Mentions

**Result: ✅ No security vulnerabilities identified**

After thorough analysis of the PR changes, no high-confidence security vulnerabilities were found. The implementation follows the established security patterns in the codebase.

---

### Analysis Performed

#### 1. Cross-Workspace Access Control ✅
- `PersonaRepository.findBySlug` properly scopes persona lookups by workspaceId
- A persona from workspace B cannot be resolved by a lookup scoped to workspace A
- The workspaceId originates from authenticated request context

#### 2. Authorization Flow ✅
- Authorization is enforced at the API entry point (message creation)
- The outbox listener operates on already-authorized events
- Thread creation uses the original message author's context

#### 3. Input Validation ✅
- Mention slugs are constrained by regex pattern `/@([\w-]+)/g` (word chars + hyphens only)
- All database queries use parameterized SQL via `squid/pg`
- No SQL injection vectors identified

#### 4. Persona Status Enforcement ✅
- Active status is checked in both the listener and agent
- No code path bypasses these checks

---

### Architecture Assessment

The security model relies on:
- **Entry-point authorization**: Message handlers validate workspace membership
- **Trusted event bus**: Outbox events represent authorized actions
- **Workspace scoping**: All persona lookups include workspace ID
- **Parameterized queries**: All database access uses safe SQL patterns

This is a sound architecture that the new feature properly inherits.

---

🤖 *Automated security review by Claude Code*

kristofferremback (commented):
## Code Review: feat(THR-28): implement agent invocation via @mentions

### Overview

This PR adds the ability to invoke AI personas by @mentioning them in any stream type. The implementation follows a clean architecture: a listener detects mentions in outbox events, dispatches jobs to a queue, and a worker delegates to a dedicated agent that handles the response generation.

### Architecture ✅

Good decisions:
- **Separate agent** - Keeping `MentionInvokeAgent` separate from `CompanionAgent` makes sense given the different trigger logic and response targeting
- **Thread creation for channels** - Responding in threads keeps channels clean
- **Reusing `ResponseGenerator`** - Consistent AI behavior across invocation methods
- **Worker is thin** - Just extracts data and delegates to agent (per INV-13)

### Issues Found

#### 1. Code Duplication: `withSession` is duplicated ⚠️

The `withSession` method in `mention-invoke-agent.ts` (lines 195-268) is a near-verbatim copy of the exported `withSession` function in `companion-agent.ts` (lines 23-116). This violates DRY.

**Recommendation:** Import and use `withSession` from `companion-agent.ts`:

```typescript
import { withSession } from "./companion-agent"
```

The only difference I can see is logging message text ("Mention invoke session" vs "Session"). That's not sufficient justification for 70+ lines of duplication.

#### 2. Missing `actorId` null check ⚠️

In `mention-invoke-listener.ts:90` and `:100`:
```typescript
createdBy: event.actorId!,
triggeredBy: event.actorId!,
```

Using non-null assertion (`!`) on `actorId` without validation is risky. Earlier code checks `event.actorType !== AuthorTypes.USER` but doesn't verify `actorId` exists.

**Recommendation:** Add explicit null check:
```typescript
if (!event.actorId) {
  logger.warn({ streamId }, "Mention invoke: message has no actorId")
  return
}
```

#### 3. `sendMessage` and `sendMessageWithSources` are identical ⚠️

In `mention-invoke-agent.ts:124-145`, both callbacks do exactly the same thing. This suggests the interface might need cleanup, or one should delegate to the other.

**Recommendation:** Have `sendMessage` delegate to `sendMessageWithSources`:
```typescript
sendMessage: async (input) => callbacks.sendMessageWithSources(input),
```

Or investigate if the `ResponseGeneratorCallbacks` interface has cruft.

### Minor Issues

#### 4. System prompt building could be shared 💡

The `buildSystemPrompt`, `buildScratchpadPrompt`, `buildChannelPrompt`, etc. functions (lines 278-400) have overlap with similar logic that might exist in companion agent. Worth checking if these can be extracted to a shared module.

#### 5. Regex pattern reset bug in `mention-extractor.ts` 💡

Using a global regex (`/g`) with `exec()` in a loop is correct here, but be aware this pattern is easy to misuse. The regex maintains state via `lastIndex`. If `extractMentions` were ever called in a way that exits early without consuming all matches, subsequent calls would behave unexpectedly.

The current implementation is fine since it always consumes all matches, but this is a known footgun.

### Good Patterns Observed ✅

1. **Clear job data interface** - `PersonaInvokeJobData` includes helpful comment explaining `targetStreamId` semantics
2. **Clean listener/worker/agent separation** - Easy to test each layer independently
3. **Proper concurrency control** - Checks for running sessions before starting new ones
4. **Graceful skipping** - Returns structured skip reasons instead of throwing
5. **Good logging** - Includes relevant context (streamId, personaId, sessionId)

### Testing Gaps 🧪

The PR notes end-to-end tests are pending (unchecked items in test plan). Unit tests for:
- `extractMentions` function (edge cases: empty string, multiple mentions, malformed patterns)
- `MentionInvokeAgent.run()` (skip conditions, error handling)

Would strengthen confidence in this feature.

### Security Considerations ✅

- Ignores persona messages (prevents infinite loops)
- Validates persona is active before invoking
- Uses workspace scoping for persona lookup

### Summary

**Verdict: Approve with requested changes**

The architecture is solid and follows project conventions. The main issue is the duplicated `withSession` code (~70 lines) which should be extracted to avoid drift between the two implementations. The null assertion on `actorId` should also be addressed.

Once those two items are fixed, this is ready to merge.

kristofferremback:
## 🔒 Security Review Results

**Status: ✅ No security vulnerabilities found**

This PR introduces a unified agent architecture for handling both companion mode and @mention invocations. A focused security review was conducted covering:

### Areas Reviewed
- **Input Validation**: Mention extraction patterns, regex handling
- **SQL Injection**: Database query parameterization
- **Authorization**: Workspace scoping for persona lookups
- **Infinite Loop Prevention**: Message author filtering
- **Data Flow**: Job queue trust boundaries

### Findings

| Category | Status | Notes |
|----------|--------|-------|
| SQL Injection | ✅ Secure | All queries use parameterized `sql` template tags |
| Authorization | ✅ Secure | `findBySlug()` correctly scopes to workspace or system personas |
| Regex Injection | ✅ Secure | `hasMention()` properly escapes slugs via `escapeRegex()` |
| Infinite Loops | ✅ Secure | Both listeners filter `actorType !== USER` |
| Trust Boundary | ✅ Secure | Job data originates from validated outbox events |

### Summary
The implementation follows secure coding practices. The workspace-scoped persona lookups, parameterized queries, and proper input escaping provide defense in depth against common attack vectors.

---
🤖 *Automated security review by Claude Code*

kristofferremback (commented):
## Code Review: feat(THR-28): implement agent invocation via @mentions

### Overview

This PR unifies companion mode and @mention invocation into a single `PersonaAgent` architecture. The approach is clean: instead of creating a separate `MentionInvokeAgent` that duplicates code, you renamed `CompanionAgent` to `PersonaAgent` and extended it to handle both triggers. Good decision.

### What's Good ✅

1. **Unified architecture** - Single agent handles both companion mode and mentions, avoiding code duplication
2. **Clean listener separation** - `companion-listener` handles companion mode checks, `mention-invoke-listener` handles @mention extraction. Each listener does one thing well.
3. **Lazy thread creation for channels** - Creates thread only on first message send, not speculatively
4. **Good test coverage for mention-extractor** - 113 lines of tests covering edge cases
5. **Proper concurrency control** - `withSession` prevents duplicate agent runs via `findRunningByStream`
6. **Clean job queue interface** - `PersonaAgentJobData` clearly documents the unified interface with optional `trigger` field

### Issues to Address 🔧

**1. Duplicated callback logic (persona-agent.ts:245-294)**

`sendMessage` and `sendMessageWithSources` have identical implementations. The thread-creation-on-first-message logic is duplicated:

```typescript
// Lines 246-267 and 271-293 are nearly identical
sendMessage: async (msgInput) => {
  // Thread creation logic...
  // Message creation logic...
}
sendMessageWithSources: async (msgInput) => {
  // Same thread creation logic...
  // Same message creation logic...
}
```

**Suggestion**: Extract thread creation into a helper, or have `sendMessage` delegate to `sendMessageWithSources`:

```typescript
sendMessage: async (input) => callbacks.sendMessageWithSources(input)
```

**2. Non-null assertion on `event.actorId` (mention-invoke-listener.ts:80)**

```typescript
triggeredBy: event.actorId!,
```

The `!` assumes `actorId` is always present for USER messages. While likely true, this is a runtime assumption. Consider:
- Adding a guard: `if (!event.actorId) return` with a warning log
- Or documenting why this is safe (e.g., "USER events always have actorId per schema constraint")

**3. Global regex with exec() (mention-extractor.ts:10,26)**

```typescript
const MENTION_PATTERN = /@([\w-]+)/g

while ((match = MENTION_PATTERN.exec(content)) !== null) {
```

Using a global regex with `exec()` relies on the regex's internal `lastIndex` state. This is fine in your implementation because you always iterate to completion, but it's a footgun if someone later calls `extractMentions` mid-iteration or reuses the result.

**Suggestion**: Either:
- Reset `lastIndex` at function start: `MENTION_PATTERN.lastIndex = 0`
- Or use `String.prototype.matchAll()` which is stateless

### Minor Observations 📝

1. **Test naming** - Tests in `mention-extractor.test.ts` follow the pattern `"should extract X"` which is good
2. **Email false positive** - The test shows `test@example.com` extracts `@example` (line 63-66). This is documented behavior but worth noting in the module doc if it matters for production
3. **System prompt modification** - Appending to system prompts works but gets harder to maintain. Consider a template approach long-term

### Architecture 🏗️

The key insight here is good: `PersonaAgent` is the unified execution unit, listeners decide *when* to invoke it. This follows INV-13 (Construct, Don't Assemble) - the agent is constructed once at startup with its deps, and jobs just provide input params.

The `trigger?: "mention"` optional field elegantly distinguishes invocation modes without breaking the interface.

### Test Plan Gaps

The test plan shows:
- [x] Server starts
- [x] Job queue registered
- [x] Listener started
- [ ] E2E tests pending

The pending E2E tests should cover:
1. Mention in channel → thread created → response in thread
2. Mention in thread → response in same thread
3. Mention inactive persona → no response
4. Multiple mentions in one message → multiple responses
5. Companion mode still works after rename

### Verdict

**Approve with minor changes** - The architecture is sound and the unification is the right call. Address the callback duplication and the non-null assertion before merging. The E2E tests can be a follow-up issue.

apps/backend/src/agents/persona-agent.ts
```diff
@@ -222,36 +232,43 @@ export class CompanionAgent {
...
+            authorType: AuthorTypes.PERSONA,
+            content: msgInput.content,
+            sources: msgInput.sources,
+          })
+          return { messageId: message.id, content: msgInput.content }
+        }
 
         // Create callbacks for the response generator
         const callbacks: ResponseGeneratorCallbacks = {
-          // Used by send_message tool (LLM-initiated)
```
kristofferremback: Given that we fetch return `stream` during the pre-check, perhaps we can simplify `buildStreamContext` to take in the stream instead of looking it up?

apps/backend/src/lib/mention-extractor.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      const result = extractMentions("")
+
+      expect(result).toEqual([])
+    })
+
+    it("should not match email addresses", () => {
+      const result = extractMentions("Contact me at test@example.com")
+
+      // Should match @example but not as an email
+      expect(result).toEqual([{ slug: "example", position: 18 }])
```
kristofferremback: Oh no, this seems bad and is oozing of mistakes. The parsing must be much stricter.

apps/backend/src/repositories/search-repository.ts
```diff
@@ -64,6 +71,68 @@ export interface HybridSearchParams {
...
+      `)
+      return result.rows.map((r) => r.id)
+    }
+
+    // With member filter: combined query using UNION for users + personas
+    const result = await client.query<{ id: string }>(sql`
+      WITH accessible AS (
+        SELECT DISTINCT s.id
+        FROM streams s
+        LEFT JOIN stream_members sm ON s.id = sm.stream_id AND sm.user_id = ${userId}
```
kristofferremback: I think this is too naive actually.

Stream visibility is (and must be) implicitly inherited by parent streams, which means a thread of a thread of a thread of a private channel the user is a member of should still be visible to a user that is a member of the parent stream.

Stream membership for threads is primarily a notification mechanism, not specifically for permissions (right now). We should model permissions like _is a member of at last one parent stream_ but right now that means the top-level stream.

We might want to revisit this model in the future, but right now we can't require direct membership for visibility.

kristofferremback: Fixed\! Updated `getAccessibleStreamsWithMembers` to properly handle thread visibility inheritance via `root_stream_id`:

```sql
LEFT JOIN streams root ON s.root_stream_id = root.id
LEFT JOIN stream_members root_sm ON root.id = root_sm.stream_id AND root_sm.user_id = $userId
WHERE ...
  AND (
    sm.user_id IS NOT NULL
    OR s.visibility = 'public'
    OR (s.root_stream_id IS NOT NULL AND (root_sm.user_id IS NOT NULL OR root.visibility = 'public'))
  )
```

Now threads are accessible if:
1. User is a direct member of the thread, OR
2. Thread itself is public, OR
3. User can access the root stream (member OR root is public)

apps/backend/tests/integration/context-builder.test.ts
```diff
@@ -82,16 +82,15 @@ describe("Context Builder", () => {
           content: "Second message",
         })
 
-        const context = await buildStreamContext(client, scratchpadId, msg2Id)
+        const context = await buildStreamContext(client, scratchpad)
```
kristofferremback: Didn't we just add an invariant that touches this?

apps/backend/tests/integration/context-builder.test.ts
```diff
@@ -232,25 +230,24 @@ describe("Context Builder", () => {
...
+        const context = await buildStreamContext(client, thread)
 
-        expect(context).not.toBeNull()
-        expect(context!.streamType).toBe(StreamTypes.THREAD)
-        expect(context!.streamInfo.name).toBe("Thread Discussion")
-        expect(context!.conversationHistory).toHaveLength(1)
-        expect(context!.threadContext).toBeDefined()
-        expect(context!.threadContext!.depth).toBe(2)
-        expect(context!.threadContext!.path).toHaveLength(2)
+        expect(context.streamType).toBe(StreamTypes.THREAD)
```
kristofferremback: Yeah, let's fix these. They are breaking the rules

apps/backend/src/agents/persona-agent.ts
```diff
@@ -325,27 +339,30 @@ export class CompanionAgent {
...
 
   let prompt = persona.systemPrompt
 
+  // Add mention invocation context if applicable
+  if (trigger === "mention") {
+    prompt += `
+
+## Invocation Context
+
+You were explicitly @mentioned by a user who wants your assistance.`
```
kristofferremback: Can we include who? I.e., the human readable name of the mentioner?

apps/frontend/src/lib/markdown/mention-renderer.tsx
```diff
@@ -77,24 +82,44 @@ export function renderMentions(text: string): ReactNode[] {
...
+      triggers.push({ index: match.index, length: match[0].length, type: "mention", slug: match[1] })
+    }
+  }
 
-  TRIGGER_PATTERN.lastIndex = 0
-  while ((match = TRIGGER_PATTERN.exec(processText)) !== null) {
-    if (match.index > lastIndex) {
-      result.push(processText.slice(lastIndex, match.index))
+  // Find channels
+  CHANNEL_PATTERN.lastIndex = 0
```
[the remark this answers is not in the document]
kristofferremback: Fixed! Now cloning the regex instead of resetting `lastIndex`:

```tsx
const channelPattern = new RegExp(CHANNEL_PATTERN.source, CHANNEL_PATTERN.flags)
```

This avoids the scary global state mutation.

apps/backend/tests/integration/context-builder.test.ts
kristofferremback: Fixed! Refactored to use `toMatchObject` instead of assert chains per INV-24:

```ts
expect(context).toMatchObject({
  streamType: StreamTypes.SCRATCHPAD,
  streamInfo: { name: "My Scratchpad", description: "Personal notes" },
  conversationHistory: [{ content: "Hello world" }, { content: "Second message" }],
})
```

apps/backend/src/agents/persona-agent.ts
kristofferremback: Fixed! Now looks up the mentioner and includes their name in the invocation context:

```
You were explicitly @mentioned by **Alice** who wants your assistance.
```

Falls back to "a user" if the mentioner cannot be looked up.

PR #52, opened 2026-01-02
feat: add MessageFormatter for author name resolution in AI prompts

kristofferremback:
## Problem

When formatting messages for AI prompts (stream naming, memo classification, memorization), we only had access to author IDs and types, not actual author names. The prompts showed content like:

```
[user:abc123]: Hello!
[persona:xyz789]: Hi there!
```

This makes it harder for LLMs to understand the conversation context, especially when the same person speaks multiple times or when author identity matters for the knowledge being extracted.

## Solution

Introduce a `MessageFormatter` class that resolves author names from the database and produces enriched message formats for AI prompts.

### How it works

1. **Batch lookup**: MessageFormatter collects all unique author IDs from messages, then makes exactly 2 queries (one for users, one for personas) regardless of message count
2. **Two output formats**:
   - XML format for StreamNamingService: `<message authorType="user" authorId="user_123" authorName="Alice" createdAt="...">content</message>`
   - Inline format for memo system: `[user] Alice: Hello!` or `[ID:msg_123] [user] Alice: Hello!`
3. **Transaction participation**: Methods accept `PoolClient` to participate in the caller's transaction rather than managing their own connections

### Key design decisions

**1. PoolClient in methods, not constructor**

The formatter could have held a `Pool` and managed its own connections. Instead, methods accept `PoolClient` so they participate in the caller's transaction. This ensures consistency when formatting happens within a larger transactional workflow (e.g., stream naming happens within a transaction that also updates the stream).

**2. Dependency injection for MemoClassifier and Memorizer**

Rather than passing MessageFormatter as a parameter to individual methods, we inject it via constructor following INV-13 (Construct, Don't Assemble). This keeps method signatures clean and makes dependencies explicit.

**3. Added findByIds to PersonaRepository**

UserRepository already had `findByIds()` for batch lookups. PersonaRepository was missing this capability, so we added it to enable efficient batch resolution.

## New files

| File | Purpose |
|------|---------|
| `src/lib/ai/message-formatter.ts` | MessageFormatter class with `formatMessages()` and `formatMessagesInline()` methods |
| `src/lib/ai/message-formatter.test.ts` | Comprehensive tests (12 test cases) |

## Modified files

| File | Change |
|------|--------|
| `src/repositories/persona-repository.ts` | Added `findByIds()` for batch lookup |
| `src/lib/ai/text-utils.ts` | Removed old `formatMessage`/`formatMessages` functions |
| `src/services/stream-naming-service.ts` | Uses `MessageFormatter.formatMessages()` |
| `src/services/stream-naming-service.test.ts` | Updated to mock MessageFormatter |
| `src/lib/memo/classifier.ts` | Accepts MessageFormatter, uses `formatMessagesInline()` |
| `src/lib/memo/memorizer.ts` | Accepts MessageFormatter, uses `formatMessagesInline()` with IDs |
| `src/services/memo-service.ts` | Passes PoolClient to classifier/memorizer methods |
| `src/server.ts` | Wires MessageFormatter to all consumers |

## Deleted files

| File | Reason |
|------|--------|
| N/A | Old formatting functions removed from `text-utils.ts` but file retained for other utilities |

## Test plan

- [x] MessageFormatter tests (12 tests)
  - Empty message handling
  - User author name resolution
  - Persona author name resolution
  - Mixed author batch lookup efficiency
  - Unknown author fallback to "Unknown"
  - Message order preservation
  - Inline format without IDs
  - Inline format with IDs
  - Double newline separation
- [x] StreamNamingService tests (12 tests) - all passing
- [x] Type checking passes for modified files

---

### DIFF
--- a/apps/backend/src/lib/ai/message-formatter.test.ts
+++ b/apps/backend/src/lib/ai/message-formatter.test.ts
@@ -74,10 +74,9 @@ describe("MessageFormatter", () => {
 
     const result = await formatter.formatMessages(mockClient, messages)
 
-    expect(result).toContain('authorName="Alice"')
-    expect(result).toContain('authorType="user"')
-    expect(result).toContain('authorId="user_123"')
-    expect(result).toContain("Hello!")
+    expect(result).toBe(
+      '<messages>\n<message authorType="user" authorId="user_123" authorName="Alice" createdAt="2024-01-01T10:00:00.000Z">Hello!</message>\n</messages>'
+    )
   })
 
   test("should format messages with persona author names", async () => {
@@ -95,32 +94,17 @@ describe("MessageFormatter", () => {
 
     const result = await formatter.formatMessages(mockClient, messages)
 
-    expect(result).toContain('authorName="Ariadne"')
-    expect(result).toContain('authorType="persona"')
-    expect(result).toContain('authorId="persona_456"')
-    expect(result).toContain("I can help with that!")
+    expect(result).toBe(
+      '<messages>\n<message authorType="persona" authorId="persona_456" authorName="Ariadne" createdAt="2024-01-01T10:00:01.000Z">I can help with that!</message>\n</messages>'
+    )
   })
 
   test("should batch lookup mixed authors efficiently", async () => {
+    const createdAt = new Date("2024-01-01T10:00:00Z")
     const messages = [
-      createMessage({
-        id: "msg_1",
-        authorId: "user_123",
-        authorType: "user",
-        content: "Question?",
-      }),
-      createMessage({
-        id: "msg_2",
-        authorId: "persona_456",
-        authorType: "persona",
-        content: "Answer!",
-      }),
-      createMessage({
-        id: "msg_3",
-        authorId: "user_123",
-        authorType: "user",
-        content: "Thanks!",
-      }),
+      createMessage({ id: "msg_1", authorId: "user_123", authorType: "user", content: "Question?", createdAt }),
+      createMessage({ id: "msg_2", authorId: "persona_456", authorType: "persona", content: "Answer!", createdAt }),
+      createMessage({ id: "msg_3", authorId: "user_123", authorType: "user", content: "Thanks!", createdAt }),
     ]
 
     mockFindUsersByIds.mockResolvedValue([{ id: "user_123", name: "Alice" }])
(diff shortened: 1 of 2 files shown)


kristofferremback (commented):
# Code Review: MessageFormatter for Author Name Resolution

## Overview

This PR introduces a `MessageFormatter` class that resolves author names (users and personas) when formatting messages for AI prompts. Previously, prompts only showed truncated author IDs like `[user:abc123]`. Now they show actual names like `[user] Alice: Hello!`.

---

## What Works Well

### ✅ **Clean batch lookup pattern**
The `resolveAuthorNames` method efficiently collects unique IDs and makes exactly 2 queries max:
```typescript
const [users, personas] = await Promise.all([
  UserRepository.findByIds(client, [...new Set(userIds)]),
  PersonaRepository.findByIds(client, [...new Set(personaIds)]),
])
```

### ✅ **Transaction participation via PoolClient**
Methods accept `PoolClient` rather than managing their own connections - correctly follows INV-5 and INV-6.

### ✅ **Good test coverage**
12 tests covering edge cases: empty lists, mixed authors, unknown authors, order preservation, both output formats.

### ✅ **Follows INV-13 (Construct, Don't Assemble)**
MessageFormatter is injected into consumers via constructor, not passed to individual methods.

---

## Issues & Suggestions

### 🔴 **Critical: Date formatting inconsistency**

`message-formatter.ts:63`:
```typescript
return `<message ... createdAt="${m.createdAt}">${m.content}</message>`
```

`m.createdAt` is a `Date` object. When coerced to string, it produces locale-dependent output like `Thu Jan 01 2024 10:00:00 GMT+0000`. The test uses:
```typescript
expect(result).toContain("createdAt=")
```

This passes because the attribute exists, but the output format is unpredictable. Use ISO format explicitly:

```typescript
createdAt="${m.createdAt.toISOString()}"
```

### 🟡 **Missing XML escaping for content**

`message-formatter.ts:63`:
```typescript
return `<message ... >${m.content}</message>`
```

If `m.content` contains `<`, `>`, or `&`, the XML becomes malformed. Consider escaping:
```typescript
function escapeXml(s: string): string {
  return s.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;')
}
```

Or document that consumers must handle this. The same issue exists for `authorName` (though less likely to have special chars).

### 🟡 **Missing XML escaping for authorName attribute**

If an author name contains `"`, it breaks the XML attribute:
```typescript
authorName="${authorName}"  // "Bob "The Builder" Smith" breaks this
```

Use proper attribute escaping:
```typescript
authorName="${authorName.replace(/"/g, '&quot;')}"
```

### 🟡 **Unnecessary Set conversion placement**

`message-formatter.ts:51-52`:
```typescript
UserRepository.findByIds(client, [...new Set(userIds)]),
PersonaRepository.findByIds(client, [...new Set(personaIds)]),
```

The deduplication happens at query time, but the arrays are built with duplicates in the loop. More efficient to use Sets from the start:

```typescript
const userIds = new Set<string>()
const personaIds = new Set<string>()

for (const m of messages) {
  if (m.authorType === "user") userIds.add(m.authorId)
  else personaIds.add(m.authorId)
}

const [users, personas] = await Promise.all([
  UserRepository.findByIds(client, [...userIds]),
  PersonaRepository.findByIds(client, [...personaIds]),
])
```

---

## Security Considerations

- **No injection risk**: The formatter only reads from repositories and concatenates strings
- **XML injection**: See issues above about escaping content/attributes

---

## Performance Implications

- **Good**: Batch queries (2 max) instead of N+1
- **Good**: Parallel fetch with `Promise.all`
- **Minor**: Set conversion could be slightly optimized (see above)

---

## Invariants Compliance

| Invariant | Status |
|-----------|--------|
| INV-5 (Repository Pattern) | ✅ Uses PoolClient first param |
| INV-9 (No Singletons) | ✅ Instance passed via DI |
| INV-12 (Pass Dependencies) | ✅ Constructed objects, not config |
| INV-13 (Construct, Don't Assemble) | ✅ Injected via constructor |

---

## Summary

**Verdict: Approve with minor fixes**

The design is sound and follows project conventions well. The main concerns:

1. **Fix**: Use `toISOString()` for date formatting
2. **Consider**: XML escaping for content and attribute values (or document the limitation)
3. **Nice-to-have**: Use Sets from the start in `resolveAuthorNames`

kristofferremback:
## Review Update

All feedback from my previous review has been addressed:

| Issue | Resolution |
|-------|------------|
| Date formatting | ✅ Now uses `toISOString()` |
| XML content escaping | ✅ Added `escapeXml()` function |
| XML attribute escaping | ✅ Added `escapeXmlAttr()` function |
| Set optimization | ✅ Uses Sets from the start in `resolveAuthorNames` |

New tests verify the escaping behavior:
- `should include createdAt in ISO format`
- `should escape XML special characters in message content`
- `should escape quotes in author names`

Ready to merge. 🚀

PR #59, opened 2026-01-06
fix: eliminate layout shift from user preferences on page load

kristofferremback:
## Problem

When users have custom appearance preferences (theme, font size, font family, high contrast, reduced motion, message display density), the page would flash default styles before applying their settings. This happened because preferences were loaded via React Query after the initial render, causing a visible layout shift.

## Solution

Apply cached appearance settings synchronously in an inline script before React mounts, eliminating the flash of unstyled content (FOUC) for returning users.

### How it works

```
Page Load Timeline:
1. Browser parses HTML
2. Inline script runs (pre-React)
   └─ Reads threa-appearance from localStorage
   └─ Applies theme, font-size, font-family, etc. to <html>
3. React mounts with correct styles already applied
4. PreferencesProvider fetches server preferences
   └─ Updates DOM if different (e.g., changed on another device)
   └─ Caches to localStorage for next load
```

### Key design decisions

**1. Single localStorage key for all appearance settings**

Consolidated `threa-theme` into a new `threa-appearance` JSON object that stores all visual preferences. Simpler than multiple keys and allows atomic updates.

**2. Cache on every preference change, not just explicit saves**

Preferences are cached to localStorage whenever they change (including initial load from server). This ensures the cache stays in sync even when preferences are updated on another device.

**3. Only apply cached values if present**

The inline script only applies settings that exist in the cache. Missing keys fall through to CSS defaults, avoiding issues when new preferences are added in the future.

## Modified files

| File | Change |
|------|--------|
| `apps/frontend/index.html` | Expanded inline script to apply all cached appearance settings |
| `apps/frontend/src/contexts/preferences-context.tsx` | Added `cacheAppearanceToLocalStorage()` to persist appearance settings |
| `apps/frontend/src/contexts/theme-context.tsx` | Updated to use `threa-appearance` JSON key instead of `threa-theme` string |
| `apps/frontend/src/components/quick-switcher/quick-switcher.integration.test.tsx` | Fixed unused variable TypeScript error (unrelated cleanup) |

## Test plan

- [x] Build passes (`bun run build`)
- [x] Manual testing: Settings persist across page reloads without flash
- [x] Manual testing: Theme changes apply immediately and persist
- [x] Manual testing: Font size/family changes apply without layout shift on reload

---


kristofferremback (commented):
## Review

Good approach - synchronous localStorage read before React mounts is the correct pattern for eliminating FOUC. The PR description is excellent.

### Concern: Missing migration for existing users

The storage key changed from `threa-theme` (string) to `threa-appearance` (JSON object). Users with existing `threa-theme` values will lose their theme preference on first load after this ships.

**Suggested fix in `index.html`:**

```javascript
var prefs
try {
  prefs = JSON.parse(localStorage.getItem("threa-appearance")) || {}
} catch (e) {
  prefs = {}
}

// Migration: check for legacy key
if (!prefs.theme) {
  var legacyTheme = localStorage.getItem("threa-theme")
  if (legacyTheme === "light" || legacyTheme === "dark" || legacyTheme === "system") {
    prefs.theme = legacyTheme
    localStorage.removeItem("threa-theme") // Clean up
  }
}
```

Same migration logic needed in `getStoredTheme()` in `theme-context.tsx`.

### Question: Dual ownership of localStorage key

Both `ThemeContext.setTheme()` and `PreferencesProvider` write to `threa-appearance`. Is `ThemeContext` still needed, or should it be deprecated in favor of the unified preferences system?

### Minor

The `_date` lint fix in the test file is unrelated - might be cleaner as a separate commit, but not blocking.

PR #62, opened 2026-01-06
feat: implement coordinated loading with delayed skeleton

kristofferremback:
## Problem

When loading a workspace, different parts of the UI (sidebar, main stream, panels) load independently and render as their data arrives. This creates a jarring experience where pieces of the UI appear at different times.

Additionally, there's no skeleton/loading UI - just "Loading..." text while waiting for data.

## Solution

Implement coordinated loading that:
1. **Loads workspace AND all open streams together** - main stream + any open panels
2. **Shows nothing for the first second** - avoids flash of skeleton on fast loads
3. **Shows animated skeleton after 1s** - if still loading, show loading state
4. **Renders content only when ALL data is ready** - single coordinated render

### How it works

```
User navigates to /w/:workspaceId/s/:streamId?panel=panelId
    ↓
WorkspaceLayout extracts all stream IDs (main + panels from URL)
    ↓
CoordinatedLoadingProvider starts:
  - Workspace bootstrap query
  - All stream bootstrap queries in parallel (via useQueries)
  - showSkeleton = false initially
    ↓
After 1s (if still loading):
  - showSkeleton = true → animated skeleton appears
    ↓
All queries complete:
  - isLoading = false
  - Gate renders actual content
```

### Key design decisions

**1. Single loading gate at workspace level**

Rather than having each component check loading state independently, a single `CoordinatedLoadingGate` wraps the entire workspace content. This keeps components dumb and loading logic centralized.

**2. Delayed skeleton (nothing → skeleton → content)**

- Fast loads (<1s): Nothing shown → content appears (no skeleton flash)
- Slow loads (>1s): Nothing → animated skeleton → content

This avoids the ugly "brief skeleton flash" on fast connections while still providing feedback on slow ones.

**3. Draft IDs filtered from server queries**

Stream IDs starting with `draft_` are local IndexedDB data and don't need server fetches. The hook filters these out to avoid unnecessary requests.

**4. Settings/QuickSwitcher outside the gate**

Overlay UI (modals, dialogs) remain outside the loading gate since they don't depend on main content being ready and use cached workspace data.

## New files

| File | Purpose |
|------|---------|
| `src/contexts/coordinated-loading-context.tsx` | Provider + Gate + hook for coordinated loading |
| `src/contexts/coordinated-loading-context.test.tsx` | Tests for loading states and gate behavior |
| `src/hooks/use-coordinated-stream-queries.ts` | Hook to batch multiple stream bootstraps via useQueries |
| `src/hooks/use-coordinated-stream-queries.test.ts` | Tests for draft filtering and loading state |
| `src/components/loading/workspace-skeleton.tsx` | Full workspace skeleton matching AppShell layout |
| `src/components/loading/index.ts` | Export file |

## Modified files

| File | Change |
|------|--------|
| `src/components/ui/skeleton.tsx` | Added `animated` prop to control pulse animation |
| `src/pages/workspace-layout.tsx` | Integrated CoordinatedLoadingProvider + Gate |
| `src/contexts/index.ts` | Export new coordinated loading context |
| `src/hooks/index.ts` | Export new useCoordinatedStreamQueries hook |

## Test plan

- [x] 6 tests for `useCoordinatedStreamQueries` hook
  - Draft ID filtering
  - Loading state calculation
  - Error state handling
  - Empty/all-draft arrays
- [x] 11 tests for `CoordinatedLoadingProvider` and gate
  - isLoading state during loading
  - showSkeleton false initially, true after 1s
  - Gate shows nothing initially, skeleton after 1s
  - Content renders when loading completes
- [ ] Manual verification: Load workspace and observe coordinated loading
- [ ] Manual verification: Open panels and verify they're included in initial load
- [ ] Manual verification: Slow network - verify skeleton appears after 1s

---

### DIFF
--- a/apps/frontend/src/components/layout/sidebar.tsx
+++ b/apps/frontend/src/components/layout/sidebar.tsx
@@ -1,10 +1,11 @@
-import { useState, useRef, useEffect } from "react"
+import { useState, useRef, useEffect, type ReactNode } from "react"
 import { Link, useParams, useNavigate } from "react-router-dom"
 import { MoreHorizontal, Pencil, Archive, Search, CheckCheck, FileEdit } from "lucide-react"
 import { ScrollArea } from "@/components/ui/scroll-area"
 import { Separator } from "@/components/ui/separator"
 import { Button } from "@/components/ui/button"
 import { Input } from "@/components/ui/input"
+import { Skeleton } from "@/components/ui/skeleton"
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -23,11 +24,88 @@ import {
   workspaceKeys,
 } from "@/hooks"
 import { useQuickSwitcher, useCoordinatedLoading } from "@/contexts"
-import { SidebarSkeleton } from "@/components/loading"
 import { UnreadBadge } from "@/components/unread-badge"
 import { StreamTypes } from "@threa/types"
 import { useQueryClient } from "@tanstack/react-query"
 import { ThemeDropdown } from "@/components/theme-dropdown"
+
+// ============================================================================
+// Shell - defines structural layout, receives content via slots
+// ============================================================================
+
+interface SidebarShellProps {
+  header: ReactNode
+  draftsLink: ReactNode
+  streamList: ReactNode
+}
+
+/**
+ * Sidebar structural shell - defines layout without content.
+ * Used by both real Sidebar and skeleton to ensure identical structure.
+ */
+export function SidebarShell({ header, draftsLink, streamList }: SidebarShellProps) {
+  return (
+    <div className="flex h-full flex-col">
+      <div className="flex h-14 items-center justify-between border-b px-4">{header}</div>
+      <div className="border-b px-2 py-2">{draftsLink}</div>
+      <ScrollArea className="flex-1">
+        <div className="p-2">{streamList}</div>
+      </ScrollArea>
+    </div>
+  )
+}
+
+// ============================================================================
+// Skeleton content for each slot
+// ============================================================================
+
+function HeaderSkeleton() {
+  return (
+    <>
+      <Skeleton className="h-5 w-32" />
+      <div className="flex items-center gap-1">
+        <Skeleton className="h-8 w-8 rounded-md" />
+        <Skeleton className="h-8 w-8 rounded-md" />
+      </div>
+    </>
+  )
+}
+
+function DraftsLinkSkeleton() {
+  return <Skeleton className="h-9 w-full rounded-md" />
+}
+
+function StreamListSkeleton() {
+  return (
+    <>
+      {/* Scratchpads section */}
+      <div className="mb-4">
+        <Skeleton className="mb-2 h-3 w-20 px-2" />
+        <div className="space-y-1">
+          <Skeleton className="h-8 w-full rounded-md" />
+          <Skeleton className="h-8 w-full rounded-md" />
+          <Skeleton className="h-8 w-full rounded-md" />
+        </div>
+      </div>
+
+      {/* Separator */}
+      <div className="my-2 h-px bg-border" />
+
+      {/* Channels section */}
+      <div>
+        <Skeleton className="mb-2 h-3 w-16 px-2" />
+        <div className="space-y-1">
+          <Skeleton className="h-8 w-full rounded-md" />
+          <Skeleton className="h-8 w-full rounded-md" />
+        </div>
+      </div>
+    </>
+  )
+}
+
+// ============================================================================
+// Main Sidebar component
+// ============================================================================
 
 interface SidebarProps {
   workspaceId: string
@@ -48,9 +126,15 @@ export function Sidebar({ workspaceId }: SidebarProps) {
   const draftCount = allDrafts.length
   const isDraftsPage = splat === "drafts" || window.location.pathname.endsWith("/drafts")
 
-  // During coordinated loading, show skeleton
+  // During coordinated loading, show skeleton using the same shell
   if (coordinatedLoading) {
-    return <SidebarSkeleton />
+    return (
+      <SidebarShell
+        header={<HeaderSkeleton />}
+        draftsLink={<DraftsLinkSkeleton />}
+        streamList={<StreamListSkeleton />}
+      />
+    )
   }
 
   const handleCreateScratchpad = async () => {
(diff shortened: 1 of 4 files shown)


kristofferremback (commented):
## Code Review

### Overview

This PR implements coordinated loading for the workspace, ensuring all required data (workspace + streams + panels) loads together before rendering content. It includes a delayed skeleton pattern (nothing for 1s, then skeleton) to avoid flash on fast loads.

### What's Good

1. **Clean abstraction** - The `CoordinatedLoadingProvider` + `CoordinatedLoadingGate` pattern separates loading coordination from UI components nicely

2. **Well-tested** - 17 tests covering the key behaviors: timing, state transitions, edge cases

3. **Thoughtful UX** - The delayed skeleton pattern avoids skeleton flash on fast loads while still providing feedback on slow ones

4. **Draft ID filtering** - Smart to skip server fetches for local-only draft IDs

5. **Reuses existing patterns** - Uses `useQueries` and reuses `streamKeys.bootstrap` for cache coordination

### Issues to Address

**1. Skeleton animation prop default inconsistency**

In `skeleton.tsx`, the default is no longer animated:
```tsx
// Old: always animated
<div className={cn("animate-pulse rounded-md bg-muted", className)} {...props} />

// New: animated=false by default
animated && "animate-pulse"
```

But `WorkspaceSkeleton` defaults to `animated = true`. This creates a breaking change for any other code using `<Skeleton />` directly - they'll suddenly stop animating.

**Suggestion**: Default `animated` to `true` to maintain backward compatibility:
```tsx
function Skeleton({ className, animated = true, ...props }: SkeletonProps) {
```

**2. Potential race with staleTime: 0**

In `use-coordinated-stream-queries.ts`:
```tsx
staleTime: 0,
```

This means every render triggers a refetch. Combined with the streaming queries already running, this could cause duplicate requests if components re-render during loading.

**Question**: Was `staleTime: 0` intentional? Consider using a short staleTime (e.g., 1000ms) or `Infinity` if the data is already fresh from the coordinated load.

**3. Missing workspaceId validation in the hook**

`useCoordinatedStreamQueries` has `enabled: !!workspaceId` but the Provider passes it through regardless. If `workspaceId` is undefined, the effect still runs and `showSkeleton` can flip states unnecessarily.

**Suggestion**: Early return in the provider if workspaceId is falsy, matching the existing pattern in `WorkspaceLayout`.

**4. Test mock location could cause issues**

The test mocks are at module scope:
```tsx
let mockWorkspaceLoading = true
vi.mock("@/hooks/use-workspaces", () => ({
  useWorkspaceBootstrap: () => ({ isLoading: mockWorkspaceLoading, ... })
}))
```

This works but can cause flaky tests if tests run in parallel or if the mock state leaks. Consider using `vi.mocked()` with per-test setup instead.

### Minor Suggestions

1. **Consider extracting the delay constant** - `SKELETON_DELAY_MS = 1000` is great, but it's buried in the context file. If this needs tuning based on UX feedback, having it in a constants file would be easier.

2. **The skeleton structure is hardcoded** - It assumes the exact layout of AppShell (sidebar width, header height, etc.). If the layout changes, the skeleton and real UI could diverge. Consider using CSS variables shared between them.

3. **Consider loading state granularity** - Right now it's all-or-nothing. Future enhancement could show sidebar skeleton first (workspace data), then main content skeleton (stream data), for progressive loading.

### Questions

1. What happens if a panel stream fails to load? Currently `isError` is set but content is still blocked. Should partial rendering be considered?

2. Is there a case where `streamIds` could contain duplicates (e.g., same stream open as main and panel)? The queries would dedupe via cache keys, but worth considering.

### Summary

**Good PR** - solves a real UX problem with a clean implementation. The delayed skeleton pattern is a thoughtful touch.

**Before merging:**
1. Fix the Skeleton animated default (breaking change)
2. Clarify the staleTime: 0 choice

kristofferremback:
## 🔒 Security Review

**No security vulnerabilities found.**

### Analysis Summary

| Area | Assessment |
|------|------------|
| **Input Validation** | ✅ URL parameters (`streamId`, `panel`) are passed to backend APIs which use parameterized queries. No SQL injection risk. |
| **Authorization** | ✅ All stream data fetches go through `validateStreamAccess()` on the backend. Frontend cannot bypass server-side authorization. |
| **XSS** | ✅ No user-controlled data rendered with `dangerouslySetInnerHTML`. Standard React rendering is safe. |
| **Path Traversal** | ✅ URL construction uses template literals with path segments. No file system operations based on user input. |
| **Data Exposure** | ✅ IndexedDB caches data already authorized by the server. No cross-workspace or cross-user data leakage possible. |

### Details

This PR implements coordinated loading in a security-appropriate manner:

- All sensitive data access is gated by backend authorization
- No new attack surfaces are introduced
- The frontend correctly delegates authorization decisions to the backend
- Input validation occurs at appropriate boundaries (backend handlers with Zod schemas, SQL queries with parameterized statements)

**Recommendation**: Safe to merge from a security perspective.

---
🤖 *Automated security review*

kristofferremback:
## Review Response

Thanks for the thorough review! I've addressed the feedback in commit 22bb732:

### Issues Addressed

**1. Skeleton animated prop default (Severity 4)** ✅
- Changed default to `animated = true` to maintain backward compatibility
- Existing code using `<Skeleton />` will continue to animate

**2. staleTime: 0 clarification (Severity 3)** ✅
- Added comment explaining this matches `useStreamBootstrap` pattern
- Intent: always refetch to ensure fresh data (socket events may be missed when not subscribed)
- React Query dedupes by key, so no duplicate requests occur

**3. workspaceId validation (Severity 3)** ✅
- Verified `useWorkspaceBootstrap` already has `enabled: !!workspaceId`
- `workspace-layout.tsx` guards with `if (!workspaceId) return null` before rendering provider
- No changes needed - already handled

**4. Skeleton ownership refactor** ✅
- Agreed this was worth addressing for maintainability
- Components now own their skeleton rendering:
  - `Sidebar` checks `useCoordinatedLoading()` and renders `SidebarSkeleton`
  - `MainContentGate` wraps Outlet and shows `StreamContentSkeleton`
- Exported `SidebarSkeleton` and `StreamContentSkeleton` for reuse
- Removed monolithic `WorkspaceSkeleton` in favor of component-owned skeletons

### Answers to Questions

**Q: What happens if a panel stream fails to load?**
A: `isError` is set but Gate still blocks until loading completes. This is intentional - we wait for all queries to settle. Partial rendering could be a future enhancement.

**Q: Could streamIds contain duplicates?**
A: Theoretically yes if same stream is main + panel. React Query dedupes via cache keys, so only one fetch occurs. No adverse effects.

### Test Coverage
Updated tests to reflect new architecture: 20 tests passing (was 17).

PR #66, opened 2026-01-07
feat(THR-73): AI cost tracking with origin tagging and admin UI

kristofferremback:
**Linear:** [THR-73](https://linear.app/threa/issue/THR-73)

## Problem

AI costs in Threa were invisible - no way to see what AI operations cost, no ability to set budgets, and no distinction between system-initiated AI work (embeddings, classification) vs user-initiated work (companion responses). This made cost management impossible and prevented planning for seat-based pricing where system vs user costs need different treatment.

## Solution

Implement comprehensive AI cost tracking with:
1. **Cost capture** - Intercept OpenRouter API responses to extract usage metadata
2. **Origin tagging** - Distinguish `system` (background jobs) from `user` (companion responses) AI calls
3. **Budget management** - Monthly budgets with configurable alerts and hard limits
4. **Admin UI** - Dashboard showing usage breakdown and budget settings

### How it works

```
AI Call → OpenRouter → Response headers (x-ratelimit-usage-tokens, x-ratelimit-usage-cost)
                          ↓
                   Cost Interceptor extracts usage
                          ↓
                   Cost Service records to ai_usage_records
                          ↓
                   Admin UI queries aggregated data
```

### Key design decisions

**1. Response header interception vs token counting**

Chose to capture costs from OpenRouter's response headers rather than counting tokens ourselves. OpenRouter provides exact cost (`x-ratelimit-usage-cost`) which accounts for their actual pricing, avoiding drift between our estimates and reality.

**2. Origin tracking (system vs user)**

Added explicit `origin` column rather than inferring from `userId === null`. This enables:
- Future separate budgets for system vs user usage
- Clearer UI that doesn't leak implementation details
- Better attribution as system jobs may run with user context

**3. Simplified system usage display**

The UI shows "Automated background processes" instead of "classification, embeddings, stream naming" to avoid exposing implementation details to end users.

**4. Per-workspace budgets (not per-user)**

Budgets are workspace-scoped. User quotas table exists but isn't exposed in UI yet - the initial focus is workspace-level cost visibility.

## New files

| File | Purpose |
|------|---------|
| `migrations/021_ai_cost_tracking.sql` | Schema for usage records, budgets, quotas, alerts |
| `migrations/022_ai_usage_origin.sql` | Add origin column to usage records |
| `lib/ai/openrouter-cost-interceptor.ts` | Fetch wrapper to capture cost from response headers |
| `repositories/ai-usage-repository.ts` | CRUD for usage records with aggregation queries |
| `repositories/ai-budget-repository.ts` | CRUD for budgets, quotas, alerts |
| `services/ai-cost-service.ts` | High-level cost recording with buffering |
| `services/ai-budget-service.ts` | Budget checking and alert logic |
| `handlers/ai-usage-handlers.ts` | REST endpoints for usage/budget API |
| `frontend/pages/ai-usage-admin.tsx` | Admin dashboard with usage charts and budget settings |
| `frontend/api/ai-usage.ts` | API client for usage endpoints |
| `frontend/hooks/use-ai-usage.ts` | React Query hooks for usage data |

## Modified files

| File | Change |
|------|--------|
| `lib/ai/ai.ts` | Add `CostRecorder` interface and `CostContext` for tracking |
| `agents/companion-runner.ts` | Add `origin: "user"` to cost recording |
| `lib/memo/classifier.ts` | Add `origin: "system"` via context |
| `lib/memo/memorizer.ts` | Add `origin: "system"` via context |
| `lib/boundary-extraction/llm-extractor.ts` | Add `origin: "system"` via context |
| `services/embedding-service.ts` | Add `origin: "system"` to embeddings |
| `services/stream-naming-service.ts` | Add `origin: "system"` to naming |
| `components/layout/sidebar.tsx` | Add AI Usage link for workspace admins |
| `packages/types/src/api.ts` | Add API types for usage/budget endpoints |

## Test plan

- [x] Unit tests for AI wrapper with cost context
- [x] Type checking passes for all packages
- [ ] Manual: Navigate to AI Usage page, verify summary cards show
- [ ] Manual: Update budget, verify changes persist
- [ ] Manual: Send messages to trigger AI calls, verify usage appears

---

### DIFF
--- a/apps/backend/src/lib/boundary-extraction/llm-extractor.ts
+++ b/apps/backend/src/lib/boundary-extraction/llm-extractor.ts
@@ -94,7 +94,7 @@ export class LLMBoundaryExtractor implements BoundaryExtractor {
             activeConversationCount: context.activeConversations.length,
           },
         },
-        context: context.workspaceId ? { workspaceId: context.workspaceId } : undefined,
+        context: context.workspaceId ? { workspaceId: context.workspaceId, origin: "system" } : undefined,
       })
 
       return this.validateResult(value, context)
--- a/apps/backend/src/lib/memo/memorizer.ts
+++ b/apps/backend/src/lib/memo/memorizer.ts
@@ -125,7 +125,9 @@ export class Memorizer {
       .replace("{{CONTENT}}", message.content)
       .replace("{{EXISTING_TAGS_SECTION}}", existingTagsSection)
 
-    const costContext: CostContext | undefined = context.workspaceId ? { workspaceId: context.workspaceId } : undefined
+    const costContext: CostContext | undefined = context.workspaceId
+      ? { workspaceId: context.workspaceId, origin: "system" }
+      : undefined
 
     const { value } = await this.ai.generateObject({
       model: this.modelId,
@@ -170,7 +172,9 @@ export class Memorizer {
       .replace("{{MESSAGES}}", messagesText)
       .replace("{{EXISTING_TAGS_SECTION}}", existingTagsSection)
 
-    const costContext: CostContext | undefined = context.workspaceId ? { workspaceId: context.workspaceId } : undefined
+    const costContext: CostContext | undefined = context.workspaceId
+      ? { workspaceId: context.workspaceId, origin: "system" }
+      : undefined
 
     const { value } = await this.ai.generateObject({
       model: this.modelId,
@@ -223,7 +227,9 @@ export class Memorizer {
       .replace("{{MESSAGES}}", messagesText)
       .replace("{{EXISTING_TAGS_SECTION}}", existingTagsSection)
 
-    const costContext: CostContext | undefined = context.workspaceId ? { workspaceId: context.workspaceId } : undefined
+    const costContext: CostContext | undefined = context.workspaceId
+      ? { workspaceId: context.workspaceId, origin: "system" }
+      : undefined
 
     const { value } = await this.ai.generateObject({
       model: this.modelId,
--- a/apps/backend/src/services/ai-cost-service.ts
+++ b/apps/backend/src/services/ai-cost-service.ts
@@ -1,6 +1,12 @@
 import type { Pool, PoolClient } from "pg"
 import { withClient, withTransaction } from "../db"
-import { AIUsageRepository, AIBudgetRepository, OutboxRepository, type UsageSummary } from "../repositories"
+import {
+  AIUsageRepository,
+  AIBudgetRepository,
+  OutboxRepository,
+  type UsageSummary,
+  type AIUsageOrigin,
+} from "../repositories"
 import { aiUsageId, aiAlertId } from "../lib/id"
 import { logger } from "../lib/logger"
 import type { UsageWithCost, ParsedModel } from "../lib/ai/ai"
@@ -19,6 +25,7 @@ export interface RecordUsageParams {
   functionId: string
   model: string
   provider: string
+  origin: AIUsageOrigin
   usage: UsageWithCost
   metadata?: Record<string, unknown>
 }
@@ -75,6 +82,7 @@ export class AICostService implements AICostServiceLike {
         completionTokens: params.usage.completionTokens ?? 0,
         totalTokens: params.usage.totalTokens ?? 0,
         costUsd: cost,
+        origin: params.origin,
         metadata: params.metadata,
       })
 
@@ -236,6 +244,17 @@ export class AICostService implements AICostServiceLike {
 
     return withClient(this.pool, (client) =>
       AIUsageRepository.getUsageByUser(client, workspaceId, periodStart, periodEnd)
+    )
+  }
+
+  /**
+   * Get usage breakdown by origin (system vs user) for current month.
+   */
+  async getUsageByOrigin(workspaceId: string) {
+    const { periodStart, periodEnd } = this.getCurrentMonthPeriod()
+
+    return withClient(this.pool, (client) =>
+      AIUsageRepository.getUsageByOrigin(client, workspaceId, periodStart, periodEnd)
     )
   }
 
--- a/apps/backend/src/agents/companion-runner.ts
+++ b/apps/backend/src/agents/companion-runner.ts
@@ -226,6 +226,7 @@ export class LangGraphResponseGenerator implements ResponseGenerator {
           functionId: "companion-response",
           model: parsedModel.modelId,
           provider: parsedModel.provider,
+          origin: "user",
           usage: {
             promptTokens: usage.promptTokens,
             completionTokens: usage.completionTokens,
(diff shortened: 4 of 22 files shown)


kristofferremback (commented):
## Code Review: PR #66 - AI Cost Tracking with Origin Tagging

### Overview

This PR implements comprehensive AI cost tracking for the Threa workspace. It captures costs from OpenRouter API responses, distinguishes between system and user-initiated AI calls, provides budget management, and adds an admin UI dashboard. The implementation is well-structured and follows the project's patterns.

### What's Good

- **Clean architecture**: The cost tracking is non-invasive - it intercepts at the fetch level and records asynchronously without blocking AI operations
- **Explicit origin tagging**: Using an explicit `origin` column rather than inferring from `userId === null` is the right design choice for future flexibility
- **Good separation of concerns**: Clear boundaries between interceptor → service → repository → handlers
- **Real response fixtures**: Capturing actual OpenRouter responses as test fixtures documents the expected API contract
- **No-op fallback**: `createNoOpCostService()` enables graceful degradation when cost tracking is disabled or in tests
- **Follows project patterns**: Uses prefixed ULIDs, repository pattern with `PoolClient` first param, no DB enums
- **Good UI/UX**: Loading skeletons, proper error states, clean breakdown between system vs user costs

### Issues & Suggestions

#### 1. **Potential race condition in alert checking** (ai-cost-service.ts:89-93)

```typescript
// Check and fire alerts (fire-and-forget, don't block on errors)
this.checkAndFireAlerts(client, params.workspaceId).catch((error) => {
  logger.error({ error, workspaceId: params.workspaceId }, "Failed to check budget alerts")
})
```

**Problem**: The `client` from `withTransaction` may be released before `checkAndFireAlerts` completes since it's fire-and-forget. This could cause "client already released" errors.

**Fix**: Either await the alert check (making it part of the transaction) or pass the `pool` and let `checkAndFireAlerts` get its own client.

#### 2. **Streaming response parsing may miss usage data** (openrouter-cost-interceptor.ts:57-75)

The streaming parser iterates in reverse looking for usage, but:
1. It breaks on the first JSON object with `usage`, which might not be the final/complete one
2. OpenRouter may include usage in a separate structure for streaming responses

**Suggestion**: Verify this works correctly with actual streaming responses. Consider logging if no usage is found for streaming responses to catch issues early.

#### 3. **AsyncLocalStorage context may not propagate to all LangChain calls** (openrouter-cost-interceptor.ts)

The `withCostTracking` wrapper relies on `AsyncLocalStorage` to track costs across the execution context. However, LangChain's internal async operations may spawn contexts where the store isn't available.

**Suggestion**: Add logging when `costStorage.getStore()` returns null for OpenRouter calls - this would surface if cost capture is missing data.

#### 4. **Missing index for user-scoped queries** (021_ai_cost_tracking.sql)

The migration creates indexes for workspace+created and user+created, but the repository's `getUserUsage` method queries by `workspace_id AND user_id AND created_at BETWEEN`. Consider adding a composite index:

```sql
CREATE INDEX idx_ai_usage_workspace_user_created 
ON ai_usage_records(workspace_id, user_id, created_at DESC);
```

#### 5. **Budget update could be more atomic** (ai-usage-handlers.ts:179-206)

The update handler does a find-then-update, which could have race conditions if two updates happen simultaneously. Consider using `INSERT ... ON CONFLICT DO UPDATE` as a single atomic operation instead of the two-step approach.

#### 6. **Frontend state not synced with server on blur** (ai-usage-admin.tsx:254-256)

```typescript
const [localBudget, setLocalBudget] = useState<string>(budget?.monthlyBudgetUsd?.toString() ?? "50")
```

The local state initializes from `budget` but doesn't update when `budget` changes (e.g., after a refetch). This could cause stale UI after server updates.

**Fix**: Either use `useEffect` to sync when `budget` changes, or use a controlled pattern where the server value is the source of truth.

#### 7. **Type safety for origin values** (various files)

The `origin` field is typed as `"system" | "user"` but stored as `TEXT` in the database. Consider adding a check constraint:

```sql
CHECK (origin IN ('system', 'user'))
```

### Minor Nits

- `test-cost-capture.ts` - Nice for manual testing, but should probably be in a `scripts/` directory rather than `lib/ai/`
- The PR description mentions response headers (`x-ratelimit-usage-cost`) but the code extracts from response body. The implementation looks correct (body is more reliable), but the description could be updated for clarity.

### Questions

1. **Cost accuracy**: Has this been validated against OpenRouter billing to ensure captured costs match actual charges?
2. **Budget enforcement**: The `degradationEnabled` and `hardLimitEnabled` flags are stored but don't appear to be enforced anywhere yet. Is this intentional for a follow-up PR?
3. **Retention policy**: Is there a plan for archiving/purging old `ai_usage_records`? At scale, this table could grow quickly.

### Summary

Overall this is a solid implementation of a complex feature. The architecture is sound and follows project patterns well. The main concerns are:

1. The fire-and-forget alert check could cause transaction issues
2. Some edge cases around streaming response parsing
3. Minor frontend state sync issues

With these addressed, this looks good to merge.

apps/backend/src/lib/boundary-extraction/llm-extractor.ts
```diff
@@ -94,6 +94,7 @@ export class LLMBoundaryExtractor implements BoundaryExtractor {
             activeConversationCount: context.activeConversations.length,
           },
         },
+        context: context.workspaceId ? { workspaceId: context.workspaceId, origin: "system" } : undefined,
```
kristofferremback: When would we not have a workspace ID?

apps/backend/src/lib/memo/memorizer.ts
```diff
@@ -123,6 +125,10 @@ export class Memorizer {
       .replace("{{CONTENT}}", message.content)
       .replace("{{EXISTING_TAGS_SECTION}}", existingTagsSection)
 
+    const costContext: CostContext | undefined = context.workspaceId
```
kristofferremback: Same comment on optionality as the others. This applies to **all** places. There should always be workspace ID

apps/backend/src/services/ai-cost-service.ts
```diff
@@ -0,0 +1,305 @@
...
+        provider: params.provider,
+        promptTokens: params.usage.promptTokens ?? 0,
+        completionTokens: params.usage.completionTokens ?? 0,
+        totalTokens: params.usage.totalTokens ?? 0,
+        costUsd: cost,
+        origin: params.origin,
+        metadata: params.metadata,
+      })
+
+      // Check and fire alerts (fire-and-forget, don't block on errors)
```
kristofferremback: We don't do fire and forget. These should be handled via the Outbox pattern.

apps/backend/src/services/search-service.ts
```diff
@@ -58,7 +58,7 @@ export class SearchService {
     let embedding: number[] = []
     if (query.trim()) {
       try {
-        embedding = await this.embeddingService.embed(query)
+        embedding = await this.embeddingService.embed(query, { workspaceId })
```
kristofferremback: I feel like we would want to be able to separate the cost of embedding via user interactivity (e.g., search) vs system handling (e.g., embedding messages). They are all system stuff, but different operations that would be valuable to capture!

PR #73, opened 2026-01-09
fix: prevent AI from mimicking timestamp format in responses

kristofferremback:
## Problem

When timestamps were added to conversation history for temporal context, Ariadne started including timestamps like `(14:30)` at the beginning of her responses. This happened because the model was learning from its own previous messages in the context, which were formatted with timestamp prefixes.

## Solution

Two changes to break the multi-shot learning pattern:

1. **Remove timestamps from assistant messages** - Only user messages now include the `(time)` prefix. The model no longer sees its own responses with timestamps.

2. **Add explicit instruction** - The system prompt now includes: "User messages are prefixed with their send time, e.g., (14:30). Do not include timestamps in your responses."

### Key design decisions

**1. Keep timestamps on user messages only**

User message timestamps provide valuable temporal context (when did users speak, how much time passed between messages). Assistant message timestamps don't add meaningful information - the model doesn't need to know when it previously responded to maintain coherent conversation.

**2. Belt-and-suspenders approach**

Both removing the multi-shot examples AND adding explicit instruction. Either alone might work, but together they're more robust.

## Modified files

| File | Change |
|------|--------|
| `apps/backend/src/agents/persona-agent.ts` | Remove `(${time})` prefix from assistant messages in `formatMessagesWithTemporal()` |
| `apps/backend/src/lib/temporal.ts` | Add `timestampInstruction` to `buildTemporalPromptSection()` explaining user timestamps and instructing model not to include them |

## Test plan

- [x] All 256 unit tests pass
- [x] `temporal.test.ts` tests still pass (use `toContain` so new instruction doesn't break assertions)
- [ ] Manual verification: send messages and confirm Ariadne no longer prefixes responses with timestamps

---

_PR by [Claude Code](https://claude.com/claude-code)_

### DIFF
--- a/apps/backend/src/agents/persona-agent.ts
+++ b/apps/backend/src/agents/persona-agent.ts
@@ -615,23 +615,20 @@ function formatMessagesWithTemporal(
   let currentDateKey: string | null = null
 
   for (const msg of messages) {
-    const msgDateKey = getDateKey(msg.createdAt, temporal.timezone)
-
-    // Insert date boundary marker when date changes
-    // Prepend to the next message's content rather than inserting a fake message
-    let dateBoundaryPrefix = ""
-    if (msgDateKey !== currentDateKey) {
-      const dateStr = formatDate(msg.createdAt, temporal.timezone, temporal.dateFormat)
-      dateBoundaryPrefix = `[Date: ${dateStr}]\n`
-      currentDateKey = msgDateKey
-    }
-
-    // Format message with timestamp
-    const time = formatTime(msg.createdAt, temporal.timezone, temporal.timeFormat)
     const role = msg.authorType === AuthorTypes.USER ? ("user" as const) : ("assistant" as const)
 
     if (msg.authorType === AuthorTypes.USER) {
-      // For user messages in multi-user contexts, include the name
+      // Check for date boundary - only on user messages to avoid model mimicking the format
+      const msgDateKey = getDateKey(msg.createdAt, temporal.timezone)
+      let dateBoundaryPrefix = ""
+      if (msgDateKey !== currentDateKey) {
+        const dateStr = formatDate(msg.createdAt, temporal.timezone, temporal.dateFormat)
+        dateBoundaryPrefix = `[Date: ${dateStr}]\n`
+        currentDateKey = msgDateKey
+      }
+
+      // Format with timestamp and optional author name for multi-user contexts
+      const time = formatTime(msg.createdAt, temporal.timezone, temporal.timeFormat)
       const authorName = authorNames.get(msg.authorId) ?? "Unknown"
       const hasMultipleUsers = context.streamType === StreamTypes.CHANNEL || context.streamType === StreamTypes.DM
       const namePrefix = hasMultipleUsers ? `[@${authorName}] ` : ""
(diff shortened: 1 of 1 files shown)


kristofferremback (commented):
## Code Review: PR #73

### Overview
This PR addresses an issue where the AI was mimicking the timestamp format `(14:30)` in its responses after timestamps were added to conversation history. The fix uses a belt-and-suspenders approach: removing timestamps from assistant messages and adding an explicit instruction.

### What I Like

**1. Root cause analysis is solid**
The PR correctly identifies that the model was learning from multi-shot examples in its own conversation history. Removing the timestamps from assistant messages breaks this pattern at the source.

**2. Belt-and-suspenders approach**
Using both structural changes (removing timestamps) AND explicit instructions is defensive. Either alone might work, but together they're more robust against model variance.

**3. Minimal, focused changes**
Only 2 files touched, 5 additions/4 deletions. The change is surgical.

**4. Good PR description**
Clear problem statement, rationale for design decisions, and test plan.

---

### Suggestions

**1. Test coverage for the new instruction**

The existing tests use `toContain` which means they'll pass, but there's no explicit test verifying the new `timestampInstruction` is included. Consider adding:

```typescript
it("should include instruction about not including timestamps in responses", () => {
  const section = buildTemporalPromptSection(baseContext)
  expect(section).toContain("Do not include timestamps in your responses")
})
```

This makes the behavior explicit in tests and guards against accidental removal.

**2. Test for `formatMessagesWithTemporal` assistant message formatting**

There don't appear to be tests for `formatMessagesWithTemporal()`. The function now has different behavior for user vs assistant messages - this is a good candidate for a test:

```typescript
it("should include timestamp for user messages but not assistant messages", () => {
  const messages = [
    { authorType: "user", content: "Hello", createdAt: new Date(), authorId: "u1" },
    { authorType: "persona", content: "Hi there", createdAt: new Date(), authorId: "p1" },
  ]
  const result = formatMessagesWithTemporal(messages, temporal, context)
  
  expect(result[0].content).toMatch(/^\(\d{1,2}:\d{2}/) // User has timestamp
  expect(result[1].content).not.toMatch(/^\(\d{1,2}:\d{2}/) // Assistant does not
})
```

**3. Minor: Instruction wording**

The instruction says "User messages are prefixed with their send time" - but it's actually the message's `createdAt` time, not necessarily when it was "sent" (though usually the same). Probably fine, but worth noting the slight semantic gap.

---

### Questions

**Q1: What about date boundary markers?**
Assistant messages still get `[Date: 2026-01-06]` prefixes when the date changes. Is there risk of the model mimicking this format too? Probably lower risk since it's less frequent and more visually distinct, but worth considering.

---

### Verdict

**Approve** - Clean fix that addresses the root cause. The suggestions above are improvements, not blockers.

kristofferremback:
## Security Review

**Result: No security vulnerabilities detected** ✅

### Changes Reviewed

| File | Lines Changed |
|------|---------------|
| `apps/backend/src/agents/persona-agent.ts` | 4 |
| `apps/backend/src/lib/temporal.ts` | 5 |

### Analysis Summary

This PR makes focused modifications to temporal prompt handling:

1. **persona-agent.ts**: Removed timestamps from assistant/persona messages to prevent the AI model from mimicking the timestamp format
2. **temporal.ts**: Added explicit instruction clarifying that user messages have timestamps and the model should not include them

### Security Checks

| Category | Status |
|----------|--------|
| Input Validation | No user-controllable data in new code paths |
| Injection Vectors | None - `timestampInstruction` uses controlled enum values only |
| Authentication/Authorization | No changes |
| Data Exposure | No sensitive data handling changes |
| Crypto/Secrets | No changes |

### Key Observations

- The `formatExample` variable derives from `temporal.timeFormat`, which is a type-safe enum (`"12h" \| "24h"`) validated at database level
- All timestamp formatting uses `Intl.DateTimeFormat` API (no string concatenation with user data)
- The change actually **improves** behavioral constraints by making them explicit rather than implicit

**Confidence: High (0.95)**

---
*Automated security review*

PR #78, opened 2026-01-13
feat: migrate to ProseMirror JSON storage with command routing

kristofferremback:
## Problem

Messages were stored as plain markdown (`TEXT`), limiting rich content support. Additionally:
- Slash commands required a separate endpoint (`POST /commands/dispatch`)
- When the server was down, the frontend hammered the backend with bootstrap requests
- Legacy code existed to handle pre-migration data formats

### 1. ProseMirror JSON as Primary Storage

Migrate from markdown to structured JSON (JSONB) for message content:
- **New column**: `content_json` (JSONB, NOT NULL) stores ProseMirror document structure
- **Keep column**: `content_markdown` (TEXT) for search, AI context, and backwards compatibility
- All clients now send `contentJson`, backend normalizes and stores both formats

### 2. Slash Command Routing via Messages

Unify the command flow - commands embedded in messages are automatically detected and executed:
- `detectCommand()` checks if first inline node is a command type
- Registered commands → dispatch with 202 response
- Unregistered commands → fall back to normal message creation
- Same guarantees as dedicated endpoint (transactional, outbox pattern)

### 3. Bootstrap Request Protection

Add defensive options to `useWorkspaceBootstrap` matching `useStreamBootstrap`:
- Error state tracking prevents re-enabling after failure
- Disable all automatic refetch triggers (`refetchOnWindowFocus`, etc.)

### 4. Legacy Code Cleanup

Remove ~180 lines of code that existed only for pre-migration data:
- Legacy outbox format handling (messageId at top level)
- Deprecated cost tracking wrapper functions
- Legacy socket.io room format (`stream:xxx`)

## Key Design Decisions

**1. Dual storage (JSON + Markdown)**
Keep both formats: JSON for rich rendering, markdown for search/AI. The backend normalizes both directions - JSON clients get markdown derived, markdown clients get JSON parsed.

**2. Command detection before normalization**
Commands must be detected from the original `contentJson` before `normalizeContent()` runs, since normalization serializes to markdown and back, losing custom node types.

**3. Consistent response shapes**
Command dispatch returns `{ command: { id, name, args, status } }` matching the message pattern `{ message: { id, ... } }`.

## New Files

| File | Purpose |
|------|---------|
| `packages/prosemirror/` | Shared ProseMirror utilities (markdown ↔ JSON) |
| `packages/types/src/prosemirror.ts` | TypeScript types for ProseMirror nodes |
| `026_message_prosemirror_content.sql` | Migration adding `content_json` column |
| `message-content-regression.test.ts` | E2E tests for content handling edge cases |

## Modified Files

| File | Change |
|------|--------|
| `message-handlers.ts` | Add command detection and routing logic |
| `event-service.ts` | Accept `contentJson` as primary, derive markdown |
| `message-repository.ts` | Store/retrieve JSON content |
| `use-workspaces.ts` | Add bootstrap request protection |
| `use-draft-*.ts` | Store drafts as JSON |
| `outbox-payload-parsers.ts` | Remove legacy format handling |
| `openrouter-cost-interceptor.ts` | Remove deprecated wrapper functions |
| `socket.ts` | Remove legacy room format |

## Test Plan

- [x] E2E tests for slash command routing (4 new tests)
- [x] E2E tests for message content handling
- [x] Integration tests updated for new format
- [x] Unit tests pass (253 tests)
- [x] Integration tests pass (215 tests)
- [x] TypeScript compiles clean

---

### DIFF
--- a/apps/frontend/src/hooks/use-draft-message.ts
+++ b/apps/frontend/src/hooks/use-draft-message.ts
@@ -2,6 +2,7 @@ import { useLiveQuery } from "dexie-react-hooks"
 import { useCallback, useEffect, useRef } from "react"
 import { db, type DraftAttachment } from "@/db"
 import type { JSONContent } from "@threa/types"
+import { isEmptyContent } from "@/lib/prosemirror-utils"
 
 // Key formats:
 // - "stream:{streamId}" for messages in existing streams
@@ -38,13 +39,8 @@ export function useDraftMessage(workspaceId: string, draftKey: string) {
       const currentDraft = await db.draftMessages.get(draftKey)
       const finalAttachments = attachments ?? currentDraft?.attachments ?? []
 
-      // Check if content is empty (only has empty paragraphs)
-      const isEmpty =
-        !contentJson.content ||
-        contentJson.content.every((node) => node.type === "paragraph" && (!node.content || node.content.length === 0))
-
       // Delete draft only if both content and attachments are empty
-      if (isEmpty && finalAttachments.length === 0) {
+      if (isEmptyContent(contentJson) && finalAttachments.length === 0) {
         await db.draftMessages.delete(draftKey)
         return
       }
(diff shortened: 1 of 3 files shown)


apps/frontend/src/hooks/use-draft-message.ts
```diff
@@ -37,16 +38,21 @@ export function useDraftMessage(workspaceId: string, draftKey: string) {
       const currentDraft = await db.draftMessages.get(draftKey)
       const finalAttachments = attachments ?? currentDraft?.attachments ?? []
 
+      // Check if content is empty (only has empty paragraphs)
+      const isEmpty =
```
kristofferremback: I've seen this piece of logic enough. Break this out and reuse it instead of having this massive chain of ands and ors. And of course find all instances of this and use the new function

PR #156, opened 2026-02-17
feat: animate thread panel open with barn door effect

kristofferremback:
## Problem

The thread panel appears and disappears instantly — no transition, just a hard cut. The sidebar has smooth CSS transitions for open/close, but the thread panel (built on `react-resizable-panels`) had no animation support.

## Solution

Replace `react-resizable-panels` with a custom flex layout that supports CSS `transition-[width]` for a barn door slide-in effect.

### How it works

The panel wrapper div is **always in the DOM** with `overflow-hidden`:
- **Closed:** `width: 0` — invisible, zero space
- **Open:** `width: 480px` — CSS transition animates from 0 → 480px
- **Inside:** content rendered at constant full width, clipped by the wrapper — content position doesn't shift during animation, only the visible area expands (barn door)

The main content area uses `flex-1`, so it smoothly shrinks/grows as the panel width changes.

### Key design decisions

**1. Suppress animation on page load**

`enableTransition` state starts `false`, flipped to `true` after `requestAnimationFrame`. On page load with `?panel=` in the URL, the wrapper mounts at full width before transitions are enabled — instant, no animation. Subsequent open/close transitions work normally.

**2. Custom resize instead of react-resizable-panels**

`react-resizable-panels` manages sizes internally with percentages and doesn't support CSS transitions on panel sizes. Replacing it with a manual mouse-drag resize handler (modeled after the sidebar's) gives full control over the width property that drives the animation. The resize handle visually matches the previous `ResizableHandle`.

**3. Single return path**

The previous implementation had two separate return blocks (`if (isPanelOpen) { ... } else { ... }`), meaning the main content remounted when toggling the panel. The new layout always renders both the main content and the panel wrapper as siblings, avoiding unnecessary remounts.

## New files

| File | Purpose |
|------|---------|
| `apps/frontend/src/hooks/use-panel-layout.ts` | Panel width state, resize drag handling, transition enable logic |

## Modified files

| File | Change |
|------|--------|
| `apps/frontend/src/pages/stream.tsx` | Replace `ResizablePanelGroup` with custom flex layout + barn door wrapper |
| `apps/frontend/src/hooks/index.ts` | Export `usePanelLayout` |

## Test plan

- [ ] Open thread panel via "Reply in thread" — panel slides in from left
- [ ] With panel open, click "Reply in thread" on a different message — panel swaps content, no animation
- [ ] With panel open, refresh page — panel is instantly open, no animation
- [ ] Resize panel via drag handle — smooth, no jank
- [ ] Close panel — content disappears, main area expands smoothly
- [ ] Conversation side panel (channels) still works independently

---

### DIFF
--- a/apps/frontend/src/components/layout/panel-resize-handle.tsx
+++ b/apps/frontend/src/components/layout/panel-resize-handle.tsx
@@ -4,6 +4,7 @@ import { cn } from "@/lib/utils"
 interface PanelResizeHandleProps {
   isResizing: boolean
   panelWidth: number
+  minWidth: number
   maxWidth: number
   onMouseDown: (e: React.MouseEvent) => void
   onKeyDown: (e: React.KeyboardEvent) => void
@@ -12,6 +13,7 @@ interface PanelResizeHandleProps {
 export function PanelResizeHandle({
   isResizing,
   panelWidth,
+  minWidth,
   maxWidth,
   onMouseDown,
   onKeyDown,
@@ -31,7 +33,7 @@ export function PanelResizeHandle({
       role="separator"
       aria-orientation="vertical"
       aria-valuenow={panelWidth}
-      aria-valuemin={300}
+      aria-valuemin={minWidth}
       aria-valuemax={maxWidth}
       aria-label="Resize thread panel"
     >
--- a/apps/frontend/src/components/layout/thread-panel-slot.tsx
+++ b/apps/frontend/src/components/layout/thread-panel-slot.tsx
@@ -7,6 +7,7 @@ interface ThreadPanelSlotProps {
   shouldAnimate: boolean
   showContent: boolean
   isResizing: boolean
+  minWidth: number
   maxWidth: number
   onTransitionEnd: (e: React.TransitionEvent) => void
   onResizeStart: (e: React.MouseEvent) => void
@@ -20,6 +21,7 @@ export function ThreadPanelSlot({
   shouldAnimate,
   showContent,
   isResizing,
+  minWidth,
   maxWidth,
   onTransitionEnd,
   onResizeStart,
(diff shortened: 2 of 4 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: claude-opus-4-6 | Reviewers: sonnet x3

Found 2 issues:

1. `use-panel-layout.ts:23-30` — **Bug:** `showContent` never becomes `false` when the panel is closed while `isResizing` is true. In that state, `shouldAnimate` is `false` (no CSS transition class applied), so the width snaps to 0 instantly without firing `transitionend`. The effect at lines 23-30 only unmounts content when `enableTransition` is `false` (page-load edge case), leaving no code path to set `showContent = false`. The `StreamPanel` stays mounted indefinitely with stale subscriptions.
   https://github.com/kristofferremback/threa/blob/d28623c12dbdddc534862fae2fef2c0576a62e86/apps/frontend/src/hooks/use-panel-layout.ts#L22-L31

2. `stream.tsx:264-280` — **Design (INV-37):** The thread panel resize handle is missing keyboard interaction and ARIA attributes that the sidebar resize handle (`app-shell.tsx:251-279`) has: no `tabIndex`, no `onKeyDown` arrow-key handler, no `aria-label`, no `aria-valuenow/min/max`. The PR extracted `useResizeDrag` as a shared hook but left the handle element duplicated with divergent accessibility contracts.
   https://github.com/kristofferremback/threa/blob/d28623c12dbdddc534862fae2fef2c0576a62e86/apps/frontend/src/pages/stream.tsx#L264-L280

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>All changes match the PR description. No missing corresponding changes.</details>

- `showContent` leak when panel closed during active resize — content stays mounted forever because no `transitionend` fires when the CSS transition class is absent.

</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No violations found.</details>

- INV-37: Two resize handle implementations now diverge — sidebar has keyboard/ARIA support, thread panel does not. Either extract a shared `ResizeHandle` component or add the missing attributes.

</details>
<details><summary>🔒 Security [CLEAN]</summary>No issues found.</details>

PR #188, opened 2026-03-01
fix: gate auto-read on tab visibility and focus

kristofferremback:
## Problem

`useAutoMarkAsRead` only checked stream activity and unread state, so an open stream in a background tab or an unfocused browser window could still emit read events. That clears unread state in the backend even though the user is not actively viewing the tab.

## Solution

Track whether the page is both visible and focused inside the auto-read hook, and only schedule `markAsRead` when that condition is true. When the tab or window becomes inactive, the effect cleanup cancels any pending debounce so no delayed read event is sent while the page is hidden or unfocused.

### How it works

- `usePageActive` derives active state from `document.visibilityState === "visible"` and `document.hasFocus()`.
- The hook listens to `focus`, `blur`, and `visibilitychange` so it reacts to both tab switches and window focus changes.
- The existing debounced auto-read effect now depends on page activity, which prevents sending read events while inactive and resumes cleanly when focus returns.
- Regression tests cover the active case, the hidden/unfocused case, and a pending debounce interrupted by blur/visibility changes.

### Key design decisions

**1. Require both tab visibility and window focus**

Using only one signal is not enough here. A visible tab in an unfocused window should not clear unread state, and a focused window with a hidden tab should not either, so the hook now requires both conditions before it sends a read event.

**2. Cancel pending debounced reads instead of only blocking new ones**

If the user switches away during the debounce window, the earlier timer must not fire later. Keeping the active state in the effect dependencies lets the existing cleanup clear pending timers without adding a second timer system.

## New files

| File | Purpose |
| --- | --- |
| `apps/frontend/src/hooks/use-auto-mark-as-read.test.ts` | Regression coverage for visible/focused gating and debounce cancellation. |

## Modified files

| File | Change |
| --- | --- |
| `apps/frontend/src/hooks/use-auto-mark-as-read.ts` | Gate automatic read events on tab visibility and window focus, and pause/resume the debounce accordingly. |

## Deleted files

None.

## Test plan

- [x] `bun scripts/test-silent.ts frontend src/hooks/use-auto-mark-as-read.test.ts`
- [x] `bun run typecheck` (in `apps/frontend`)
- [ ] Manual verification in the browser

---

## Summary by CodeRabbit

* **New Features**
  * Enhanced page activity tracking (visibility/focus/active) for more reliable connection heartbeats and auto-read behavior.
  * Heartbeat behavior improved to emit immediately on reconnect and respond to page activity transitions.

* **Tests**
  * Added comprehensive tests for auto-mark-as-read covering visibility, focus, suppression, and recovery scenarios.

### DIFF
--- a/apps/frontend/src/contexts/socket-context.tsx
+++ b/apps/frontend/src/contexts/socket-context.tsx
@@ -164,13 +164,21 @@ export function SocketProvider({ workspaceId, children }: SocketProviderProps) {
     const previousPageActivity = previousPageActivityRef.current
     const gainedFocus = !previousPageActivity.isFocused && pageActivity.isFocused
     const becameVisible = !previousPageActivity.isVisible && pageActivity.isVisible
+
+    if (!socket || status !== "connected") {
+      previousPageActivityRef.current = pageActivity
+      return
+    }
+
     previousPageActivityRef.current = pageActivity
 
-    if (!socket || status !== "connected") return
     if (!gainedFocus && !becameVisible) return
 
+    // Focus gains always emit — bypassing the throttle prevents a becameVisible
+    // heartbeat (focused=false) from blocking the closely-following gainedFocus
+    // heartbeat (focused=true) via the 10s throttle window.
     const now = Date.now()
-    if (now - lastInteractionHeartbeatRef.current > 10_000) {
+    if (gainedFocus || now - lastInteractionHeartbeatRef.current > 10_000) {
       lastInteractionHeartbeatRef.current = now
       socket.emit("heartbeat", { focused: pageActivity.isFocused })
     }
--- a/apps/frontend/src/hooks/use-page-activity.ts
+++ b/apps/frontend/src/hooks/use-page-activity.ts
@@ -26,27 +26,29 @@ export function getPageActivityState(): PageActivityState {
 }
 
 export function usePageActivity(): PageActivityState {
-  const [pageActivity, setPageActivity] = useState(getPageActivityState)
+  const initial = getPageActivityState()
+  const [isVisible, setIsVisible] = useState(initial.isVisible)
+  const [isFocused, setIsFocused] = useState(initial.isFocused)
 
   useEffect(() => {
     if (typeof window === "undefined" || typeof document === "undefined") return
 
-    const updatePageActivity = () => {
-      setPageActivity(getPageActivityState())
-    }
+    const updateVisibility = () => setIsVisible(document.visibilityState === "visible")
+    const updateFocus = () => setIsFocused(document.hasFocus())
 
-    updatePageActivity()
+    updateVisibility()
+    updateFocus()
 
-    window.addEventListener("focus", updatePageActivity)
-    window.addEventListener("blur", updatePageActivity)
-    document.addEventListener("visibilitychange", updatePageActivity)
+    document.addEventListener("visibilitychange", updateVisibility)
+    window.addEventListener("focus", updateFocus)
+    window.addEventListener("blur", updateFocus)
 
     return () => {
-      window.removeEventListener("focus", updatePageActivity)
-      window.removeEventListener("blur", updatePageActivity)
-      document.removeEventListener("visibilitychange", updatePageActivity)
+      document.removeEventListener("visibilitychange", updateVisibility)
+      window.removeEventListener("focus", updateFocus)
+      window.removeEventListener("blur", updateFocus)
     }
   }, [])
 
-  return pageActivity
+  return { isVisible, isFocused, isActive: isVisible && isFocused }
 }


kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: claude-opus-4-6 | Reviewers: sonnet x3

Found 3 issues:

1. `use-auto-mark-as-read.test.ts:144` — INV-23 violation (Testing, "Do not assert event counts; assert presence/content of specific events"). `toHaveBeenCalledTimes(1)` is a count assertion; the content assertion on line 145 (`toHaveBeenCalledWith(...)`) already verifies presence and correctness, making the count check redundant and against INV-23.
   https://github.com/kristofferremback/threa/blob/0ed35a0c16be56efd82dacc68840a108bcb5765c/apps/frontend/src/hooks/use-auto-mark-as-read.test.ts#L144

2. `use-auto-mark-as-read.test.ts:93-103` — Missing test for partial inactive case. The "hidden and unfocused" test sets both `visibilityState = "hidden"` and `hasFocus = false`, but the production guard is `visibilityState === "visible" && document.hasFocus()`. The case where `visibilityState` is `"visible"` but `hasFocus()` is `false` (user switches window without hiding tab) should also suppress reads but is not covered.
   https://github.com/kristofferremback/threa/blob/0ed35a0c16be56efd82dacc68840a108bcb5765c/apps/frontend/src/hooks/use-auto-mark-as-read.test.ts#L93-L103

3. `use-auto-mark-as-read.ts:16-40` — `usePageActive` implements the same `focus`/`blur`/`visibilitychange` event listener pattern already present in `socket-context.tsx`. INV-37 requires reusing existing abstractions instead of creating parallel implementations. Consider extracting to a shared hook.
   https://github.com/kristofferremback/threa/blob/0ed35a0c16be56efd82dacc68840a108bcb5765c/apps/frontend/src/hooks/use-auto-mark-as-read.ts#L16-L40

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>No plan found. PR description aligns with implementation.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No logic errors, race conditions, or null hazards found. The AND gate logic is correct, debounce cleanup works properly, and state transitions are sound.</details>
<details><summary>📋 CLAUDE.md Compliance [2 issues]</summary>

1. **INV-23** — Count assertion `toHaveBeenCalledTimes(1)` in test (line 144)
2. **INV-37** — Parallel implementation of focus/visibility listener pattern (lines 16-40)

</details>

Missing test coverage for the partial inactive case (visible but unfocused). This is an important edge case given the PR explicitly requires both conditions.

</details>
<details><summary>🔒 Security [CLEAN]</summary>No security issues found.</details>

kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: claude-opus-4-6 | Reviewers: sonnet x3

Found 3 issues:

1. `socket-context.tsx:163-177` — `previousPageActivityRef.current` is updated (line 167) before the `!socket || status !== "connected"` guard (line 169). When focus is gained while disconnected, the transition is silently consumed: when the socket later connects and the effect re-runs with the same `pageActivity`, `gainedFocus` computes to `false` and the throttled focus heartbeat is skipped. The first-connect `emitHeartbeat()` in the other effect still fires, so this is not catastrophic, but the ref update should move to after both guards for correct intent.
   https://github.com/kristofferremback/threa/blob/2378840c11634cd83dabef6019577b90af161278/apps/frontend/src/contexts/socket-context.tsx#L163-L177

2. `use-page-activity.ts:31-49` — `usePageActivity` is instantiated independently in both `SocketProvider` and `useAutoMarkAsRead`, attaching two separate sets of `focus`/`blur`/`visibilitychange` listeners and holding independent state. The two instances can transiently hold different states across render batches. A shared `PageActivityContext` above `SocketProvider` or exposing `isActive` from socket context would deduplicate both listener overhead and state (INV-37: "Reuse existing helpers and abstractions instead of parallel implementations").
   https://github.com/kristofferremback/threa/blob/2378840c11634cd83dabef6019577b90af161278/apps/frontend/src/hooks/use-page-activity.ts#L31-L49

3. `use-auto-mark-as-read.test.ts:11-22` — INV-48 violation: "Avoid `mock.module()` for shared modules; prefer scoped `spyOn` patterns." Uses `vi.mock("./use-unread-counts", ...)` and `vi.mock("./use-activity-counts", ...)` to replace two shared hooks with module-level mocks. The correct approach would be wrapping the hook render with a real `QueryClient` provider seeded with test data, or using `vi.spyOn` on the module exports.
   https://github.com/kristofferremback/threa/blob/2378840c11634cd83dabef6019577b90af161278/apps/frontend/src/hooks/use-auto-mark-as-read.test.ts#L11-L22

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>No plan found. PR description aligns with implementation. The new shared use-page-activity.ts hook and socket-context.tsx refactor are consistent with the stated goal.</details>

1. **Ref-before-guard in heartbeat effect** — `previousPageActivityRef` updated before the socket/connected guard silently consumes focus transitions that occur while disconnected. Not catastrophic since the first-connect heartbeat still fires, but the ordering is misleading.

</details>
<details><summary>📋 CLAUDE.md Compliance [2 violations]</summary>

1. **INV-37** — Duplicate `usePageActivity` instantiation in socket context and auto-read hook creates parallel implementations with separate state and listeners.
2. **INV-48** — Test uses `vi.mock()` for shared modules (`use-unread-counts`, `use-activity-counts`) instead of scoped `spyOn` patterns.

</details>

Dual `usePageActivity` instances create duplicate event listener sets and can transiently diverge across render batches. A single context-provided instance would be more robust.

</details>
<details><summary>🔒 Security [CLEAN]</summary>No security issues found.</details>

PR #237, opened 2026-03-18
feat: infinite scroll for messages with bidirectional pagination

kristofferremback:
## Summary

- **Backward pagination**: Added `beforeSequence` to the event repository and `before` query param to the events API, enabling proper infinite scroll for older messages
- **Jump-to-message from search**: New `events/around` endpoint loads a window of events around any message (accepts both event IDs and message IDs). Frontend enters bidirectional pagination mode with a "Jump to latest" button to return to live tail
- **Thread panel auto-loading**: Threads automatically fetch all older messages on mount since they're typically small
- **Pull-to-refresh suppression**: Streams suppress pull-to-refresh via a data attribute, preventing conflicts with upward scrolling
- **Scroll position preservation**: When older messages are prepended, scroll position is maintained so the user doesn't lose their place
- **Configurable fetch trigger**: Loads next page after scrolling through half the page size (25 of 50 messages), easily adjustable via constants

### Bug fix
The existing `getEvents` frontend API was mapping `before` → `after` query param, making backward pagination non-functional. This is now fixed with proper `before`/`after` param handling.

## Design decisions

- **No virtual scroll** — preserves native browser search (Cmd+F)
- **Cursor-based pagination** using sequence numbers (not offset) for consistency with existing patterns
- **`hasOlderEvents` in bootstrap** derived from `events.length === 50` — avoids an extra COUNT query
- **Message-count-based scroll trigger** instead of pixel threshold — more predictable across different message sizes
- **`data-suppress-pull-refresh` attribute** — lightweight coordination between stream containers and pull-to-refresh without React context overhead

## File changes

| Area | Files | What changed |
|------|-------|-------------|
| Backend repo | `event-repository.ts` | `beforeSequence` filter, `listAround()`, `findByMessageId()` |
| Backend service | `event-service.ts` | `listEventsAround()` with event/message ID fallback |
| Backend handlers | `handlers.ts`, `routes.ts` | `before` param on events, `events/around` endpoint, `hasOlderEvents` in bootstrap |
| Shared types | `api.ts`, `index.ts` | `StreamBootstrap.hasOlderEvents`, `EventsAroundResponse` |
| Frontend API | `streams.ts`, `services-context.tsx` | Fixed `getEvents`, added `getEventsAround` |
| Frontend hooks | `use-events.ts` | Bidirectional pagination, jump-to mode, `loadAll` for threads |
| Frontend scroll | `use-scroll-behavior.ts` | Item-count triggers, bidirectional, scroll preservation |
| Frontend UI | `stream-content.tsx` | Wired up bidirectional scroll, jump-to-message, "Jump to latest" button |
| Frontend misc | `use-pull-to-refresh.ts`, `constants.ts` | Suppress in streams, shared page size constant |

## Test plan

- [ ] Open a stream with >50 messages, scroll up — older messages load seamlessly
- [ ] Verify scroll position doesn't jump when older messages prepend
- [ ] Search for an old message, click result — jumps to message with highlight, can scroll both directions
- [ ] Click "Jump to latest" button — returns to live tail with latest messages
- [ ] Open a thread with >50 messages — all messages auto-load
- [ ] On mobile, pull-to-refresh works on sidebar/non-stream views but NOT inside streams
- [ ] New messages via socket still appear at bottom and auto-scroll works

https://claude.ai/code/session_01J5cZVz8WD1prAxSfmPuFrS

### DIFF
--- a/tests/browser/infinite-scroll.spec.ts
+++ b/tests/browser/infinite-scroll.spec.ts
@@ -27,7 +27,7 @@ async function seedMessages(
   const BATCH_SIZE = 5
   for (let start = 1; start <= count; start += BATCH_SIZE) {
     const end = Math.min(start + BATCH_SIZE - 1, count)
-    const promises = []
+    const promises: Promise<void>[] = []
     for (let i = start; i <= end; i++) {
       promises.push(
         page.request
@@ -160,7 +160,7 @@ test.describe("Infinite Scroll", () => {
   })
 
   test("should not make pagination requests when all messages fit in bootstrap", async ({ page }) => {
-    const channelName = `scroll-nopage-${testId}`
+    const channelName = `scroll-no-page-${testId}`
     await createChannel(page, channelName)
 
     const { workspaceId, streamId } = extractIds(page)


kristofferremback:
Both issues fixed in 2ac104c:

**1. `hasNewer` off-by-one in `listAround`**
Now fetches `half + 1` newer events and checks `> half`, mirroring the `hasOlder` probe pattern. The extra probe event is trimmed from results via `newerEvents.slice(0, half)`.

**2. `loadAll` infinite loop**
`hasOlderEvents` now checks whether the infinite query has produced any pages (`olderData?.pages.length > 0`). Once a query has run, the bootstrap hint is ignored — only `hasOlderPage` (from `getNextPageParam`) determines if more pages exist. This prevents the re-seed/re-fetch loop for threads with 51–99 events.

PR #342, opened 2026-04-12
feat(memos): upgrade models, add confidence floor, resolve author names

kristofferremback:
## Problem

Memo generation (the GAM knowledge extraction pipeline) uses `openrouter:openai/gpt-oss-120b` for both classification and memorization. This model has two compounding problems:

1. **Low quality** — as a ~5B active-param MoE, it struggles with the nuanced tasks the memo pipeline demands: distinguishing real knowledge from noise (classifier), and writing self-contained abstracts with pronoun resolution and date anchoring (memorizer). The result is high false-positive gem rates and poorly-written memos.
2. **Unreliable latency** — OpenRouter's patchwork provider backing for this model causes 60–120s tail latency on some calls (same issue that motivated the researcher model swap in #333).

Additionally, message memos see no author identity — the classifier prompt shows `From: user (3fa8b2c1)` (a ULID suffix) and the memorizer prompt shows just `From: user` with no identifying information at all. This degrades pronoun resolution and makes memos less self-contained.

## Solution

Three targeted improvements to the memo pipeline, all in the same layer:

### 1. Model split — purpose-matched models for classifier vs memorizer

The single `MEMO_MODEL_ID` is replaced with two constants:

| Component | Old model | New model | Rationale |
|-----------|-----------|-----------|-----------|
| **Classifier** | `gpt-oss-120b` ($0.04/$0.19) | `claude-haiku-4.5` ($0.25/$1.25) | Binary+enum structured output. Haiku 4.5 is already recommended for "memo classification" in `model-reference.md` and proven on the researcher pathway in #333. |
| **Memorizer** | `gpt-oss-120b` ($0.04/$0.19) | `gpt-5-mini` ($0.40/$1.60) | Abstractive writing, pronoun resolution, date anchoring. Needs higher quality than classification but doesn't require Sonnet-tier cost. |

The infrastructure already had separate `COMPONENT_PATHS.MEMO_CLASSIFIER` and `COMPONENT_PATHS.MEMO_MEMORIZER` — they just both pointed to the same model. This change makes them independent.

### 2. Confidence floor — skip low-confidence gems

The classifier schema already returns a `confidence: 0.0–1.0` score, but the service layer ignored it. Now gems with `confidence < 0.7` are skipped with an `info`-level log for observability. Applied to both message and conversation classification paths.

This directly reduces false-positive memo volume — the biggest source of "garbage memos" — and partially offsets the model cost increase by running the memorizer less often.

### 3. Author name resolution for message memos

The service's Phase 1 data fetch already calls `UserRepository.findByIds()` to get author timezones. The same result set now also extracts `member.name`, which is passed to the classifier (`From: user (Alice)` instead of `From: user (3fa8b2c1)`) and memorizer (`From: user (Alice)` instead of `From: user`).

Zero additional database queries — just reading an extra field from the existing result.

### Key design decisions

**1. Split constants rather than env var overrides**

The existing `AI_MEMO_MODEL` env var in `env.ts` was dead code (zero consumers). Rather than adding two more env vars, the config.ts constants are the source of truth (INV-44), and the eval CLI supports `--model` overrides for experimentation. Removed the dead env var as cleanup.

**2. `classification.confidence != null` guard**

The classifier schema defines confidence as `z.number().min(0).max(1).nullable()`. The null check ensures we don't accidentally skip memos when the model omits the confidence field — only skip when we have an explicit low-confidence score.

**3. Author name fallback behavior**

- Classifier: falls back to ULID suffix (`authorId.slice(-8)`) when name unavailable — preserves existing behavior.
- Memorizer: falls back to `"Unknown"` — the prompt template now uses `From: {{AUTHOR_TYPE}} ({{AUTHOR_NAME}})` and "Unknown" is better than a meaningless ID for the LLM's pronoun resolution.

**4. Eval suites pass synthetic `authorName: "Alex"`**

So eval prompts contain realistic data instead of exercising the fallback path. This makes eval results representative of production behavior with the new prompt template.

## Modified files

| File | Change |
| --- | --- |
| `apps/backend/src/features/memos/config.ts` | Split `MEMO_MODEL_ID` → `MEMO_CLASSIFIER_MODEL_ID` + `MEMO_MEMORIZER_MODEL_ID`. Add `MEMO_GEM_CONFIDENCE_FLOOR`. Update `MEMORIZER_MESSAGE_PROMPT` with `{{AUTHOR_NAME}}` placeholder. |
| `apps/backend/src/features/memos/index.ts` | Update barrel exports for new constants |
| `apps/backend/src/features/memos/classifier.ts` | Add `authorName?: string` to `ClassifierContext`, use resolved name in prompt |
| `apps/backend/src/features/memos/memorizer.ts` | Add `authorName?: string` to `MemorizerContext`, use in message memo prompt |
| `apps/backend/src/features/memos/service.ts` | Phase 1: collect author names alongside timezones. Phase 2: pass to classifier/memorizer, add confidence floor checks on both message and conversation paths. |
| `apps/backend/src/lib/ai/static-config-resolver.ts` | Use split model constants for memo component configs |
| `apps/backend/src/lib/env.ts` | Remove dead `memoModel` / `AI_MEMO_MODEL` (zero consumers) |
| `apps/backend/src/lib/ai/config-resolver.test.ts` | Update expected model IDs in assertions |
| `apps/backend/evals/suites/memo-classifier/suite.ts` | Use `MEMO_CLASSIFIER_MODEL_ID`, pass `authorName: "Alex"` |
| `apps/backend/evals/suites/memorizer/suite.ts` | Use `MEMO_MEMORIZER_MODEL_ID`, pass `authorName: "Alex"` |

## New files

| File | Purpose |
| --- | --- |
| `.claude/plans/nifty-dazzling-cherny.md` | Implementation plan |

## Test plan

- [x] `bun run typecheck` — all 9 projects pass
- [x] `bun run --cwd apps/backend test:unit` — 788 tests pass, 0 failures
- [x] `bun run --cwd apps/frontend test` — 1081 tests pass, 0 failures
- [x] Pre-commit hooks pass (lint, typecheck, OpenAPI check)
- [x] Zero remaining references to `MEMO_MODEL_ID` or `gpt-oss-120b` in production code (verified via grep)
- [ ] Run eval comparison to validate model picks before deploying:
  ```bash
  cd apps/backend
  bun run evals/run.ts -s memo-classifier -m "openrouter:openai/gpt-oss-120b,openrouter:anthropic/claude-haiku-4.5" -p 2
  bun run evals/run.ts -s memorizer -m "openrouter:openai/gpt-oss-120b,openrouter:openai/gpt-5-mini" -p 2
  ```

## Follow-ups (not in this PR)

- **Provenance tracking** — add `generated_with_model` / `generated_with_prompt_version` columns to memos table so we can identify which pipeline version generated each memo
- **Backfill script** — re-run memo generation on historical data generated with gpt-oss-120b
- **Kill single-message memo path** — always defer to conversation path to fix Q-without-A memos
- **Completeness gating** — skip memo generation until conversation boundary extraction marks the conversation as sufficiently complete
- **Self-verification** — two-pass pattern where a critic model reviews draft memos before persistence

---

https://claude.ai/code/session_01Am4QjWr2yvFyGphAQBQePw

### DIFF
--- a/.claude/plans/nifty-dazzling-cherny.md
+++ b/.claude/plans/nifty-dazzling-cherny.md
@@ -1,0 +1,129 @@
+# Improve Memo Generation Quality (Phase 1)
+
+## Context
+
+Threa's GAM (General Agentic Memory) extracts knowledge from conversations into memos. The memo pipeline currently uses `openrouter:openai/gpt-oss-120b` -- a cheap, low-quality model that produces garbage memos in production. PR #333 (now merged) already migrated the *researcher* pathway from the same model to `claude-haiku-4.5` with documented quality + latency wins. That PR's plan file explicitly deferred memorizer quality as "Option H -- separate PR; user wants to focus on it properly." This is that PR.
+
+Three improvements, shipped together:
+
+1. **Model split** -- Haiku 4.5 for classifier, GPT-5 Mini for memorizer
+2. **Confidence floor** -- skip memos when classifier confidence < 0.7
+3. **Author name resolution** -- message memos currently see "user (3fa8b2c1)" instead of "user (Alice)"
+
+## Changes
+
+### 1. Model split
+
+**`apps/backend/src/features/memos/config.ts`** (line 20)
+- Replace `MEMO_MODEL_ID` with two constants:
+  ```
+  MEMO_CLASSIFIER_MODEL_ID = "openrouter:anthropic/claude-haiku-4.5"
+  MEMO_MEMORIZER_MODEL_ID  = "openrouter:openai/gpt-5-mini"
+  ```
+- Remove stale comment about `AI_MEMO_MODEL` env var override
+
+**`apps/backend/src/features/memos/index.ts`** (line 25)
+- Replace `MEMO_MODEL_ID` export with `MEMO_CLASSIFIER_MODEL_ID` and `MEMO_MEMORIZER_MODEL_ID`
+
+**`apps/backend/src/lib/ai/static-config-resolver.ts`** (lines 19, 52-59)
+- Update import: `MEMO_CLASSIFIER_MODEL_ID, MEMO_MEMORIZER_MODEL_ID, MEMO_TEMPERATURES`
+- Use `MEMO_CLASSIFIER_MODEL_ID` for `COMPONENT_PATHS.MEMO_CLASSIFIER`
+- Use `MEMO_MEMORIZER_MODEL_ID` for `COMPONENT_PATHS.MEMO_MEMORIZER`
+
+**`apps/backend/evals/suites/memo-classifier/suite.ts`** (lines 11, ~137)
+- Import `MEMO_CLASSIFIER_MODEL_ID` instead of `MEMO_MODEL_ID`
+- Use in `defaultPermutations`
+
+**`apps/backend/evals/suites/memorizer/suite.ts`** (lines 19, ~224)
+- Import `MEMO_MEMORIZER_MODEL_ID` instead of `MEMO_MODEL_ID`
+- Use in `defaultPermutations`
+
+**`apps/backend/src/lib/env.ts`** (lines 14-15, 137)
+- Remove dead `memoModel` property from `AIConfig` interface and `createEnv()` -- confirmed zero consumers via grep
+
+### 2. Confidence floor
+
+**`apps/backend/src/features/memos/config.ts`**
+- Add `MEMO_GEM_CONFIDENCE_FLOOR = 0.7`
+
+**`apps/backend/src/features/memos/service.ts`** (after line 217, after line 314)
+- After classifier returns `isGem: true`, check `confidence < MEMO_GEM_CONFIDENCE_FLOOR`
+- If below threshold: `logger.info(...)` with messageId/confidence/threshold, then `continue`
+- Apply same check for conversation classification path
+
+**`apps/backend/src/features/memos/index.ts`**
+- Export `MEMO_GEM_CONFIDENCE_FLOOR`
+
+### 3. Author name resolution for message memos
+
+**Problem**: `memorizeMessage()` receives a raw `Message` and shows `From: user` with no name. `classifyMessage()` shows `From: user (3fa8b2c1)` -- last 8 chars of a ULID. Conversation memos already get proper names via `MessageFormatter.formatMessages()`.
+
+**`apps/backend/src/features/memos/service.ts`** -- Phase 1 data fetch (lines 155-173)
+- The code already calls `UserRepository.findByIds(client, workspaceId, ...)` to get timezones
+- Also extract `member.name` into a new `authorNames: Map<string, string>` alongside `authorTimezones`
+- Zero additional DB queries -- reuse the existing result set
+- Add `authorNames` to `fetchedData` return object
+
+**`apps/backend/src/features/memos/service.ts`** -- Phase 2 message processing (lines 203-268)
+- Resolve: `const authorName = fetchedData.authorNames.get(message.authorId) ?? undefined`
+- Pass to classifier: `this.classifier.classifyMessage(message, { workspaceId, authorName })`
+- Pass to memorizer: `this.memorizer.memorizeMessage({ ..., authorName })`
+
+**`apps/backend/src/features/memos/classifier.ts`** (lines 22-24, 55-83)
+- Add `authorName?: string` to `ClassifierContext` interface
+- In `classifyMessage()`: use `context.authorName ?? message.authorId.slice(-8)` as the author label in the prompt. No template change needed -- `{{AUTHOR_ID}}` placeholder now receives a name instead of a ULID suffix.
+
+**`apps/backend/src/features/memos/memorizer.ts`** (lines 32-41, 50-91)
+- Add `authorName?: string` to `MemorizerContext` interface
+- In `memorizeMessage()`: use `context.authorName ?? "Unknown"` as author label
+
+**`apps/backend/src/features/memos/config.ts`** -- prompt template (line 195)
+- Change `MEMORIZER_MESSAGE_PROMPT` line from:
+  `From: {{AUTHOR_TYPE}}`
+  to:
+  `From: {{AUTHOR_TYPE}} ({{AUTHOR_NAME}})`
+- Add `{{AUTHOR_NAME}}` replacement in `memorizer.ts` line 65
+
+## Implementation order
+
+1. `config.ts` -- model constants, confidence floor constant, prompt template update
+2. `index.ts` -- update exports
+3. `static-config-resolver.ts` -- use split model constants
+4. `classifier.ts` -- accept `authorName` in context
+5. `memorizer.ts` -- accept `authorName` in context, use in prompt
+6. `service.ts` -- collect author names in Phase 1, pass to classifier/memorizer in Phase 2, add confidence floor checks
+7. `env.ts` -- remove dead `memoModel`
+8. Eval suites -- update imports to per-component model IDs
+
+## Verification
+
+**Typecheck**: `bun run typecheck` must pass
+
+**Unit tests**: `bun run --cwd apps/backend test` -- existing tests must pass
+
+**Eval comparison** (before merging, to validate model picks):
+```bash
+cd apps/backend
+
+# Classifier: compare old vs new
+bun run evals/run.ts -s memo-classifier \
+  -m "openrouter:openai/gpt-oss-120b,openrouter:anthropic/claude-haiku-4.5" -p 2
+
+# Memorizer: compare old vs new
+bun run evals/run.ts -s memorizer \
+  -m "openrouter:openai/gpt-oss-120b,openrouter:openai/gpt-5-mini" -p 2
+
+# Run with new defaults (after code change)
+bun run evals/run.ts -s memo-classifier
+bun run evals/run.ts -s memorizer
+```
+
+**Eval suite updates**: update eval task functions to pass a synthetic `authorName: "Alex"` so prompts contain realistic data instead of exercising the fallback path.
+
+## Not in this PR
+
+- Kill the single-message memo path (Approach 2) -- architectural, separate PR
+- Completeness gating / context windowing (Approach 3) -- depends on boundary extraction changes
+- Self-verification / critic pass (Approach 5) -- deferred until we measure quality after model upgrade
+- Provenance tracking migration (`generated_with_model`, `generated_with_prompt_version`) -- separate PR, enables backfill
+- Backfill script for historical memos -- depends on provenance tracking
--- a/apps/backend/evals/suites/memo-classifier/suite.ts
+++ b/apps/backend/evals/suites/memo-classifier/suite.ts
@@ -8,7 +8,7 @@
 import type { EvalSuite, EvalContext, RunEvaluator, CaseResult } from "../../framework/types"
 import { binaryClassificationEvaluator, categoricalEvaluator } from "../../framework/evaluators/classification"
 import { MemoClassifier, type MessageClassification } from "../../../src/features/memos"
-import { MEMO_MODEL_ID, MEMO_TEMPERATURES } from "../../../src/features/memos"
+import { MEMO_CLASSIFIER_MODEL_ID, MEMO_TEMPERATURES } from "../../../src/features/memos"
 import { MessageFormatter } from "../../../src/lib/ai/message-formatter"
 import { classifierCases, createTestMessage, type ClassifierInput, type ClassifierExpected } from "./cases"
 import { messageId } from "../../../src/lib/id"
@@ -23,8 +23,8 @@ async function classifyMessage(input: ClassifierInput, ctx: EvalContext): Promis
   // Create a test message from the input
   const message = createTestMessage(input, messageId(), ctx.userId)
 
-  // Classify the message
-  return classifier.classifyMessage(message, { workspaceId: ctx.workspaceId })
+  // Classify the message (pass synthetic author name for realistic prompt content)
+  return classifier.classifyMessage(message, { workspaceId: ctx.workspaceId, authorName: "Alex" })
 }
 
 /**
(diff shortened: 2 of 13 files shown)


kristofferremback:
## Follow-up 1: Provenance tracking migration

**Goal:** Track which model + prompt version generated each memo so we can query "all memos generated with the old pipeline" for selective backfill.

### Schema change

New migration file `apps/backend/src/db/migrations/YYYYMMDDHHMMSS_memo_provenance.sql`:

```sql
ALTER TABLE memos ADD COLUMN generated_with_model TEXT;
ALTER TABLE memos ADD COLUMN generated_with_prompt_version TEXT;

CREATE INDEX idx_memos_prompt_version ON memos(workspace_id, generated_with_prompt_version)
  WHERE status = 'active';
```

### Code changes

1. **`apps/backend/src/features/memos/config.ts`** — Add `MEMO_PROMPT_VERSION = "v2-2026-04"`. Bump this whenever prompts or models change materially.

2. **`apps/backend/src/features/memos/repository.ts`** — Add `generatedWithModel` and `generatedWithPromptVersion` to `InsertMemoParams` and the `insert()` method. These map to the new columns via the existing snake_case ↔ camelCase convention.

3. **`apps/backend/src/features/memos/service.ts`** — When building `MemoToCreate` objects (lines ~239-253 for message memos, ~340-355 and ~404-418 for conversation memos), populate:
   ```ts
   generatedWithModel: config.modelId,  // from configResolver
   generatedWithPromptVersion: MEMO_PROMPT_VERSION,
   ```
   This requires threading the resolved model ID from the classifier/memorizer back to the service. One option: have the classifier/memorizer return the model ID they used alongside the result. Another: resolve the config in the service and pass it down.

4. **Backfill existing rows:** Run a one-time SQL update:
   ```sql
   UPDATE memos SET generated_with_model = 'openrouter:openai/gpt-oss-120b', generated_with_prompt_version = 'v1-legacy' WHERE generated_with_model IS NULL;
   ```

### Dependencies
- None — can be done independently.

### Verification
- `bun run --cwd apps/backend test:unit` must pass
- Manually insert a memo via the pipeline and verify the new columns are populated in the DB

---

kristofferremback:
## Follow-up 2: Backfill script for historical memos

**Goal:** Re-run memo generation on memos created with the old `gpt-oss-120b` pipeline to improve their quality, without rerunning everything.

### Pattern to follow

Copy the structure from `apps/backend/scripts/backfill-thread-names.ts` — it queries rows needing work, queues jobs, and lets the existing worker process them.

### New file: `apps/backend/scripts/backfill-memos.ts`

**CLI interface:**
```bash
bun run scripts/backfill-memos.ts \
  --workspace=ws_xxx \              # optional: scope to one workspace
  --from-version=v1-legacy \        # required: which prompt version to replace
  --to-version=v2-2026-04 \         # informational, logged
  --knowledge-type=decision \       # optional: filter by type
  --limit=100 \                     # optional: cap batch size
  --dry-run                         # required first run: show what would change
```

**What it does:**

1. Query `memos WHERE generated_with_prompt_version = $fromVersion AND status = 'active'` (optionally filtered by workspace, knowledge_type, date range)
2. For each memo, find its source — either `source_message_id` or `source_conversation_id`
3. Re-queue that **source** into `memo_pending_items` table (not the memo itself). The existing uniqueness constraint `(workspace_id, item_type, item_id)` handles deduplication.
4. Reset the relevant `memo_stream_state` row's `last_processed_at` so the batch worker picks it up on next tick
5. When the new memo gets written, the existing revision flow in `service.ts:318-382` (`MemoRepository.supersede`) marks the old memo as `superseded`. **Never delete old memos** — supersede them.

**Rate limiting:** Enqueue in chunks of ~50, sleep between chunks, so the batch worker (polls every 30s) isn't overwhelmed.

**Dry-run mode is mandatory first.** Print count of memos that would be touched, broken down by `knowledge_type`.

### Selective strategies (all use the same script, different flags)

- **Version-based:** `--from-version=v1-legacy` — regenerate everything old
- **Classifier-driven purge:** For each old memo, run only the NEW classifier. If it says "not a gem" with high confidence, archive the memo (don't regenerate). Much cheaper — classifier only, no memorizer. Add `--mode=purge` flag.
- **Conversation-completeness:** `--incomplete-only` — find memos whose source conversation's `completeness_score` changed since creation. Fixes the Q-without-A bug retroactively.

### Dependencies
- **Requires Follow-up 1 (provenance tracking)** — needs `generated_with_prompt_version` column to query against.

### Verification
- `--dry-run` must print accurate counts
- After a small backfill run, verify old memos are `superseded` (not deleted), new memos have the new `generated_with_prompt_version`, and `memo:revised` outbox events fire so the frontend auto-updates

---

kristofferremback:
## Follow-up 3: Kill the single-message memo path (or hard-gate it)

**Goal:** Eliminate the primary source of "question without answer" garbage memos. Currently, if boundary extraction creates a 1-message conversation (just the question), the memo pipeline skips the conversation path (`MIN_CONVERSATION_MESSAGES = 2` at `service.ts:19`) and falls through to the single-message path — which has zero surrounding context.

### Root cause analysis

The problem is a timing interaction between two pipelines:
1. **Boundary extraction** (`features/conversations/boundary-extraction-service.ts`) runs eagerly on `message:created` — when the question arrives, there's no response yet, so it creates a 1-message conversation
2. **Memo accumulator** (`features/memos/accumulator-outbox-handler.ts`) queues both a `message` item and a `conversation` item
3. **Memo service** processes the conversation item first — sees `messageIds.length < 2`, skips it. Then processes the message item — classifies and memorizes the lone question with no context.

### Option A: Context expansion for message memos (conservative)

In `service.ts:203-269`, before classifying a lone message, fetch 3 messages before and 5 after from the same stream using `MessageRepository.findSurrounding()` (already exists — used by boundary extraction at `boundary-extraction-service.ts:70-76`). Pass the window as context to the classifier.

**Changes:**
- `service.ts` Phase 1: For each message pending item, also fetch surrounding messages via `MessageRepository.findSurrounding(client, item.itemId, streamId, 3, 5)`
- Add surrounding context to `fetchedData`
- Update `CLASSIFIER_MESSAGE_PROMPT` in `config.ts` to include a `## Surrounding Context` section: *"Here is the surrounding conversation for context. Classify only the target message, but use the context to judge whether it's self-contained."*
- **Critical INV-41 constraint:** The surrounding message fetch MUST happen in Phase 1 (withClient), not Phase 2

### Option B: Remove message memos entirely (aggressive, recommended)

Stop writing `MemoTypes.MESSAGE` memos in the hot path. Always defer to the conversation path:
1. In `accumulator-outbox-handler.ts`, on `message:created`, queue the **conversation** (not the message) for memo processing. If the message doesn't belong to a conversation yet, don't queue anything — the conversation will be queued when boundary extraction creates/updates it.
2. Drop `MIN_CONVERSATION_MESSAGES` guard from 2 to 1, BUT only for conversations older than X minutes with no new messages (stale soliloquy check).
3. Add a completeness gate: `if (conversation.completenessScore < 0.5) continue` — the boundary extractor already computes `completenessScore` via `CompletenessUpdate` in `features/conversations/boundary-extraction/types.ts`. We just never use it in the memo pipeline.

**Key files:**
- `apps/backend/src/features/memos/accumulator-outbox-handler.ts` — change what gets queued on `message:created`
- `apps/backend/src/features/memos/service.ts:19` — `MIN_CONVERSATION_MESSAGES` guard
- `apps/backend/src/features/memos/service.ts:281` — add completeness check
- `apps/backend/src/features/conversations/boundary-extraction/types.ts` — `CompletenessUpdate` interface (already exists)
- `apps/backend/src/features/conversations/repository.ts` — `completenessScore` field (already persisted)

**Migration note:** Existing `memo_type = 'message'` memos in the DB stay valid. We just stop creating new ones. No schema migration needed.

### Dependencies
- None — can be done independently of provenance tracking / backfill.

### Verification
- Send a question message in a stream. Verify NO message memo is created immediately.
- Wait for a response. Verify a conversation memo is created containing BOTH question and answer.
- Send a standalone knowledge message (not a Q&A). Verify it still gets memoed (via the conversation path once boundary extraction groups it).
- `bun run --cwd apps/backend test:unit` must pass
- Run memo-classifier and memorizer evals to ensure quality holds

---

kristofferremback:
## Follow-up 4: Completeness gating for conversation memos

**Goal:** Don't generate memos for conversations that are still in-progress. The current debounce (30s quiet / 5min cap in `batch-worker.ts:8-9`) helps but doesn't distinguish "the conversation paused" from "the conversation is done."

### How it works

The boundary extractor already tracks conversation completeness:
- `CompletenessUpdate` type in `features/conversations/boundary-extraction/types.ts` — includes `score: number` and `status: string`
- `ConversationRepository.update()` persists `completenessScore` and `status` on each boundary extraction pass
- Values: `status` can be `"active"` / `"closed"`, `completenessScore` is 0.0-1.0

The memo pipeline never checks these fields. It should.

### Changes

**`apps/backend/src/features/memos/config.ts`:**
```ts
export const MEMO_COMPLETENESS_THRESHOLD = 0.7
export const MEMO_MAX_WAIT_BEFORE_FORCE_MS = 10 * 60 * 1000 // 10 minutes
```

**`apps/backend/src/features/memos/service.ts`** — In the conversation processing loop (around line 281), after fetching the conversation, add:

```ts
const isComplete = conversation.status === "closed" || conversation.completenessScore >= MEMO_COMPLETENESS_THRESHOLD
const isStale = Date.now() - new Date(conversation.updatedAt).getTime() > MEMO_MAX_WAIT_BEFORE_FORCE_MS

if (!isComplete && !isStale) {
  // Re-queue for later — don't mark as processed
  logger.info({ conversationId: conversation.id, completenessScore: conversation.completenessScore }, "Conversation not yet complete, deferring memo")
  continue  // leave the pending item unprocessed so next batch picks it up
}
```

**Important:** The `continue` must skip the `markProcessed` call at the end of the batch. Currently `markProcessed` runs on ALL pending items at line 469. This needs to change — only mark items as processed that were actually handled. Track which items were deferred in a `deferredItemIds` set and exclude them from the `markProcessed` call.

**`apps/backend/src/features/memos/batch-worker.ts`** — Consider increasing the cap interval from 5min to 10min so deferred items get more time before being force-processed.

### Dependencies
- None — can be done independently. Pairs well with Follow-up 3 (killing single-message path).

### Verification
- Send a question in a channel. Verify no memo is created while `completenessScore < 0.7`.
- Send the answer. Verify boundary extraction bumps completeness. Verify memo is created on next batch tick.
- Wait 10 minutes without answering. Verify the `MEMO_MAX_WAIT_BEFORE_FORCE_MS` safety valve fires and creates a memo anyway (with whatever context is available).
- Check logs for `"Conversation not yet complete, deferring memo"` messages during the deferral window.

---

kristofferremback:
## Follow-up 5: Self-verification (critic pass)

**Goal:** Add a second LLM pass that reviews draft memos before persistence, catching hallucinations, meta-commentary, unresolved pronouns, and incomplete information. This is the most ambitious follow-up — only build if quality issues persist after Follow-ups 1-4.

### Architecture

Split memorization into two passes:

1. **Pass 1 — Extract** (`gpt-5.4-nano`): Generate draft memo with existing schema (fast, cheap)
2. **Pass 2 — Critique** (`claude-haiku-4.5` or `gpt-5.4-nano`): Review draft against source messages. Three outcomes:
   - **Approve** — persist as-is
   - **Reject** — log reason, do NOT persist
   - **Revise** — return corrected memo content

### New file: `apps/backend/src/features/memos/critic.ts`

```ts
export interface CriticResult {
  verdict: "approve" | "reject" | "revise"
  revisedContent?: MemoContent  // only when verdict === "revise"
  reason: string                // always: why this verdict
}

export class MemoCritic {
  constructor(private ai: AI, private configResolver: ConfigResolver) {}

  async critique(draft: MemoContent, sourceMessages: string, context: CriticContext): Promise<CriticResult> {
    // ...
  }
}
```

### Prompt design

System prompt for the critic:
```
You are a quality reviewer for knowledge memos. You receive a draft memo and the original source messages it was generated from.

Review the draft against these criteria:
1. SELF-CONTAINED: Can a reader understand this without seeing the original messages?
2. FACTUAL: No meta-commentary ("this memo captures...", "the team discussed...")
3. COMPLETE: Does the source material actually contain enough information for this memo? If the source is a question without an answer, REJECT.
4. PRONOUNS: Are pronouns resolved to actual names where possible?
5. DATES: Are relative dates anchored to absolute dates?

Output your verdict as JSON matching the schema.
```

### Config: `apps/backend/src/features/memos/config.ts`

```ts
export const MEMO_CRITIC_MODEL_ID = "openrouter:openai/gpt-5.4-nano"
export const MEMO_CRITIC_TEMPERATURE = 0.1
```

Add `COMPONENT_PATHS.MEMO_CRITIC = "memo:critic"` in `config-resolver.ts`.

### Integration: `apps/backend/src/features/memos/service.ts`

After `this.memorizer.memorizeMessage()` / `memorizeConversation()` returns (around lines 226-232 and 391-397), add:

```ts
const critique = await this.critic.critique(content, formattedMessages, { workspaceId })
if (critique.verdict === "reject") {
  logger.info({ messageId, reason: critique.reason }, "Memo rejected by critic")
  continue
}
if (critique.verdict === "revise") {
  content = critique.revisedContent!
}
// proceed to embedding and persistence
```

### Audit table for rejected memos

New migration adding `memo_rejections` table:
```sql
CREATE TABLE memo_rejections (
  id TEXT PRIMARY KEY,
  workspace_id TEXT NOT NULL,
  source_type TEXT NOT NULL,          -- 'message' | 'conversation'
  source_id TEXT NOT NULL,
  draft_title TEXT NOT NULL,
  draft_abstract TEXT NOT NULL,
  rejection_reason TEXT NOT NULL,
  critic_model TEXT NOT NULL,
  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
);
```

This lets you analyze what the critic is catching and tune thresholds.

### Cost estimate

At ~500 output tokens per critique using `gpt-5.4-nano` ($0.20/$1.25):
- Per memo: ~$0.000725
- Combined (extract + critique): ~$0.001 per memo
- At 10k memos/month: ~$10/month total

### Eval additions

Add eval cases in `apps/backend/evals/suites/` for the critic:
- Known-bad memos (meta-commentary, Q-without-A, unresolved pronouns) → expect `reject`
- Known-good memos → expect `approve`
- Borderline memos with fixable issues → expect `revise` with specific corrections

### Dependencies
- None technically, but should only be built after Follow-ups 1-4 are shipped and quality is measured. If GPT-5.4 nano + confidence floor + completeness gating already produce good memos, the critic adds cost without enough value.

---

PR #42, opened 2025-12-26
feat(THR-45): add web search and URL reading tools for agents

kristofferremback:
**Linear:** [THR-45](https://linear.app/threa/issue/THR-45)

## Problem

Agents currently have no way to access current information from the web. When users ask about recent events, current prices, or other time-sensitive data, the agent can only rely on its training data which may be outdated.

## Solution

Add two new tools for agents:
- **`web_search`** - Search the web using Tavily API, returns results with titles, URLs, and content snippets
- **`read_url`** - Fetch and extract text content from a specific URL for deeper analysis

Tools are configurable per persona via the existing `enabledTools` field, allowing fine-grained control over which agents have web access.

### How it works

```
User asks question → Agent decides if web search needed → Calls web_search tool
                                                       → Gets results with snippets
                                                       → Optionally calls read_url for full content
                                                       → Cites sources in response
```

### Key design decisions

**1. Tavily for search**

Tavily's API is optimized for LLM consumption - it returns clean snippets and can include a direct answer. Simple REST API with no SDK needed.

**2. Per-persona tool enablement**

Tools are enabled via `persona.enabledTools` array. If `null`, all tools are enabled (backward compatible). This allows creating personas with/without web access.

**3. Conditional system prompt**

The system prompt only includes tool documentation for tools that are actually enabled, avoiding confusion for the LLM about unavailable capabilities.

**4. Simple HTML extraction**

`read_url` uses regex-based HTML-to-text extraction rather than a heavy dependency like cheerio. Sufficient for reading articles and docs.

## New files

| File | Purpose |
|------|---------|
| `src/agents/tools/web-search-tool.ts` | Tavily-powered web search tool |
| `src/agents/tools/web-search-tool.test.ts` | Unit tests (5 tests) |
| `src/agents/tools/read-url-tool.ts` | URL content fetcher with HTML extraction |
| `src/agents/tools/read-url-tool.test.ts` | Unit tests (8 tests) |
| `src/db/migrations/016_ariadne_enabled_tools.sql` | Enable tools for Ariadne |

## Modified files

| File | Change |
|------|--------|
| `packages/types/src/constants.ts` | Add `AgentToolNames` constant |
| `packages/types/src/index.ts` | Export new constants |
| `src/lib/env.ts` | Add `TAVILY_API_KEY` config |
| `src/agents/companion-runner.ts` | Create tools conditionally based on persona |
| `src/agents/companion-agent.ts` | Add conditional tool docs to system prompt |
[...]

PR #367, opened 2026-04-14
feat(messaging): message metadata for external references

kristofferremback:
## Problem

Slack-style notification flows (GitHub PR events, CI, incident pings) have no way to
attach structured external references to a message. Two concrete pain points:

1. **Dedup is ad-hoc.** When a GitHub webhook fires twice — or when an integration
   wants to post "PR #42 opened" exactly once — there's no primitive for "has a
   message already been posted for this PR event?".
2. **Callers can't query by external IDs.** Even if the content happens to contain
   the PR number, there's no durable way to list all messages tied to a given
   external entity.

## Solution

Add an opt-in `metadata: Record<string, string>` map on messages, queryable via the
public API with **AND-containment** semantics. Callers attach arbitrary key/value
references on send and later look them up by a filter object.

### How it works

```
Send:     POST /streams/:id/messages { content, metadata: { "github.pr.id": "42" } }

Query:    POST /messages/find-by-metadata { metadata: { "github.pr.id": "42" } }
          → returns every non-deleted, accessible message whose metadata contains
            ALL the supplied key/value pairs (JSONB @> containment, GIN-indexed)
```

Storage is a `JSONB NOT NULL DEFAULT '{}'` column on `messages` plus a GIN
`jsonb_path_ops` index — the narrowest opclass that supports `@>` and is materially
smaller/faster than default `jsonb_ops` when we only do containment.

### Key design decisions

**1. Metadata is for association, `clientMessageId` is for idempotency.**
Resist the urge to bake dedup into the send endpoint. Metadata can't enforce
uniqueness generically (you'd need per-subset unique indexes). Instead, integrations
derive a deterministic `clientMessageId` — e.g. `gh:${prId}:${eventType}:${deliveryId}` —
which reuses the existing race-safe `ON CONFLICT DO NOTHING` path (INV‑20). Metadata
is then free to carry *semantic* references (pr.id, repo, event, labels…) that may
overlap across messages and are purely for lookup.

**2. Reserved `threa.*` key prefix.**
System-generated metadata (e.g. `threa.source=github`) must not be spoofable by
API callers. Writes reject keys starting with `threa.`; reads/queries allow them so
callers can still lookup system-tagged messages.

**3. AND semantics, not OR.**
`@>` containment naturally gives AND. For the dedup use case you almost always want
"PR #42 **and** this event type" — if someone later needs OR, it can be expressed as
two queries or a dedicated filter grammar; we didn't want to ship an overbuilt query
DSL in v1.

**4. Dedicated `POST /messages/find-by-metadata` endpoint, not query-params on
`listMessages`.** Dedup is cross-stream by nature. JSON-in-body beats encoding
nested objects in URL query params, and `listMessages` is sequence-paginated inside
one stream — it doesn't compose with containment lookups.

**5. Always-present `metadata` field (empty object when unset).**
Every message response — send, list, find-by-metadata, search — returns
`metadata: {}` when nothing is set. Predictable wire shape beats "omit-when-empty".

**6. Conservative limits: 20 keys × 64-char keys × 256-char values × 4 KB serialized.**
Big enough for real integration payloads (the GitHub example uses 3 keys), small
enough that a rogue caller can't blow up message sizes.

**7. Access control pushed into SQL.**
`findByMetadata` takes the caller's accessible stream IDs and `stream_id = ANY(...)`
is applied in the query, so even if a handler forgets to pre-check access, the repo
can't leak messages from private streams. An optional `streamId` param is intersected
with the accessible set.

**8. Metadata is immutable post-send in v1.**
No PATCH on metadata. If you need to change references, resend with a new
`clientMessageId` or reply. Keeps the surface small; can add later.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260414160920_add_message_metadata.sql` | Add `metadata JSONB` column + GIN `jsonb_path_ops` index. |
| `apps/backend/src/features/messaging/metadata-schema.ts` | Shared Zod validators (`messageMetadataSchema` for writes, `messageMetadataFilterSchema` for queries) + exported limit constants. |
| `apps/backend/src/features/messaging/metadata-schema.test.ts` | Unit tests for the validators (reserved prefix, key/value limits, empty-filter rejection on query side). |
| `apps/backend/tests/integration/message-metadata.test.ts` | Integration tests: round-trip, default `{}`, AND semantics, access-scoping, `streamId` intersection, soft-delete exclusion, empty-input guards. |

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/domain.ts` | Add `Message.metadata: Record<string, string>`. |
| `packages/types/src/api.ts` | Add optional `metadata` to the four `CreateMessage*Input` shapes. |
| `apps/backend/src/features/messaging/repository.ts` | Row/domain/insert/SELECT_FIELDS/mapRowToMessage plumbing; new `findByMetadata(db, { streamIds, filter, streamId?, limit? })` using `@>` containment, access-scoped, newest-first, excludes soft-deleted. |
| `apps/backend/src/features/messaging/event-service.ts` | Thread `metadata` through `CreateMessageParams` + `MessageCreatedPayload` so it reaches the outbox/socket/frontend cache. |
| `apps/backend/src/features/messaging/handlers.ts` | Internal create schemas accept `metadata`. |
| `apps/backend/src/features/messaging/index.ts` | Barrel exports for the metadata schema module. |
| `apps/backend/src/features/public-api/schemas.ts` | `sendMessageSchema` accepts `metadata`; new `findMessagesByMetadataSchema`. |
| `apps/backend/src/features/public-api/routes.ts` | `messageSchema` + `searchResultSchema` now declare `metadata`; new `findMessagesByMetadata` route entry. |
| `apps/backend/src/features/public-api/handlers.ts` | `serializeMessage` always returns `metadata`; `sendMessage` forwards it; new `findMessagesByMetadata` handler. |
| `apps/backend/src/features/search/repository.ts` | Add `metadata` to `SearchResult` + all four SQL variants (full-text, hybrid, exact, empty-query). |
| `apps/backend/src/features/search/handlers.ts` | `serializeSearchResult` always returns `metadata`. |
| `apps/backend/src/routes.ts` | Wire `POST /api/v1/workspaces/:workspaceId/messages/find-by-metadata` (scope: `MESSAGES_READ`). |
| `apps/backend/tests/e2e/public-api-crud.test.ts` | Send-with-metadata round-trip, always-present empty, reserved-prefix rejection, oversize value rejection, find-by-metadata AND semantics, stream filter, cross-stream scope, 400/404 paths. |
| `apps/backend/tests/e2e/public-api-openapi.test.ts` | Response-shape check for the new `find-by-metadata` endpoint. |
| `docs/public-api/openapi.json` | Regenerated spec (new endpoint + `metadata` field on message responses). |
| `apps/backend/src/features/agents/researcher/researcher.ts` | Fill in `metadata` on locally-constructed `SearchResult`s after the type widened. |
| Test fixtures (`apps/backend/evals/...`, `apps/backend/src/features/agents/*.test.ts`, `apps/backend/src/features/conversations/...`, `apps/backend/src/lib/ai/message-formatter.test.ts`, `apps/frontend/src/hooks/use-stream-or-draft.test.tsx`) | Add `metadata: {}` to inline `Message` fixtures to match the widened wire type. |

## Test plan

- [x] `bun test apps/backend/src` — 816 / 816 pass
- [x] `bun run typecheck` across the monorepo — green
- [x] `bun run --cwd apps/backend lint` — clean
- [x] `bun run generate:api-docs:check` — spec up to date
- [x] Migration is append-only with the next timestamp (INV-17)
- [ ] Integration tests run in CI with a live Postgres (`tests/integration/message-metadata.test.ts` — not runnable in the sandbox)
- [ ] E2E tests run in CI against a booted backend (`tests/e2e/public-api-crud.test.ts` + `public-api-openapi.test.ts`)
- [ ] Manual smoke: issue an API key with `messages:write` + `messages:read`, send a
      message with `metadata: { source: "github", "github.pr.id": "42" }`, call
      `find-by-metadata` with the same filter and confirm the message comes back.

## Invariants respected

INV-1 (no FKs), INV-3 (no DB enums; string-map shape validated in code), INV-5
(data access via repository), INV-8 (workspace/stream-scoped queries), INV-17
(append-only migration), INV-20 (idempotency via `clientMessageId`, not check-then-act
on metadata), INV-47 (no nested ternaries), INV-51/52 (feature colocation + barrel
imports — the public-api layer imports from `../messaging`, not the internal file),
INV-55 (Zod validation), INV-58 (contentJson remains canonical; metadata is orthogonal).

---

🤖 *PR by [Claude Code](https://claude.com/claude-code)*


kristofferremback:
### Code review
**Review models:** Orchestrator: opus | Reviewers: sonnet x3

One layering issue found by self-review and already addressed in `3010fcf5` before posting:

1. `apps/backend/src/features/public-api/handlers.ts:797` — `findMessagesByMetadata` was calling `MessageRepository.findByMetadata(pool, ...)` directly while every other message read in this file (`getMessages`, `getMessageById`) goes through `eventService` (historical / INV‑34). Fixed by adding a thin `EventService.findByMetadata` passthrough that mirrors the existing `getMessages`/`getMessageById` pattern and dropping the direct `MessageRepository` import from the handler.
   https://github.com/kristofferremback/threa/blob/3010fcf5c9734c8f2a3ec69db722fa2e1d091a1c/apps/backend/src/features/public-api/handlers.ts#L794-L806

---

All PR-body commitments shipped: `metadata` plumbed through send/list/search/find-by-metadata responses; new endpoint has dedicated Zod schema + route entry + router wiring + OpenAPI regeneration; limits enforced (20 keys × 64/256 chars × 4 KB); reserved `threa.*` prefix rejected on writes, allowed on queries; migration is append-only at the next timestamp; `findByMetadata` uses `@>` containment with `jsonb_path_ops` GIN index and AND semantics; access scoping pushed into SQL via `stream_id = ANY($)`.
</details>

Parameterized JSONB writes/queries (`${JSON.stringify(metadata)}`, `${JSON.stringify(params.filter)}::jsonb`) via `sql` template tag are injection-safe. `metadata ?? {}` fallback on the wire matches the DB's `NOT NULL DEFAULT '{}'`. Empty-input guards on repo (`streamIds.length === 0`, `Object.keys(filter).length === 0`) return `[]` early. Sequential reaction lookup after the first query is consistent with the rest of the repo (INV-30 doesn't apply — two reads, not one).
</details>

INV‑1 (no FKs), INV‑3 (shape in code, not DB enum), INV‑5 (repo pattern), INV‑8 (stream‑scoped — `messages` has no `workspace_id` column; scoping flows through `stream_id`, matching the codebase pattern), INV‑17 (append‑only migration), INV‑20 (idempotency still via `clientMessageId` + existing `ON CONFLICT DO NOTHING`; metadata isn't used for dedup), INV‑47 (no nested ternaries), INV‑51/52 (feature colocation + barrel imports — `public-api/schemas.ts` now imports `from "../messaging"`, not the internal file), INV‑55 (Zod validation everywhere), INV‑58 (contentJson unchanged).
</details>

<details><summary>🏗️ Design — 1 concern (fixed)</summary>

Layering issue described above, now fixed in `3010fcf5`.

Noted but not flagged: `findByMetadata` query doesn't include a `workspace_id` filter — relies entirely on `stream_id = ANY(accessibleStreamIds)` for tenant isolation. This matches the established pattern (the `messages` table has no `workspace_id` column; isolation rides on the stream FK-less relationship per INV‑1/INV‑8), so no action required, but worth knowing if anyone refactors `getAccessibleStreamIds` in the future.
</details>

New endpoint requires `MESSAGES_READ` scope and runs through the standard public-api auth + rate‑limit middleware chain. Access control pushed into SQL so handler oversight can't leak cross‑stream messages. Write validator rejects reserved `threa.*` prefix; query validator allows it (intentional — for looking up system-tagged messages). Size caps (20 keys × 256‑char values × 4 KB serialized) bound worst-case request size.
</details>

---

kristofferremback:
### Code review (re-run on `255ab876`)
**Review models:** Orchestrator: opus | Reviewers: sonnet x3

No issues found in the two new commits since the prior review (`3010fcf` layering fix + `255ab87` workflow dogfood). Supersedes [#issuecomment-4246287568](https://github.com/kristofferremback/threa/pull/367#issuecomment-4246287568).

---

- `3010fcf` resolves the prior-review layering finding: `findMessagesByMetadata` now calls `eventService.findByMetadata(...)`, matching the `getMessages`/`getMessageById` pattern.
- `255ab87` dogfoods the primitive exactly as described in the PR body: `notify.yml` tags messages with `clientMessageId` + structured metadata; `staging.yml` uses `staging-deploy:pr-<N>` as a per-PR deterministic `clientMessageId`; the new teardown step uses `POST /messages/find-by-metadata` + `PATCH /messages/:id` — the exact flow the PR body promises.
</details>

- `EventService.findByMetadata` forwards all params to the repo with no mutation or dropped arguments.
- Handler still calls `getAccessibleStreamIds(req)` before the service call; repo-level `streamIds` intersection provides defense-in-depth.
- Teardown step guards on empty `MESSAGE_ID` via `jq -r '.data[0].id // empty'` + `[[ -z "$MESSAGE_ID" ]]` — graceful no-op when no deploy message exists (pre-existing PRs, skipped notify step).
- All workflow user-controlled inputs flow through `jq --arg` which JSON-escapes; `clientMessageId` components (`WORKFLOW_RUN_ID`, `CONCLUSION`, `PR_NUMBER`) are GitHub-assigned and non-attacker-controlled.
</details>

One agent raised INV-30 ("avoid `withClient` for single-query paths; pass `pool` directly") against the new `EventService.findByMetadata` passthrough since the underlying repo method runs two sequential queries (messages + reactions). Cross-checked against the existing file: `getMessages` and `getMessageById` also pass `this.pool` directly for the same messages+reactions two-query pattern; only `getMessagesByIds` uses `withClient`. INV-30 is phrased as a negative constraint on `withClient`, not a positive mandate for it on multi-query paths. The new method matches the majority precedent — stylistic at most, not a clear violation. Dropped.

All other invariants preserved from the prior review: INV-1, INV-3, INV-5, INV-8, INV-17, INV-20, INV-47, INV-51/52, INV-55, INV-58.
</details>

- New `EventService.findByMetadata` mirrors the existing `getMessages`/`getMessageById` passthrough shape (thin, pool-direct).
- Workflow jq payloads use `--arg` consistently, no shell interpolation into JSON values.
- `resolveAuthorDisplayNames` in the new handler is the standard batch call; no N+1.
- No duplication warranting extraction between `notify.yml` and `staging.yml` jq payloads (different field sets, ≤10 lines each).
- Teardown uses `continue-on-error: true` scoped to the notification step, so notification failures never fail the teardown job.
</details>

No new attack surface. New endpoint still requires `MESSAGES_READ` scope behind the standard public-api auth + rate-limit chain. Workflow secrets (`THREA_BOT_API_KEY`) not exposed in logs; `continue-on-error` on notification avoids leaking state via failure traces.
</details>

---

PR #375, opened 2026-04-16
feat(saved-messages): Save + reminder system for messages

kristofferremback:
Per-user save + reminder (todo) system for messages. All five phases landed.

## What's in it

- **Save a message** for later via hover bookmark button, context menu, or popover preset.
- **Reminders at minute granularity** delivered through the existing `queue_messages` pipeline.
- **Statuses `saved | done | archived`** — distinct intent, distinct ordering.
- **Cross-device reactivity** via user-scoped socket events; all open tabs/devices update instantly.
- **Offline-first** with a new Dexie store — the Saved view renders from IDB, server syncs in the background.
- **Live-resolved message content** (INV-58) — deleted/access-lost messages surface as "unavailable" without denormalising content.
- **Timezone-aware reminder math** (INV-42) via `Intl.DateTimeFormat` — no extra deps.

## Phases

- [x] **Phase 1** — schema, repository, id helpers, `SavedStatus` type (23 repo tests)
- [x] **Phase 2** — service + handlers + outbox (`saved:upserted`, `saved:deleted`) (12 service tests + 9 handler tests)
- [x] **Phase 3** — reminder queue (`saved.reminder_fire`) + worker + push delivery (13 service tests + 3 worker tests)
- [x] **Phase 4** — sidebar + Saved view + IDB store v24 + TanStack hooks + socket sync (8 `formatFutureTime` tests)
- [x] **Phase 5** — hover bookmark + reminder popover + context menu entry

## Key design choices

**Race-safe upsert via CTE.** `INSERT … ON CONFLICT DO UPDATE` with a `WITH old AS (…)` that snapshots the pre-update `reminder_queue_message_id`. Required because `RETURNING` on the conflict path reads post-update values, and the UPDATE clause nulls the column — without the CTE the caller could never tombstone the stale queue row.

**Tombstone skips in-flight claims.** `QueueRepository.tombstoneById` only fires when `claimed_by IS NULL`. If a worker already claimed the row, the tombstone noops; the worker's own `markReminderFired` checks domain state (status + `reminder_sent_at IS NULL`) and correctly noops when the saved row has moved off `saved`. This avoids clobbering the worker's completion UPDATE (which matches on `claimed_by`).

**Outbox payload built from tx client.** `resolveSavedView` accepts a `Querier` so calls from inside `withTransaction` use the tx client, not a fresh pool connection. The outbox event reflects committed state under concurrency (INV-4/7).

**PATCH enforces one field per request.** Zod `refine` rejects requests that combine `status` and `remindAt` — each mutation is exactly one transaction and one socket event.

**Reminder presets timezone-aware.** "Tomorrow 9am" / "Next Monday 9am" resolve calendar boundaries in the user's stored IANA timezone using `Intl.DateTimeFormat`. Wall-clock math via a short converge pass handles DST.

**Offline Saved view.** Dexie store v24 (`savedMessages`) with compound `(workspaceId+status+_savedAtMs)` and `(workspaceId+status+_statusChangedAtMs)` indexes so each tab hits a narrow partial index. `useLiveQuery` is the render source; TanStack runs in the background.

**Cross-device reactivity.** Three new user-scoped outbox events (`saved:upserted`, `saved:deleted`, `saved_reminder:fired`) fan out to every socket in the `ws:{wsId}:user:{userId}` room. Frontend handlers write through to IDB and invalidate TanStack caches. `refetchOnReconnect: true` closes the offline-gap (INV-53).

## Test plan

- [x] `bun test apps/backend/src/features/saved-messages/` — 57 pass (repo + service + worker + handler)
- [x] `bun run typecheck` — clean across all packages
- [x] `bun run lint` — clean
- [x] 33 `formatFutureTime` tests covering timezone edge cases (LA/Tokyo boundaries)
- [ ] Manual run-through of the saved flow in a local dev server
- [ ] Playwright E2E covering hover → save → reminder flow (future follow-up)

## Reviewer notes

- Migration `20260416173513_saved_messages.sql` is append-only (INV-17), adds one table + partial indexes per tab.
- No FK constraints (INV-1), no DB enums (INV-3), all ULID-prefixed (INV-2: `saved_`, `remq_`).
- The backend `service.ts` has a side-effect `import "../workspaces"` to preload the public-api → schemas → messaging chain before its own imports would re-enter messaging mid-evaluation. Activity's service relies on the same ordering.

https://claude.ai/code/session_01LE4WCBVeFLLiQbzayRPh8d


kristofferremback:
# Code review — self-review of recent iteration

## Overview

Per-user Save + Reminder system for messages: new `saved_messages` table, reminder queue pipeline through existing `queue_messages` infra, three outbox events (`saved:upserted`, `saved:deleted`, `saved_reminder:fired`), a `saved_reminder` activity kind, offline-first Saved view backed by Dexie v24, hover/drawer bookmark + reminder UI, and an inline "Saved" indicator in the timeline.

## Strengths

- **Race-safe upsert** via `WITH old AS (…)` CTE correctly captures pre-update `reminder_queue_message_id` for tombstoning.
- **Invariants honoured**: no FKs (INV-1), ULIDs (INV-2), TEXT over enums (INV-3), outbox in same tx as writes (INV-4/7), tx-aware outbox payload resolver, append-only migrations (INV-17).
- **Tombstone-skips-claim** avoids clobbering the worker's `complete()` UPDATE.
- **IDB reconciliation** (`40a1100`) matches the stream-sync `fetchStartedAt` guard pattern so concurrent socket writes survive.
- **Honest menu labels** (`5a46d53`) — the prior silent toggle-without-label-flip was exactly the kind of bug that ships and bothers users forever.

## Concerns

### Correctness
- **`actorId = savedId`** for saved_reminder dedup is correct but semantically stretchy — `actor_id` is conventionally a user/persona/bot id. Worth a comment near the `user_activity` table if a future contributor relies on that invariant.
- **`useLiveSavedCount`** loads full `toArray()` then filters in JS. Fine for current scale (< 100 reminded rows per workspace), but scans every saved row on each trigger. If this grows, persist `_reminderFiredAtMs` as a compound index.
- **`replaceSavedPage`** reads `existing` outside the `rw` transaction, then deletes/puts inside. Low-probability concurrent-tab clobber acceptable for a cache.

### UX
- **Done/archived items have no visual indicator** in the timeline — users can unknowingly "re-save" an archived message (correctly bumps it back to saved per spec, but no feedback).
- **Saved + ReminderBadge** could wrap awkwardly on narrow screens with long stream names; no `max-w` / `truncate` applied.
- **Native `<input type="date">`/`type="time"`** on older iOS may degrade to text inputs.

### Tests
- Zero new unit tests for the iteration fixes: `SavedIndicator`, `replaceSavedPage` reconciliation, `useLiveSavedCount` filter, activity dedup fix (`actorId = savedId`), save/unsave menu split.
- The activity dedup change is a silent correctness fix — a regression test (two fires with different `savedId`, same message → two rows) would pin the behaviour.

### Security
- No new attack surface. `saved_reminder:fired` outbox payloads carry `contentMarkdown` but the activity row stores only a 200-char preview, matching how mentions/reactions work.
- Workspace scoping enforced via `workspace_id` filter in every saved_messages query.

## Follow-ups (addressing in a follow-up commit)

1. Service-level test: re-saving after delete creates a **new** saved_reminder activity.
2. Dexie reconciliation test: local row + absent from server + old `_cachedAt` → deleted; fresh `_cachedAt` → kept.
3. Visible "Saved · Done" pill on done items in the timeline.

`useLiveSavedCount` optimisation deferred until the reminded set grows.

---

kristofferremback:
### Code review
**Confidence: 5/7** — Good

Found 4 issues:

1. `apps/backend/src/features/saved-messages/service.ts:L329-L342` — ULID collision in `enqueueReminder` writes the colliding queue id to `reminder_queue_message_id` on the saved row. The `INSERT` fails on PK conflict, the `isUniqueViolation` branch logs and falls through, then `SavedMessagesRepository.updateReminder(... queueMessageId)` persists the pointer anyway — so the saved row now points at another queue row (likely the one that won the PK race). A later tombstone would cancel the wrong job or noop. ULID collisions are cryptographically rare, but the collision branch explicitly pretends to handle this and doesn't. Fix: skip the `updateReminder` on collision (or re-fetch the existing queue row and reuse its id). (bug)
   https://github.com/kristofferremback/threa/blob/9abb6060b4d2ad3cad68ea7b7d5b4648a27d3c4e/apps/backend/src/features/saved-messages/service.ts#L329-L342

2. `apps/frontend/src/hooks/use-saved.ts:L73-L97` — `replaceSavedPage` reconciles against the full IDB set for `(workspaceId, status)` but the server only returns the first page (`limit: 50`). Cached rows 51+ are absent from `serverIds` and, if their `_cachedAt < fetchStartedAt`, they're bulk-deleted on every first-page refetch. No pagination UI loads page 2 today, so this is latent, but the reconciliation logic is incorrect by design. Fix: gate the delete on `nextCursor === null` (we have the full set), or scope deletion to the server page's savedAt/statusChangedAt window. (bug)
   https://github.com/kristofferremback/threa/blob/9abb6060b4d2ad3cad68ea7b7d5b4648a27d3c4e/apps/frontend/src/hooks/use-saved.ts#L73-L97

3. `apps/frontend/src/hooks/use-saved.ts:L104-L112` — Comment claims "reconnect hook in workspace-sync invalidates `savedKeys.all` so offline-missed socket events get filled in on the next render — INV-53" but no such invalidation exists. `refetchOnReconnect: true` only fires on browser network online/offline, not Socket.io reconnect (which is the gap INV-53 targets — "bootstrap must be invalidated on reconnect/resubscribe to close event gaps"). Activity queries are invalidated inside the socket handlers; saved queries aren't. Fix: wire a `savedKeys.all` invalidation to the socket reconnect handler in `workspace-sync.ts`, or correct the comment to reflect actual coverage. (reactivity)
   https://github.com/kristofferremback/threa/blob/9abb6060b4d2ad3cad68ea7b7d5b4648a27d3c4e/apps/frontend/src/hooks/use-saved.ts#L104-L112

4. `apps/frontend/src/components/timeline/message-event.test.tsx:L99-L105` — INV-48 ("Testing and Reliability Expectations"): "Avoid `mock.module()` for shared modules; prefer scoped `spyOn` patterns." This PR adds a new `vi.mock("@/hooks/use-saved", ...)` call — the vitest equivalent flagged by the repo's own `no-restricted-syntax` rule. The file has pre-existing `vi.mock` usage, but extending the pattern to a newly-stubbed shared module is the thing the invariant calls out. A scoped `spyOn` (or wrapping `<ServicesProvider>` in the test's `Wrapper`) would avoid it. (claude-md)
   https://github.com/kristofferremback/threa/blob/9abb6060b4d2ad3cad68ea7b7d5b4648a27d3c4e/apps/frontend/src/components/timeline/message-event.test.tsx#L99-L105

---

No standalone plan file for this feature (iterated directly with the user across 19 commits). Diff matches commit intent: schema → repository → service → handlers → outbox → worker → push → frontend + UI iterations + URL-driven tabs (INV-59) + E2E.
</details>

<details><summary>🔍 Bugs — 2 issues</summary>

- **ULID collision path (#1)**: collision handler falsely advertises safety; writes bad pointer.
- **First-page-only reconciliation (#2)**: `replaceSavedPage` is incorrect by design when paginated data exists locally.
</details>

- **INV-48 (#4)**: new `vi.mock` on a shared-module boundary; scoped `spyOn` or `ServicesProvider` wrapping preferred.

Other invariants checked and clean: INV-1 (no FKs), INV-2 (ULID prefixes `saved_`, `remq_`), INV-3 (TEXT over enums), INV-4/7 (outbox in same tx as writes), INV-8 (workspace scoping on every query), INV-14 (Shadcn primitives), INV-17 (append-only migrations), INV-20 (race-safe upsert via CTE), INV-30 (pool for single queries), INV-32 (HttpError), INV-40 (Link for nav / button for actions), INV-42 (timezone-aware date math), INV-47 (no nested ternaries — resolved in reminder-picker-sheet.tsx), INV-55 (Zod input validation), INV-58 (contentJson canonical, markdown at boundary), INV-59 (URL-driven tabs — newly added and applied to Saved + Activity).
</details>

- **Reactivity gap (#3)**: socket-reconnect invalidation promised by the code comment isn't wired up.

No other concerns. Race-safe upsert via CTE is good. Tombstone-skips-claim handles the worker race correctly. `actorId = savedId` dedup trick for saved_reminder is documented on `InsertActivityParams`. Dexie v25 upgrade function correctly populates `_reminderFiredAtMs` for pre-existing caches.
</details>

No new attack surface. `saved_reminder:fired` outbox payload carries `contentMarkdown` but activity context stores only a 200-char preview, matching mentions/reactions precedent. All saved-message queries filter by `workspace_id` + `user_id`. No `dangerouslySetInnerHTML`, no injection sinks, no new auth paths.
</details>

PR #389, opened 2026-04-19
fix(timeline): never leave timeline blank while IDB is unresolved

kristofferremback:
## Summary

- Stream timeline could render a terminal blank scroll area (no skeleton, no "No messages yet") when `useLiveQuery` returned `undefined` for an extended window — observed on mobile as a DM with no messages until pull-to-refresh.
- Both `isLoading` and `isConfirmedEmpty` were gated on `idbResolved`, so any stretch where it stayed false produced a dead-looking page with no indicator.
- Add a 200ms grace window before flipping to a skeleton: fast resolves still render briefly blank (preserving the original flicker-free intent), slow/stuck resolves now show progress.
- Extracted `computeTimelineLoadState` as a pure helper with unit tests pinning the state machine.

## Root cause notes

I couldn't isolate a single deterministic logic bug that keeps `idbResolved=false` for seconds. Likely contributors:

- Dexie cold-start under iOS memory pressure (IDB open + first range scan can stretch to seconds after Safari evicts the page).
- `dexie-react-hooks` returns `undefined` across every dep change of `useLiveQuery`. Compounding dep churn (`streamId` change, bootstrap floor flipping, coordinated `useQueries` rebuilds) chains undefined windows.

The fix is defensive — it doesn't depend on the root cause. The state machine now guarantees a blank scroll area is never a terminal state.

## Test plan

- [x] Unit tests for `computeTimelineLoadState` (5 input permutations) — `apps/frontend/src/hooks/use-events.test.ts`
- [x] Full frontend test suite (`bun run test` in apps/frontend) — 1281 passing
- [x] Lint + typecheck via lint-staged on commit
- [ ] Manual verification on mobile: navigate to a DM with messages, confirm no blank flash and skeleton appears if IDB stalls

https://claude.ai/code/session_01Mqb74KEmnWuhR8ZJMxv2D9

### DIFF
--- a/apps/frontend/src/hooks/use-events.test.ts
+++ b/apps/frontend/src/hooks/use-events.test.ts
@@ -135,7 +135,7 @@ describe("computeTimelineLoadState", () => {
     ).toEqual({ isLoading: false, isConfirmedEmpty: false })
   })
 
-  it("reports bootstrap-settled via isConfirmedEmpty even when events are present (call site gates on item count)", () => {
+  it("does not claim confirmed-empty once events are present after bootstrap settles", () => {
     expect(
       computeTimelineLoadState({
         idbResolved: true,
@@ -143,6 +143,6 @@ describe("computeTimelineLoadState", () => {
         isBootstrapLoading: false,
         idbResolveTimedOut: false,
       })
-    ).toEqual({ isLoading: false, isConfirmedEmpty: true })
+    ).toEqual({ isLoading: false, isConfirmedEmpty: false })
   })
 })
--- a/apps/frontend/src/hooks/use-events.test.ts
+++ b/apps/frontend/src/hooks/use-events.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest"
 import {
+  computeTimelineLoadState,
   filterEventsForDisplay,
   getCachedWindowFloor,
   getDisplayFloor,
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (few non-blocking concerns)
**Review models:** Orchestrator: opus-4.7 | Reviewers: sonnet x3

Found 2 issues:

1. `apps/frontend/src/hooks/use-events.ts:68-97` — `isConfirmedEmpty` is a semantic misnomer. The field returns `true` whenever `idbResolved && !isBootstrapLoading` — including when events exist (`hasAnyEvents=true`). The JSDoc says *"Render 'No messages yet' — both data sources confirmed empty"*, but the implementation actually signals "bootstrap has settled." The call site in `stream-content.tsx` guards correctly with `visibleItems.length === 0 && isConfirmedEmpty`, and the test at line 138 explicitly documents this with *"call site gates on item count"* — but this is a footgun, not a fix. Any future caller that reads `isConfirmedEmpty` without the item-count guard will mis-render on a populated stream. Either rename to `isBootstrapSettled`, or fold the `hasAnyEvents` guard into the helper so the name matches the value. Relevant: [INV-25, §7 Maintainability and Scope Control]: *"Comments explain why the current implementation works; they do not narrate the change history."* — the JSDoc describes a contract the implementation doesn't honor.
   https://github.com/kristofferremback/threa/blob/30db8c1777eeffa2244c3ad5b9cc6aeac1158842/apps/frontend/src/hooks/use-events.ts#L68-L98

2. `apps/frontend/src/hooks/use-events.ts:51` — `IDB_SKELETON_DELAY_MS = 200` is a parallel implementation of `LOADING_DELAY_MS = 300` in `coordinated-loading-context.tsx` (INV-35). Both are "delay before showing skeleton to avoid flicker," tuned independently. If one is retuned the other silently drifts. Additional concrete concern: during the initial load the `CoordinatedLoadingGate` holds `phase="loading"` for 300ms, but this hook's 200ms timer can fire **inside** that window, latching `idbResolveTimedOut=true`. When the gate finally releases, the timeline mounts with the skeleton already on — bypassing the fast-path blank the constant was designed to preserve. Consider either (a) importing `LOADING_DELAY_MS` and offsetting from it, or (b) resetting the timer when the coordinated-loading gate flips, or (c) accepting this and documenting why the two constants diverge.
   https://github.com/kristofferremback/threa/blob/30db8c1777eeffa2244c3ad5b9cc6aeac1158842/apps/frontend/src/hooks/use-events.ts#L46-L52

---

No plan file for this branch in `.claude/plans/`. Diff matches the PR description: 200ms grace window, pure helper extraction, unit tests pinning 6 input permutations. No missing corresponding changes.
</details>

- Timer correctly resets on stream switch (effect depends on `idbResolved`, which flips via `useStreamEvents` when `streamId` changes).
- Timer cleanup path (`setIdbResolveTimedOut(false)` on resolve) prevents stuck skeleton after late IDB resolution.
- State machine covers all 4 input axes sensibly; no off-by-one or null hazards.
</details>

<details><summary>📋 CLAUDE.md Compliance [1 concern, folded into issue #1 above]</summary>

INV-25 JSDoc accuracy flagged under issue #1 — the interface doc describes semantics the implementation doesn't honor.
</details>

<details><summary>🏗️ Design [2 concerns, folded into issues above]</summary>

- Leaky abstraction on `isConfirmedEmpty` (issue #1).
- Parallel timing constant divergence (issue #2).
- `computeTimelineLoadState` is cheap and pure — not memoizing is fine.
- Hardcoded 200ms is acceptable for a single call site (INV-36).
</details>

Pure React state logic. No user input sinks, no dangerouslySetInnerHTML, no data-exposure surface. Nothing to flag.
</details>

---

PR #390, opened 2026-04-20
feat(timeline): denser Comfortable density with author-grouping

kristofferremback:
## Summary

Redesign the Comfortable message density so more conversation fits on screen without losing the airy feel. Body typography (`text-sm leading-relaxed`) is preserved exactly — this is chrome reduction + grouping + affordance polish, not a typography change.

- **Author-grouping** — consecutive same-author messages within 5 minutes collapse into a "head + continuations" run. Continuations drop the avatar/name row and show only a gutter time on hover, saving ~50px per extra message.
- **Denser baseline** — `py-4` → `py-3`, `gap-[14px]` → `gap-3`, avatar `h-9 w-9 rounded-[10px]` → `h-8 w-8 rounded-[8px]`.
- **ThreadCard** — replaces the tiny "N replies" text link with a tappable card showing participant avatars, reply count, last-reply timestamp, and a markdown-stripped preview of the latest reply. Adds a 2px gold left-line that literally extends Ariadne's thread.
- **ActivityPill** — inline "is thinking…" pill next to the author name, replacing the redundant footer link.
- **Reactions polish** — 26px min-height pills, gold `primary/0.14` + border for the "you reacted" state, cleaner `+` button.
- **Floating hover toolbar** — moves out of the header row into an `absolute`-positioned sibling so heads and continuations share the same affordance. Mobile keeps the long-press drawer (no hover-only actions).
- **New `threadSummary` on the message read model** (cap 3 participants, raw markdown on wire stripped on the client — INV-60).

Plan: [.claude/plans/giggly-inventing-rose.md](.claude/plans/giggly-inventing-rose.md)

## Virtuoso interaction notes

Author-grouping raises a concern: when pagination loads an older same-author message, the previously-first-loaded message flips from head to continuation and shrinks ~40px. Could cause scroll jitter at the seam.

Measured empirically in dev:test with an 80-message same-author run, scrolling up across ~14 pagination rounds. Across 656 frames tracking 91 rendered messages: exactly **1 flip**, and it happened at `top: -486` — entirely inside Virtuoso's `increaseViewportBy: {top: 600}` pre-render zone, never visible to the user. Zero involuntary scrollTop jumps. The `rangeChanged`-based pagination trigger fires 3 items from the edge, well before any head enters the viewport, so heads flip while still virtualized.

No stable-head cache or scrollBy compensation needed — Virtuoso handles it cleanly on its own.

## Test plan

- [x] `bun run test` passes (new: `annotateAuthorGroups` boundary matrix — 10 cases covering author change, 5-min window, membership/command/session breaks, deletion heads, editing force-head, companion_response grouping)
- [x] Backend: `threadSummary` populated for 0 / 1 / N replies, participant stack caps at 3, raw markdown on wire
- [x] Manual — Comfortable density visibly fits more messages per screen
- [x] Manual — pagination scroll-up smooth across the 80-msg same-author seam (no jitter)
- [x] Manual — ThreadCard preview strips markdown correctly (INV-60)
- [x] Manual — reactions show gold "you reacted" state and are tappable on mobile
- [x] Manual — hover toolbar appears on heads AND continuations, hidden on mobile
- [ ] Staging — end-to-end verification on `app.threa.io` once staging label lands


kristofferremback:
### Code review — superseded

> Superseded by a more recent review. See the latest unified-review comment on this PR.

<details><summary>Previous review content (for history)</summary>

### Code review

Found 4 issues:

1. `apps/backend/src/features/messaging/event-service.ts:~325-332, ~436-441` / `apps/frontend/src/sync/stream-sync.ts` — **The real-time `message_updated` event for `updateType: "reply_count"` carries only `replyCount`; it never includes an updated `threadSummary`.** After a reply arrives live, the frontend patches `replyCount` correctly but `payload.threadSummary` stays frozen at its bootstrap value (often `undefined` if the message had zero replies at bootstrap). `ThreadCard` then renders with stale/absent preview + participants until the stream is re-bootstrapped. The outbox emission must also compute and attach an updated `ThreadSummary` so the sync handler can apply it alongside `replyCount`. (reactivity)
   https://github.com/kristofferremback/threa/blob/f85b355e639acb06f816aed3a07efe345303c856/apps/backend/src/features/messaging/event-service.ts#L320-L340

2. `apps/frontend/src/components/timeline/message-event.tsx:89-94` — **INV-25 violation.** The JSDoc on `groupContinuation` in `MessageEventProps` says *"Rendering detail lands in a follow-up commit — currently unused so the timeline keeps its existing layout while the grouping pipeline is validated."* The prop is in fact used in this same PR (wired through `SentMessageEvent → MessageLayout → renderAsContinuation`). The comment narrates a prior commit-sequence state, not the current implementation. INV-25: *"Comments explain why the current implementation works; they do not narrate the change history."* (claude-md)
   https://github.com/kristofferremback/threa/blob/f85b355e639acb06f816aed3a07efe345303c856/apps/frontend/src/components/timeline/message-event.tsx#L87-L96

3. `apps/backend/src/features/streams/repository.ts:~810-815` — **`ARRAY_AGG(author_id)` in the `participants` CTE has no `ORDER BY`.** PostgreSQL does not guarantee order for unordered aggregates, so the three participant IDs surfaced in the avatar stack are non-deterministic — the order can vary across page loads or parallel bootstraps for the same thread. Add `ORDER BY` (e.g. `ORDER BY first_reply_at` via a join, or at minimum `ORDER BY author_id`) for stable output. (performance)
   https://github.com/kristofferremback/threa/blob/f85b355e639acb06f816aed3a07efe345303c856/apps/backend/src/features/streams/repository.ts#L800-L820

4. `apps/frontend/src/components/timeline/event-list.test.ts` / `apps/backend/src/features/streams/` — **Plan verification section skipped.** The plan's commit #8 calls out unit tests for `ThreadCard` (with/without summary, 1-reply vs N-replies noun, null when no replies), `ActivityPill` (visibility + pulse), `MessageReactions` (reacted/unreacted/empty; ≥26px tap target), plus backend coverage for `threadSummary` population at 0/1/N replies and ≥4 participants (stack caps at 3). Only the `annotateAuthorGroups` boundary matrix tests landed. (plan)

All 4 issues were addressed in subsequent commits on this PR.

</details>

kristofferremback:
Addressed all 4 issues from the code review in commits cc1cfe1b and 75e910da:

1. **Real-time `threadSummary` reactivity** — backend now recomputes the summary via a new single-parent `findThreadSummaryByParentMessage` query in the same transaction as the reply-count update, and attaches it to the `message:updated` outbox payload. Frontend `stream-sync.ts` patches `threadSummary` onto the parent message alongside `replyCount`. `null` is a meaningful value (last reply deleted) so the handler only skips the patch when the field is `undefined` (older server compatibility).
2. **Non-deterministic participant ordering** — `ARRAY_AGG` in the participants CTE now orders by first-reply sequence, so the three surfaced avatars are stable across loads and earliest participants appear first.
3. **INV-25 stale comment** — replaced the "follow-up commit" comment on `groupContinuation` with a description of current behavior (including the pending/failed/editing head-layout override).
4. **Plan-listed tests** — added ThreadCard, ActivityPill, MessageReactions unit tests (timeline/*.test.tsx) and a backend thread-summary integration test (0/1/N replies, ≥4 participants cap, deterministic ordering, raw markdown on wire, batch vs single-parent parity).

All new tests green: 5 ThreadCard, 4 ActivityPill, 6 MessageReactions, 8 thread-summary integration cases. 165/165 timeline tests pass.

kristofferremback:
### Known follow-up (not fixing here)

Edge case: if the socket disconnects between `agent_session:started` and `agent_session:completed` (or the corresponding `message_created` for Ariadne's reply), the ThreadSlot's "thinking" state can stick around until the page is refreshed. The live `progressBySession` map in `useAgentActivity` doesn't get reconciled against the refreshed events array on reconnect/resubscribe.

Likely fix (separate PR):
- On stream resubscribe, clear `progressBySession` entries for sessions whose trigger message is in the freshly-bootstrapped events and whose `agent_session:completed/failed/deleted` is now present in those events.
- Or more robust: re-derive `runningSessions` from the post-reconnect bootstrap and intersect `progressBySession` against it, dropping any that have since terminated.

This is related to INV-53 ("bootstrap must be invalidated on reconnect to close event gaps") but applied specifically to session lifecycle state, not just event presence.

kristofferremback:
### Code review
**Confidence: 5/7** — Good

Found 2 issues (both non-blocking):

1. `apps/backend/src/features/streams/service.ts:997-999` — **`StreamService.getThreadSummaryForMessage` is dead on arrival.** The service wrapper is defined but never called — `event-service.ts` reaches into `StreamRepository.findThreadSummaryByParentMessage` directly, bypassing the service method. Either route the outbox emission through the service wrapper for consistency with the rest of the stream service API, or delete the unused wrapper. (CLAUDE.md INV-38 [Code hygiene and maneuverability]: *"Delete dead code immediately"*)
   https://github.com/kristofferremback/threa/blob/dc4f73d5a2a595d982df648bde0716bceb25e173/apps/backend/src/features/streams/service.ts#L996-L1000

2. `apps/backend/src/features/streams/repository.ts` — **`findThreadSummaryByParentMessage` duplicates all four CTEs from `findThreadSummaries`.** The batch query uses `DISTINCT ON` + a join on `parent_stream_id`; the single-parent query uses `ORDER BY … LIMIT 1` + a filter on `parent_message_id`. Otherwise the CTEs, column aliases, row shape, and `ThreadSummary` construction logic are identical. When the `ThreadSummary` wire type gains a field, both queries must be updated in sync or silently drift — and no test currently asserts the two return identical shapes for a parent that's visible through either path. Low immediate risk (thin slice of the surface) but worth either a shared helper (e.g. a `buildThreadSummary` row mapper) or a parity test. (abstraction)
   https://github.com/kristofferremback/threa/blob/dc4f73d5a2a595d982df648bde0716bceb25e173/apps/backend/src/features/streams/repository.ts#L780-L860

---
All plan-listed tests are present: `annotateAuthorGroups` boundary matrix, ThreadCard, ActivityPill (→ ThreadSlot after refactor), MessageReactions, backend threadSummary integration (0/1/N replies, ≥4 participants cap, deterministic ordering, raw markdown on wire). The ThreadSlot unification goes beyond the plan but is a clear improvement that was discussed on-thread.
</details>
Re-checked the reactivity fix, `annotateAuthorGroups` edge cases, `ThreadSlot` useRef/useEffect gating, pending-message continuation handoff. No high-confidence bugs found.
</details>
INV-38 dead code (see item 1 above). INV-25 stale comment, INV-56/INV-20 set-based race-safe updates, INV-60 preview stripping, INV-47 no nested ternaries — all clean.
</details>
Duplicate SQL between batch and single-parent thread-summary queries (see item 2 above). ThreadSlot architecture with `ownsLeftLine` prop is a clean shared-ownership pattern; `grid-template-rows` animation has graceful degradation for older browsers.
</details>
All SQL uses parameterized `sql` tagged templates. Preview rendering routes raw markdown through `truncateContent` → `stripMarkdownToInline` per INV-60. No `dangerouslySetInnerHTML`, no reflection of user input into query parameters.
</details>

Note: the 4 issues from the previous review are all addressed in the current diff. A known follow-up (reconnect-gap for active sessions) is tracked in a separate comment below and is explicitly deferred.

kristofferremback:
Two follow-up findings from the latest review pass:

1. `apps/backend/src/features/messaging/event-service.ts:384-389` — latest-reply previews stay stale after thread reply edits. The new thread card renders `threadSummary.latestReply`, but the edit path only emits `message:edited` inside the thread. We never recompute/send a parent-stream `message:updated` with a refreshed `threadSummary`, so editing the latest reply leaves the parent card showing the old snippet until the next bootstrap or reload.

2. `apps/backend/src/features/streams/repository.ts:785-794` — bootstrap can resurrect deleted-only threads. This query still derives `replyCount` from historical `message_created` events. `findThreadSummaries()` excludes deleted replies, so after deleting the last reply a fresh bootstrap can return `replyCount > 0` with no summary. Because the new thread card renders whenever `replyCount` is non-zero, a reload will show a dangling reply card for an otherwise empty thread.

kristofferremback:
Follow-up on the two findings from https://github.com/kristofferremback/threa/pull/390#issuecomment-4279784338:

Accepted and fixed in `e63016b9`.

- **Latest-reply preview staleness after thread edits**: `EventService.editMessage()` now emits the same parent-stream `message:updated` refresh used by thread create/delete, via a shared helper that recomputes `threadSummary` and republishes the parent message’s thread metadata.
- **Deleted-only threads reappearing after bootstrap**: `findThreadsWithReplyCounts()` now counts live rows from `messages` (`deleted_at IS NULL`) instead of historical `message_created` events, so bootstrap `replyCount` stays aligned with `findThreadSummaries()`.
- **Regression coverage**: added integration tests for both cases in `apps/backend/tests/integration/thread-summary.test.ts`.

Verified locally:
- `bun test ./apps/backend/tests/integration/thread-summary.test.ts`
- `bun test ./apps/backend/tests/integration/event-sourcing.test.ts`

`bun test ./apps/backend/tests/e2e/threads.test.ts` still needs a local backend on `localhost:3001`; it fails immediately with connection refused in this shell.

PR #399, opened 2026-04-21
feat: /invite slash command with user and bot support

kristofferremback:
## Problem

There was no way to invite users or bots to channels via the chat input. The only way to add members was through the stream settings UI. Additionally, bot mentions were not properly scoped — all workspace bots appeared in the @mention autocomplete regardless of whether they were actually members of the stream.

## Solution

Add a `/invite` slash command that supports inviting both users and bots, with proper frontend autocomplete filtering and reactive member lists. Also wire up `member_left` timeline events so removals appear in the stream.

### How it works

**`/invite` command flow:**
1. User types `/invite @slug` in a channel or channel-rooted thread
2. The frontend filters the @mention autocomplete to show only non-members (users + bots)
3. On dispatch, the backend resolves slugs to users and bots, validates permissions, and adds them
4. A `member_added` timeline event is created and rendered via the existing `MembershipEvent` component

**Bot mention filtering (normal chat):**
1. The stream bootstrap now includes `botMemberIds` — bots granted access via `bot_channel_access`
2. In normal chat mode, only bots in `botMemberIds` appear in @mentions
3. In threads, bot membership is checked against the **root stream** (channel/DM/scratchpad)
4. In `/invite` mode, all non-member bots are shown (admin/owner only)

**Reactive updates:**
1. Socket `stream:member_added` adds new members/bots to the bootstrap cache
2. Socket `stream:member_removed` removes members/bots from the bootstrap cache
3. The mention dropdown updates live without refresh

### Key design decisions

**1. Bot membership uses a separate table from human members**

Bots use `bot_channel_access` while humans use `stream_members`. We surface bot IDs in a new `botMemberIds` field on `StreamBootstrap` so the frontend has a single source of truth for filtering mentions. We considered joining bot data into the `members` array, but that would require extending `StreamMember` with bot-specific fields and complicating the membership model.

**2. `member_added` events have `actorType: "bot"` for bot invites**

The existing `MembershipEvent` component already supported `actorType: "bot"` through `useActors`. By setting `actorType: "bot"` on the event, bot invites render "Botty was added to the conversation" just like human invites, with correct names and avatars.

**3. Command events broadcast to the stream room, not the author room**

[...]

PR #403, opened 2026-04-22
feat(agents): ContextBag primitive + Discuss with Ariadne

kristofferremback:
## Summary

Introduces a reusable **ContextBag primitive** — a typed collection of context refs attached to a stream and resolved on every AI turn — and wires up its first consumer: **"Discuss with Ariadne"** from a message's context menu and the `/discuss-with-ariadne` slash command.

Starting a side-conversation with Ariadne about an existing thread used to mean copy-pasting content into a scratchpad or @mentioning her inline (which pollutes the source thread). The new flow: right-click any message → "Discuss with Ariadne" → private scratchpad opens with the thread pre-loaded and Ariadne already oriented.

## Architecture

Four layers, each independently extensible:

1. **Primitive** — `ContextBag` types + two new tables: `stream_context_attachments` and `context_summaries`
2. **Resolver registry** — one resolver per `ref.kind` (first: `thread`); returns `{ items, inputs, fingerprint, tailMessageId }`
3. **Intent config** — intent drives system preamble + per-kind summarize-vs-inline strategy (first: `discuss-thread`)
4. **Entry points + pipeline** — message-menu action + slash command → server-side scratchpad create with bag → outbox `stream:created` → new `ContextBagOrientationHandler` dispatches a `CONTEXT_BAG_ORIENT` job → orientation worker wraps the AI call in `withCompanionSession` so the activity indicator + trace dialog populate live, then posts Ariadne's kickoff message

### Key properties

- **Shared, access-gated summary cache** — `context_summaries` is keyed by `(workspace, ref_kind, ref_key, fingerprint)`. Fingerprint hashes the explicit `inputs` manifest (messageId + contentFingerprint + editedAt + deleted), so any drift flips the key → cache miss → fresh summary. No silent-drift failure mode. Access is re-verified on read (INV-8).
- **Live-follow semantics** — on every turn the bag is re-resolved; a one-pass diff against the previous `last_rendered` snapshot surfaces appends, edits and deletes.
- **Cache-friendly rendering** — output splits into a stable (append-only) region + a volatile "since last turn" delta. Previously-rendered content is never mutated in the stable region; edits flow through the delta with a "previous fingerprint" breadcrumb. Lets a future `cache_control` step light up Anthropic prompt-cache reuse without rendering changes.
- **Intent-driven prompts** — `discuss-thread` intent ships a neutral orientation template; summaries via `openrouter:openai/gpt-5.4-nano` above an 8k-char threshold, inline below.
- **Thread root included** — when the source is a thread, the parent (root) message is prepended to the fetched replies so Ariadne never loses the message that spawned the conversation. Matches the `buildThreadContext` pattern.
- **Auto-naming kicks in** — the scratchpad is created with `displayName: null`, so the `NamingHandler` generates a per-thread title from Ariadne's orientation message (instead of every Discuss-with-Ariadne scratchpad showing up as the same generic label).
- **Slash command gated on send** — `/discuss-with-ariadne` inserts a chip into the composer like any other slash command and fires on send; does not execute immediately when picked from autocomplete.

## Invariants honoured

- **INV-1** no FKs · **INV-2** prefixed ULIDs (`sca_`, `cs_`) · **INV-3** `intent`/`ref_kind` are TEXT, not enums · **INV-7** bag insert + `stream:created` in the same transaction · **INV-8** both tables workspace-scoped · **INV-17** append-only migration · **INV-19** `createAI` calls carry telemetry (`functionId: "context-bag.summarize"`, `"context-bag.orient"`) · **INV-20** `SummaryRepository.upsert` uses `ON CONFLICT DO NOTHING` · **INV-28** all AI through `createAI` · **INV-30** repos take `Querier` first · **INV-31** `AuthorType` imported from `@threa/types` · **INV-41** DB connections released before AI calls · **INV-44** intent config + summarizer model colocated · **INV-51/52** feature colocation + barrel exports · **INV-55** Zod on the route · **INV-57** bag lives in its own tracking table, not on `streams`

## Scope deviations from the plan

One deliberate v1 deviation: `buildStreamContext` integration for subsequent user turns was **deferred**. The orientation turn gives Ariadne enough context for the user's first follow-up, and Ariadne's orientation response is in conversation history for later turns. What's missing v1 is live-follow awareness on *subsequent* turns (edits/deletes on the source thread after the orientation don't flow into later turns' prompts yet). The primitive is shaped so adding this is a focused follow-up — the bag persists, the resolver works, rendering is ready; we just need to call `resolveBagForStream` inside `buildAgentContext` and append `resolvedBag.stable`/`.delta` to the system prompt.

Also deferred (per plan): `cache_control` markers in `createAI`, additional ref kinds (memo, message, stream, attachment), additional intents, a "refresh context" UI chip, and the "absorb delta into summary" maintenance job.

## Suggested follow-ups

These surfaced during implementation + review but are deliberately kept out of v1 to keep this PR focused. Each is small and well-scoped:

- **`MessageRepository.listConversation(db, streamId)`** — the central "thread root gets forgotten" footgun keeps recurring because `MessageRepository.list(threadStreamId)` naturally excludes the parent message (which lives in the parent stream). Every context-building path has to remember to do a second `findById(parentMessageId)` and prepend it, and every new path that forgets is a bug. Consolidate into a single canonical read method that returns `[parent?, ...streamMessages]` when the stream is a thread. Migrate `buildThreadContext` + `ThreadResolver.prependParentMessage` to use it. Optional: add a `no-restricted-imports` lint rule that steers agent/context callers away from the lower-level `list`. Retires this class of bug without the dual-ownership refactor the data model would otherwise imply.
- **`withCompanionSession` → `withAgentSession`** — the function is now a shared agent-session primitive used by both the persona-agent and the orientation worker; the current name misleads future callers. Mechanical rename, best in its own PR so the churn is contained.
- **Client-action command routing on `clientActionId` (not `name`)** — today `message-input.tsx` routes client-actions by command `name`, which works because `name === clientActionId === "discuss-with-ariadne"`. When a second client-action command lands with a different `clientActionId`, this will silently fall through to the server dispatch. Add `clientActionId` as a second attribute on the `slashCommand` node before the second such command is introduced.
- **`prependThreadRoot` as a shared helper** — the parent-prepend logic is now inlined in two places (`buildThreadContext` + `ThreadResolver.prependParentMessage`). Subsumed by the `listConversation` follow-up above.
- **Soft-delete filter on prepended parent** — narrow follow-up flagged in review: `MessageRepository.findById` returns soft-deleted messages; the orientation bag should skip them.

## Test plan

- [x] 30+ new unit tests covering diff detection, render stability, fingerprint invariants, thread resolver (access, fingerprint, anchors, root prepend), summary cache race-safe upsert, orientation dispatch (6 paths), discuss-thread intent preamble + inline/summary branching, `extractCommandName`.
- [x] Full monorepo typecheck (`bun run typecheck`) — all 9 workspaces pass
- [x] Lint (`bun run lint`) — all pass, including feature-barrel enforcement (INV-52)
- [x] `bun run test:unit` — 971 backend unit tests pass
- [x] `bun run test` (frontend) — 1353 frontend tests pass
- [ ] Manual: open a long thread, trigger "Discuss with Ariadne", confirm the activity indicator appears immediately, the trace dialog populates live, and Ariadne's orientation message posts without leaking `[msg_*]` ids to the user.
- [ ] Manual: trigger Discuss-with-Ariadne on a thread root message; confirm the parent message is part of the context (Ariadne's orientation can reference it).
- [ ] Manual: type `/discuss-with-ariadne` in the composer; confirm a chip inserts and nothing fires until send.
- [ ] Manual: after orientation, check the sidebar — the scratchpad should be renamed from "New scratchpad" to a per-thread title within ~10 seconds.
- [ ] Manual: type a follow-up to Ariadne *while* her orientation is in flight; confirm the message is answered (not dropped by the session-slot collision that was fixed in `b16d645`).

https://claude.ai/code/session_01XkS3K9fikTd7C4CtRSXKsc

---
_Generated by [Claude Code](https://claude.ai/code/session_01XkS3K9fikTd7C4CtRSXKsc)_


kristofferremback:
### Code review

Found 5 issues:

1. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:232-244` — INV-7 violation + orientation worker is not idempotent. `createMessage` (own `withTransaction`) and `persistSnapshot` are two independent writes. If the process crashes between them (or `persistSnapshot` fails), the queue retries the whole job (maxRetries=5) and posts a duplicate Ariadne kickoff message; on the inverse failure mode, the snapshot is never written and the next turn re-emits the entire thread as "new" in the delta. All three reviewers independently flagged this, and `persistSnapshot`'s own JSDoc says it must be called "in the same transaction as the agent's reply write (INV-7)". Fix: either wrap both in one transaction (pass a `PoolClient`-accepting createMessage path), or guard the worker with `if (bag.lastRendered !== null) return` so retries are no-ops. (claude-md [INV-7, "Architecture Boundaries and Dependency Flow"]: "Event-source updates and read projections are committed together")
   https://github.com/kristofferremback/threa/blob/e7498ffa037fe91ea1c067683267df97315116ea/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L230-L246

2. `apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts:19, 95-104` — `applyAnchors` silently falls back to the full (500-message) window when either `fromMessageId` or `toMessageId` isn't found in the fetch. `MessageRepository.list` pulls the 500 *most recent* messages descending, so any anchor pointing into older history returns -1 from `findIndex`, triggers `return messages` unchanged, and the caller gets the full recent window — not the scoped slice they asked for. v1 doesn't exercise anchored bags in the UI, but the route schema accepts `fromMessageId`/`toMessageId` and it's a footgun for the next consumer. Either throw on anchor-miss, or extend the repository to fetch deterministically by anchor range. (bug: silent scope inversion + unbounded `MAX_FETCH=500` magic number)
   https://github.com/kristofferremback/threa/blob/e7498ffa037fe91ea1c067683267df97315116ea/apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts#L93-L106

3. `apps/backend/src/features/agents/context-bag/registry.ts:7`, `resolve.ts:125-150`, `streams/handlers.ts:338-351` — The `Resolver<TRef extends ContextRef>` abstraction leaks because the registry stores `Resolver` (erased base), forcing `ThreadResolver as Resolver` at registration and `ref.kind as typeof ContextRefKinds.THREAD` at three call sites in the route handler. `loadOrCreateSummary` further widens `refKind: string` and then casts back to `"thread"` twice. After Zod parsing `ref.kind` is already the string literal, so the casts are redundant; more importantly the pattern will break silently once a second ref kind lands, because `getResolver` returns a base `Resolver` whose `fetch(db, ref)` internally assumes `ThreadRef`. Fix: type `getResolver` with an overloaded signature keyed on `ContextRefKind`, narrow `loadOrCreateSummary`'s `refKind` param to `ContextRefKind`, and drop the manual re-assembly in the handler — pass the Zod-parsed ref directly. (abstraction: leaky kind erasure)
   https://github.com/kristofferremback/threa/blob/e7498ffa037fe91ea1c067683267df97315116ea/apps/backend/src/features/agents/context-bag/registry.ts#L1-L23

4. `apps/backend/src/features/agents/context-bag/` — Plan section "Verification / Unit tests" called for four test files that are absent: `summary-repository.test.ts` (access-gated read + race-safe upsert), `thread-resolver.test.ts` (fingerprint stability), `intents/discuss-thread.test.ts` (inline-vs-summarize threshold), and a dispatch test for `ContextBagOrientationHandler` asserting a `stream:created` event with a matching bag queues a `CONTEXT_BAG_ORIENT` job and one without a bag is skipped. The three render/diff/fingerprint tests that ship are solid but leave the access-gating and dispatch paths uncovered. (plan)

5. `apps/frontend/src/components/editor/triggers/use-command-suggestion.tsx:22-55` — The filter path now mixes two mechanisms: server-driven `clientActionId` (this PR) alongside a hard-coded client-side `isInviteAllowed(streamId, streams)` stream-type check. PR #399 established that server metadata (`WorkspaceBootstrap.commands`) is the source of truth for what shows up in the slash list; this PR's new `clientActionId` path respects that, but the `invite` filter duplicates stream-type knowledge (`StreamTypes.CHANNEL`, `stream.rootStreamId`) in the frontend. Consider extending `CommandInfo` with a `when`-style predicate hint (e.g. `supportedStreamTypes?: StreamType[]`) so both `invite` and future contextual commands go through one mechanism. Not blocking; flagging because the parallel filtering is new in this PR. (historical)
   https://github.com/kristofferremback/threa/blob/e7498ffa037fe91ea1c067683267df97315116ea/apps/frontend/src/components/editor/triggers/use-command-suggestion.tsx#L22-L58

---

<details><summary>📐 Plan Adherence [1 gap]</summary>

Four of the six unit-test files called out in the plan's Verification section are missing: `summary-repository.test.ts`, `thread-resolver.test.ts`, `intents/discuss-thread.test.ts`, and a dispatch test for the new `ContextBagOrientationHandler`. The diff/render/fingerprint tests that ship are good, but the access-gated cache read and the orientation dispatch path (both highlighted in the plan) have no coverage.

The `buildStreamContext` integration gap is documented in the PR body as a deliberate v1 deviation and is not flagged here.
</details>

**Non-atomic orientation → duplicate messages on retry** (see #1 above). High-confidence correctness bug.

**`applyAnchors` silent fallback** (see #2 above). v1 callers don't set anchors so this doesn't fire today, but the fallback turns a narrowing hint into an unbounded fetch, and the route schema already accepts the anchors.
</details>

**INV-7** — "Event-source updates and read projections are committed together" — violated by `context-bag-orientation-handler.ts:232-244`. The `createMessage` call opens its own `withTransaction`; `persistSnapshot` is a subsequent standalone UPDATE. `persistSnapshot`'s own JSDoc acknowledges the constraint: _"Callers should invoke this in the same transaction as the agent's reply write (INV-7)."_ See issue #1 for the fix.
</details>

**Leaky `Resolver` kind abstraction** (see #3 above). The registry stores the erased base `Resolver`, which forces casts at registration (`ThreadResolver as Resolver`), at summary cache callers (`refKind: string` then `as "thread"` twice), and at route-level ref assembly (`as typeof ContextRefKinds.THREAD` × 3). Each cast is individually small; together they hide that `getResolver(kind)` returns something whose `fetch(db, ref)` is kind-specific. Narrowing with an overloaded `getResolver` signature + narrower param types eliminates all five casts and makes adding a second ref kind structurally safe.
</details>

No high-confidence vulnerabilities found. Reviewed: access gating on bag creation (caller's access to every ref is verified via `resolver.assertAccess` in `streams/handlers.ts` before the bag is persisted), shared summary cache (access is re-verified on read — cache hit without access behaves as miss), Ariadne persona resolution server-side via slug (no client-sent persona id).
</details>

---

kristofferremback:
Thanks for the review. Addressed all five in `6c29498`:

1. **INV-7 + duplicate-retry** (`context-bag-orientation-handler.ts`) — added an early return when `bag.lastRendered !== null` so queue retries are no-ops, and threaded `clientMessageId: orient:<bagId>` through the `createMessage` adapter so the message insert itself is idempotent via the existing `ON CONFLICT DO NOTHING` path in `event-service`. Either side of a crash between message-write and snapshot-write is now safe: the retry's guard fires once the snapshot is set, and the message insert dedupes either way.

2. **`applyAnchors` silent fallback** (`thread-resolver.ts`) — now throws `HttpError("CONTEXT_ANCHOR_NOT_FOUND", 422)` when `fromMessageId`/`toMessageId` can't be located in the fetched window, instead of silently returning the full list.

3. **`Resolver` kind abstraction** (`registry.ts`, `resolve.ts`, `streams/handlers.ts`) — `getResolver<K extends ContextRefKind>` now returns `Resolver<ContextRefByKind[K]>` via a mapped type. Zod enums for `intent` and `kind` carry their concrete unions (`ContextIntent`/`ContextRefKind`) instead of `string`, which let me drop all five redundant `as` casts: the handler now passes the Zod-parsed `contextBag` straight through to the service, and `loadOrCreateSummary` takes `refKind: ContextRefKind` directly.

4. **Missing planned tests added** — 25 new unit tests (966 total backend unit tests, was 941):
   - `context-bag-orientation-handler.test.ts` — 6 tests (dispatch paths: bag present, no bag, non-scratchpad, companion off, no persona, unrelated event)
   - `context-bag/resolvers/thread-resolver.test.ts` — 7 tests (access gating: public, private, root-inherited, cross-workspace; fingerprint stability; anchor-miss throw)
   - `context-bag/summary-repository.test.ts` — 5 tests (cache hit/miss + race-safe upsert fallback)
   - `context-bag/intents/discuss-thread.test.ts` — 6 tests (intent config + renderer inline-vs-summary branches)

5. **Invite client-side stream-type filter** — left as-is for this PR. Agreed the duplication with `clientActionId` is mildly awkward but consolidating it is a broader refactor (extending `CommandInfo` with a `when`-style predicate) that touches the `/invite` path too. Opened as a follow-up.

Full monorepo typecheck + lint + 966 backend unit tests + 1350 frontend tests all green.

---

kristofferremback:
### Code review (re-review after `6c29498`)
**Confidence: 5/7** — Good (few non-blocking issues)

_Supersedes [the earlier unified-review comment](https://github.com/kristofferremback/threa/pull/403#issuecomment-4299350784)._

### Status of prior review (all 5 items)

| Prior issue | Status |
| --- | --- |
| INV-7 + orientation non-idempotency | ✅ **Resolved.** `bag.lastRendered` guard + `clientMessageId: orient:<bagId>` for ON CONFLICT dedup. The two writes are still separate transactions, but INV-41 blocks single-txn atomicity (can't hold a connection across the AI call), so compensating controls are the right shape here. One ergonomic nit in the implementation — see issue 1 below. |
| `applyAnchors` silent fallback | ✅ **Resolved.** Now throws `HttpError(422, "CONTEXT_ANCHOR_NOT_FOUND")`. One follow-on clarity issue — see issue 4. |
| `Resolver` kind erasure (5 casts) | ✅ **Resolved.** `getResolver<K>` returns `Resolver<ContextRefByKind[K]>` via mapped type; all 5 casts gone; type generalizes cleanly when a second kind lands. |
| Missing planned tests | ✅ **Resolved.** 4 new test files covering dispatch, access gating, fingerprint stability, cache miss/upsert race, intent config. 25 new tests. |
| Invite FE filter duplication | ⏭️ Acknowledged, deferred to follow-up. |

### New findings (4 non-blocking)

1. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:260-262` — The comment above the `createMessage` call inverts the actual safety model. It reads _"the retry's guard fires only once snapshot succeeds, and the message insert is a no-op either way"_, but the guard actually does NOT fire in the crash window (between `createMessage` success and `persistSnapshot` failure) — `lastRendered` is still null, so the retry re-runs the AI and relies purely on `clientMessageId` dedup to avoid a duplicate post. The two mechanisms cover different failure modes: the guard avoids a wasted AI call when the whole job succeeded before, the clientMessageId dedupes the crash-window case. End-state is correct, but the comment is load-bearing for future maintainers who might e.g. remove `clientMessageId` thinking the guard covers it. Rewriting to name each mechanism's role would prevent a subtle regression. (bug: misleading safety invariant comment)
   https://github.com/kristofferremback/threa/blob/6c294988b6c6355c372821fff5caa844154d1e8f/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L255-L275

2. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:189,215` — `ContextBagRepository.findByStream` runs twice per invocation: once for the idempotency guard at L189, once again inside `resolveBagForStream` at `resolve.ts:46`. Single indexed PK lookup so cost is bounded, but could be avoided by having `resolveBagForStream` return the loaded bag (or `bag.lastRendered`) alongside `ResolvedBag` so the caller makes the idempotency decision with one fewer round-trip. (design: redundant DB lookup)
   https://github.com/kristofferremback/threa/blob/6c294988b6c6355c372821fff5caa844154d1e8f/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L184-L220

3. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:153` — `ContextBagOrientationWorkerDeps.createMessage.authorType` is declared as an inline string literal union `"user" | "persona" | "system" | "bot"` instead of importing `AuthorType` from `@threa/types`. `PersonaAgentDeps.createMessage` at `persona-agent.ts:68` uses the imported type. If `AUTHOR_TYPES` gains a member the orientation interface silently diverges. (INV-31, Quick Invariant Lookup "API and backend contracts": _"Type derivation should flow from schemas/constants to inferred types to avoid drift"_)
   https://github.com/kristofferremback/threa/blob/6c294988b6c6355c372821fff5caa844154d1e8f/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L146-L163

4. `apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts:105-117` — The `CONTEXT_ANCHOR_NOT_FOUND` 422 is raised whenever `findIndex` misses. Because the fetch cap is `MAX_FETCH = 500` most-recent messages, an anchor that genuinely exists in the DB but precedes that window triggers the same error with the same code and message. The inline comment _"usually means the anchor predates the MAX_FETCH tail"_ is accurate; the user-visible error message `"{from|to}MessageId anchor not found in the fetched window"` at least hints at this, but there's no distinguishable code a caller could switch on to backfill the window vs reject the reference. v1 doesn't surface anchors in the UI yet; worth a note or a distinct code (e.g. `CONTEXT_ANCHOR_OUT_OF_WINDOW`) when the anchor exists but predates the fetch. (correctness: unactionable error semantics)
   https://github.com/kristofferremback/threa/blob/6c294988b6c6355c372821fff5caa844154d1e8f/apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts#L96-L121

---

All four previously-missing test files now exist: `summary-repository.test.ts`, `thread-resolver.test.ts`, `intents/discuss-thread.test.ts`, `context-bag-orientation-handler.test.ts`. Coverage is sufficient for the plan's Verification section. The `buildStreamContext` integration gap remains deliberately deferred per the PR body.
</details>

<details><summary>🔍 Bugs [1 clarity issue]</summary>

**Misleading comment on orientation worker's crash-window safety** (see #1). No correctness bug — the code is correct — but the comment will mislead the next maintainer about which mechanism covers which failure mode.
</details>

<details><summary>📋 CLAUDE.md Compliance [1 drift risk]</summary>

**INV-31 drift** — orientation worker's `authorType` inline union bypasses the shared `AuthorType` type (see #3). Prior INV-7 concern is resolved in spirit by the guard + clientMessageId compensating-controls approach; the aspirational JSDoc on `persistSnapshot` (_"Callers should invoke this in the same transaction as the agent's reply write"_) is now stale and should be updated to reflect the idempotency-key design, but that's a documentation refresh, not a violation.
</details>

<details><summary>🏗️ Design [2 minor concerns]</summary>

- Redundant `findByStream` call (see #2). Bounded-cost but avoidable with an API tweak to `resolveBagForStream`.
- `CONTEXT_ANCHOR_NOT_FOUND` conflates "doesn't exist" with "outside fetch window" (see #4). Matters once anchors are surfaced in the UI.
</details>

No high-confidence vulnerabilities found. `clientMessageId: orient:<bagId>` is per-stream (keyed on `(stream_id, client_message_id)`) and bagId is an unguessable ULID, so the new dedup key isn't a practical collision vector.
</details>

---

kristofferremback:
Pushed `6284e4e` addressing both the re-review and the UX feedback.

**User-reported issues:**

- **No activity indicator while orientation runs.** Fixed by wrapping the orientation AI call in `withCompanionSession` (same primitive persona-agent uses). The worker now emits `agent_session:started` → the activity indicator appears the moment the user lands in the new scratchpad, `SessionTrace.startStep` emits `context_received` / `thinking` / `message_sent` live so the trace dialog populates in real-time, and `agent_session:completed` clears the indicator when done. This also flips session idempotency to the primary guard: a retry with the same bag's synthetic `triggerMessageId: orient:<bagId>` sees a COMPLETED session and returns `skipped` with no second AI call.

- **Raw `[msg_…]` ids leaking into Ariadne's orientation message.** Fixed in the `discuss-thread` intent:
  - System preamble now explicitly frames `[msg_…]` tags as internal-only grounding and forbids echoing them to the user.
  - Orientation user-prompt rewritten from "structured headers + citations" to "natural, conversational three-or-four-sentence note — no headers, no message ids, refer to people by name". Added a regression test on the preamble asserting the id-ban instruction is present.

**Re-review items (4 of 4):**

1. `orientation-handler.ts:260-262` misleading safety comment → rewritten. The session skip, `clientMessageId` dedup, and idempotent `persistSnapshot` are now named separately with what each covers.
2. Double `findByStream` → `resolveBagForStream` now takes `{ skipIfAlreadyRendered: true }` and short-circuits inside its existing single transaction; worker makes one bag load total.
3. Inline `authorType` union (INV-31) → imports `AuthorType` from `@threa/types`.
4. `CONTEXT_ANCHOR_NOT_FOUND` conflation → split into `CONTEXT_ANCHOR_NOT_FOUND` (id doesn't exist in this stream) vs `CONTEXT_ANCHOR_OUT_OF_WINDOW` (id exists but predates `MAX_FETCH`). Two new tests cover both branches; one asserts cross-stream ids stay `NOT_FOUND`.

969 backend unit tests pass; full monorepo typecheck + lint green.

---

kristofferremback:
### Code review (re-review after `6284e4e`)
**Confidence: 3/7** — Needs Work

_Supersedes [the prior unified-review](https://github.com/kristofferremback/threa/pull/403#issuecomment-4302323486)._

The session+trace rewrite delivers the activity-indicator UX cleanly, and the four prior items are properly addressed — but it introduces **two real regressions** worth fixing before merge: a concurrent-session-slot collision that drops the user's first message during orientation, and a trace-step schema mismatch that leaves the trace dialog visually broken for orientation sessions (the very feature this commit added).

### Top issues

1. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:228-331` — **Concurrent session-slot collision drops the user's first message during orientation.** `withCompanionSession` holds the partial unique slot `(stream_id) WHERE status = 'running'` for the entire orientation AI call (potentially 30s+ on large bags). If the user types their first message while orientation is in flight, `CompanionHandler` sees an active session and skips dispatching the persona-agent job (`companion-outbox-handler.ts:176-188`: _"Session already active for stream, new message will be handled in-flight"_). But the orientation worker is a one-shot — it doesn't poll for new messages the way `persona-agent` does via `newMessages.check`, and it doesn't call `checkForUnseenMessages` on completion. The result: the user's first message arrives during the orientation window, gets marked seen, and is **silently dropped on the floor** — never replied to. Repro: open a bag-attached scratchpad, immediately start typing — Ariadne's orientation arrives and the user's first message is ignored. Fix options: (a) call `checkForUnseenMessages` (or an equivalent) at the end of the orientation work block so the post-orientation cursor catches up, or (b) skip `withCompanionSession` for orientation and emit just the trace lifecycle directly so we don't take the lock. (a) is the smaller change. (bug, 88)
   https://github.com/kristofferremback/threa/blob/6284e4e1ad4848bc16ceb4e7ddd83818bd0c5fe3/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L210-L335

2. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:261-270` — **`context_received` trace-step content schema doesn't match the frontend renderer.** The orientation worker stores `{ intent, stableChars, deltaChars }` as the step's content; the trace-step renderer (`apps/frontend/src/components/agent-trace/trace-step.tsx:228`) only branches on `"messages" in structured` and falls through to a `"Processing messages..."` placeholder otherwise. Net effect: the trace dialog for every orientation session permanently shows "Processing messages..." instead of any real context preview — the very UX this commit was supposed to fix. Either (a) add a renderer branch for `"intent" in structured`, or (b) emit a `{ messages: [...] }` shape from the worker (e.g. the resolved bag items). (bug, 88)
   https://github.com/kristofferremback/threa/blob/6284e4e1ad4848bc16ceb4e7ddd83818bd0c5fe3/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L260-L271

3. `apps/backend/src/features/agents/context-bag-orientation-handler.ts:235-258` — **Session record is permanently inconsistent in the rare crash-window where snapshot succeeded but session-completion failed.** Sequence: prior run posted the message AND wrote `lastRendered` AND failed before the completion txn → session is FAILED. Retry resumes, `resolveBagForStream` returns null because of the `skipIfAlreadyRendered` guard, work returns `{ messagesSent: 0, sentMessageIds: [] }`, `completeSession` records `response_message_id = null` and `sent_message_ids = []`. The actual orientation message is in the DB but unreferenced by its session row, so any analytics or UI joining sessions to messages misses it. Fix: when the early-exit branch fires, look up the existing message via the `clientMessageId: orient:<bagId>` key and return its id in `sentMessageIds`. (bug, 82)
   https://github.com/kristofferremback/threa/blob/6284e4e1ad4848bc16ceb4e7ddd83818bd0c5fe3/apps/backend/src/features/agents/context-bag-orientation-handler.ts#L235-L260

4. `apps/backend/src/features/agents/context-bag/resolve.ts:14-23` — **`ResolvedBag.nextSnapshot` JSDoc still claims atomicity that's no longer attempted.** The field comment reads _"Snapshot the caller must persist in the SAME transaction as the agent reply (INV-7)."_ The `persistSnapshot` JSDoc was correctly rewritten in this commit to acknowledge the compensating-controls design, but this type-level contract was missed. The orientation worker calls `persistSnapshot(pool, ...)` outside `createMessage`'s internal transaction, so the field's stated requirement is structurally violated. Update the JSDoc to match reality (or rename for clarity). (claude-md INV-7 — _"Event-source updates and read projections are committed together"_; documentation drift) (claude-md, 80)
   https://github.com/kristofferremback/threa/blob/6284e4e1ad4848bc16ceb4e7ddd83818bd0c5fe3/apps/backend/src/features/agents/context-bag/resolve.ts#L14-L23

5. `apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts:106-146` — `assertAnchorExists` does an extra `MessageRepository.findById` round-trip on every anchor-miss to disambiguate `CONTEXT_ANCHOR_NOT_FOUND` from `CONTEXT_ANCHOR_OUT_OF_WINDOW`. There's no frontend code today that branches on these two codes differently (no current consumer for the disambiguation). Adds DB cost on a failure path with no immediate payoff. Either defer the split until the frontend renders different recovery UX, or accept the cost as forward-looking — flag here so the trade-off is intentional. (design, 82)
   https://github.com/kristofferremback/threa/blob/6284e4e1ad4848bc16ceb4e7ddd83818bd0c5fe3/apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts#L106-L146

6. `apps/backend/src/features/agents/companion/session.ts:28` — `withCompanionSession` is now called from a non-companion-triggered worker (orientation). The function is purely a session-lifecycle primitive and contains nothing companion-specific; the name will mislead future maintainers. Renaming to `withAgentSession` is contained (one barrel re-export, two call sites). Low urgency, but worth flagging since the second call site landed in this commit. (naming, 80)
   https://github.com/kristofferremback/threa/blob/6284e4e1ad4848bc16ceb4e7ddd83818bd0c5fe3/apps/backend/src/features/agents/companion/session.ts#L28-L46

### Status of prior review

| Prior issue | Status |
| --- | --- |
| Misleading crash-window safety comment | ✅ Resolved — comment now names each mechanism (session COMPLETED guard / clientMessageId dedup / idempotent UPDATE) correctly |
| Double `findByStream` load | ✅ Resolved — `{ skipIfAlreadyRendered: true }` short-circuits inside the existing transaction |
| Inline `authorType` union (INV-31) | ✅ Resolved — imports `AuthorType` |
| `CONTEXT_ANCHOR_NOT_FOUND` conflation | ✅ Resolved (with the cost noted in #5) |

---

The PR body's only declared deviation (deferred `buildStreamContext` integration) still holds and is unrelated to the session+trace rewrite. The withCompanionSession addition is purely UX — subsequent user turns still go through `CompanionHandler` → `PersonaAgent` → `buildStreamContext`, which means the bag is still not visible on later turns. Documented and acknowledged.
</details>

<details><summary>🔍 Bugs [3 issues, 2 user-facing]</summary>

- **Concurrent session slot collision** drops the user's first message during orientation (#1). User-facing.
- **`context_received` trace step shape mismatch** breaks the trace dialog for the new orientation flow (#2). User-facing.
- **Session record inconsistency on rare crash-window retry** (#3). Edge case, not user-facing but breaks analytics joins.
</details>

<details><summary>📋 CLAUDE.md Compliance [1 documentation drift]</summary>

**INV-7** — Documentation drift on `ResolvedBag.nextSnapshot` (#4). The field still claims SAME-transaction semantics that the orientation worker can't honor. The compensating-controls design is fine; the JSDoc just needs to reflect it.
</details>

<details><summary>🏗️ Design [2 minor concerns]</summary>

- `assertAnchorExists` extra round-trip with no current FE consumer (#5).
- `withCompanionSession` naming now misleading at the new call site (#6).
</details>

No high-confidence vulnerabilities. The synthetic `triggerMessageId: orient:<bagId>` is an unguessable ULID-derived string and the column has no FK; access checks are unchanged from the previous review.
</details>

---

kristofferremback:
Pushed `b16d645` fixing both regressions plus the edge-case + JSDoc drift:

**1. Session-slot collision → user's first message getting dropped (#1).**
On orientation completion we now call `checkForUnseenMessages` — the same follow-up-dispatch helper `persona-agent-worker` uses for exactly this race. Exported it from `persona-agent-worker.ts` + the agents barrel so we don't fork the logic. `initialSequence` on the session is now captured from `StreamEventRepository.getLatestSequence` *before* taking the slot, so the follow-up check correctly picks up anything the user typed during the AI call.

**2. `context_received` trace-step schema mismatch → "Processing messages..." placeholder (#2).**
Orientation now emits a persona-agent-style `{ intent, messages: [...], totalMessageCount, truncated }` payload that the `trace-step.tsx` `"messages" in structured` branch actually renders. `ResolvedBag` gained an `items: RenderableMessage[]` field so the worker can grab them without re-running the resolvers. Preview capped at 15 items so large threads don't bloat the trace row.

**3. Session-record inconsistency on skip path (#3).**
On the `skipIfAlreadyRendered` branch inside an active (replayed) session — the rare case where a prior `persistSnapshot` committed but the session-completion txn failed — the worker now looks up the orientation message via `MessageRepository.findByClientMessageId(pool, streamId, "orient:<bagId>")` and returns its id in `sentMessageIds`, so the replayed session row preserves the link instead of recording `response_message_id = null` with an orphan message.

**4. Stale `ResolvedBag.nextSnapshot` JSDoc (#4).**
Rewritten to name the compensating-controls model (session skip + `clientMessageId` dedup + idempotent UPDATE) that's actually in force.

**Not addressed (calling out for transparency):**
- #5 `assertAnchorExists` extra round-trip — intentionally kept for forward FE differentiation. Flagged as a minor optimization; no current FE consumer means the cost is paid on the rare anchor-miss path only.
- #6 `withCompanionSession` naming — the function is now genuinely a shared agent-session primitive. Agreed the name is awkward; leaving as a follow-up rename PR so the churn is contained.

969 backend unit tests pass; full monorepo typecheck + lint green.

---

kristofferremback:
### Code review (re-review after `5837acb`)

_Supersedes [the prior unified-review](https://github.com/kristofferremback/threa/pull/403#issuecomment-4302660996)._

Two new commits address prior feedback cleanly (auto-naming, thread root inclusion, send-gated slash command). Most findings are minor follow-ups, but **one real correctness bug** landed with the thread-root fix: a soft-deleted parent is silently fed to the AI as live context.

### Top issues

1. `apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts:159-176` — **Soft-deleted parent message leaks into the AI bag.** `prependParentMessage` calls `MessageRepository.findById(db, stream.parentMessageId)`, which has no `deleted_at IS NULL` guard (`messaging/repository.ts:146-156`). A soft-deleted root retains its `content_markdown` until hard-delete runs, so when the user deletes the message that spawned a thread, the next Discuss-with-Ariadne invocation on that thread pulls the deleted text back in with `deleted: false` in the `SummaryInput` manifest. The helper's comment at line 155 even says _"parent has been hard-deleted"_ — but the guard only handles the null (hard-deleted) case. Fix: one-liner — `if (parent.deletedAt) return items` after the null check. (bug, 85)
   https://github.com/kristofferremback/threa/blob/5837acb931bf29686a46beef9519749ac1e35ae4/apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts#L155-L180

2. `apps/frontend/src/components/timeline/message-input.tsx:384-391` + `apps/frontend/src/hooks/use-discuss-with-ariadne.ts:30-34` — **Double error notification when the discussion mutation fails.** The hook toasts on catch and rethrows; the composer's `handleSubmit` catches the rethrow and sets the inline composer error with the same meaning. The user sees a sonner toast _and_ a persistent composer error banner for one failure. Fix: either remove the `toast.error` from the hook and let the caller be the single source (consistent with server-command error surfacing), or have the composer skip its `setError` for client-action rejections (the hook already handled UX). (bug, 82)
   https://github.com/kristofferremback/threa/blob/5837acb931bf29686a46beef9519749ac1e35ae4/apps/frontend/src/hooks/use-discuss-with-ariadne.ts#L26-L39

3. `apps/frontend/src/components/timeline/message-input.tsx:383` — **Client-action routing keys on `name`, not `clientActionId`.** The switch `commandName === DISCUSS_WITH_ARIADNE_COMMAND` works today because `name === clientActionId === "discuss-with-ariadne"`, but `CommandNodeAttrs` only persists `name` — `clientActionId` isn't on the node. If a second client-action command ever lands with `name !== clientActionId`, the composer will silently miss it and dispatch to the server. Add `clientActionId` as a second attribute on the `slashCommand` node (or rely solely on `name` and document the invariant) before the second client-action command lands. Not blocking v1. (abstraction, 81)
   https://github.com/kristofferremback/threa/blob/5837acb931bf29686a46beef9519749ac1e35ae4/apps/frontend/src/components/timeline/message-input.tsx#L377-L397

4. `apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts:159-177` + `apps/backend/src/features/agents/context-builder.ts:322-329` — `prependParentMessage` duplicates the parent-prepend pattern already inline in `buildThreadContext`. Known concern (already called out in-session); extracting a shared helper is the right follow-up. Not blocking. (abstraction, 82)
   https://github.com/kristofferremback/threa/blob/5837acb931bf29686a46beef9519749ac1e35ae4/apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts#L152-L177

5. `apps/frontend/src/lib/commands.ts` + `apps/frontend/src/components/timeline/message-input.tsx:368-369` — `hasCommandNode` + `extractCommandName` walk the same ProseMirror tree back-to-back in `handleSubmit`. Since `extractCommandName` returning non-null already tells you there's a command node, the `hasCommandNode` call is redundant. Tiny cleanup. (design, 80)
   https://github.com/kristofferremback/threa/blob/5837acb931bf29686a46beef9519749ac1e35ae4/apps/frontend/src/lib/commands.ts#L17-L41

### Status of prior review (5 items)

| Prior issue | Status |
| --- | --- |
| Session-slot collision (user's first message dropped) | ✅ Resolved — `checkForUnseenMessages` exported from `persona-agent-worker` and called on orientation completion |
| `context_received` trace shape mismatch | ✅ Resolved — emits `{ intent, messages: [...], totalMessageCount, truncated }` matching the renderer |
| Session-record inconsistency on skip path | ✅ Resolved — crash-window skip path looks up the existing message via `clientMessageId` |
| Stale `ResolvedBag.nextSnapshot` JSDoc | ✅ Resolved — describes compensating-controls model |
| `withCompanionSession` naming | ⏭️ Acknowledged, deferred to follow-up rename |

Plus the two UX regressions surfaced in-session (orientation held a session slot & trace shape) are both fixed; auto-naming now kicks in per-thread; slash command gated on send; thread root prepended.

---

Only the declared `buildStreamContext` integration deviation remains (per PR body). New commits are tactical fixes that don't conflict with the plan.
</details>

- **Soft-deleted parent leak** — #1, real correctness concern.
- **Double error notification** — #2, user-visible but non-blocking.
</details>

INV-35 parallel implementation concern on `prependParentMessage` scored below the ≥80 threshold (the existing inlined logic in `buildThreadContext` isn't a named export, so there's nothing to reuse today). Covered as a design follow-up in #4.
</details>

<details><summary>🏗️ Design [3 concerns]</summary>

- `prependParentMessage` duplication (#4) — best fixed via a shared `prependThreadRoot(db, stream, messages)` helper in the messaging barrel, applied to both the bag resolver and `buildThreadContext`.
- Client-action routing on `name` (#3) — latent gap if a second client-action command lands with a different clientActionId.
- `hasCommandNode` + `extractCommandName` double-walk (#5) — trivial.
</details>

No new vulnerabilities. Note that the soft-delete bug (#1) has a data-exposure flavor — the deleted content re-surfaces in an AI prompt — but since the bag creator always has access to the stream (verified at creation), this isn't a cross-user disclosure, just a violation of the user's "I deleted that" expectation.
</details>

---

kristofferremback:
### Code review (re-review after `b842434`)

_Supersedes [the prior unified-review](https://github.com/kristofferremback/threa/pull/403#issuecomment-4311723914)._

No issues found. All three reviewers returned `ISSUES: none`.

### Resolution of prior findings (5 items)

| Prior issue | Status |
| --- | --- |
| Soft-deleted parent leak (bug, 85) | ✅ Resolved via new `MessageRepository.findThreadRoot` — filters `deletedAt` as part of the canonical helper |
| Double error notification (bug, 82) | ✅ Resolved — composer's client-action branch keeps a silent catch; hook's `toast.error` is the single failure surface |
| Client-action routing on `name` (abstraction, 81) | ✅ Resolved — `CommandNodeAttrs.clientActionId: string \| null` added + persisted via `mapPropsToAttrs`; composer branches on `clientActionId === DISCUSS_WITH_ARIADNE_COMMAND` |
| `prependParentMessage` duplication (abstraction, 82) | ✅ Resolved — `findThreadRoot` is now the canonical helper; used by `buildThreadContext`, `buildThreadPath`, AND `ThreadResolver.fetch`. The bespoke `prependParentMessage` was deleted |
| `hasCommandNode` + `extractCommandName` double-walk (design, 80) | ✅ Resolved — single `extractCommandNode(content) -> { name, clientActionId } \| null` replaces both; dead code removed per INV-38 |

Plus two items the prior review surfaced mid-round and were closed in `b842434`:
- `buildThreadPath` bypass of the soft-delete filter → migrated to `findThreadRoot` so `threadPath[*].anchorMessage` gets the same scrubbing as `conversationHistory`.
- Test coverage gap → added 4 direct `findThreadRoot` tests in new `messaging/repository.test.ts` (null short-circuit, hard-delete, soft-delete, happy path) + rewrote thread-resolver spies to intercept at the contract boundary.

And the INV-25 JSDoc narration cleaned up in `ae14e0f`.

---

Deferred `buildStreamContext` integration still the only declared deviation; unchanged by this round of commits.
</details>

No issues found. `findThreadRoot`'s soft-delete filter is exercised by a direct unit test; `buildThreadPath` and `ThreadResolver.fetch` both route through it now, so deleted roots can't leak via either conversation history or the structural breadcrumb path.
</details>

INV-7, INV-25, INV-30, INV-31, INV-35, INV-38, INV-48, INV-51 all checked. No violations. No `vi.mock`/`mock.module` usage, no `.skip`/`.todo` tests, no dead imports, feature colocation preserved (new `repository.test.ts` placed flat in `messaging/` matching the existing `event-service.test.ts` pattern).
</details>

The `findThreadRoot` consolidation retires a recurring bug class without the dual-ownership refactor the data model would otherwise imply. Call sites (`buildThreadContext`, `buildThreadPath`, `ThreadResolver.fetch`) are consistent; the three remaining `findById(parentMessageId)` calls in the codebase are structurally different uses (outbox payload construction, reply-count updates, thread creation validation) and don't build AI prompt context, so they correctly don't need the helper.
</details>

No new vulnerabilities. Soft-delete filter closes the "deleted parent re-surfaces in prompt" exfil vector flagged earlier.
</details>

---

kristofferremback:
Closed in favor of #411

PR #404, opened 2026-04-23
feat(frontend): add stashed-drafts picker + Cmd+S save

kristofferremback:
## Problem

The composer has always kept exactly one auto-saved draft per scope (`DraftMessage` keyed `stream:<id>` / `thread:<parentId>`). That works for resuming a single in-flight reply, but not for the "I was typing one thing, now I want to respond to something else first" flow: the user has to cut the in-progress text to the clipboard (losing attachments, formatting, and quote-replies), type the new message, send, then paste the original back. There's no UI for explicitly parking a draft and coming back to it.

## Solution

Keep the existing auto-saved `DraftMessage` (one per scope, ambient) and add a **parallel stash** — a sibling pile where any number of explicitly-saved drafts can coexist for the same scope. The composer gets a small bookmark-icon popover in the toolbar for browsing/restoring stashed drafts, and **Cmd/Ctrl+S** from inside the composer stashes the current content, clears the editor, and fires a `toast.success("Saved as draft")`. Restoring a stashed draft while the composer already holds content auto-stashes the current content first, so swapping drafts never loses work silently.

### How it works

```
User types in composer          ─▶ DraftMessage (auto-save, 500ms debounce)
User presses Cmd+S              ─▶ snapshot current contentJson + attachments
                                ─▶ insert row into `stashedDrafts` table (ULID)
                                ─▶ clear composer + DraftMessage
                                ─▶ toast "Saved as draft"
User opens picker popover        ─▶ useLiveQuery by [workspaceId+scope]
                                   list (newest-first) with preview + relative time
User clicks a stashed row       ─▶ auto-stash current (if any)
                                ─▶ pop the chosen row (get + delete)
                                ─▶ load contentJson + restore attachments
                                ─▶ toast "Draft restored"
```

### Key design decisions

**1. Stash as a sibling pile, not a refactor of `DraftMessage`**

Alternative considered: collapse everything into a single `drafts` table with an "active" flag. Rejected as too invasive — touches the auto-save hook, the `/drafts` aggregate page, and the workspace-scoped draft store cache. The stash is purely additive, the ambient auto-save keeps working exactly as before, and `DraftMessage` stays one-per-scope (no new concurrency surface). The tradeoff is two coexisting notions of "draft" (ambient vs. stashed), which the UI disambiguates via label: auto-save has no UI, stashed drafts live behind the bookmark icon.

**2. ULID prefix `stash_` (not `draft_`)**

`draft_` is already claimed by `DraftScratchpad` and several code paths branch on `id.startsWith("draft_")` to distinguish draft scratchpads from real streams (`use-draft-scratchpads.ts`, `coordinated-loading-context.tsx`, `sync-engine.ts`, `use-coordinated-stream-queries.ts`). Reusing it would false-positive those checks. `stash_` is distinct and makes rows self-identifying in IDB.

**3. Compound index `[workspaceId+scope]` in Dexie v26**

The picker always queries by `(workspaceId, scope)`. Without the compound index, Dexie would scan the workspace and filter in JS. With it, the picker does an index range query regardless of how many stashed drafts exist across unrelated streams.

**4. Cmd+S handled at the composer root via capture-phase**

TipTap's ProseMirror plugins own the contentEditable's keydown in the bubble phase. Attaching `onKeyDownCapture` on the composer wrapper div runs first, so we get Cmd+S before the editor sees it and before the browser's default "save page" fires (which we `preventDefault`). This also scopes the shortcut correctly: only the composer that received the key press stashes — if both the main stream and a thread panel are open, only the focused one responds, no cross-talk.

Intentionally not wired through `useKeyboardShortcuts` yet: that hook is document-level and user-remappable, and sharing one action between two simultaneously-mounted composers requires more plumbing than this PR should carry. The key is hardcoded; follow-up can register `draftStash` in `SHORTCUT_ACTIONS` once the dispatch model is decided.

**5. Auto-stash-before-restore**

Clicking a stashed row while the composer holds content could silently destroy work. Instead, we stash the current content first (a no-op if empty), then pop the chosen row. The cost is one extra stash row per swap; the benefit is "nothing is ever lost silently" matches the mental model the user described.

**6. Two picker slots (`stashedDraftsTrigger` + `stashedDraftsTriggerFab`)**

Desktop inline and mobile use a 7×7 ghost button next to send; the expanded-mode FAB drawer uses a 30×30 outline-shadow button to match the other drawer buttons. Rather than encode size-branching into `MessageComposer`, hosts pass both trigger nodes — explicit and lets the picker component stay a pure presentation unit.

**7. Mutations extracted as module-level pure functions**

`createStashedDraft`, `popStashedDraft`, and `deleteStashedDraftById` are top-level exports; the hook wraps them with `useCallback`. This lets the unit tests exercise mutations directly without `renderHook` or `useLiveQuery` (the library's ESM export isn't spy-able, and per INV-48 we avoid `vi.mock`).

**8. Silent no-op on empty composer**

Per the product decision during scoping: Cmd+S on an empty composer does nothing — no toast, no row. The picker's "Save current" button is disabled in that state. Prevents spamming the stash with blank rows on stray key presses.

## New files

| File | Purpose |
| --- | --- |
| `apps/frontend/src/hooks/use-stashed-drafts.ts` | `useStashedDrafts` hook + pure mutation helpers (`createStashedDraft`, `popStashedDraft`, `deleteStashedDraftById`, `generateStashId`). Re-exports `StashedDraft` so components don't import `@/db` (INV-15). |
| `apps/frontend/src/hooks/use-stashed-drafts.test.ts` | 10 unit tests covering id format, empty-content no-op, attachment-only stash, scope-undefined no-op, restore get+delete, idempotent delete. |
| `apps/frontend/src/components/composer/stashed-drafts-picker.tsx` | Bookmark-icon popover with count badge, "Save current" action, per-row preview (via `stripMarkdownToInline` per INV-60) + terse relative time, trash affordance on hover. Compact and FAB size variants. |

## Modified files

| File | Change |
| --- | --- |
| `apps/frontend/src/db/database.ts` | Added `StashedDraft` interface and v26 migration (`stashedDrafts` table, `[workspaceId+scope]` index). Added to `clearAllCachedData`. |
| `apps/frontend/src/db/index.ts` | Re-export `StashedDraft` type. |
| `apps/frontend/src/hooks/use-draft-composer.ts` | Exposed `restoreAttachments` so stashed drafts round-trip their attachments on restore. |
| `apps/frontend/src/hooks/index.ts` | Barrel-export the new hook + types. |
| `apps/frontend/src/components/composer/index.ts` | Barrel-export `StashedDraftsPicker`. |
| `apps/frontend/src/components/composer/message-composer.tsx` | Added `onStashDraft`, `stashedDraftsTrigger`, `stashedDraftsTriggerFab` props. Cmd+S handled via `onKeyDownCapture` on both expanded and inline roots. Picker rendered in desktop toolbar, mobile action bar, and FAB drawer. All new props optional — existing callers (edit forms) unaffected. |
| `apps/frontend/src/components/timeline/message-input.tsx` | Wire stash/restore/delete into the main stream composer via `useStashedDrafts`; pass both trigger variants. |
| `apps/frontend/src/components/thread/stream-panel.tsx` | Same wiring for the draft-thread composer (both expanded and inline instances). Picker hidden when `draftKey` is empty (non-draft panel). |

## Test plan

- [x] 10 new unit tests for the stash helpers (id format, empty/scope no-ops, attachment-only stash, pop, idempotent delete) — passing
- [x] Frontend typecheck — clean
- [x] Frontend lint — clean (INV-15 respected via hook re-export; INV-48 respected via module-level helpers + scoped `spyOn`)
- [x] Full frontend test run (1365 tests) — all passing
- [x] Monorepo typecheck — clean
- [ ] Manual: Cmd+S on desktop stashes current draft, clears composer, shows toast
- [ ] Manual: bookmark popover lists saved drafts with preview + relative time; click restores into composer
- [ ] Manual: restoring with dirty composer auto-stashes the current content first (no data loss)
- [ ] Manual: empty-composer Cmd+S is a silent no-op (no toast, no row)
- [ ] Manual: trash icon on a stashed row deletes without restoring
- [ ] Manual: mobile — picker trigger tappable in action bar, popover legible
- [ ] Manual: expanded (fullscreen) composer — picker in FAB drawer, Cmd+S still works
- [ ] Manual: thread draft panel — picker scoped to that thread, not leaking into main stream
- [ ] Manual: attachments round-trip (stash a draft with an image, restore, image reappears)

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01WAo5EA1v2ViY9WxC5Rbck7)_

### DIFF
--- a/apps/frontend/src/components/composer/index.ts
+++ b/apps/frontend/src/components/composer/index.ts
@@ -1,2 +1,3 @@
 export { MessageComposer, type MessageComposerProps } from "./message-composer"
 export { FloatingComposerShell } from "./floating-composer-shell"
+export { StashedDraftsPicker } from "./stashed-drafts-picker"
--- a/apps/frontend/src/components/composer/message-composer.tsx
+++ b/apps/frontend/src/components/composer/message-composer.tsx
@@ -12,6 +12,8 @@ import {
 } from "react"
 import { ArrowUp, X, Plus, AtSign, Slash, Paperclip, Maximize2 } from "lucide-react"
 import { useIsMobile } from "@/hooks/use-mobile"
+import { usePreferencesOptional } from "@/contexts"
+import { getEffectiveKeyBinding, matchesKeyBinding } from "@/lib/keyboard-shortcuts"
 import { RichEditor, EditorToolbar, EditorActionBar } from "@/components/editor"
 import type { RichEditorHandle } from "@/components/editor"
 import { Button } from "@/components/ui/button"
@@ -355,23 +357,25 @@ export function MessageComposer({
     [onCollapse]
   )
 
-  // Cmd/Ctrl+S stashes the current draft. Attached in the capture phase on
+  // `draftStash` stashes the current draft. Attached in the capture phase on
   // the composer root so it runs before TipTap's contentEditable sees the
-  // event, and before the browser's default "save page" behavior. The
-  // browser default is always suppressed inside a mounted composer — even
-  // when `onStashDraft` is absent, Cmd+S saving a blank HTML page of the
-  // app has no useful outcome.
+  // event, and before the browser's default "save page" behavior. Capture
+  // scopes the shortcut to whichever composer actually received focus — if
+  // main + thread are both mounted, only the focused one fires. Registered
+  // via `SHORTCUT_ACTIONS`, so the user can remap it in settings.
+  const preferencesCtx = usePreferencesOptional()
+  const stashBinding = getEffectiveKeyBinding("draftStash", preferencesCtx?.preferences?.keyboardShortcuts ?? {})
+
   const handleStashKeyDown = useCallback(
     (event: ReactKeyboardEvent<HTMLDivElement>) => {
-      if (event.key !== "s" && event.key !== "S") return
-      if (!(event.metaKey || event.ctrlKey)) return
-      if (event.altKey || event.shiftKey) return
+      if (!stashBinding) return
+      if (!matchesKeyBinding(event.nativeEvent, stashBinding)) return
 
       event.preventDefault()
       event.stopPropagation()
       onStashDraft?.()
     },
-    [onStashDraft]
+    [onStashDraft, stashBinding]
   )
 
   const sharedEditor = (
--- a/apps/frontend/src/components/composer/message-composer.tsx
+++ b/apps/frontend/src/components/composer/message-composer.tsx
@@ -1,6 +1,7 @@
 import {
   type ChangeEvent,
   type KeyboardEvent as ReactKeyboardEvent,
+  type ReactNode,
   type RefObject,
   useMemo,
   useCallback,
@@ -144,6 +145,31 @@ export interface MessageComposerProps {
   streamContext?: MentionStreamContext
   /** Imperative handle ref for programmatic focus from parent */
   composerRef?: React.MutableRefObject<{ focus: () => void; focusAfterQuoteReply: () => void } | null>
+
+  /**
+   * Triggered when the user presses Cmd/Ctrl+S with focus inside the composer,
+   * or when they click "Save current" in the stashed-drafts picker. The host
+   * is responsible for snapshotting the current content/attachments, adding a
+   * row to the stash, clearing the active draft, and showing a toast. An
+   * empty composer should no-op; the picker disables its own button when
+   * `canStashCurrent` is false.
+   */
+  onStashDraft?: () => void
+
+  /**
+   * Slot for the stashed-drafts picker trigger used in the desktop inline
+   * toolbar and the mobile action bar (compact size). Omit to hide the
+   * affordance entirely (used by edit forms and other non-draft consumers).
+   */
+  stashedDraftsTrigger?: ReactNode
+
+  /**
+   * Separate slot for the expanded-mode FAB drawer, where the trigger needs
+   * to match the 30x30 outline-shadow style of the other drawer buttons.
+   * Hosts pass both slots because the picker is rendered fresh in each
+   * context rather than shared by reference.
+   */
+  stashedDraftsTriggerFab?: ReactNode
 }
 
 export function MessageComposer({
@@ -173,6 +199,9 @@ export function MessageComposer({
   onCollapse,
   streamContext,
   composerRef,
+  onStashDraft,
+  stashedDraftsTrigger,
+  stashedDraftsTriggerFab,
 }: MessageComposerProps) {
   // Controls (buttons, file input) are disabled during both external disable and sending.
   // The editor itself stays editable during sending so mobile keyboards don't close/reopen.
@@ -326,6 +355,25 @@ export function MessageComposer({
     [onCollapse]
   )
 
+  // Cmd/Ctrl+S stashes the current draft. Attached in the capture phase on
+  // the composer root so it runs before TipTap's contentEditable sees the
+  // event, and before the browser's default "save page" behavior. The
+  // browser default is always suppressed inside a mounted composer — even
+  // when `onStashDraft` is absent, Cmd+S saving a blank HTML page of the
+  // app has no useful outcome.
+  const handleStashKeyDown = useCallback(
+    (event: ReactKeyboardEvent<HTMLDivElement>) => {
+      if (event.key !== "s" && event.key !== "S") return
+      if (!(event.metaKey || event.ctrlKey)) return
+      if (event.altKey || event.shiftKey) return
+
+      event.preventDefault()
+      event.stopPropagation()
+      onStashDraft?.()
+    },
+    [onStashDraft]
+  )
+
   const sharedEditor = (
     <RichEditor
       ref={setRichEditorHandle}
@@ -412,6 +460,7 @@ export function MessageComposer({
           className={cn("relative flex flex-col h-full bg-background", className)}
           tabIndex={-1}
           onKeyDown={handleExpandedShellKeyDown}
+          onKeyDownCapture={handleStashKeyDown}
         >
           <p id={instructionsId} className="sr-only">
             {screenReaderInstructions}
@@ -470,7 +519,8 @@ export function MessageComposer({
           {/* Floating action drawer + send button — bottom-right corner */}
           <div className="absolute bottom-4 right-4 z-10 flex items-center gap-1.5 group/fab">
             {/* Action drawer — slides out from behind the + button on hover or focus-within */}
-            <div className="flex items-center gap-1 overflow-hidden max-w-0 opacity-0 group-hover/fab:max-w-[200px] group-hover/fab:opacity-100 group-focus-within/fab:max-w-[200px] group-focus-within/fab:opacity-100 transition-all duration-200 ease-out">
+            <div className="flex items-center gap-1 overflow-hidden max-w-0 opacity-0 group-hover/fab:max-w-[240px] group-hover/fab:opacity-100 group-focus-within/fab:max-w-[240px] group-focus-within/fab:opacity-100 transition-all duration-200 ease-out">
+              {stashedDraftsTriggerFab}
               <Tooltip>
                 <TooltipTrigger asChild>
                   <Button
@@ -619,6 +669,7 @@ export function MessageComposer({
         )}
         onFocusCapture={isMobile ? handleFocusCapture : undefined}
         onBlurCapture={isMobile ? handleBlurCapture : undefined}
+        onKeyDownCapture={handleStashKeyDown}
       >
         <p id={instructionsId} className="sr-only">
           {screenReaderInstructions}
@@ -690,7 +741,16 @@ export function MessageComposer({
                     onMobileExpandedChange={setMobileExpanded}
                     showAttach
                     onAttachClick={handleAttachClick}
-                    trailingContent={sendButton}
+                    trailingContent={
+                      stashedDraftsTrigger ? (
+                        <div className="flex items-center gap-1">
+                          {stashedDraftsTrigger}
+                          {sendButton}
+                        </div>
+                      ) : (
+                        sendButton
+                      )
+                    }
                   />
                 ) : (
                   <div className="flex items-center gap-1">
@@ -808,6 +868,7 @@ export function MessageComposer({
                         Attach files
                       </TooltipContent>
                     </Tooltip>
+                    {stashedDraftsTrigger}
                     {sendButton}
                   </div>
                 )}
--- a/apps/frontend/src/components/composer/stashed-drafts-picker.tsx
+++ b/apps/frontend/src/components/composer/stashed-drafts-picker.tsx
@@ -1,0 +1,181 @@
+import { useCallback, useMemo, useState } from "react"
+import { Bookmark, BookmarkPlus, Trash2 } from "lucide-react"
+import { serializeToMarkdown } from "@threa/prosemirror"
+import { Button } from "@/components/ui/button"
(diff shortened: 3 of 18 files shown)


kristofferremback:
> [!NOTE]
> **Superseded by [a later review](https://github.com/kristofferremback/threa/pull/404#issuecomment-4311700791)** — the author fixed all 7 findings from this pass in commit `ae1f131`; the follow-up review verified them and flagged one new non-blocking issue. Original review preserved below for reference.

<details>
<summary>Original review (click to expand)</summary>

### Code review

Found 7 issues — 1 CLAUDE.md violation, 2 correctness, 2 design (refactor), 2 reactivity/perf. None block the feature's core correctness path; the most impactful is the INV-60 violation on the `/drafts` page.

1. `apps/frontend/src/pages/drafts.tsx:72` — stashed-row `label` assigned from `draft.preview`, which is raw markdown from `serializeToMarkdown` (via `getContentPreview` in `use-all-drafts.ts`). `ItemList` renders `label` in a plain `<span>`, so `**bold**`, fenced code, `[text](url)`, and `:emoji:` ship to users as characters. (CLAUDE.md [INV-60, *Frontend Composition and UX Semantics*]: "Preview surfaces must strip markdown before rendering… never pass raw markdown into a plain `<span>`"). The composer-side picker already strips correctly; pipe `draft.preview` through `stripMarkdownToInline` here too.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/pages/drafts.tsx#L70-L80

2. `apps/frontend/src/components/timeline/message-input.tsx:248-300` + `apps/frontend/src/components/thread/stream-panel.tsx:155-202` — `handleStashDraft`, `handleRestoreStashed`, `handleDeleteStashed` are copy-pasted between both composer hosts (same attachment-filter predicate, same swap-before-restore, same toasts). They'll drift: the comment explaining the `handleContentChange` debounce path is already present in one copy but absent in the other. Extract to a `useStashComposerHandlers(composer, stashedDrafts, emptyDoc)` hook.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/components/timeline/message-input.tsx#L248-L300
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/components/thread/stream-panel.tsx#L155-L202

3. `apps/frontend/src/hooks/use-stashed-drafts.ts:77-82` — `popStashedDraft` is `get()` then `delete()` as two separate IDB requests. With two tabs open to the same workspace, both can read the row before either deletes, so both restore the same draft into their respective composers. Dexie won't wrap multi-step helper calls for you — wrap in `db.transaction('rw', db.stashedDrafts, async () => { const row = await …; if (!row) return null; await …delete(id); return row })`.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/hooks/use-stashed-drafts.ts#L76-L83

4. `apps/frontend/src/hooks/use-all-drafts.ts:175-183` — the workspace-wide stash query uses the simple `workspaceId` index and pulls every row into React state on each Dexie change anywhere in `stashedDrafts`. A power user with ~200 stashed drafts across many streams re-materializes all 200 on every write. The `[workspaceId+scope]` compound index is used correctly by the per-scope picker but the explorer path has no ceiling. Add a `.limit(N)` or (better) paginate; note the trade-off in a comment either way.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/hooks/use-all-drafts.ts#L172-L184

5. `apps/frontend/src/hooks/use-all-drafts.ts:188-192` — `hasThreadDrafts` `useMemo` depends on the `stashedDrafts` array identity. `useLiveQuery` returns a fresh array ref on every re-fire regardless of content, so this memo recomputes on every stash write anywhere in the workspace (even stream-scoped writes). The file already uses a stable-key derivation for `cachedStreams` (`streamIdKey`) for this exact reason — same mitigation applies here.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/hooks/use-all-drafts.ts#L185-L194

6. `apps/frontend/src/components/timeline/message-input.tsx:246` + `apps/frontend/src/components/thread/stream-panel.tsx:153` — two new `const emptyDoc = useMemo<JSONContent>(() => ({ type: "doc", content: [{ type: "paragraph" }] }), [])`. This literal already appears 6+ times across the frontend; `prosemirror-utils.ts` (home of `isEmptyContent`) is the natural place for a module-level `EMPTY_DOC` export. The `useMemo` wrap is also unneeded — a module-level object is already referentially stable.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/components/timeline/message-input.tsx#L244-L248

7. `apps/frontend/src/components/timeline/message-input.tsx:309-323` + `apps/frontend/src/components/thread/stream-panel.tsx:232-244` — in `?stash=<id>` auto-restore, `setSearchParams` strips the param before `handleRestoreStashed` runs. If the inner `stashDraft` (pre-restore auto-stash of current content) throws (IDB quota, etc.), the outer promise rejects and the chosen stash row is never popped — but the URL param is already gone, so a refresh won't retry. Data isn't lost (the row remains in the picker) but the deep link silently fails. Strip the param only on success, or catch-and-retry.
   https://github.com/kristofferremback/threa/blob/77316588ee4498a2e39961fb516a886f6b821170/apps/frontend/src/components/timeline/message-input.tsx#L302-L324

---

All items from the PR description are present: Dexie v26 table + index, `stash_` ULID prefix, capture-phase Cmd+S, picker with two slots, auto-stash-before-restore, empty-composer no-op, `restoreAttachments` exposure, 10 unit tests, `/drafts` grouping, `?stash=` deep link.

</details>

Non-atomic `popStashedDraft` (#3) and URL-param strip before error-prone restore (#7). One earlier TDZ claim was a false positive — `handleRestoreStashed` is defined lexically before every `useEffect` that references it in both hosts.

</details>

INV-60 on `/drafts` stashed-row label (#1). INV-15/INV-48 respected (type re-exported through hooks, module-level pure fns + scoped `spyOn`). Dexie migration is append-only (INV-17). `stash_` prefix is distinct from `draft_` (keeps the existing `isDraftId` discriminator intact).

</details>

Copy-pasted stash-handler trio across the two composer hosts (#2) and `EMPTY_DOC` duplication (#6). The two hosts also share nearly-identical `?stash=` auto-restore `useEffect` blocks — likely falls out naturally if you extract the handler hook.

</details>

Previews strip via `stripMarkdownToInline` before rendering in the composer picker. `?stash=<id>` param read from URL is user-scoped by Dexie — `popStashedDraft` is keyed by id but `stashedDrafts` is IndexedDB-local (per-browser), so cross-workspace restore via crafted URL is not a realistic surface. No secret/token leakage introduced.

</details>

<details><summary>🔄 Reactivity/Perf [2 concerns]</summary>

Unbounded workspace-wide stash scan in the `/drafts` explorer (#4) and `stashedDrafts`-array-identity churn re-triggering `hasThreadDrafts` + downstream event lookup (#5).

</details>

---
</details>

kristofferremback:
### Code review (re-review, HEAD `ae1f131`)

Supersedes [the prior review](https://github.com/kristofferremback/threa/pull/404#issuecomment-4311516226) — all 7 prior findings verified fixed by `ae1f131`. One new non-blocking finding on this pass.

**Prior findings — all fixed and verified:**
- ✅ INV-60: `getContentPreview` now strips markdown at source
- ✅ Duplicate stash handlers: extracted `useStashComposer` hook, both hosts consume it
- ✅ `popStashedDraft` atomicity: wrapped in `db.transaction("rw", db.stashedDrafts, …)`
- ✅ Unbounded workspace scan: `.limit(500)` added (but see new finding below)
- ✅ `stashedDrafts` identity churn: stable `stashedScopesSignature` via `Set`→sort→join
- ✅ `EMPTY_DOC` centralised to `prosemirror-utils.ts`
- ✅ URL `?stash=` param stripped only after restore resolves; pre-stash errors caught

---

**New findings (1):**

1. `apps/frontend/src/hooks/use-all-drafts.ts:194` — the `.limit(500)` ceiling intended to bound the worst case silently returns the **oldest** 500 stashes, not the newest. `.where("workspaceId").equals(workspaceId)` iterates the `workspaceId` index in key order, and within a single workspace the primary key (ULID) is the tiebreaker. Without an explicit sort or reverse, the limit truncates from the old end. For power users who cross the cap, recently-stashed rows disappear from `/drafts` with no warning (the picker is unaffected because it's scoped). Fix either:
   - `.where("workspaceId").equals(workspaceId).reverse().limit(500).toArray()` — simplest, ULIDs sort chronologically so reverse = newest-first
   - add a `[workspaceId+createdAt]` compound index in a follow-up migration and use it here

   https://github.com/kristofferremback/threa/blob/ae1f1310abe748164c9aa0ca2599ebc54ab1b1e6/apps/frontend/src/hooks/use-all-drafts.ts#L186-L200

---

All items from the PR description present. Refactor commit `ae1f131` is additive over the plan (hook extraction + safety rails) without changing the user-facing behavior.

</details>

`.limit` without sort returns oldest rows (#1). Remaining verification points from the re-review: `db.transaction` wrap confirmed covering both ops; URL-strip inside `.then()` confirmed race-free (ref-based dedup absorbs effect re-runs between mount and searchParams stabilizing); pre-stash errors correctly caught and logged without aborting the restore.

</details>

INV-60 satisfied at the source (`getContentPreview` strips). INV-48 still OK — no `vi.mock` reintroduced. Dexie migration still append-only (INV-17). No new violations introduced by the refactor commit.

One spec-reviewer flag on test coverage of the transaction wrapper (the spies mock `.get`/`.delete` but not `.transaction`, so atomicity isn't directly exercised) was downgraded below threshold — the invariant cited didn't match and the limitation is inherent to testing Dexie transactions without fake-indexeddb ceremony.

</details>

<details><summary>🏗️ Design [CLEAN — 5 minor notes]</summary>

Below-threshold observations for when you have time (none blocking):
- `EMPTY_DOC` migration is "partial across the repo" — 3 pre-existing sites remain in `use-draft-composer.ts`, `message-edit-form.tsx`, `unsent-message-edit-form.tsx`. The PR centralised the 5 sites in the files it was already touching; the others are leftover from before this PR.
- `useStashComposer(composer, …)` takes the full `DraftComposerState` when it only reads ~8 members. Narrower structural type would be slightly cleaner but this hook only has two call sites.
- `getPreview` in `stashed-drafts-picker.tsx` re-serialises `contentJson` at render time, duplicating work already done by `getContentPreview` in `use-all-drafts.ts`. Different data flows (the picker receives `StashedDraft` directly, the explorer receives `UnifiedDraft.preview`), so no regression — just a coordination gap.
- `stashedScopesSignature.includes("thread:")` uses substring match on a sorted-joined string. An exact `Array.some(s => s.startsWith("thread:"))` on the original scopes would be safer against contrived scope shapes, but current scopes are controlled (only `stream:` / `thread:` prefixes).
- Hosts still render two `<StashedDraftsPicker>` instances (compact + fab). `size` could move inside the picker via container query / `useIsMobile`. Intentional for now per the PR description ("explicit, lets the picker component stay a pure presentation unit").

</details>

No new surface introduced by the refactor. Preview rendering still strips markdown before `<span>`. URL-param auto-restore still Dexie-local (per-browser IDB) — no cross-workspace restore via crafted URL.

</details>

---

PR #411, opened 2026-04-24
feat(context-bag): ContextBag as composer attachment

kristofferremback:
## Summary

Reshapes "Discuss with Ariadne" away from an eager kickoff message toward a **composer-attachment UX that follows the same lifecycle as a file upload**: the chip lives in the composer pre-send (sidecar pill above `MessageInput`), and on send "moves" onto the first message in the timeline as an inline badge — same mental model the rest of the app already uses for image/file uploads.

The user-facing change: when you trigger "Discuss with Ariadne" from a thread message, the new scratchpad opens with a small pill above an empty composer reading **"12 messages in #intro"**. You type your first question, send, and the pill becomes a small breadcrumb on that message. No auto-greeting from Ariadne, no chip stuck above the composer forever.

## Architecture

| Concern | Where it lives |
|---|---|
| Server-side bag (persistent) | `stream_context_attachments` table (unchanged from #403) |
| Bag fetch | `GET /streams/:id/context-bag` |
| Draft sidecar (pre-send only) | `DraftMessage.contextRefs?: DraftContextRef[]` |
| Composer surface (pre-send) | `<ContextRefStrip>` reads sidecar, renders nothing once draft clears |
| Timeline surface (post-send) | `<MessageContextBadge>` rendered on the first message via `useStreamContextBag` |
| Label generation | `formatContextRefLabel({source, anchors})` |
| Cache warming | `ContextBagPrecomputeHandler` (outbox) + `POST /context-bag/precompute` (client) |
| Per-turn injection | `resolveBagForStream` inside `PersonaAgent.run` |

### Backend

- `ContextBagOrientationHandler` → `ContextBagPrecomputeHandler`. Drops kickoff message, keeps cache warming + snapshot persist.
- `POST /context-bag/precompute` — Zod-validated, stream-agnostic, summarizes refs over the inline threshold.
- `GET /streams/:id/context-bag` — returns `{bag, refs[]}` with each ref carrying `source: {displayName, slug, type, itemCount}`. Drives both strip and badge label rendering.
- `MessageRepository.countByStream` — single COUNT(*) for the chip's "12 messages" label.
- `loadOrCreateSummary` extracted from `resolve.ts` and shared with the standalone service (INV-35).
- `appendBagToSystemPrompt` + bag integration in `PersonaAgent.run`: every turn calls `resolveBagForStream`, folds `stable + delta` into the system prompt, persists the next snapshot. Live-follow on edits/deletes lands automatically. Resolves the deferred-from-#403 item.
- `DiscussThreadIntent.orientationUserPrompt` deleted (INV-38).

### Frontend

- **`<ContextRefStrip>`** — pills above the composer. Renders only when `draft.contextRefs` is non-empty. Once the draft clears on send, the strip disappears.
- **`<MessageContextBadge>`** — pill rendered inline on the first message of any bag-attached scratchpad. Reads `useStreamContextBag` for source-stream metadata.
- **`event-list.tsx`** computes `firstMessageId` once from `timelineItems` and threads `isFirstMessage` through `TimelineItemRenderContext` → `EventItem` → `MessageEvent` variant inner fns → `MessageLayout`. Badge renders inside `messageBody` alongside attachments, before link previews. Pending/failed/editing variants get the badge too — chip is part of the message lifecycle, not just the sent state.
- `DraftContextRef` sidecar on `DraftMessage.contextRefs` (additive optional field, no Dexie migration). Type lives in `lib/context-bag/types.ts` so view layers can `import type` without violating INV-15.
- `formatContextRefLabel({source, anchors})` — produces "12 messages in #intro" / "Slice of #intro" / "Thread in #intro".
- `useStreamContextBag(workspaceId, streamId)` — TanStack Query wrapper, 60s staleTime.
- `useDraftMessage` returns `contextRefs`; `useDraftComposer` exposes them through `DraftComposerState.contextRefs` and gates `canSend` on every ref being `ready` or `inline` (same model as in-flight uploads).
- `useDiscussWithAriadne` writes a sidecar entry to the destination scratchpad's draft (status optimistic `"ready"`) before navigation.

### Removed during the refactor arc

- `ContextRefChipExtension`, `ContextRefChipView` (the original ProseMirror chip).
- `lib/context-bag/extract.ts` (chip walker + `canSendContextBag` gate).
- `editor-markdown.ts` chip-strip handling + chip atom branch.
- `seedDraftWithContextRefChip` helper that wrote chip nodes into composer content.

## Invariants honoured

INV-1, INV-2, INV-3, INV-7, INV-8, INV-14, INV-15 (component types live outside `@/db`), INV-17, INV-18, INV-19, INV-20, INV-21, INV-28, INV-30, INV-32 (`STREAM_NOT_FOUND` / `STREAM_FORBIDDEN` codes), INV-33, INV-35, INV-38, INV-41, INV-44, INV-48, INV-51/52, INV-55, INV-56, INV-57 (bag stays stream-level; first-message anchor is render-only).

## Deliberate product change

Eager orientation greeting from #403 is gone. Phase 1 is independently revertable — the summary table + snapshot contract don't change.

## Known limitations / follow-ups

- **No remove × on the strip pill yet** — the only way to "remove" attached context today is to abandon the scratchpad before sending. Small follow-up.
- **Anchored slices say "Slice of #foo"** rather than "Messages X–Y in #foo" — needs sequence resolution server-side, deferred.
- **No dedicated rate limiter on the new endpoints** — they fall under `rateLimits.globalBaseline`.
- **Strip not rendered inside the expanded fullscreen composer overlay** — kept the typing-focused mode uncluttered.
- **Badge anchored to the first message** — if that message is later deleted, the badge has nowhere to go. The bag still applies server-side; only the visual breadcrumb is lost.

## Test plan

- [x] `bun run typecheck` — all 9 workspaces pass
- [x] `bun run lint` — all pass (including the INV-15 component-database-import rule)
- [x] `bun run test:unit` — **993 backend unit tests pass** (was 971 on #403 base)
- [x] `bun run test` (frontend) — **1381 frontend tests pass**
- [x] OpenAPI spec regenerated cleanly
- [ ] Manual: click "Discuss with Ariadne" → strip "N messages in #foo" appears above empty composer; no auto-greeting.
- [ ] Manual: send first message → strip disappears, badge "N messages in #foo" appears on the just-sent message in the timeline.
- [ ] Manual: send subsequent messages → badge stays on the first message only; later messages don't carry it.
- [ ] Manual: reload → badge still visible on the first message (server fetch via GET endpoint).
- [ ] Manual: mobile collapsed view → strip visible above composer pre-send; badge visible on first message post-send.
- [ ] Manual: edit a message in the source thread → Ariadne's next reply reflects the edit via the "Since last turn" delta block.

https://claude.ai/code/session_01HXN3PR5EmmkQKAK5Bu1jiL


kristofferremback:
@coderabbitai review

apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts
```diff
@@ -0,0 +1,172 @@
...
+ * the stream or on the root stream (threads inherit root membership).
+ */
+export const ThreadResolver: Resolver<ThreadRef> = {
+  kind: ContextRefKinds.THREAD,
+
+  canonicalKey(ref) {
+    return `thread:${ref.streamId}`
+  },
+
+  async assertAccess(db, ref, userId, workspaceId) {
```
kristofferremback, line 29: This is just incorrectly re-implementing `StreamService.checkAccess(...)`

kristofferremback: Good catch — and on a second look this was duplicated all over: `fetchStreamBag`, `ThreadResolver.assertAccess`, and `StreamService.checkAccess` were each rolling their own version of the membership-vs-visibility-vs-thread-inheritance check, so each surface drifted from the others. Same story for the user+persona name lookup (was inlined in `ThreadResolver` and `companion/context.ts`).

Fixed in 4b7c4a1 by extracting two reusable primitives instead of patching call sites:

- `apps/backend/src/features/streams/access.ts` — `checkStreamAccess(db, streamId, workspaceId, userId)`. Single source of truth for access; handles workspace boundary, public-channel skip, and threads-inherit-from-root. `StreamService.checkAccess` now delegates to it; `fetchStreamBag` and `ThreadResolver.assertAccess` use it directly. Also collapses missing / wrong-workspace / no-access into one FORBIDDEN response so the error doesn't confirm existence of streams the caller can't see.
- `apps/backend/src/features/agents/actor-names.ts` — `resolveActorNames(db, workspaceId, ids)`. Batched user+persona lookup (INV-56), reused by `ThreadResolver.fetch` and `companion/context.ts`.

Also renamed `FetchStreamBagOptions.skipMembershipCheck` → `skipAccessCheck` for accuracy.

---

kristofferremback:
### Code review
**Confidence: 5/7** — Good

Found 2 issues:

1. `apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts:128-145` — `assertAnchorExists` mis-classifies a soft-deleted anchor message as `CONTEXT_ANCHOR_OUT_OF_WINDOW`. `MessageRepository.list` filters `deleted_at IS NULL` so a soft-deleted anchor is absent from the fetched window and `findIndex` returns -1; the assertion then calls `MessageRepository.findById` (which does **not** filter deleted) at line 134, sees `anchor.streamId === streamId` is true, and throws OUT_OF_WINDOW — telling the client to widen the fetch window when widening can't help. Add `&& !anchor.deletedAt` to the `existsInStream` predicate so deleted anchors fall through to the NOT_FOUND branch (bug).
   https://github.com/kristofferremback/threa/blob/35ec47e7b36c98dc74194e3ad58aa737037c9bab/apps/backend/src/features/agents/context-bag/resolvers/thread-resolver.ts#L128-L145

2. `apps/frontend/src/components/composer/context-ref-strip.tsx:61-86` — `refKey` is defined as a closure (lines 61-67) and used to build the server-by-key Map at line 68, but the React `key` prop at line 86 manually inlines a near-identical expression. They have **already drifted**: the closure has `r.refKind ?? r.kind` (handles both shapes) while the inline `key` only reads `ref.refKind` — if a draft ref ever lands without `refKind`, the Map lookup uses `undefined|...` while the React key uses `undefined|...` too, but the two stay in sync only by accident. Hoist `refKey` to module scope and call it at line 86 (abstraction).
   https://github.com/kristofferremback/threa/blob/35ec47e7b36c98dc74194e3ad58aa737037c9bab/apps/frontend/src/components/composer/context-ref-strip.tsx#L61-L86

---

The reviewer flagged a missing orientation-message handler against the plan's §6/§7. The PR description's "Deliberate product change" section explicitly removes the eager kickoff: *"Eager orientation greeting from #403 is gone."* Filtered as a false positive — the change is intentional and documented.
</details>

See issue #1 above.
</details>

No invariant violations met the threshold. Recent commits visibly improved compliance: extracted `checkStreamAccess` (INV-8 access vs. membership), `resolveActorNames` (INV-56 batched lookups), `countByStreams` (INV-56), `context-bag/config.ts` (INV-44 colocated config), shared `context-bag/schemas.ts` (INV-31, INV-33), `ON CONFLICT … DO UPDATE` upsert (INV-20).
</details>

See issue #2 above.

Considered but not flagged:
- `ContextBagRepository.insert` `ON CONFLICT (stream_id, intent)` lacks a `workspace_id` guard. Defense-in-depth only — `stream_id` ULIDs are globally unique, the upstream `createScratchpad` flow generates the streamId in the calling workspace, and the `updateLastRendered` path already filters by workspace. No concrete vuln.
- `precomputeRefSummaries` Phase 1 loop runs per-ref `assertAccess` + `fetch` sequentially under `withClient`. `Promise.all` over a single `PoolClient` would be serialized at the protocol level by node-postgres — the agent's suggested fix wouldn't actually parallelize. Would need to swap the access-checked single connection for `pool` to fan out, which trades correctness guarantees for latency on a path bounded to 10 refs. Not flagging.
</details>

Traced data flow from `req.body`/`req.params` through the new `POST /context-bag/precompute` and `GET /streams/:id/context-bag` endpoints. Inputs are Zod-validated via the centralized `contextBagSchema`. Per-ref access goes through `checkStreamAccess` + `resolver.assertAccess`. No injection, auth bypass, or data exposure surfaced.
</details>

---

kristofferremback:
## Investigation: timeline order regression after offline → stream switch → return

Logging this here so the underlying issue isn't lost — the badge-anchor mitigation just shipped (be749a1) only addresses the surface symptom.

### Repro
1. Open a bag-attached scratchpad with one user message + one Ariadne reply (correctly rendered: user → session-complete → Ariadne).
2. Go offline, switch to a different stream, switch back.
3. Timeline order is scrambled.

**Observed (offline-reconnect render):**
1. `Session complete` (07:35)
2. Ariadne reply (07:35)
3. **`Ariadne is working… 0 steps · 0 messages sent`** (phantom)
4. User message (07:35)

**Expected:** user → session-complete → Ariadne (chronological).

The user message is rendered *after* the reply that responded to it, and a fresh "in progress" session card appears *after* the completed one with `0 steps · 0 messages sent`.

### Scope of cause

Not introduced by this PR. The branch's diff against `main` for the relevant files:

- `apps/frontend/src/hooks/use-events.ts` — only added `computeTimelineLoadState` (skeleton-vs-empty decision logic). No changes to `sortBySequence`, dedupe, or pagination.
- `apps/frontend/src/sync/stream-sync.ts` — only added `contextBag` mirroring into IDB.
- `apps/frontend/src/sync/workspace-sync.ts` — only added `contextBag` preservation through `bulkPut` (doesn't touch the events table).
- `apps/frontend/src/components/timeline/event-list.tsx` — added author-grouping annotation + `findFirstMessageId`. No event reordering.

None of those touch event sequencing, agent-session reconciliation, or reconnect handling. The bug almost certainly reproduces on `main` against any agent stream with the same offline → switch → return path; this PR didn't introduce it but does make it visible because the badge then mis-anchors.

### Likely culprits (for whoever picks this up)

1. **Phantom in-progress session card** — `0 steps · 0 messages sent` after a completed one suggests a stale or duplicate active-session shell. Possibly the WS re-broadcast on reconnect emits a `session_started` that's not deduped against the existing `agent_sessions` row, or the live-session derivation key (sessionId + version?) doesn't match what the persisted card uses.
2. **Out-of-order events in IDB merge** — Dexie `bulkPut` is keyed on `id`, so a re-broadcast of the same event id won't duplicate it; but if reconnect-bootstrap ships events without `_sequenceNum` populated, `useLiveQuery` ordering by `_sequenceNum` could place them at the head.
3. **agent_sessions vs messages timing** — session events may carry a different sequence basis from messages, and the timeline grouping (`groupTimelineItems`) might splice them in by an attribute other than the global sequence. Worth checking what key the session_group items use and whether it's stable across the reconnect path.

### Mitigation already shipped (be749a1)

`findFirstMessageId` now matches `message_created` only, not `companion_response`. Rationale: in a bag-attached scratchpad the conversation always opens with the user's first question (auto-greeting was removed deliberately), so a `companion_response` landing first in render order can only mean a transient ordering glitch. With the tighter check the badge either anchors on the real opening message or hides until the user message renders — never on Ariadne's reply.

This does not fix the underlying ordering bug; it just refuses to mis-anchor on it.

### Recommended follow-up

Separate branch / issue. Repro path:
1. Open any agent-active stream that has a completed session.
2. Throttle network to offline.
3. Switch streams, switch back online.
4. Inspect Dexie `events` records by `_sequenceNum` ASC and compare to `timelineItems` render order; check whether a phantom `agent_session_started` lands without a paired completion key matching the existing one.

Filing here rather than as a separate issue so the context-bag work can ship without holding for the deeper sync investigation.

🤖 Posted by Claude based on the in-session investigation.

---

PR #435, opened 2026-04-27
feat(messaging): slice 3 — attachments on shared messages

kristofferremback:
## Summary

A "look at this" + image share previously rendered as text only. The hydration backend's `ok` variant only carried `contentJson` / `contentMarkdown`, and the shared-message card-body never reached for `<AttachmentList>`. This slice extends the wire shape end-to-end so a shared image, video, or file appears inside the card the same way it appears in the source stream.

Privacy is implicit — attachments only ride on `ok` payloads, where the access resolver above the mapping has already established viewer access to the source. `private` / `deleted` / `truncated` placeholders never surface attachments.

## What changed

**Backend — `hydration.ts`:**
- `HydratedSharedMessage.ok` carries `attachments: AttachmentSummary[]`.
- `hydrateSharedMessageIds` batches `AttachmentRepository.findByMessageIds` alongside the author-name lookup in a single `Promise.all` across every ok-state message in the chain — one extra round-trip regardless of recursion depth (INV-56). The lookup runs **outside** the BFS while-loop against the accumulated `okMessages` map, so messages that fail the per-level access check never reach the attachment fetch.
- The mapping uses the shared `toAttachmentSummary` helper, mirroring the `message_created` event payload exactly.

**Backend — `attachments/summary.ts` (new, refactor commit):**
- `toAttachmentSummary(a: Attachment): AttachmentSummary` — single source of truth for `Attachment` row → wire-shape mapping, including the video-only `processingStatus` carve-out.
- `event-service.ts` drops its local `interface AttachmentSummary` (which had a looser `processingStatus?: string`) in favor of the canonical `@threa/types` one and routes through the shared mapper. Resolves the INV-37 duplication that slice 3's second copy would otherwise have introduced.

**Wire types — `packages/types/src/api.ts`:**
- `SharedMessageHydration.ok` adds `attachments: AttachmentSummary[]`.

**Frontend:**
- `context.tsx` and `use-shared-message-source.ts` thread the new field through their respective type mirrors.
- `SharedMessageCardBody` renders `<AttachmentList>` inside `<AttachmentProvider>` for the ok branch. The provider mounts only when there are attachments to render so isolated card-body unit tests (no `MediaGalleryProvider` in scope) still cover the no-attachment ok case.

## Tests

- 4 new `hydrateSharedMessageIds` cases:
  - ok-state inlines attachment summaries
  - empty array when source has none (uniform wire shape)
  - one batched lookup across multiple ok messages (INV-56)
  - skipped entirely when no ok-state messages survive
- Existing fixtures updated for the new required `attachments` field (5 frontend test files).
- A module-scope `beforeEach` stub for `AttachmentRepository.findByMessageIds` (returning an empty Map) keeps the existing ok-state tests passing without per-test boilerplate. Uses scoped `spyOn` against the namespace import per INV-48.
- Backend hydration suite 16/16, full frontend suite 1466/1466.

## Test plan

- [ ] Pull a private channel + DM with an image attachment, share the channel message into the DM, confirm the image renders inside the DM card.
- [ ] Share a message into a re-share chain (A → B → C), confirm `ok`-state pointers along the chain show attachments and the truncated/private pointers don't.
- [ ] Share a video attachment, confirm `processingStatus` rides on the wire and the video player gates correctly on the `completed` / `skipped` states.
- [ ] Share a deleted message, confirm no attachments leak (deleted placeholder).
- [ ] Backend tests: `bun test src/features/messaging/sharing/hydration.test.ts` and `src/features/messaging/event-service.test.ts`.

https://claude.ai/code/session_01FRG5ZrJRUbsG4bpuxCMDKU

---
_Generated by [Claude Code](https://claude.ai/code/session_01FRG5ZrJRUbsG4bpuxCMDKU)_

### DIFF
--- a/apps/backend/src/features/attachments/index.ts
+++ b/apps/backend/src/features/attachments/index.ts
@@ -2,6 +2,9 @@
 export { AttachmentRepository } from "./repository"
 export type { Attachment, InsertAttachmentParams, AttachmentWithExtraction } from "./repository"
 
+// Wire-shape mappers
+export { toAttachmentSummary } from "./summary"
+
 export { AttachmentExtractionRepository } from "./extraction-repository"
 export type {
   AttachmentExtraction,
--- a/apps/backend/src/features/attachments/summary.ts
+++ b/apps/backend/src/features/attachments/summary.ts
@@ -1,0 +1,20 @@
+import type { AttachmentSummary } from "@threa/types"
+import type { Attachment } from "./repository"
+import { isVideoAttachment } from "./video"
+
+/**
+ * Map a stored `Attachment` row to the lightweight `AttachmentSummary` wire
+ * shape used in `message_created` event payloads and shared-message
+ * hydration. The `processingStatus` carve-out for videos is the only
+ * conditional bit — keeping it in one place avoids drift between the
+ * timeline and shared-message paths (INV-37).
+ */
+export function toAttachmentSummary(a: Attachment): AttachmentSummary {
+  return {
+    id: a.id,
+    filename: a.filename,
+    mimeType: a.mimeType,
+    sizeBytes: a.sizeBytes,
+    ...(isVideoAttachment(a.mimeType, a.filename) && { processingStatus: a.processingStatus }),
+  }
+}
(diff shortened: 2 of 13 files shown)


kristofferremback:
@coderabbitai review

PR #421, opened 2026-04-26
feat(messaging): slice 2 — cross-stream share picker + per-viewer recursive hydration

kristofferremback:
## Problem

Slice 1 of message sharing (#409) shipped share-to-parent only — a thread message could be shared up to its parent stream, with single-level pointer hydration. Anything more (cross-stream pickers, private re-share chains) was deferred. This PR ships **Slice 2** of the plan in `docs/plans/message-sharing-streams.md`: the cross-stream picker, recursive per-viewer hydration with `private` and `truncated` placeholder states, and the wire to send + confirm shares that cross a privacy boundary.

### Picker → handoff → composer

Reuses the same hand-off pattern Slice 1 used for share-to-parent. The new `ShareMessageModal` filters the workspace's accessible top-level streams (channel / DM / scratchpad, no threads, no archived), and on select queues a `ThreaSharedMessage` node for the target stream's composer via `queueShareHandoff`. Commentary and send happen in the target's normal composer — there is no second editor surface.

### Per-viewer recursive hydration

`hydrateSharedMessageIds` now walks each pointer chain level-by-level up to `MAX_HYDRATION_DEPTH = 3`. At every level the viewer's access is resolved via two batched queries:

- **Stream membership** (`listAccessibleStreamIds`) — workspace boundary, public visibility, direct member, or thread inheriting from root. Same three rules as `checkStreamAccess`, in one set-based SQL round-trip.
- **Share-grant lookup** (`SharedMessageRepository.listSourcesGrantedToViewer`) — a viewer can read a source iff a share with that source exists in a target stream the viewer can read. Composes with `listAccessibleStreamIds` so the access predicate stays in one place.

Per-message classification:

- `ok` — viewer has access; current source content inlined.
- `deleted` — accessible source, tombstoned.
- `missing` — source row never existed.
- `private` — viewer has no read path. Reveals only the source stream's `kind` + `visibility`. Threads resolve to their parent's kind so the placeholder vocabulary stays in {channel, dm, scratchpad}.
- `truncated` — pointer collected past the depth cap. Renders a "open in source stream" link using the cached attrs streamId — no extra DB hit.

### Privacy-block toast

[...]

PR #437, opened 2026-04-28
fix(messaging): keep shared_messages and pointer hydration consistent across moves

kristofferremback:
## Problem

Three logical conflicts surface when batch move-to-thread (#420) lands on top of the share-message slices (#409, #421, #435). They're independent on the wire but they all leak the same root cause: a few denormalized columns and one cached attr value record where a message lives, and a move doesn't update them.

> Stacks on top of #420 (`add-batch-move-messages-to-thread`) — fix 1 modifies code that doesn't exist on main yet. Once #420 merges this rebases onto main directly.

### 1. `shared_messages` is missing from `updateStreamScopedReferences`

`MessageRepository.updateStreamScopedReferences` re-stamps `stream_id` on every other table that denormalizes it after a move (`attachments`, `saved_messages`, `user_activity`, `researcher_cache`, `memo_pending_items`, `link_previews.target_stream_id`, `agent_sessions.stream_id`) but didn't touch `shared_messages`. That row has **two** denormalized columns:

- `source_stream_id` — stale when a moved message is the source of one or more shares.
- `target_stream_id` — stale when a moved message is the share message itself (its body contains a `sharedMessage` node).

Concrete failure: after moving a share message into a thread, edits to its source still emit `pointer:invalidated` to the **old** target stream's room (because the broadcaster reads `target_stream_id`), so viewers in the new thread don't get a refresh signal until the next bootstrap.

Fix mirrors the existing `link_previews` block — two `UPDATE shared_messages` statements, one per side.

### 2. Cached `streamId` on `sharedMessage` node attrs goes stale

The pointer node carries `{ messageId, streamId }` baked at share time. `hydrateSharedMessageIds` was using that cached `streamId` for the **truncated** branch past `MAX_HYDRATION_DEPTH = 3`. After a source move, the cache is wrong and the deep-link points to the wrong stream.

Fix: resolve current streamIds for past-cap entries in one batched lookup post-loop. Past-cap entries also now surface `deleted` and `missing` correctly instead of handing out a truncated link to a tombstone or a deleted row.

Cost: one extra `findByIdsInWorkspace` only when the chain hits the cap. Hot path (chains ≤ 3) is unchanged.

### 3. No `pointer:invalidated` fan-out from `messages:moved`

`invalidatePointersForEvent` only handled `message:edited` and `message:deleted`. A move changes `source.streamId` (which hydration returns to clients), so pointer consumers in target streams should re-fetch on `messages:moved` too — even though content/author/createdAt are unchanged, the "open in source stream" link should follow the move.

Fix: extend the extractor to a multi-id form and look up shares for every moved id. Emit one event per `(targetStreamId, sourceMessageId)` pair so each pointer in a room knows which source to refetch — collapsing across sources would force every pointer to refetch on every per-source change.

### Key design decisions

**Stacked on #420, not on main.** Fix 1 modifies `updateStreamScopedReferences`, which is part of the move flow and lives only on the batch-move branch. Targeting main would require carrying the move code in this PR too.

**No move-time content rewrite for the cached `streamId` attr.** The straightforward alternative for fix 2 is to walk every share-message body that references a moved source and rewrite the cached `streamId` attr in place. That would touch every target message's contentJson + emit `message:edited` for each — heavy work for a cache that hydration can refresh at read time anyway. Resolving at hydration is one batched query, off the hot path, and self-correcting after any future move.

**Per-source pointer-invalidate emit.** Clients subscribe to streams, not to sources. Collapsing the fan-out to one emit per target stream (regardless of source) would force every pointer in that room to refetch on every per-source change. The plural form keeps the existing single-source contract — clients can route the cache-bust to the specific pointer that needs it.

## Modified files

| File | Change |
| --- | --- |
| `apps/backend/src/features/messaging/repository.ts` | `updateStreamScopedReferences` re-stamps both columns on `shared_messages` |
| `apps/backend/src/features/messaging/sharing/hydration.ts` | Past-cap entries resolve current streamId via a batched lookup; emit `deleted`/`missing` for moved sources past the cap |
| `apps/backend/src/features/messaging/sharing/outbox-handler.ts` | Extractor returns `string[]`; handle `messages:moved`; emit one event per `(target, source)` pair |

## Tests

| File | Coverage |
| --- | --- |
| `apps/backend/src/features/messaging/repository.test.ts` | Pins both `UPDATE shared_messages` statements with their WHERE clauses; pins the empty-messageIds no-op |
| `apps/backend/src/features/messaging/sharing/hydration.test.ts` | Truncated entry uses the live row's streamId, not the cached attr; deleted/missing past-cap states surface correctly |
| `apps/backend/src/features/messaging/sharing/outbox-handler.test.ts` | `messages:moved` fans out per `(target, source)` pair; share lookup receives every moved id; existing tests updated to include `sourceMessageId` on mock rows |

## Test plan

- [x] `bun test apps/backend/src/features/messaging/` — 74 pass, 0 fail
- [x] `bun run --cwd apps/backend typecheck` — clean
- [x] `bun run --cwd apps/backend lint` — clean
- [ ] Backend integration / E2E — would require a DB; the unit tests pin the SQL shape and the broadcast contract directly.
- [ ] Manual: share a message, then move it into a thread → verify the source-side `shared_messages.source_stream_id` row reflects the destination, and editing the moved source still pushes `pointer:invalidated` to viewers of the share.

## Notes on what's intentionally NOT fixed

- **Privacy boundary re-evaluation on move.** `moveMessagesToThread` always creates the destination thread under the same root, inheriting visibility, so the audience for any moved message is unchanged. `crossesPrivacyBoundary` results from share time stay correct. If the thread-inheritance rule ever changes, that needs to be revisited.
- **`SOURCE_STREAM_MISMATCH` during a concurrent move.** A client racing a move into a fresh share gets a clean 400 from `validateAndRecordShares`. That's a correctness check, not a bug — the client retries with the current attrs.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NbwyBgCwjUXzJkHwKkvMaH)_

### DIFF
--- a/apps/backend/src/features/messaging/repository.test.ts
+++ b/apps/backend/src/features/messaging/repository.test.ts
@@ -1,6 +1,19 @@
 import { afterEach, describe, expect, it, mock, spyOn } from "bun:test"
 import { MessageRepository } from "./repository"
 
+function fakeDb() {
+  const queries: Array<{ text: string; values: unknown[] }> = []
+  const db = {
+    async query(textOrConfig: any, values?: unknown[]) {
+      const text = typeof textOrConfig === "string" ? textOrConfig : textOrConfig.text
+      const vals = typeof textOrConfig === "string" ? (values ?? []) : (textOrConfig.values ?? [])
+      queries.push({ text, values: vals })
+      return { rows: [], rowCount: 0 }
+    },
+  }
+  return { db: db as any, queries }
+}
+
 function makeMessage(overrides: Record<string, any> = {}): any {
   return {
     id: "msg_root",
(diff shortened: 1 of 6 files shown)


kristofferremback:
@coderabbitai review

PR #442, opened 2026-04-28
feat: lossless message copy/paste roundtrip

kristofferremback:
## Summary

Closes the gaps that prevented copying a message and pasting it back into the composer as a working draft. The "Copy as Markdown" context-menu action already wrote `contentMarkdown`; this PR makes the rest of the surfaces lossless too.

### What was broken

1. **Resending a message with attachments returned a 500.** `event-service.ts` validation rejected attachment IDs whose `messageId` was not `null`, and `attachToMessage` only updated rows where `message_id IS NULL`. Resending `[Image #1](attachment:id)` always failed.
2. **`sharedMessage` nodes were silently dropped on in-editor copy.** The frontend's `editor-markdown.ts` lacked a serializer case for `sharedMessage` (only the shared `@threa/prosemirror` package had it), so copying a draft that contained a shared message lost the node.
3. **`shared-message:` URLs weren't reconstructed on paste.** Neither parser turned `Shared a message from [Author](shared-message:streamId/messageId)` back into a `sharedMessage` node, so paste landed it as a plain paragraph + link and the backend share-recording step never saw it.
4. **Native browser Ctrl+C stripped structural URLs.** The rendered DOM intentionally hides the `quote:`/`shared-message:`/`attachment:` URLs, so a select-all + Ctrl+C captured only visible text and paste reconstructed as a flat paragraph.

### What this changes

- **Backend: attachment re-referencing.** `createMessage` now splits `attachmentIds` into "to attach" (still goes through `attachToMessage`) and "to reference" (already linked elsewhere). Re-references are gated by the same read-access chain `getDownloadUrl` already uses (direct stream access → shared-message grant → inline reference). A new `attachment_references` table records every inline usage so recipients of the new message can resolve download access without overwriting `attachments.message_id` on the original. Backfill copies one row per existing attached attachment so the new lookup works for legacy data.
- **Backend: download access chain extended.** `getDownloadUrl`, `AttachmentService.getAccessible`, and the new `isAttachmentReadableByAuthor` consult the references table as a third tier after stream access and shared-message grants — keeps the rule in one place across handler, service, and validator.
- **Markdown roundtrip.** Added the `sharedMessage` serializer case to the frontend `editor-markdown.ts` (mirroring the shared package) and added a `parseSharedMessageLine` rule in both parsers so paste produces a `sharedMessage` node, which then re-triggers backend share validation on send.
- **Native copy of whole messages.** New `useMessageMarkdownCopy` hook attaches a `copy` listener on the rendered markdown wrapper and writes `contentMarkdown` to `text/plain` when the selection covers the whole body. Partial selections fall through to the browser default.

### Design decisions

- **Why a join table over scanning `messages.contentJson`?** Resolving "is this attachment visible to a viewer of stream X?" via JSON traversal would be O(messages × attachments) per request. A small join table indexed on `(attachment_id, message_id)` mirrors how `shared_messages` works and stays cheap as the dataset grows.
- **Why preserve original `message_id` on re-reference?** Overwriting it would orphan the inline `attachment:id` link in every other message that still references the original (e.g. the source message itself). Keeping ownership stable and tracking references separately means resending a message into stream B doesn't break the version of it that lives in stream A.
- **Why scope the copy hook to MarkdownContent only?** The wrapping div is checked against `contentMarkdown` text — including AttachmentList / link previews / context badge would mean the equality check never passes on attachment-bearing messages, silently disabling the lossless path.
- **Why default `authorId: ""` / `actorType: "user"` on parsed `sharedMessage` nodes?** The serializer encodes only `streamId`/`messageId` (the `authorName` comes from the link text). The display path always hydrates the live author via `useSharedMessageSource`, and `ShareService.validateAndRecordShares` only reads `messageId`/`streamId`, so the defaults are fallbacks — they never reach a render.

### File changes

- `apps/backend/src/db/migrations/20260428120000_attachment_references.sql` — new table + backfill
- `apps/backend/src/features/attachments/reference-repository.ts` — new repo with `insertMany`, `findByAttachmentId`, `hasViewerAccessByReference`
- `apps/backend/src/features/attachments/{handlers,service,index}.ts` — chain in references-based access
- `apps/backend/src/features/messaging/event-service.ts` — split validation + record references
- `packages/prosemirror/src/markdown.ts` + `apps/frontend/src/components/editor/editor-markdown.ts` — parse `shared-message:` lines + serialize `sharedMessage` (frontend only)
- `apps/frontend/src/hooks/use-message-markdown-copy.ts` + `apps/frontend/src/components/timeline/message-event.tsx` — native copy override

## Test plan

- [x] `bun test` (apps/backend) — 139 attachment + messaging tests passing, including new cases for re-reference allowed / rejected / inline-reference download access
- [x] `bun run test` (apps/frontend) — full suite (1479 tests) passing, including `editor-markdown` roundtrip + `useMessageMarkdownCopy` whole-vs-partial selection cases
- [x] `bun test ./packages/prosemirror/src/markdown.test.ts` — 20 tests passing, including `sharedMessage` lossless roundtrip
- [x] `bun run typecheck` clean across all packages
- [x] `bun run lint` clean across all apps
- [ ] Manually verify: copy a message with an attachment, paste back into composer, send → no 500
- [ ] Manually verify: copy a message containing a `sharedMessage` node, paste back → composer renders the pointer card (not a plain link)
- [ ] Manually verify: copy a message containing a quote reply via Ctrl+C → paste reconstructs the quoteReply node (not just the snippet text)
- [ ] Manually verify: select half a message, Ctrl+C → browser default text is copied (no regression on partial copies)

https://claude.ai/code/session_01PppQTyVFMzQq4T7N9VaENC

---
_Generated by [Claude Code](https://claude.ai/code/session_01PppQTyVFMzQq4T7N9VaENC)_


kristofferremback:
### Code review
**Confidence: 5/7** — Good

Found 4 issues worth surfacing. None are blockers — the highest-impact ones are an INV-2 ULID-shape inconsistency in the new migration and a missing index for the workspace-scoped query path.

1. `apps/backend/src/db/migrations/20260428120000_attachment_references.sql:36-44` — **INV-2 violation** (Data Model and Persistence Safety: *"All entities use prefixed ULIDs like `stream_xxx`"*). The backfill mints ids as `'aref_' || replace(gen_random_uuid()::text, '-', '')` — a UUID-derived hex string, not a ULID. Runtime code uses `generateId("aref")` which goes through the `ulid` package, so existing backfilled rows will be non-ULID-shaped and break the time-sortability invariant for that prefix.
   https://github.com/kristofferremback/threa/blob/262a1f1d12b1474af33aea9c2f8f5d618d7f7d4c/apps/backend/src/db/migrations/20260428120000_attachment_references.sql#L36-L44

2. `apps/backend/src/features/attachments/reference-repository.ts:71-99` — **Missing index for the workspace-filtered lookup.** `findByAttachmentId` filters on `(workspace_id, attachment_id)` but the migration only declares `attachment_references_pair_idx ON (attachment_id, message_id)`. Postgres can use the leftmost prefix for `attachment_id` but resolves the workspace filter heap-side. `hasViewerAccessByReference` runs in the `getDownloadUrl` hot path on every attachment serve, then issues a second query (`listAccessibleStreamIds`). Either add a composite `(attachment_id, workspace_id)` index or fold both lookups into a single JOIN before this grows.
   https://github.com/kristofferremback/threa/blob/262a1f1d12b1474af33aea9c2f8f5d618d7f7d4c/apps/backend/src/features/attachments/reference-repository.ts#L71-L99

3. `apps/backend/src/features/attachments/service.ts:172-191` — `getAccessible` reimplements the "direct stream OR any reference stream" rule with a caller-supplied `accessibleStreamIds` set, while `hasViewerAccessByReference` encodes the same rule via `listAccessibleStreamIds`. Two paths now encode the same access semantics in different shapes — when the rule changes (thread-root inheritance, share-grant chaining, etc.) only one will be updated.
   https://github.com/kristofferremback/threa/blob/262a1f1d12b1474af33aea9c2f8f5d618d7f7d4c/apps/backend/src/features/attachments/service.ts#L172-L191

4. **Missing test coverage for the changed `getAccessible` method.** The PR added `handlers.test.ts` cases for the download endpoint and `event-service.test.ts` cases for `createMessage`, but `apps/backend/src/features/attachments/service.test.ts` wasn't updated. The new "direct stream OR reference fallback" branch on a security-sensitive method has no direct unit-level coverage. Either add cases here or note they're indirectly covered.

---

- INV-2 ULID prefix invariant — see issue 1. Fix: change the backfill to use a ULID generator. Postgres can call out to `gen_random_uuid()` but ULIDs need an extension or app-side generation; consider an app-script backfill instead of inline SQL, or accept that one-time backfilled rows are exempt from the time-sortability guarantee and document it on the migration.
</details>

- Missing `service.test.ts` coverage for `getAccessible` (issue 4). The PR description's test plan lists "139 attachment + messaging tests passing, including new cases for re-reference allowed / rejected / inline-reference download access" — those land in `handlers.test.ts` and `event-service.test.ts`. The `getAccessible` change is exercised indirectly but not directly.
</details>

Reviewed `deleteAdjacentInlineAtom` / `handleBeforeInputAtomDelete` / `handleBeforeInputKeyboardPaste` (composer side), the access-control fallback chain, and the new pointer-URL parsers. One agent flagged a `\r\n` non-normalisation in the keyboard-paste fallback path (`multiline-blocks.ts:632-635`); on verification the fallback only fires when `getParsedContent` returns empty, which requires whitespace-only post-normalisation input that wouldn't reach this branch via the styling-char heuristic. Not landed.
</details>

- Issue 2 (missing composite index, hot-path two-query fan-out).
- Issue 3 (parallel encoding of the access rule between `getAccessible` and `hasViewerAccessByReference`).
</details>

One agent flagged the removal of `!attachment.streamId` from `getAccessible`'s early-return guard as a defense-in-depth regression. On verification: a pending upload (`streamId === null`) cannot have an `attachment_references` row because refs are only inserted in `createMessage`, which also sets `streamId` via `attachToMessage`. The runtime behavior remains "return null for unanchored uploads." Not a concrete vulnerability — just less-explicit control flow. Worth adding a comment explaining the implicit guarantee, but not a blocker.
</details>

---

PR #443, opened 2026-04-30
[plan] Ariadne lossless message references (forward / quote / resurface attachments)

kristofferremback:
Draft PR carrying the implementation plan only — no code changes yet.

The full plan lives at `.claude/plans/we-recently-merged-a-ethereal-sedgewick.md`. Read it there for the structured form (TOC links, slice anchors, code blocks).

## TL;DR

PR #442 gave human users three pointer-URL formats inside the markdown wire format (`shared-message:`, `quote:`, `attachment:`). Ariadne's pipeline already runs her output through `parseMarkdown` and `ShareService.validateAndRecordShares`, so **forwards and cross-stream quotes already work end-to-end** — but the agent `createMessage` adapter at `apps/backend/src/server.ts:300-324` never extracts `attachmentIds` from the parsed JSON, so the access check + `attachment_references` projection are silently skipped for attachment resurfacings.

Three slices:

1. **Slice A — Backend (must-do, ~10 lines).** New `collectAttachmentReferenceIds` helper in `@threa/prosemirror`; wire it into the `createMessage` and `editMessage` adapters in `server.ts` so attachment references go through the same access gate as user posts. Without this, copy-paste resends of Ariadne's responses fail the access check on recipients without source-stream access.
2. **Slice B — Prompt.** Surface `[msg:… stream:… author:…]` tags in `formatMessagesWithTemporal`, add a "Referring to messages and attachments" section to `system-prompt.ts` with three literal-markdown examples, carve a pointer-URL exception into `discuss-thread.ts:44-47`, expose memo IDs + resolved source-message IDs in `formatRetrievedContext`.
3. **Slice C — `describe_memo` tool.** Thin wrapper over the existing `MemoExplorerService.getById()` so Ariadne can pull source-message IDs out of a memo surfaced by `workspace_research`. Register in `tool-set.ts`, add `DESCRIBE_MEMO` to `AgentToolNames`, enable on Ariadne's built-in persona.

## Key nuances surfaced during planning

- **Most infra already exists.** `parseMarkdown` is called on Ariadne's output, `ShareService.validateAndRecordShares` walks her `contentJson`, memos already carry `sourceMessageIds`. The "flimsy" version (Slice B alone) would *appear* to work but silently break attachment access for anyone outside the source stream.
- **Access scope is already correct.** Ariadne's tool deps carry `accessibleStreamIds` derived from the invoking user, and `isAttachmentReadableViaShareOrReference` accepts persona authorship via the same chain — so loud failure on out-of-scope IDs is the desired behavior, not a leak surface.
- **Same-stream `quoteReply` writes no DB row by design** — purely presentational. Cross-stream quotes go through `shared_messages` with `flavor='quote'`. Ariadne's case is identical to a user's, no special branching needed.
- **`discuss-thread.ts` already says "NEVER include `[msg_…]` ids".** That instruction is currently correct for raw IDs as prose but blocks pointer URLs. The carve-out is *"the structural pointer formats from the system prompt are the exception"* — minimal prompt diff.
- **Attachment numbering.** Image references serialize as `[Image #N](attachment:att_xxx)`. Number images on the emit side in conversation order; renderer reads `imageIndex` from node attrs so any consistent numbering roundtrips.

## Verification

Unit test for `collectAttachmentReferenceIds`, integration test for persona-authored `createMessage` writing the right `attachment_references` / `shared_messages` rows, snapshot test for the new ID tags in formatted history, manual cross-stream forwarding test where the recipient lacks source-stream access (relies on the recorded share grant).

## Out of scope

- @mentions / #channels from Ariadne (parser accepts them but no use case yet)
- Mobile lossless paste of pointer URLs (unrelated; user-side, reverted in #442)
- `describe_memo` write paths (governance discussion, separate change)

https://claude.ai/code/session_01GSsdtLRbnvRDA1QqTVmowp

---
_Generated by [Claude Code](https://claude.ai/code/session_01GSsdtLRbnvRDA1QqTVmowp)_


kristofferremback:
@coderabbitai review summary — addressed in commits `a0610f2` + `0dc123e`:

**Applied:**
- ✅ Persona messages now also carry `[msg:… author:…]` so `quote:` pointer-URL contract roundtrips for self-quotes (`message-format.ts`).
- ✅ Trailing space in no-temporal idTag concatenation (`message-format.ts:47`).
- ✅ `safetyStatus !== CLEAN` check in strip helper, mirroring event-service step 1's malware-scan gate. New `attachment-not-clean` drop reason + test (`strip-inaccessible-refs.ts`).
- ✅ Reference-projection lookups batched via `Promise.all` (`strip-inaccessible-refs.ts:103-122`).
- ✅ `[msg_…]` → `[msg:…]` in discuss-thread preamble.
- ✅ `describe_memo` catch block returns stable generic error; raw `error.message` no longer echoed to model / persisted trace.
- ✅ Pointer-tag prefixes (`msg:`, `stream:`, `author:`, `type:`, `attach:`, `memo:`) centralized in new `apps/backend/src/features/agents/pointer-tags.ts` per INV-33. Both `message-format.ts` and `context-formatter.ts` now build tags through the same helpers.
- ✅ Plan file: corrected access-model wording (scope-restricted `AgentAccessSpec`, not invoking-user-full); added `text` fence language tag for MD040.

**Plus a self-review safety fix not in your comments:** bot-triggered personas (no `invokingUserId`) were forwarding `accessibleStreamIds: undefined`, which bypassed strip and crashed event-service's userId-keyed gate. Now defaults to `[]` so the persona path always uses the set-membership gate (commit `a0610f2`).

**Deferred to follow-up:** the `editMessage` attachment_references refresh (event-service.ts ~587-652) is a real INV-7 gap, but it pre-dates this PR (introduced when the projection table landed in #442) and affects both user and persona edits symmetrically. Out of scope here — this PR's create path now writes those rows correctly via the agent adapter, which is the primary user-visible regression. Will track the edit-time refresh as a separate change that touches `EditMessageParams` for both the user and agent paths.

Test status: 302/303 agents tests pass; the one failure is a pre-existing PersonaRepository batch-resolve flake unrelated to these changes (verified by stashing my diff and re-running).

---

PR #447, opened 2026-04-30
test(activity): synchronize on messageId to fix mention-detection flake

kristofferremback:
## Summary

Fixes the intermittent CI failure on `Activity Feed E2E > Mention Detection > should create activity when user is mentioned` (PR #445 originally surfaced it).

The flake's root cause: `joinStream` writes a `MEMBER_ADDED` activity row for the joiner. Tests then sent a mention message and called `waitForActivity(client, workspaceId)`, which polled for "any row in the feed". The pre-existing `MEMBER_ADDED` row already satisfied that condition, so the helper returned **before** the outbox handler had processed the new mention event. The subsequent `find((a) => a.activityType === "mention")` then returned `undefined` and the assertion failed.

## Approach

Rather than papering over the race with a longer timeout (timeouts always rot), restructure the synchronization so it can't trigger on a stale row:

- Replace `waitForActivity` (any-row poll) with `waitForMessageActivity`, keyed on the `messageId` returned by `sendMessage`.
- Mention/message activity rows store the source message's id in the `messageId` column. `MEMBER_ADDED` rows store a *stream-event* id (a different ULID namespace). The freshly-minted message id can never collide with anything that pre-existed in the feed, so the wait can only resolve when the outbox handler has actually written rows for *this* message.
- `processMessageCreated` runs mention → notification → self-message inserts sequentially, so waiting for the author's self-row is also a valid synchronization point for "no mention should exist for X" assertions (used in the self-mention and non-member tests, replacing two `setTimeout(1500)` hard waits).

## Changes

- `waitForActivity` → `waitForMessageActivity(client, workspaceId, messageId | messageId[], opts?)`
- All 9 call sites updated; multi-message test ("mark all activity as read") passes both ids.
- Two hard `setTimeout(1500)` waits replaced with the deterministic synchronization too.
- No production code changes — pure test stabilization.

## Test plan

- [ ] CI `Tests` job passes on this branch.
- [ ] Re-run the `Tests` job a few times to confirm the flake is gone (the failing test passed once on main after squash-merge of #445, so a single green run is necessary but not sufficient).

🤖 _Authored by [Claude Code](https://claude.com/claude-code)_

### DIFF
--- a/apps/backend/tests/e2e/activity.test.ts
+++ b/apps/backend/tests/e2e/activity.test.ts
@@ -69,25 +69,34 @@ async function markAllActivityAsRead(client: TestClient, workspaceId: string): P
 }
 
 /**
- * Polls until activity items appear for the given member, or times out.
- * The outbox handler processes events asynchronously, so we need to wait.
+ * Polls the activity feed until rows referencing every `messageId` are
+ * visible for `client`'s user, then returns the full feed.
+ *
+ * Activity rows are written asynchronously by the outbox handler, so a
+ * read immediately after `sendMessage` can miss them. Synchronizing on
+ * the freshly-minted message id (a unique ULID) — rather than "any row
+ * in the feed" — sidesteps pre-existing rows like the MEMBER_ADDED that
+ * `joinStream` writes for the joiner: those carry a stream-event id, a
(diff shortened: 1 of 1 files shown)


kristofferremback:
@coderabbitai review

PR #457, opened 2026-05-03
feat: scheduled messages

kristofferremback:
## Summary

Adds scheduled messages — users can queue a message from any composer, manage their pending sends from a dedicated page, and edit them up until the worker fires. Built on the existing cron + queue infrastructure used by saved-message reminders, the existing `EventService.createMessage` send pipeline, and the offline-first Dexie + TanStack stack used by the rest of the app.

Plan: `.claude/plans/schedule-messages-feature-MVdyL.md`

## What's in

**Backend** (`apps/backend/src/features/scheduled-messages/`)
- Migration `scheduled_messages` with workspace-scoped indexes for the To send / Sent / Failed tabs and the per-stream popover. All queries filter by `workspace_id` (INV-8), including primary-key lookups and the worker's CAS.
- Race-safe lock CAS (INV-20) — single UPDATE that takes the editor lock or flips status to `sending` for the worker.
- Service handles schedule / claim / heartbeat / release / update / send-now / cancel / fire (worker entry). The worker invokes `EventService.createMessage` with `clientMessageId: scheduled:<id>` so an interrupted finalize doesn't double-send.
- Past-time PATCH performs the createMessage atomically inside the same tx as the status flip — Save = Send semantics.
- Worker registered on the LIGHT tier in `server.ts`.
- Outbox events `scheduled_message:upserted | sent | cancelled` registered as user-scoped (delivered to the author only).

**Frontend**
- Dexie v27 with a `scheduledMessages` table.
- API client + hooks (`useScheduledList`, `useScheduleMessage`, `useUpdateScheduled`, `useCancelScheduled`, `useSendScheduledNow`, `useClaimScheduled`, `useReleaseScheduled`, `useHeartbeatScheduled`).
- Workspace sync registers the three socket handlers and invalidates `scheduledKeys.all` at the top of `registerWorkspaceSocketHandlers` so the reconnect event-gap closes (INV-53).
- `SchedulePicker` popover next to the composer's send button (reuses the reminder presets).
- `/w/:wsId/scheduled` page with URL-driven tabs (INV-59), inline cancel / send-now / edit, and an edit dialog that claims the lock on open, heartbeats every 30s while open, and releases on close. Past-time edits flip the action labels to Send / Send unchanged.

## Lock semantics

Single CAS on `(edit_lock_owner_id, edit_lock_expires_at, status)`. Whoever wins the CAS owns the next 60s (editor) or 10s (worker) of the row's life. The `sync` hint on `/claim` lets the client UI-block within 30s of `scheduled_for` (with a 300ms wait budget); beyond that the editor opens optimistically.

## Self-review fixes already applied

- Past-time PATCH no longer crashes — flips status to `sending` via the same CAS shape the worker uses, with the editor's lockToken as ownerId.
- Parent message lookup workspace-scopes via the parent's stream (was a leak).
- Edit dialog re-derives `contentJson` from edited markdown before sending PATCH (INV-58).

## What's NOT included (deferred)

- Per-stream composer popover (`ScheduledMessagesPicker`).
- Offline `pendingOperations` queue extensions for `schedule_message`, `cancel_scheduled_message`, `reschedule_scheduled_message` — Journey 11 in the plan.
- `scheduled_message:lock_changed` cross-device awareness event.
- In-stream edit mode on `MessageComposer` (the v1 surface for editing is the page modal).
- Optimistic IDB write on `useScheduleMessage`.

These are tracked in the plan's "Open Decisions" / status checklist for follow-up.

## Test plan

- [x] Backend unit tests: repository (race-safe CAS, workspace scoping, list ordering, idempotent transitions) and service (schedule guard cases, claim threshold logic, cancel-with-queue-tombstone, worker fire path including retry-budget-exhausted, past-time PATCH atomic-send, lock-token verification).
- [x] All 1140 existing backend unit tests still pass (one pre-existing flaky test in `persona-repository.test.ts` unrelated to this branch).
- [x] All 1528 existing frontend tests still pass; `useScheduleMessage` stub added to `message-input.test.tsx` so the new composer wiring doesn't require a ServicesProvider in those tests.
- [x] Typecheck + lint clean across all workspaces.
- [ ] E2E coverage (deferred — happy paths covered by unit + integration tests).
- [ ] Manual QA: schedule from the composer, see it in the page, reschedule, cancel, edit content, edit past-time row to verify Save = Send, click sent row to navigate to live message.

https://claude.ai/code/session_01BU8bmr7a6Tvvg1CcZoMbZp

---
_Generated by [Claude Code](https://claude.ai/code/session_01BU8bmr7a6Tvvg1CcZoMbZp)_


kristofferremback:
## Self-review: reuse + duplication audit

Following @kristofferremback's feedback that this PR added far more code than warranted given the existing patterns, I ran a four-angle audit (reinvented wheels, internal duplication, div soup, platform reusables). Findings + the planned cleanup are recorded here as a paper trail. Phases A, B, and C are landing in this PR; phase D is deferred to follow-ups.

### Phase A — mechanical fixes (landing in this PR)

1. **Stream name resolution.** `scheduled-item.tsx:43-44` and `scheduled-edit-dialog.tsx:411-414` use `streams.find(...)?.displayName ?? "stream"`. The canonical helpers are `getStreamName` + `streamFallbackLabel` from `apps/frontend/src/lib/streams.ts:27,76` (used by quick-switcher, share-message-modal, breadcrumb-ellipsis-dropdown, thread-header, breadcrumb-helpers, stream-panel). Fix: replace with `getStreamName(stream) ?? streamFallbackLabel(stream.type, "generic")`.

2. **Adopt `ResponsiveDialog`.** `apps/frontend/src/components/ui/responsive-dialog.tsx` already implements the Dialog↔Drawer swap with snap points and is used by `share-message-modal`, `stream-settings-dialog`, `workspace-settings-dialog`, `create-channel-dialog`, `document-editor-modal`, `trace-dialog`. `scheduled-edit-dialog.tsx:349-381` rolls its own `useIsMobile()` branch. Fix: drop the manual ladder; use `ResponsiveDialog/Content/Header/Title/Description`. The bespoke `mobileExpanded` toggle (L357 `!h-[100dvh]`) maps onto `DEFAULT_SNAP_POINTS = [0.8, 1]`.

3. **Adopt `<Empty>` primitive.** `scheduled-empty.tsx:27-33` rolls a flex-column. Project standard is `Empty/EmptyHeader/EmptyMedia/EmptyTitle/EmptyDescription` from `components/ui/empty.tsx` (used in `error-view`, `stream-error-view`, `error-boundary`, `thread/stream-panel`, `timeline/stream-content`).

4. **Bug — missing `has-[[data-state=open]]:opacity-100`** on `ScheduledItem`'s desktop hover cluster (`scheduled-item.tsx:192`). `SavedItem` has it (`saved-item.tsx:106-185`); without it, a popover opened from a scheduled row's actions hides its own trigger as soon as the popover opens. Fix: add the variant.

### Phase B — extractions that pay back in this PR and elsewhere (landing)

5. **`<DateTimeField>`.** `apps/frontend/src/components/timeline/reminder-picker-sheet.tsx:158-178` + helpers `:200-208` and `apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx:248-272` + helpers `:385-393` are an exact duplicate (the comment in scheduled-edit-dialog even cites the reminder picker as prior art). Extract to `apps/frontend/src/components/forms/date-time-field.tsx` with API `<DateTimeField value onChange min disabled layout="grid"|"stacked" />`. Both call sites collapse. Not under `components/ui/` since INV-14 reserves that for Shadcn primitives.

6. **`useMentionStreamContext`.** `scheduled-edit-dialog.tsx:406-448` is line-for-line `apps/frontend/src/components/timeline/message-input.tsx:209-249` (same `useStreamBootstrap` for current+root, same workspaceUsers role lookup, same THREAD/rootStreamType branch, same `memberIds`/`botMemberIds`/`canInviteBots` assembly). Extract into `apps/frontend/src/hooks/use-mentionables.ts` (which already owns `MentionStreamContext` at `:22`). Both call sites consume it.

7. **`formatSendCountdown` helper + threshold consolidation.** The `≤1m → "Sending soon"` collapse is duplicated verbatim in `scheduled-item.tsx:56` and `scheduled-messages-picker.tsx:448`: `/^\d+m$/.test(rawLabel) && Number(rawLabel.replace("m","")) <= 1 ? "Sending soon" : rawLabel`. The regex round-trips a string we just formatted. Add `formatSendCountdown(date, now, prefs)` to `apps/frontend/src/lib/dates.ts` next to `formatFutureTime` so the threshold lives in one place.

8. **`toDateTimeLocal` reuse.** `scheduled-messages-picker.tsx:46-50` (`buildLocalDatetimeOneHourAhead`) reimplements `apps/frontend/src/components/timeline/reminder-popover-content.tsx:188-192` (`toDateTimeLocal`). Export `toDateTimeLocal` from `lib/dates.ts` and call `toDateTimeLocal(new Date(Date.now() + 3_600_000))`.

### Phase C — judgement-call cleanups (landing, light variants only)

9. **Unify the action triplet.** `ScheduledRowActions` (hover cluster, `scheduled-item.tsx:185-222`) + `ScheduledActionDrawer` (mobile sheet) + the picker's drawer wiring (`scheduled-messages-picker.tsx:231-245`) all render the same Send-now / Edit / Cancel triplet, with the same `canEdit/canSendNow/onCancel` gating, the same `Pencil`/`Send`/`Trash2` icons, and the same `status === "failed" ? "Remove" : "Cancel"` label split duplicated at `scheduled-item.tsx:215` and `scheduled-action-drawer.tsx:78`. Going with the **light variant**: a single `<ScheduledActions variant="hover-cluster" | "drawer-list" />` driven by an action descriptor. The heavier alternative — modeling these as `MessageAction[]` and reusing `DrawerActionItem` from `message-action-drawer.tsx` — is deferred to phase D because it would force the messaging-actions framework to model lock state and other scheduled-only concerns.

10. **Backend `assertPendingOrThrow` helper.** The find-by-id + "not pending → 409 with SENDING vs NOT_PENDING code split" block appears at `service.ts:222-238` (claim), `:330-348` (update — also inlines lock-token check), `:424-439` (sendNow), `:467-482` (cancel). All but update are byte-identical except the verb. Extract to a private method on the service. ~50 LOC saved.

11. **Backend `publishUpsert` helper.** `const view = toScheduledMessageView(...); await OutboxRepository.insert(client, "scheduled_message:upserted", { workspaceId, targetUserId, scheduled: view })` recurs at `service.ts:178-183, 256-265, 306-311, 411-416, 573-580, 597-604` — six identical pairs. Private `publishUpsert(client, row, userId)` (and `publishCancelled`, `publishSent`) on the service. ~25 LOC saved.

12. **Backend `listByUser` parametrisation.** `repository.ts:194-227` runs two near-identical SELECTs differing only by ORDER BY column/direction and the cursor comparison. Parametrise `orderColumn`, `direction`, `cmp` and use `sql.raw` for the column. ~20 LOC saved.

13. **Backend `enqueueQueuedJob` helper in `lib/queue/`.** `scheduled-messages/service.ts:675-703` (`enqueueSendJob`) and `saved-messages/service.ts:306-348` (`enqueueReminder`) share the build-payload + retry-on-PK-collision loop. Extract to `apps/backend/src/lib/queue/enqueue-with-id-retry.ts` with signature `enqueueQueuedJob(client, { queueName, workspaceId, payload, processAfter, generateId, maxRetries })`. The "write queueId back to the parent row" step stays in the feature since it's table-specific. Both feature services drop ~30 lines.

### Phase D — follow-ups (NOT in this PR, deliberately deferred)

These are real wins but each touches working surface area or other features and deserves its own PR. Tracking them here so they don't get lost:

- **D1 — `useIDBSyncedList<T>` factory.** `apps/frontend/src/hooks/use-scheduled.ts:140-193` and `apps/frontend/src/hooks/use-saved.ts:126-173` are ~80% structural twins. Same `replace*Page` reconcile-by-cached-at, same `useLiveQuery` + Dexie range scan, same `staleTime: Infinity` + `refetchOnReconnect` block (with INV-53 comment), same `isLoading && rows.length === 0` mapping. Proposed factory in `apps/frontend/src/lib/idb-sync.ts`: `useIDBSyncedList<View, Cached, StatusEnum>({ table, keyFor, toCached, fromCached, primaryIndex, fetcher })` returning `{ useList, useLiveCount, persistRows, removeRow, replacePage }`. Real reuse — memos and threads will be next.
- **D2 — `<ListRow>` / `<ListRowSkeleton>` shared geometry.** `SavedItem` and `ScheduledItem` share row layout (header chip + 1-line preview + footer + hover cluster) and skeleton variants. Defer per INV-36 — no third consumer yet, and decomposition without one risks a wrong abstraction.
- **D3 — Adopt the message-actions framework for scheduled rows.** Heavy variant of finding 9. Would model scheduled actions as `MessageAction[]` and reuse `DrawerActionItem` + `groupVisibleActions` from `message-action-drawer.tsx`. Requires extending the framework to model lock state and idempotency keys. Worth doing once a third feature wants identical action semantics.
- **D4 — Generic `tryAcquireLock` CAS in `lib/`.** Currently in `scheduled-messages/repository.ts:298-331`. With N=1 it's premature (saved-messages doesn't take an editor lock). Revisit when the second consumer lands (likely scheduled documents or pinned-AI tasks).
- **D5 — INV-61 candidate.** "When a queue row schedules action against a mutable row, due-time must be re-asserted inside the CAS." This is the `requireDueNow` pattern from `scheduled-messages/repository.ts:292-296` + `service.ts:531-541`. The hazard is a stale leased queue row firing after the row was rescheduled into the future. Worth promoting to an invariant once a second example exists; for now it's a comment in this feature.
- **D6 — `editorBody` IIFE in `scheduled-edit-dialog.tsx:235-347`.** Borderline INV-18 — closes over render state and returns JSX, so it behaves like an inline component. Falls out naturally when adopting `ResponsiveDialog` in phase A; if any of it remains after phase A, lift it to a top-level `ScheduledEditForm`.
- **D7 — `ScheduledMessagesPicker` PickingMode preset/custom branch** (`:367-410`). Two branches in one component (preset list vs custom datetime form). Split into `PresetList` and `CustomTimeForm` siblings. Low-risk decomposition; defer since it's contained.

### Estimated impact of phases A + B + C

- ~80 LOC removed from frontend (action triplet collapse + duplicate date helpers + regex collapse).
- ~95 LOC removed from backend (assertPendingOrThrow + publishUpsert + listByUser parametrise + enqueueQueuedJob extraction).
- One real bug fixed (has-data-state-open opacity rule).
- Two genuinely shared abstractions added (`<DateTimeField>`, `useMentionStreamContext`) that the codebase needs anyway.

Will follow up here when the cleanup commits land.

---

kristofferremback:
## Cleanup landed (phases A + B + C)

Three commits on top of the existing branch — see `4d09079`, `d0579da`, `b9b88b7`.

### Diffstat across the cleanup

```
18 files changed, 622 insertions(+), 557 deletions(-)
```

Net +65 LOC, but that hides ~250 LOC of removed duplication offset by ~315 LOC of new shared abstractions (`<DateTimeField>`, `useMentionStreamContext`, `formatSendCountdown`, `<ScheduledActions>`, `enqueueQueuedJob`). The shared abstractions absorb today's duplicate code AND give other features a place to land.

### What landed per phase

**Phase A** (`4d09079`) — mechanical fixes
- `scheduled-item.tsx` uses `getStreamName` + `streamFallbackLabel` from `lib/streams`; manual `#` prefix dropped (helper handles it for channels)
- `scheduled-edit-dialog.tsx` adopts `ResponsiveDialog`; the bespoke `useIsMobile` ladder + `mobileExpanded` state are gone (vaul snap points handle resize via gesture)
- `scheduled-empty.tsx` uses the project's `<Empty>` primitive
- `ScheduledItem` hover cluster picks up the `has-[[data-state=open]]:opacity-100` rule for parity with `SavedItem`

**Phase B** (`d0579da`) — extractions that pay back here AND elsewhere
- `apps/frontend/src/components/forms/date-time-field.tsx` — `<DateTimeField>` with `density: "compact" | "comfortable"`. `ReminderPickerSheet` and `ScheduledEditDialog` both consume it; the duplicate JSX + helpers are gone.
- `apps/frontend/src/hooks/use-mentionables.ts` — new `useMentionStreamContext` hook. `MessageInput` and `ScheduledEditDialog` both consume it; the line-for-line clone in scheduled-edit-dialog (`useDestinationStreamContext`) is gone.
- `apps/frontend/src/lib/dates.ts` — `toDateInputValue` / `toTimeInputValue` / `toDateTimeLocalValue` / `parseLocalDateTime` / `formatSendCountdown`. Two `pad+slice` clones, one local `buildLocalDatetimeOneHourAhead`, and the verbatim `≤1m → "Sending soon"` regex from two files are all gone.

**Phase C** (`b9b88b7`) — judgment-call cleanups (light variants)
- `apps/frontend/src/components/scheduled/scheduled-actions.tsx` — `<ScheduledActions variant="hover-cluster" | "drawer-list" />`. The icon/label/destructive/onClick descriptors live in one place; `ScheduledRowActions` and `ScheduledActionDrawer` are now thin wrappers around it.
- `service.ts` private `assertPendingOrThrow` (replaces 4 find+check blocks; cancel keeps its own message via the `notPendingMessage` override).
- `service.ts` private `publishUpsert` (replaces 6 outbox-insert pairs).
- `repository.ts` `listByUser` parametrised — pending/non-pending branches collapsed into one query that picks ORDER BY column/direction and the cursor comparison op via `sql.raw` based on status.
- `apps/backend/src/lib/queue/enqueue-with-id-retry.ts` — `enqueueQueuedJob<Q>` shared helper. Both `scheduled-messages` and `saved-messages` use it; the per-feature wrappers do only the table-specific write-back of the queue id.

### Verification

- Typecheck + lint + dockerfile + OpenAPI spec — all pass via the lint-staged pre-commit pipeline (the build pipeline ran on each commit).
- Frontend: 1528/1528 tests pass (only test touched: `message-input.test.tsx` got a `useMentionStreamContext` stub mirroring the existing `useStreamBootstrap` stub pattern).
- Backend: 1148/1149 tests pass; the remaining failure (`PersonaRepository built-in agent config > batch-resolves built-in overrides in findByIds`) is the pre-existing flake noted in the PR description and is unrelated to this branch.
- `scheduled-messages` + `saved-messages` feature suites: 92/92 pass.

### What I did NOT do — phase D follow-ups

Each of these is real reuse but touches surface area beyond this PR's scope. Tracking here so they don't get lost:

- **D1 — `useIDBSyncedList<T>` factory.** `use-scheduled.ts` and `use-saved.ts` are ~80% structural twins (`replace*Page` reconcile-by-cached-at, `useLiveQuery` + Dexie range scan, `staleTime: Infinity` + `refetchOnReconnect` + INV-53 comment). Real reuse — memos and threads will be next consumers.
- **D2 — `<ListRow>` / `<ListRowSkeleton>` shared geometry.** Defer per INV-36 — no third consumer yet.
- **D3 — Adopt the message-actions framework for scheduled rows.** Heavier alternative to phase C1; would require extending the framework to model lock state and idempotency keys.
- **D4 — Generic `tryAcquireLock` CAS in `lib/`.** N=1; revisit on the second consumer.
- **D5 — Promote the `requireDueNow` pattern to an invariant** ("when a queue row schedules action against a mutable row, due-time must be re-asserted inside the CAS"). Wait for a second example.
- **D6 — `editorBody` IIFE in `scheduled-edit-dialog.tsx`.** Fell out partially with `ResponsiveDialog` adoption; what remains is borderline INV-18 — lift to a top-level `ScheduledEditForm` if it grows further.
- **D7 — `ScheduledMessagesPicker` PickingMode preset/custom branch.** Two branches in one component. Low-risk decomposition; defer since it's contained.

Branch is at `b9b88b7`.

---

kristofferremback:
@coderabbitai review

kristofferremback:
## Concurrency redesign + offline-first scheduling landed (`91b3acb`)

Pushed the coordinated F1 + F2 + F3 plan as one commit since the changes are interlocked.

### F1 — offline-first scheduling

The composer no longer blocks on the schedule POST. Pattern mirrors `sendMessage`:

- `useScheduleMessage` writes a `sched_local_<ulid>` placeholder row to `db.scheduledMessages` immediately, enqueues a `schedule_message` operation in `db.pendingOperations`, kicks the queue, returns. UI unblocks the same instant.
- `PendingOperation['type']` extended with `schedule_message` / `update_scheduled_message` / `cancel_scheduled_message` / `send_scheduled_now`.
- `operation-queue.ts` executor handles each type. The schedule executor runs `replaceLocalScheduledRow(placeholderId, serverRow)` in a single Dexie transaction so the live query never observes a frame with both rows.
- `SyncEngine` accepts a new `scheduledService` dep so the queue can replay ops on reconnect.
- Workspace-sync's `scheduled_message:upserted` handler now sweeps any placeholder sharing the row's `clientMessageId` — closes the race where the socket event arrives before the executor returns.
- Cancel and send-now keep their `try-then-enqueue` pattern (online → fast path; offline → enqueue, retry on reconnect).

The user-visible win: scheduling is instant. The whole "I'm unable to use the app until the scheduled message is persisted" complaint is gone.

### F2 — drop the exclusive editor lock; worker-fence + first-save-wins

**Migration** `20260504123556_scheduled_messages_fence_lock.sql`:
- `DROP COLUMN edit_lock_owner_id`
- `RENAME edit_lock_expires_at → edit_active_until`

**Repository:**
- `tryAcquireLock` / `heartbeatLock` / `releaseLock` deleted.
- `bumpEditFence` (anonymous, no owner) — used by both claim and heartbeat.
- `tryStartSend` (worker-only) — CAS with `status = 'pending' AND scheduled_for <= NOW() AND (edit_active_until IS NULL OR edit_active_until <= NOW())`.
- `update` CAS now requires `updated_at = expectedUpdatedAt`. First save wins; second save's `UPDATE … RETURNING` returns zero rows → service throws `409 SCHEDULED_MESSAGE_STALE_VERSION`.

**Service:**
- `claim` is read-only + bump fence. Never 409s on lock contention. Two devices can both open the editor.
- `heartbeat` takes no token, just bumps the fence.
- `release` endpoint deleted entirely.
- Worker `fire` defers (with `reschedule: true`) when `edit_active_until` is in the future.
- Past-time save still does atomic Save = Send via `tryStartSend` + `finalizeSendInTx`.

**Wire format:**
- `ScheduledMessageView` drops `editLockOwnerId` / `editLockExpiresAt`; gains `editActiveUntil` + `clientMessageId`.
- `UpdateScheduledMessageInput` drops `lockToken`; requires `expectedUpdatedAt`.
- `ClaimScheduledMessageResponse` drops `lockToken` / `lockExpiresAt` / `sync`; returns just `{ scheduled, editActiveUntil }`.
- `SCHEDULED_MESSAGE_SYNC_LOCK_THRESHOLD_SECONDS` constant deleted (no more sync lock; the dialog opens optimistically).

**Frontend:**
- `useReleaseScheduled` deleted.
- The dialog tracks `expectedUpdatedAt` instead of `lockToken`. On save, `409 SCHEDULED_MESSAGE_STALE_VERSION` surfaces a "edited elsewhere — refreshing…" toast and closes the dialog; the next list refresh pulls the latest content.
- Cross-device: open device A and device B; both see the editor; first save wins; second device sees the toast.

The user-visible win: the "I can't edit because the lock is stuck" complaint is gone. There's no exclusive lock to be stuck.

### F3 — clone `MessageEditForm` for the modal composer

Replaced the bespoke `ResponsiveDialog` ladder + 5-deep `flex-1 min-h-0` chain with the proven `MessageEditForm` shape:

- Mobile: `<Drawer>` with `max-h-[85dvh]` (or `!h-[100dvh]` when `mobileExpanded`).
- Editor wrapper: `flex-1 min-h-0 overflow-y-auto` with `[&_.tiptap]:max-h-none [&_.tiptap]:min-h-full`. Long messages fill the drawer instead of scrolling inside a ~288px cap.
- Action bar pinned at the bottom of the drawer (separate sibling, not stuffed inside the bordered card).
- Save / Cancel sit in `EditorActionBar`'s trailing slot at `h-8 px-3` — same shape as `MessageEditForm`. No more 360px-phone truncation.
- `mobileExpanded` toggle restored on the action bar — vaul drag is no longer the only way to grow the drawer.
- `autoFocus={!isMobile}` stays so the keyboard doesn't pop before the user reviews date/time.
- Local placeholder rows (`sched_local_…`) seed the editor from IDB rather than calling `claim` (which would 404). Save is disabled on placeholders with a "Scheduling… try again in a moment" hint.

The user-visible win: the modal is usable. Editor takes the full viewport on mobile, Save button is reachable, OS native selection menu works, quote-reply atoms have a tap target after them.

### Verification

- Frontend: 1528 / 1528 tests pass.
- Backend: 1144 / 1145 (the 1 failure is the pre-existing flaky `PersonaRepository built-in agent config` test, unrelated to this branch).
- Lint + typecheck + dockerfile + OpenAPI green via the lint-staged pre-commit pipeline.

### Diffstat for this commit

```
21 files changed, 878 insertions(+), 755 deletions(-)
```

### What's not included (deliberate)

- `update_scheduled_message` op type is reserved but the executor throws — the editor surfaces `STALE_VERSION` synchronously rather than retrying silently. If we want offline edits to queue, that's a follow-up that needs a UX call (do we replay with the user's stale content, or merge?).
- The dialog's `claim` still blocks on the network for non-local rows. That's fine because it's just an immediate "GET + bump fence" round-trip and the user is actively opening a modal — a brief loading state is expected. If the network is bad, the dialog falls back to "Could not start editing" and the user can retry.
- The fence is anonymous, so a malicious client could keep the worker out indefinitely by heartbeating a row that's not theirs — but `bumpEditFence` is workspace-scoped (INV-8) and the existing auth middleware gates the route by user, so they can only stall their own rows. Acceptable.

---

kristofferremback:
@coderabbitai resume

kristofferremback:
@coderabbitai review

PR #468, opened 2026-05-05
feat: link-based workspace invitations

kristofferremback:
## Summary

Adds a second invitation kind alongside the existing email-first flow. Admin generates a single-use link with a role + optional admin-only note, ships it through any channel (Signal, SMS, etc.), and the recipient binds their email at claim time. From the claim onward we hand off to WorkOS via the existing `sendInvitation` rails so verification semantics are preserved.

Plan: [`docs/plans/link-based-invitations.md`](docs/plans/link-based-invitations.md)

## Why

Today, inviting someone to a workspace is "type their email, wait for WorkOS to email them." That's the wrong fit for the smallest version of the product — for a two-person scratchpad, a five-person hobby project, a startup that hasn't bought a domain yet, the natural way to invite someone is to message them: "drop your email here." This feature inverts the flow so admins can do exactly that.

WorkOS does **not** ship this primitive (verified via docs + API ref — `email` is required on `sendInvitation`), so we build the link layer ourselves and hand off to WorkOS at claim time.

## Architecture

```
admin creates link (regional)
  → workspace_invitations row { kind='link', email=NULL, token_hash, note }
  → outbox 'invitation:link-created'
  → shadow-sync mirrors to CP shadow

friend opens /join/<token>  (CP, unauthenticated)
  → GET /api/invitations/lookup?token=…  → { workspaceName, expiresAt }
                                            (never email/role/note/inviter)

friend submits email
  → POST /api/invitations/claim { token, email }
  → CP forwards to regional via internal API (CP knows region from shadow)
  → regional atomic single-use claim: UPDATE … SET email = $1
       WHERE token_hash = $2 AND status = 'pending' AND email IS NULL
       AND expires_at > NOW() RETURNING …
  → regional emits 'invitation:link-claimed'
  → CP shadow-sync triggers WorkOS sendInvitation(email, organizationId, role)

friend gets WorkOS email, signs in
  → existing /api/auth/callback → existing acceptPendingForEmail rails
  → workspace member created
```

- **Single use** is enforced by the existing `status` lifecycle + the atomic claim's `WHERE` clause (INV-20 — no select-then-act).
- **Role** is carried on the link at create time, exactly like email invites today.
- **Note** is admin-only — never returned by the public lookup endpoint.
- **Token storage**: only `sha256(token)` is persisted. Plaintext is returned exactly once on creation.

## What changed

**Schema (both regional + CP)** — append-only migrations add `kind TEXT NOT NULL DEFAULT 'email'`, nullable `email`, `token_hash`, and (regional only) `note`. Partial unique index on `token_hash`.

**Backend (regional)** — `InvitationService.createLink` + `claimLinkByToken`, two new outbox events, new admin route `POST /api/workspaces/:id/invitations/links`, new internal route `POST /internal/invitations/claim-link`. `tokenHash` is stripped from wire responses.

**Control-plane** — new public unauthenticated endpoints `GET /api/invitations/lookup` and `POST /api/invitations/claim` (rate-limited via `authLimit`). `RegionalClient.claimInvitationLink` forwards to the right region. `InvitationShadowService.acceptLinkClaim` triggers WorkOS `sendInvitation` after claim.

**workspace-router** — two new path patterns route the public link surface to CP.

**Frontend (designed to match the existing app style — same shells, fonts, spacing, borders as `LoginPage` / `WorkspaceSelectPage` / `UserSetupPage`)**:

- Workspace settings → Members tab: "Invite" is now a `DropdownMenu` with "Invite by email" and "Create invite link" options (`Mail` / `LinkIcon` icons). Pending list renders both kinds in the same `rounded-md border` row container, with kind-aware icons, role badges, expiry, and the admin note rendered as `text-xs text-muted-foreground`. Per-row actions: "Resend"/"Revoke" for email kind, "Copy link"/"Revoke" for link kind. The plaintext token is held in a parent-scoped `ref` so it can be re-copied until page refresh; once cleared, the action shows "Link sent" in a disabled state.
- New `CreateInviteLinkDialog` — form view (role + optional note), then a swap to a "copy link" view with a one-time read-only display + check ✓ animation.
- New public `/join/:token` page — lazy-loaded, lives outside `WorkspaceLayout`. Three states: lookup loading → email-input form → submitted ("check your inbox" or "you're already a member"). Friendly error states for not-found / revoked / expired / already-claimed.

## Invariants applied

INV-1 (no FKs), INV-3 (no DB enums; `kind` is TEXT validated in code), INV-4 (cross-service via outbox), INV-6 (services own transactions), INV-8 (workspace_id on all writes), INV-17 (append-only migration), INV-20 (atomic single-use claim, no select-then-act), INV-30/INV-41 (no DB connection held during WorkOS calls), INV-32 (`HttpError` with `code`), INV-55 (Zod for inputs), INV-14/INV-40 (Shadcn primitives, links for nav, buttons for actions).

## Test plan

- [x] `bun run typecheck` — clean across the monorepo
- [x] `bun test apps/backend/src/features/invitations/service.test.ts` — 26/26 (17 existing + 9 new for `createLink` and `claimLinkByToken`)
- [x] `bun test apps/workspace-router` — 35/35
- [x] `bun run test` (frontend) — 1543/0
- [ ] Manual: admin creates a link → copies → opens in fresh incognito → submits email → receives WorkOS verification → signs in → lands on workspace setup
- [ ] Manual: open the same link twice — second attempt shows "Link already used"
- [ ] Manual: revoke a link before claim → recipient sees "Invitation revoked"
- [ ] Manual: pending list renders both email and link invites with correct icons + actions

## Out of scope (future, not blocking this PR)

- **Multi-use links** (`max_uses` column + per-recipient audit trail).
- **WorkOS webhook ingester.** We currently have no inbound WorkOS event channel. The link flow doesn't need one — membership is created when the user signs in and our app sees them — but a webhook receiver for `invitation.accepted` / `user.created` / `organization_membership.created` would unlock eager membership creation, better revoke sync, and richer per-claim audit (e.g. tracking the claiming WorkOS user ID independent of the email join). Filed as a follow-up because it's its own architectural addition.
- Optional name field on the join page that pre-fills `UserSetupPage`.

https://claude.ai/code/session_01F4z5rTxvJJ298nugVCfpug


kristofferremback:
@coderabbitai review

PR #471, opened 2026-05-06
fix: emit stream:member_added for thread parent message author

kristofferremback:
## Problem

When person A writes a message in a private channel and person B starts a thread on it, person A (the parent message author) is correctly added to `stream_members` in the database — but they never receive a `stream:member_added` socket event.

The frontend's `stream:created` handler only adds private streams to the sidebar for the thread creator (`isCreator` check in `workspace-sync.ts`). Other members are expected to be notified via `stream:member_added`. Without that event being emitted, the parent message author:

- Never sees the thread appear in their sidebar in real-time
- Must reload the page to discover the thread they should be aware of
- Miss real-time messages sent in the thread

Public threads are unaffected (they're visible to all via `visibility = 'public'`). The bug only impacts private threads in private channels or DMs.

## Solution

Replace the raw `StreamMemberRepository.insert()` call for the parent message author in `createThread` with `this.addToStream()` — the same internal method already used by `addMember`.

`addToStream` does three things the raw insert was skipping:
1. Creates a `member_added` timeline event in the thread (sets the read cursor so it's not shown as unread)
2. Emits a `stream:member_added` outbox event

The broadcast handler delivers `stream:member_added` to both the thread's socket room **and** the added user's personal room (`ws:{workspaceId}:user:{memberId}`), so the author receives it even before subscribing to the thread socket room. The frontend `handleStreamMemberAdded` handler then adds the stream to their sidebar and subscribes them to the room.

### Key design decisions

**1. Use `addToStream` instead of a raw insert**

`addToStream` is already used by `addMember` and encapsulates the correct membership + notification pattern. The alternative (manually emitting the outbox event after the insert) would duplicate logic and risk drift.

**2. Don't use `addToStream` for the thread creator**

The creator is already notified via the `stream:created` event, which triggers sidebar addition + socket subscription in the frontend. Using `addToStream` for them would produce a redundant `member_added` timeline entry in the thread.

## Modified files

| File | Change |
| --- | --- |
| `apps/backend/src/features/streams/service.ts` | Replace raw `StreamMemberRepository.insert` with `this.addToStream` for parent message author in `createThread` |
| `apps/backend/src/features/streams/service.test.ts` | Add 4 tests for `createThread` membership and notification behavior |

## Test plan

- [x] Author (≠ creator) receives `stream:member_added` outbox event
- [x] Author = creator: no `stream:member_added` emitted (creator path unchanged)
- [x] Bot-authored parent message: no `stream:member_added` emitted
- [x] `stream:created` is always emitted to parent stream room
- [x] All 16 service unit tests pass
- [x] TypeScript type-checks pass (pre-commit hook)

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/471"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/streams/service.test.ts
+++ b/apps/backend/src/features/streams/service.test.ts
@@ -6,6 +6,7 @@ import { StreamMemberRepository } from "./member-repository"
 import { StreamEventRepository } from "./event-repository"
 import { OutboxRepository } from "../../lib/outbox"
 import { UserRepository } from "../workspaces"
+import { MessageRepository } from "../messaging"
 import * as idModule from "../../lib/id"
 import * as db from "../../db"
 import { HttpError } from "../../lib/errors"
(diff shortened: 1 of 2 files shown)


kristofferremback:
@coderabbitai review

PR #498, opened 2026-05-11
feat: WorkosAuthzAdminService (Phase 2 / PR-5a)

kristofferremback:
## Problem

Phase 2 of the WorkOS authorization rollout needs write paths that translate workspace-owner intent ("promote Alice to admin", "remove Bob") into WorkOS organization-membership mutations. The original PR-5 attempted to land write paths, backoffice UI, owner backfill, regional API surface, and invitation `role_slug` migration in one ~6.7k-LOC change — too large to review safely. We're splitting it into five mergeable sub-PRs; this is the first.

See `docs/plans/phase2-pr5-split.md` for the full split (5a → 5e) and dependency graph.

## Solution

Introduce the **`WorkosAuthzAdminService`** in the control plane plus three new methods on `WorkosOrgService`. No callers yet — handlers and regional fan-out land in 5b/5d. This PR is purely the service layer, its dependencies, and an integration-test harness, so we can land write paths in isolation before wiring anything up.

### Key design decisions

**1. `AdminActor { workosUserId, isPlatformAdmin }` instead of a boolean flag.**

Two surfaces will call this service: the backoffice (platform admins acting on any workspace) and the regional API (workspace owners acting on their own). Passing the actor as a value object lets both share the same data-integrity guards (last-owner, self-demote) while only the backoffice bypasses the owner-action gate. The owner-action gate is defense-in-depth — handlers still apply `requireWorkspacePermission('workspace:owner')` middleware first. The service is the second line.

**2. `SELECT COUNT` for last-owner guard, not `listByOrganization` + filter.**

A workspace with hundreds of members shouldn't hydrate every row to answer "is there another owner?". Added `WorkosAuthzRepository.countByRoleExcludingUser` — a single bounded `SELECT COUNT(*)::text` filtered by role and excluding the target user.

**3. SDK calls key by membership id, not `(orgId, userId)`.**

`updateOrganizationMembership` and `deleteOrganizationMembership` in the WorkOS SDK take a membership id. The mirror has it (`organization_membership_id`), so the service reads the mirror first to resolve the id, then mutates WorkOS. The mirror is the source of truth for "what does WorkOS currently believe?"; mutations flow through WorkOS, and the event poller replays them back into the mirror asynchronously. Callers see eventual consistency, not read-your-write.

**4. No transaction on the admin service.**

The local mirror is read-only on this path (reads to enforce guards). The actual mutation lives in WorkOS. There's nothing to wrap in a transaction. INV-6 still applies — services own transactions when they exist — but this one doesn't need one.

**5. Stub realism.**

`StubWorkosOrgService.ensureOrganizationMembership` previously logged-and-returned. Made it actually populate `membershipsByOrg` so the integration tests can drive the full WorkOS round-trip locally. Verified no existing test relied on the no-op behaviour.

## New files

| File | Purpose |
| --- | --- |
| `apps/control-plane/src/features/workos-authz/admin-service.ts` | `WorkosAuthzAdminService` — `assignRole`, `changeRole`, `removeMember` with owner-action / last-owner / self-demote guards |
| `apps/control-plane/tests/integration/workos-authz-admin-service.test.ts` | Integration tests against real Postgres + `StubWorkosOrgService`; covers happy paths and every guard rejection (FORBIDDEN, NOT_FOUND, LAST_OWNER, SELF_DEMOTE, INVALID_ROLE) |
| `docs/plans/phase2-pr5-split.md` | Sub-PR split plan (5a–5e) with dependencies and merge order |

## Modified files

| File | Change |
| --- | --- |
| `packages/backend-common/src/auth/workos-org-service.ts` | Add `ensureOrganizationMembership`, `changeOrganizationMembershipRole`, `removeOrganizationMembership` to the interface and `WorkosOrgServiceImpl` (wraps `userManagement.{create,update,delete}OrganizationMembership`) |
| `packages/backend-common/src/auth/workos-org-service.stub.ts` | Make the three methods actually mutate the in-memory store; extract `findMembershipById` helper |
| `apps/control-plane/src/features/workos-authz/repository.ts` | Add `countByRoleExcludingUser` — bounded `SELECT COUNT(*)` for the last-owner guard |
| `apps/control-plane/src/features/workos-authz/index.ts` | Barrel exports (INV-52) |

## Test plan

- [x] `apps/control-plane/tests/integration/workos-authz-admin-service.test.ts` — 14 cases covering every guard outcome and happy path
- [x] Type-check passes locally
- [x] Lint passes locally
- [ ] CI: integration tests run against Postgres
- [ ] Manual: spin up CP locally with stub WorkOS, exercise the three methods via a temporary handler — deferred to 5b when the handler lands

## Follow-ups

- **5b** — Backoffice UI + handler (`POST/PATCH/DELETE /api/backoffice/workspaces/:id/members[/:userId]`) calling this service with `isPlatformAdmin: true`
- **5c** — Owner backfill (existing workspaces' creators get the `owner` role slug) + new-workspace creator flip from `admin` to `owner`
- **5d** — Regional API surface (`/api/workspace/members/*`) + CP internal endpoint for fan-out
- **5e** — Invitation `role_slug` end-to-end

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/498"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/apps/control-plane/src/features/workos-authz/admin-service.ts
+++ b/apps/control-plane/src/features/workos-authz/admin-service.ts
@@ -161,11 +161,12 @@ export class WorkosAuthzAdminService {
   }
 
   private async assertNotLastOwner(organizationId: string, targetUserId: string): Promise<void> {
-    const rows = await WorkosAuthzRepository.listByOrganization(this.pool, organizationId)
-    const remainingOwners = rows.filter(
-      (r) => r.workos_user_id !== targetUserId && r.role_slugs.includes(WORKSPACE_ROLE_SLUGS.OWNER)
-    )
-    if (remainingOwners.length === 0) {
+    const remainingOwners = await WorkosAuthzRepository.countByRoleExcludingUser(this.pool, {
+      workosOrganizationId: organizationId,
+      roleSlug: WORKSPACE_ROLE_SLUGS.OWNER,
+      excludeWorkosUserId: targetUserId,
+    })
+    if (remainingOwners === 0) {
       throw new HttpError("Cannot leave the workspace without an owner", {
         status: 422,
         code: "LAST_OWNER",
--- a/apps/control-plane/src/features/workos-authz/admin-service.ts
+++ b/apps/control-plane/src/features/workos-authz/admin-service.ts
@@ -1,0 +1,181 @@
+import type { Pool } from "pg"
+import { HttpError, logger, type WorkosOrgService } from "@threa/backend-common"
+import { WORKSPACE_ROLE_SLUGS, WORKSPACE_USER_ROLES, type WorkspaceRoleSlug } from "@threa/types"
+import { WorkosAuthzRepository, type WorkosOrgMembershipRow } from "./repository"
+
+interface Dependencies {
+  pool: Pool
+  workosOrgService: WorkosOrgService
+}
+
+/**
+ * The user performing an admin action. `isPlatformAdmin` lets the backoffice
+ * surface bypass the owner-action gate while still respecting data-integrity
+ * guards (last-owner, self-demote). Regional surfaces pass
+ * `isPlatformAdmin: false`; their `requireWorkspacePermission('workspace:owner')`
+ * middleware is the first gate and this service is the second.
+ */
+export interface AdminActor {
+  workosUserId: string
+  isPlatformAdmin: boolean
+}
+
+export interface AssignRoleParams {
+  actor: AdminActor
+  organizationId: string
+  targetUserId: string
+  roleSlug: WorkspaceRoleSlug
+}
+
+export interface ChangeRoleParams {
+  actor: AdminActor
+  organizationId: string
+  targetUserId: string
+  roleSlug: WorkspaceRoleSlug
+}
+
+export interface RemoveMemberParams {
+  actor: AdminActor
+  organizationId: string
+  targetUserId: string
+}
+
+/**
+ * Write paths for WorkOS organization memberships. Reads the mirror to enforce
+ * data-integrity guards and delegates the actual WorkOS mutation to
+ * `WorkosOrgService`. Mirror updates land asynchronously through the regular
+ * event poller — callers see eventual consistency, not read-your-write.
+ *
+ * INV-6: service owns the orchestration; no transaction is required because
+ * the local mirror is read-only on this path.
+ */
+export class WorkosAuthzAdminService {
+  private pool: Pool
+  private workosOrgService: WorkosOrgService
+
+  constructor({ pool, workosOrgService }: Dependencies) {
+    this.pool = pool
+    this.workosOrgService = workosOrgService
+  }
+
+  async assignRole(params: AssignRoleParams): Promise<void> {
+    assertKnownRole(params.roleSlug)
+    await this.assertActorMayManage(params.actor, params.organizationId)
+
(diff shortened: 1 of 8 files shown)


kristofferremback:
Pushed a race-fix and switched this PR to draft. Wanted to surface what it addresses since the bug isn't visible in the existing tests.

**The hole:** the last-owner and actor-is-owner guards read from the local mirror (`workos_organization_memberships`), which is poller-fed and async. Two concurrent demote calls would each see the pre-write state, both pass the guard, and WorkOS would end up with zero owners. The canonical scenario is two co-owners demoting each other simultaneously — credit to whoever flagged this in conversation; it's a clean INV-20 violation that the current tests don't exercise.

**The fix** (commit 65668ff1):

1. New `withOrganizationAdminLock` — per-org `pg_try_advisory_lock` on a dedicated pool client, polled with a 10s wait budget; returns 503 `ADMIN_LOCK_HELD` on contention so a stuck caller can't block requests indefinitely. The lock client sits idle holding the kernel-level lock; no DB or pool work happens on it while held.
2. Rewrote `assignRole` / `changeRole` / `removeMember` to derive every guard from a single `listOrganizationMemberships` snapshot taken **inside** the lock. WorkOS is the source of truth on the write path; the mirror is no longer consulted (it remains the read surface for the backoffice etc.).
3. Added two integration tests:
   - Two co-owners demoting each other in parallel → exactly one succeeds, the other gets `LAST_OWNER`, ending state has one owner.
   - Platform admin removing both co-owners in parallel → same shape.

**Why advisory locks vs. the time-based lease pattern used elsewhere** (`WorkosEventPollerLock`, `CursorLock`): leases are for long-lived leader election. This is per-request short-lived mutual exclusion where postgres's built-in lock queue is the right primitive. The `pg_advisory_lock` rejection in `apps/backend/docs/distributed-cron-design.md:1167-1170` is specifically scoped to cron scheduling; it doesn't apply here.

**INV-41 note:** the lock-holding connection is bound for the duration of the WorkOS call. This is a deliberate narrow exception, documented in `org-admin-lock.ts`. Admin writes are low-frequency human actions; WorkOS membership writes are sub-second p50. The alternative — outbox + compensating transactions — is significantly more code for the same effect, and has a brief observable window of zero-owner state which the lock approach avoids.

Will re-open once green.

kristofferremback:
CI is green on the race-fix commits — flipped back to ready for review.

Summary of what changed since the original review:
- `withOrganizationAdminLock` (per-org `pg_try_advisory_lock`, 10s wait budget, 503 `ADMIN_LOCK_HELD` on contention).
- `assignRole` / `changeRole` / `removeMember` now derive every guard from a single `listOrganizationMemberships` snapshot taken **inside** the lock. WorkOS is source of truth on the write path; mirror is no longer consulted there.
- Two integration tests cover the concurrent path: mutual-demote (loser hits `FORBIDDEN` because the winner already demoted them) and platform-admin co-owner removal (loser hits `LAST_OWNER`). Both demonstrate the race is closed (exactly one owner remains).

Ready for another pass when you have time.

---

PR #507, opened 2026-05-13
fix(timeline): stop async cache flips and 0px composer footer from jumping Virtuoso

kristofferremback:
## Problem

When loading a stream with dynamic content — code blocks, blockquotes, link previews, attachment images — Virtuoso shifts on-screen content around for several frames after first paint. The bug is reliably reproducible on hard refresh: scrolled-to-end position becomes unstable while sibling rows resize underneath. UX is a core pillar of the app; the timeline is one of the most-used surfaces, and this jumping made the whole experience feel jittery.

Two compounding sources, both kicking in on the first commit and resolving asynchronously after Virtuoso's height bookkeeping has already settled:

**1. Async IDB collapse-state flip.** `useBlockCollapse` (code blocks, blockquotes, quote-replies) and `useLinkPreviewCollapse` (link preview cards) read persisted toggle state through Dexie's `useLiveQuery`. `useLiveQuery` returns `undefined` synchronously and resolves the persisted value on a later microtask. The hook fell back to `defaultCollapsed` (or `false` for previews) on first paint, then flipped to the persisted value once IDB resolved — every persistently-toggled card or fenced block resized after the timeline had already drawn itself. Inside Virtuoso's `stickToBottom` mode that translates to cascading row shifts.

**2. Composer footer growing from 0 → real height.** The timeline's footer spacer reserves space for the floating composer pill via `var(--composer-height, 0px)`. The variable was only set inside the composer's `ResizeObserver` callback (a `useEffect`), so first paint always rendered the spacer at 0px and then grew it to ~80px after the effect ran. Virtuoso compensates for footer growth by shifting content up.

### Sync, hydrated-at-boot collapse cache

A module-level `Map<string, boolean>` mirrors the `markdownBlockCollapse` and `linkPreviewCollapse` IDB tables. `hydrateCollapseCache()` runs once from `main.tsx` before React mounts; consumers read via `useSyncExternalStore` so the very first render reflects persisted state. Toggles update the in-memory map synchronously and write through to IDB.

### Persisted composer height applied to `:root` pre-mount

Every ResizeObserver measurement is mirrored to `localStorage`, and `applyPersistedComposerHeight()` reads it back and applies it as a global `--composer-height` fallback on `:root` before React mounts. The editor zone's local override still kicks in once the composer mounts, but the first paint already approximates the right reservation instead of starting at 0px.

### Key design decisions

**1. Boot-time hydration in `main.tsx`, not lazy hook-level reads.**

Lazy IDB reads inside the hook would either keep the first-paint flip (the original bug) or require Suspense plumbing the timeline doesn't otherwise need. Front-loading the single bulk read at app boot is cheaper than a network round-trip for workspace bootstrap, so by the time the timeline first renders the cache is reliably ready.

**2. `useSyncExternalStore`, not Zustand or a custom subscription.**

`useSyncExternalStore` is the React-built-in primitive for exactly this case (a module-scoped store with external mutation). It handles concurrent rendering and tearing detection without needing a context provider or another dependency.

**3. Cross-tab Dexie change feed deliberately dropped.**

`useLiveQuery` subscribed to Dexie's change feed, which means toggling collapse state in tab A propagated to tab B. The new cache hydrates once and doesn't re-read IDB. That's an intentional trade — collapse state is a per-tab UX preference, and the live subscription is exactly what made the first paint unstable. The trade-off is documented inside `collapse-cache.ts` so the next reader doesn't reach for `useLiveQuery` and reintroduce the bug.

**4. Hydration loop preserves in-flight writes.**

A user toggle that lands between hydration's `toArray()` and its loop would otherwise be overwritten by the stale snapshot. The hydration loop skips keys already present in the map so a recent in-memory write always wins over the IDB read it raced. Practical risk is tiny — Dexie hydration finishes well before the user can interact — but the guard is one line.

**5. Composer height stored at `:root`, not on `[data-editor-zone]`.**

CSS variable inheritance means the editor zone's own override (set by the composer's ResizeObserver) wins for descendants within the zone, while `:root` carries the persisted fallback elsewhere. Same `var(--composer-height, 0px)` consumer, two layers of overrides, no JSX changes needed.

**6. Image attachment width reflow and "Show more" button toggle intentionally not fixed.**

Image attachment width grows from 128px (loader) to whatever the natural width is, and the link-preview "Show more" button only renders once `useLayoutEffect` measures `scrollHeight`. Both fall into "legitimately shifts when content is received the first time" — your carve-out — and would each need their own pass (an IDB-backed image dimensions cache, or measuring before commit). Out of scope here.

## New files

| File                                                | Purpose                                                                                                                                    |
| --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
| `apps/frontend/src/lib/markdown/collapse-cache.ts`  | Module-level sync mirror of the two collapse IDB tables. Exposes `hydrateCollapseCache`, `setBlockCollapse`/`setLinkPreviewExpand`, and `useBlockCollapseStore`/`useLinkPreviewExpandStore` via `useSyncExternalStore`. |
| `apps/frontend/src/lib/composer-height-storage.ts`  | `applyPersistedComposerHeight` (called before React mounts) + `persistComposerHeight` (called from the ResizeObserver). Reads/writes a single `localStorage` value with sanity bounds. |

## Modified files

| File                                                          | Change                                                                                                                                                |
| ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
| `apps/frontend/src/lib/markdown/use-block-collapse.ts`        | Swapped `useLiveQuery` for `useBlockCollapseStore`; toggles call `setBlockCollapse` directly. Public hook signature unchanged.                       |
| `apps/frontend/src/hooks/use-link-preview-collapse.ts`        | Swapped `useLiveQuery` for `useLinkPreviewExpandStore`; toggles call `setLinkPreviewExpand`. Public hook signature unchanged.                        |
| `apps/frontend/src/hooks/use-composer-height-publish.ts`      | Inside the ResizeObserver write path, also mirrors the height to `localStorage` via `persistComposerHeight`.                                          |
| `apps/frontend/src/main.tsx`                                  | Calls `hydrateCollapseCache()` and `applyPersistedComposerHeight()` before `createRoot()`.                                                            |
| `apps/frontend/src/test/setup.ts`                             | Global `beforeEach` resets the collapse cache so module-scoped state doesn't leak between tests.                                                      |
| `apps/frontend/src/lib/markdown/code-block.test.tsx`          | Pre-seeded test now calls `await hydrateCollapseCache()` after the IDB put so the new sync cache picks up the seed.                                   |
| `apps/frontend/src/lib/markdown/blockquote-block.test.tsx`    | Same pattern as code-block test.                                                                                                                      |

## Test plan

- [x] `bun run typecheck` — repo-wide, clean
- [x] `bunx vitest run` (full frontend suite) — 140 files, 1594 tests passing
- [x] Targeted run of the four collapse-related suites (`code-block`, `blockquote-block`, `quote-reply-block`, `use-link-preview-collapse`) — 24/24 passing
- [ ] Hard-refresh the current stream with a mix of long code blocks, expanded link previews, and image attachments; confirm no visible row jumps after first paint
- [ ] Toggle a collapsed code block, refresh, confirm the persisted state still applies on first paint (no flip)
- [ ] Verify composer height is preserved across reload (footer spacer doesn't briefly collapse to 0)

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01UWmTEwHwL3nNG1se1AiwdK)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/507"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/hooks/use-composer-height-publish.ts
+++ b/apps/frontend/src/hooks/use-composer-height-publish.ts
@@ -1,4 +1,5 @@
 import { useEffect } from "react"
+import { persistComposerHeight } from "@/lib/composer-height-storage"
 
 /**
  * Measures the element referenced by `ref` and publishes its height (in px) as
@@ -10,7 +11,11 @@ import { useEffect } from "react"
  * Pass `active: false` (e.g. while the expand-to-fullscreen overlay is open)
  * to disconnect the observer. The CSS variable is intentionally *not* cleared
  * on cleanup so that stream navigation preserves the last-known height; the
- * next composer mount overwrites it with its own measurement.
+ * next composer mount overwrites it with its own measurement. The same value
+ * is also mirrored to localStorage so the next hard refresh starts with a
+ * sensible global default on `:root` instead of falling back to 0px — that
+ * fallback grew the timeline's footer spacer mid-paint and caused Virtuoso
+ * to shift content up on every reload.
  */
 export function useComposerHeightPublish(
   ref: React.RefObject<HTMLElement | null>,
@@ -24,7 +29,9 @@ export function useComposerHeightPublish(
     if (!zone) return
 
     const write = (h: number) => {
-      zone.style.setProperty("--composer-height", `${Math.ceil(h)}px`)
+      const px = Math.ceil(h)
+      zone.style.setProperty("--composer-height", `${px}px`)
+      persistComposerHeight(px)
     }
 
     write(el.getBoundingClientRect().height)
--- a/apps/frontend/src/hooks/use-link-preview-collapse.test.ts
+++ b/apps/frontend/src/hooks/use-link-preview-collapse.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, beforeEach } from "vitest"
 import { renderHook, act, waitFor } from "@testing-library/react"
 import { useLinkPreviewCollapse } from "./use-link-preview-collapse"
+import { hydrateCollapseCache, __resetCollapseCacheForTests } from "@/lib/markdown/collapse-cache"
 import { db } from "@/db"
 
 describe("useLinkPreviewCollapse", () => {
@@ -32,6 +33,11 @@ describe("useLinkPreviewCollapse", () => {
     expect(stored?.previewId).toBe("preview_1")
 
     unmount()
+
+    // Drop the in-memory cache and rehydrate from IDB so the remount exercises
+    // the actual persisted-state restore path, not just the still-warm Map.
+    __resetCollapseCacheForTests()
+    await hydrateCollapseCache()
 
     const { result: remounted } = renderHook(() => useLinkPreviewCollapse("msg_1", "preview_1"))
     await waitFor(() => {
--- a/apps/frontend/src/hooks/use-link-preview-collapse.ts
+++ b/apps/frontend/src/hooks/use-link-preview-collapse.ts
@@ -1,6 +1,5 @@
 import { useCallback } from "react"
-import { useLiveQuery } from "dexie-react-hooks"
-import { db } from "@/db"
+import { setLinkPreviewExpand, useLinkPreviewExpandStore } from "@/lib/markdown/collapse-cache"
 
 export interface LinkPreviewCollapseState {
   expanded: boolean
@@ -15,29 +14,23 @@ export interface LinkPreviewCollapseState {
  * messages. Mirrors the pattern used by `useBlockCollapse` for collapsible
  * markdown blocks.
  *
+ * Reads are synchronous via the shared `collapse-cache` so the first paint
+ * already reflects the persisted state — important inside the timeline so
+ * Virtuoso doesn't see preview cards resize after mount.
+ *
  * A missing `messageId` (e.g. tests or transient render contexts) disables
  * persistence — toggles become no-ops and `canToggle` reports false.
  */
 export function useLinkPreviewCollapse(messageId: string | undefined, previewId: string): LinkPreviewCollapseState {
   const id = messageId ? `${messageId}:${previewId}` : null
 
-  const persistedOverride = useLiveQuery(async () => {
-    if (!id) return undefined
-    const row = await db.linkPreviewCollapse.get(id)
-    return row?.expanded
-  }, [id])
+  const persistedOverride = useLinkPreviewExpandStore(id)
 
   const expanded = persistedOverride ?? false
 
   const toggle = useCallback(() => {
     if (!id || !messageId) return
-    void db.linkPreviewCollapse.put({
-      id,
-      messageId,
-      previewId,
-      expanded: !expanded,
-      updatedAt: Date.now(),
-    })
+    setLinkPreviewExpand(id, messageId, previewId, !expanded)
   }, [id, messageId, previewId, expanded])
 
   return {
--- a/apps/frontend/src/lib/composer-height-storage.ts
+++ b/apps/frontend/src/lib/composer-height-storage.ts
@@ -1,0 +1,73 @@
+/**
+ * Persists the last-measured composer height and applies it as a global
+ * fallback CSS variable on `:root` before the editor zone mounts.
+ *
+ * Why: the timeline's footer spacer reserves vertical space for the floating
+ * composer pill via `var(--composer-height, 0px)`. That variable is set by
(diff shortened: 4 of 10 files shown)


kristofferremback:
### Code review

**Confidence: 5/7** — Good (a few non-blocking findings, two fixed in `9453d97`)

**Review models:** Orchestrator: Claude Opus | Reviewers: sonnet x3

Three findings from spec-compliance, correctness/security, and design passes. Two have been addressed in the follow-up commit `9453d97`; one is left as a noted-but-not-recommended design observation.

1. `apps/frontend/src/lib/markdown/collapse-cache.ts` — module-level singletons (`blockCollapse`, `linkPreviewExpand`, `hydrated`, listener Sets) introduced without an `INV-9 exception:` comment, while `apps/frontend/src/lib/draft-promotions.ts:8` and `apps/frontend/src/hooks/use-link-preview-dismissals.ts:14` establish the convention. CLAUDE.md [INV-9, *Architecture Boundaries and Dependency Flow*]: "No hidden singletons (except approved logger, Langfuse/OTEL, and web-push bootstrap constraints)". **Fixed in `9453d97`** — banner now carries an explicit `INV-9 exception:` note.
   https://github.com/threahq/threa/blob/749adcb3f23274cc05fc0f62eeb85c3c9e8ad3bc/apps/frontend/src/lib/markdown/collapse-cache.ts#L5-L25

2. `apps/frontend/src/hooks/use-link-preview-collapse.test.ts:18-40` — the "persists … and survives remount" test wasn't updated to match the pattern applied to `code-block.test.tsx` and `blockquote-block.test.tsx`. With the new module-level cache the in-memory `Map` survives the unmount, so the test passes trivially without ever exercising the IDB → hydration restore path. A bug where `setLinkPreviewExpand` quietly drops the IDB write would go undetected. **Fixed in `9453d97`** — the test now calls `__resetCollapseCacheForTests()` + `await hydrateCollapseCache()` between unmount and remount.
   https://github.com/threahq/threa/blob/749adcb3f23274cc05fc0f62eeb85c3c9e8ad3bc/apps/frontend/src/hooks/use-link-preview-collapse.test.ts#L18-L40

3. `apps/frontend/src/lib/markdown/collapse-cache.ts` — the two stores (`blockCollapse` + `blockListeners` + `useBlockCollapseStore`, `linkPreviewExpand` + `linkPreviewListeners` + `useLinkPreviewExpandStore`) are structural near-duplicates of the same `Map<string, boolean>` + listener `Set` pattern. INV-35/INV-37 argue for consolidating into one factory. **Not applied.** CLAUDE.md also says "Three similar lines is better than a premature abstraction"; with only two instances and semantically distinct booleans (`collapsed` vs `expanded`), a factory would force readers to mentally invert one of the two flags at every call site. Flagging for awareness if a third store is ever added.
   https://github.com/threahq/threa/blob/749adcb3f23274cc05fc0f62eeb85c3c9e8ad3bc/apps/frontend/src/lib/markdown/collapse-cache.ts#L27-L131

---

No plan file backed this PR; all PR-body-documented changes are implemented and exercised.

</details>

No correctness issues at score ≥80. `useSyncExternalStore` invariants hold (snapshot returns primitives from a `Map.get`; listeners detach on unmount). `localStorage` is bounds-checked and try/catch-wrapped. Hydration's `finally`+`notify` correctly wakes subscribers after `toArray()`. Module-state idempotency is preserved across StrictMode double-invocation.

</details>

<details><summary>📋 CLAUDE.md Compliance — 1 violation, fixed</summary>

INV-9 exception comment missing on a new module-level singleton — fixed in `9453d97`.

</details>

<details><summary>🏗️ Design — 1 noted observation</summary>

Parallel implementations for the two boolean stores. Left unconsolidated by design; see #3 above.

</details>

No user-controlled data reaches a dangerous sink. `localStorage` numerics are `parseInt`+bounds-checked before being interpolated into a CSS variable; IDB rows are boolean-typed and never flow into HTML.

</details>

---

PR #513, opened 2026-05-13
fix: register HorizontalRule extension so paste of --- does not silently fail

kristofferremback:
## Problem

Pasting markdown content that includes `---` (horizontal rules) into the message composer silently fails — nothing appears. This happens because `parseMarkdown` correctly produces `{ type: "horizontalRule" }` nodes, but the TipTap editor has no `HorizontalRule` extension registered. When `insertContent` tries to create nodes from the parsed JSON, `schema.nodeType("horizontalRule")` throws `RangeError: Unknown node type: horizontalRule`.

TipTap's `createNodeFromContent` catches this error internally (when `errorOnInvalidContent` is false) and returns empty content. The command then succeeds with nothing to insert, and the paste handler calls `event.preventDefault()` thinking it handled the paste. The user sees nothing pasted — no error, no feedback.

## Solution

Import and register `@tiptap/extension-horizontal-rule` in the editor extensions alongside other block-level extensions (Blockquote, CodeBlock, etc.).

### How it works

- The `parseMarkdown` parser already handles `---` → `{ type: "horizontalRule" }` correctly (with round-trip tests)
- The serializer already handles `horizontalRule` → `---` correctly
- The only gap was the missing extension registration in the editor schema
- Adding the extension means `insertContent` can create the node, and paste works end-to-end

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/package.json` | Added `@tiptap/extension-horizontal-rule` dependency |
| `apps/frontend/src/components/editor/editor-extensions.ts` | Imported and registered `HorizontalRule` extension |
| `bun.lock` | Lockfile update |

## Test plan

- [x] TypeScript typecheck passes across all packages
- [x] All 104 markdown editor tests pass (including horizontal rule parsing, serialization, and round-trip tests)
- [x] All lint checks pass
- [x] Pre-existing frontend date test failures are timezone-related and unrelated

## Notes

No new CSS, keyboard shortcuts, or parser changes needed — the TipTap extension handles rendering and the existing prose styling covers `<hr>` elements.

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/513"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/.claude/plans/fix-markdown-copy-paste.md
+++ b/.claude/plans/fix-markdown-copy-paste.md
@@ -1,0 +1,38 @@
+# Fix: Markdown paste silently fails when content contains horizontal rules
+
+## Goal
+
+Fix a bug where pasting markdown content containing `---` (horizontal rules) into the message composer results in nothing being pasted — the entire paste content is silently dropped.
+
+## What Was Built
+
+### Added HorizontalRule extension to the TipTap editor
+
+`parseMarkdown` from `@threa/prosemirror` correctly parses `---` into `{ type: "horizontalRule" }` nodes. However, the TipTap editor had no `HorizontalRule` extension registered, so when `insertContent` tried to create nodes from the parsed JSON, `schema.nodeType("horizontalRule")` threw `RangeError: Unknown node type: horizontalRule`.
+
+TipTap's `createNodeFromContent` catches this error internally and returns empty content (when `errorOnInvalidContent` is false). The command then succeeds with nothing to insert, and the paste handler calls `event.preventDefault()` thinking it handled the paste. The user sees nothing pasted.
+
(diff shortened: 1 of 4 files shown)


kristofferremback:
Closed in favor of #512 despite this being a better solution. I'm having codex take the fix from this PR and apply its test that it added. I'm actually really impressed with DSv4-Flash here

PR #516, opened 2026-05-13
feat(backoffice): per-workspace members re-sync button

kristofferremback:
## Problem

When the regional permission mirror drifts (event poller wedge, missed WorkOS event, fan-out stalled), the existing remedy was a full re-runnable backfill script across every org. That's a CLI, blast-radius-wide operation; mobile operators can't trigger it, and it does far more work than the situation usually needs.

Concrete case that prompted this: a PAT was 401-ing with `OWNER_INACTIVE` because the regional `workspace_user_permissions` row was missing even though the WorkOS membership was live. The owner was online, the WorkOS dashboard showed them active — the mirror just hadn't caught up. With no per-workspace refresh available, the only fix was the global backfill script.

## Solution

A small operator surface that re-runs the existing backfill for a single workspace, exposed in the backoffice as a button on the workspace Members tab.

### Backend

- `WorkosAuthzBackfill.runForOrganization(workosOrganizationId)` — per-org variant that reuses the same transactional reconcile + outbox path as `run()`. Refactor extracts the common body into a private `runForOrganizationInternal`; `run()` becomes a loop over `WorkspaceRegistryRepository.listWorkosOrganizationIds()`. Does **not** stamp `last_backfill_at` — a single-org refresh isn't equivalent to a full backfill and shouldn't unblock the first-boot guard for other orgs.
- `BackofficeService.resyncWorkspaceMembers(workspaceId)` resolves workspace → org id, throws `404 NOT_FOUND` if the workspace doesn't exist and `400 NOT_LINKED` if the workspace has no WorkOS org, then delegates to the backfill. Server wiring constructs `BackofficeService` after `authzBackfill` so it can be injected.
- `POST /api/backoffice/workspaces/:id/members/resync`, behind `requirePlatformAdmin`. Returns `{ membershipsUpserted, membershipsRemoved }` — structured, no display strings (INV-46).

### Frontend

- "Re-sync members" button in the Members `Section` header (uses the existing `actions` slot), with success/error banner. Disabled when the workspace isn't linked to WorkOS or while a re-sync is in flight. Success banner reports the upsert/remove counts; the empty-diff case says "Already up to date — no changes."
- Self-review extracted two duplicated helpers — `InlineBanner` and `readApiError` — that the resync page would have copy-pasted from `invite-workspace-owner.tsx`. `InlineBanner` → `components/inline-banner.tsx`, `readApiError` → next to `ApiError` in `api/client.ts`. Updates both pages.

### Key design decisions

**1. Per-org backfill emits outbox events unconditionally**
The operator-triggered case wants a "definitively re-sync regional state" effect, so we always emit `OUTBOX_AUTHZ_MEMBERSHIP_CHANGED` for every membership returned by WorkOS (and `_REMOVED` for reconciled rows), not just for diffs. Matches `run()`'s behavior — no special-casing.

**2. No new lock around per-org resync**
The repository's timestamp-guarded upsert (INV-20) already protects against races with the regular event poller — backfill stamps `last_event_at = now()`, the poller has its own `>` guard on incoming events. A fresh advisory lock for the operator path would be redundant.

**3. Does NOT stamp `last_backfill_at`**
The first-boot guard in `server.ts` uses `last_backfill_at` to decide whether to run the global backfill. A per-org re-sync can't substitute for that, so leaving the timestamp untouched is the safe default.

## New files

| File | Purpose |
| --- | --- |
| `apps/backoffice/src/components/inline-banner.tsx` | Shared success/error banner |

## Modified files

| File | Change |
| --- | --- |
| `apps/backoffice/src/api/backoffice.ts` | `resyncWorkspaceMembers` + `ResyncWorkspaceMembersResult` |
| `apps/backoffice/src/api/client.ts` | Adds `readApiError(error)` next to `ApiError` |
| `apps/backoffice/src/pages/workspace-detail-members.tsx` | Re-sync button + success/error banner + mutation |
| `apps/backoffice/src/pages/invite-workspace-owner.tsx` | Uses shared `InlineBanner` + `readApiError` |
| `apps/control-plane/src/features/workos-authz/backfill.ts` | New public `runForOrganization`; `run()` refactored to loop |
| `apps/control-plane/src/features/workos-authz/index.ts` | Re-export the new result type |
| `apps/control-plane/src/features/backoffice/service.ts` | `resyncWorkspaceMembers`; constructor takes `authzBackfill` |
| `apps/control-plane/src/features/backoffice/handlers.ts` | New handler |
| `apps/control-plane/src/routes.ts` | New route under `requirePlatformAdmin` |
| `apps/control-plane/src/server.ts` | `BackofficeService` built after `authzBackfill` |
| `apps/control-plane/tests/e2e/backoffice.test.ts` | Resync endpoint cases (401/403/404/200 shape) |
| `apps/control-plane/tests/integration/workos-authz-backfill.test.ts` | `runForOrganization` cases (targets named org only, reconciles deletes) |

## Test plan

- [x] `bun run typecheck` across all 9 packages
- [x] `bun run lint` + Dockerfile workspace check + OpenAPI check (pre-commit hooks)
- [ ] CI: new `runForOrganization` integration tests
- [ ] CI: new `POST /api/backoffice/workspaces/:id/members/resync` e2e tests
- [ ] Manual: platform admin → workspace detail → Members tab → "Re-sync members" → banner shows counts; member list refreshes
- [ ] Manual: hit `POST .../members/resync` with a non-admin session → 403
- [ ] Manual (the original bug): with a regional mirror missing a known-good WorkOS member, click re-sync, then retry a PAT call → no more `OWNER_INACTIVE`

---


kristofferremback:
@coderabbitai review

PR #519, opened 2026-05-13
feat(backoffice): poll outbox event status for resync progress

kristofferremback:
## Problem

Follow-up to #516. When a platform admin clicks "Re-sync members" on a workspace, the handler responds the moment the backfill commits, and the banner says "Re-synced N memberships" — even though fan-out to the regional backends is still pending in the outbox. If the dispatcher dead-letters one of the events, the operator never finds out from this UI; they just see a green banner and have to know to go check logs.

## Solution

Return the outbox event ids from the resync handler and add a small status endpoint the frontend can poll. The banner updates progressively while events drain ("3 of 5 propagated") and ends in one of three terminal states: fully propagated, partially propagated after timeout, or dead-lettered.

### How it works

1. `WorkosAuthzBackfill.runForOrganization` now captures the rows returned by `OutboxRepository.insertMany` and exposes their ids as decimal strings on the result.
2. The resync HTTP handler passes those ids through to the client.
3. New `OutboxRepository.getEventStatuses(listenerId, ids)` reports each id as `processed | pending | dead_lettered`:
   - **processed** if the id is at-or-below the listener cursor, OR present in the sliding-window dedup map.
   - **dead_lettered** if a row exists in `outbox_dead_letters` for the same `(listener_id, outbox_event_id)`. DLQ wins over cursor, because `CursorLock.moveFirstEventToDLQ` advances the cursor past poison events while also writing to the DLQ — without the DLQ check we'd falsely report it as processed.
   - **pending** otherwise.
4. New `GET /api/backoffice/outbox-events/status?ids=...` (max 200) returns `{ statuses: [{ id, status }] }`. Hardcoded to the `control-plane` listener — the CP today only runs one outbox listener, and clients shouldn't need to know that name.
5. The members page polls every 1.5s for up to 15s after a successful resync and renders one of:
   - "No changes — mirror was already up to date." (zero events)
   - "Propagating to regional backends: X of N processed."
   - "Fan-out timed out with N pending. Pending events will continue to retry in the background."
   - "1 event failed to propagate (dead-lettered). Check the dead-letter queue."
   - "Re-synced N memberships. Fan-out propagated to all regional backends."

### Key design decisions

**1. Single shared `OutboxRepository.getEventStatuses` instead of a CP-local query.** The cursor + sliding-window semantics live in `packages/backend-common`; replicating that logic in the CP handler would drift the moment we touch dedup. The repository method takes a `listenerId` so a future page (regional admin?) can reuse it.

**2. DLQ overrides cursor, not the other way around.** When `moveFirstEventToDLQ` advances the cursor past a poison event, the cursor check would falsely report "processed". The repo runs both queries in parallel and lets the DLQ result win.

**3. Two-track stop conditions for polling.** `refetchInterval` returns `false` once everything is terminal — that handles the happy path. A separate `setTimeout` flips `didTimeout` after 15s so the banner can show a distinct "fan-out timed out" message instead of staying in "Propagating…" forever. The button stays disabled while polling so the operator doesn't pile up resyncs.

**4. BigInt ids serialise as decimal strings on the wire.** Outbox ids are `bigint` in the DB. Frontend treats them as opaque strings; the Zod schema on the endpoint regex-validates each as `^\d+$` and the service re-parses with `BigInt(...)` so a non-numeric id returns 400 instead of crashing pg.

**5. Hardcoded listener id.** The handler doesn't take a listener-id query param. The CP runs exactly one listener; exposing it to the client would be a footgun (different listeners → different cursors → different "processed" answers).

## New files

| File | Purpose |
| --- | --- |
| `apps/control-plane/src/lib/outbox-listeners.ts` | Single-source constant `CONTROL_PLANE_LISTENER_ID` so the server bootstrap and the backoffice service agree on the listener name. |
| `apps/control-plane/tests/integration/outbox-event-statuses.test.ts` | Integration tests for `OutboxRepository.getEventStatuses`: cursor, sliding window, DLQ-override, missing listener, mixed-order preservation. |

## Modified files

| File | Change |
| --- | --- |
| `packages/backend-common/src/outbox/repository.ts` | Add `getEventStatuses` + `OutboxEventStatus` / `OutboxEventProcessingStatus` types. |
| `packages/backend-common/src/outbox/index.ts`, `packages/backend-common/src/index.ts` | Re-export the new types. |
| `apps/control-plane/src/features/workos-authz/backfill.ts` | `runForOrganization` returns `outboxEventIds: string[]` from `insertMany`. |
| `apps/control-plane/src/features/backoffice/service.ts` | Add `getOutboxEventStatuses(eventIds)`: validates as BigInt, hardcodes listener. |
| `apps/control-plane/src/features/backoffice/handlers.ts` | Add Zod-validated `getOutboxEventsStatus` handler (max 200 ids). |
| `apps/control-plane/src/routes.ts` | Register `GET /api/backoffice/outbox-events/status` behind `requirePlatformAdmin`. |
| `apps/control-plane/src/server.ts` | Use shared `CONTROL_PLANE_LISTENER_ID` constant. |
| `apps/backoffice/src/api/backoffice.ts` | Add `outboxEventIds` to `ResyncWorkspaceMembersResult`, add `getOutboxEventsStatus`. |
| `apps/backoffice/src/pages/workspace-detail-members.tsx` | Poll status, render progressive banner with 5 states, disable button while polling. |
| `apps/control-plane/tests/integration/workos-authz-backfill.test.ts` | Assert `outboxEventIds` shape. |
| `apps/control-plane/tests/e2e/backoffice.test.ts` | Update resync test for new return shape; 5 new tests for the status endpoint (401, 403, empty, 400 invalid, 200 happy path). |

## Test plan

- [x] `bun run typecheck` across all workspaces
- [x] `bun run lint` across all workspaces
- [x] All 153 control-plane tests pass (`bun run test` in `apps/control-plane`), including 7 new integration tests for `getEventStatuses` and 5 new e2e tests for the status endpoint
- [ ] Manual: trigger resync on a workspace, observe banner progress while regional backends drain the outbox
- [ ] Manual: deliberately break a regional fan-out, trigger resync, confirm dead-lettered terminal state surfaces in the banner

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01WLeMwKzjMr8iztp8do67dP)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/519"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/apps/backoffice/src/api/backoffice.ts
+++ b/apps/backoffice/src/api/backoffice.ts
@@ -159,12 +159,21 @@ export function resyncWorkspaceMembers(id: string): Promise<ResyncWorkspaceMembe
     .then((r) => r.result)
 }
 
-export function getOutboxEventsStatus(ids: string[]): Promise<OutboxEventStatus[]> {
-  if (ids.length === 0) return Promise.resolve([])
-  const params = new URLSearchParams({ ids: ids.join(",") })
-  return api
-    .get<{ statuses: OutboxEventStatus[] }>(`/api/backoffice/outbox-events/status?${params.toString()}`)
-    .then((r) => r.statuses)
+const OUTBOX_STATUS_BATCH_SIZE = 200
+
+export async function getOutboxEventsStatus(ids: string[]): Promise<OutboxEventStatus[]> {
+  if (ids.length === 0) return []
+  const batches: string[][] = []
+  for (let i = 0; i < ids.length; i += OUTBOX_STATUS_BATCH_SIZE) {
+    batches.push(ids.slice(i, i + OUTBOX_STATUS_BATCH_SIZE))
+  }
+  const responses = await Promise.all(
+    batches.map((batch) => {
+      const params = new URLSearchParams({ ids: batch.join(",") })
+      return api.get<{ statuses: OutboxEventStatus[] }>(`/api/backoffice/outbox-events/status?${params.toString()}`)
+    })
+  )
+  return responses.flatMap((r) => r.statuses)
 }
 
 export function listWorkspaceInvitations(id: string): Promise<WorkspaceInvitation[]> {
--- a/apps/backoffice/src/api/backoffice.ts
+++ b/apps/backoffice/src/api/backoffice.ts
@@ -67,6 +67,19 @@ export interface WorkspaceMember {
 export interface ResyncWorkspaceMembersResult {
   membershipsUpserted: number
   membershipsRemoved: number
+  /**
+   * Decimal-encoded outbox event ids emitted by the re-sync. Empty when the
+   * mirror was already up to date. Poll {@link getOutboxEventsStatus} with
+   * these to surface fan-out progress.
+   */
+  outboxEventIds: string[]
+}
+
+export type OutboxEventProcessingStatus = "processed" | "pending" | "dead_lettered"
+
+export interface OutboxEventStatus {
+  id: string
+  status: OutboxEventProcessingStatus
 }
 
 export interface WorkspaceInvitation {
@@ -144,6 +157,14 @@ export function resyncWorkspaceMembers(id: string): Promise<ResyncWorkspaceMembe
       result: ResyncWorkspaceMembersResult
     }>(`/api/backoffice/workspaces/${encodeURIComponent(id)}/members/resync`)
     .then((r) => r.result)
+}
+
+export function getOutboxEventsStatus(ids: string[]): Promise<OutboxEventStatus[]> {
+  if (ids.length === 0) return Promise.resolve([])
+  const params = new URLSearchParams({ ids: ids.join(",") })
+  return api
+    .get<{ statuses: OutboxEventStatus[] }>(`/api/backoffice/outbox-events/status?${params.toString()}`)
+    .then((r) => r.statuses)
 }
 
 export function listWorkspaceInvitations(id: string): Promise<WorkspaceInvitation[]> {
--- a/apps/backoffice/src/pages/workspace-detail-members.tsx
+++ b/apps/backoffice/src/pages/workspace-detail-members.tsx
@@ -260,10 +260,11 @@ function ResyncBanner({
   const pending = total - processed - deadLettered
 
   if (deadLettered > 0) {
+    const eventsWord = `event${total === 1 ? "" : "s"}`
     return (
       <InlineBanner tone="error">
-        Re-sync committed ({summary}) but fan-out failed for {deadLettered} of {total} event
-        {total === 1 ? "" : "s"} — check control-plane logs for the dead-letter queue.
+        Re-sync committed ({summary}) but fan-out failed for {deadLettered} of {total} {eventsWord}
+        {isPolling ? " so far — still checking the rest." : " — check control-plane logs for the dead-letter queue."}
       </InlineBanner>
     )
   }
--- a/apps/backoffice/src/pages/workspace-detail-members.tsx
+++ b/apps/backoffice/src/pages/workspace-detail-members.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query"
 import { useParams } from "react-router-dom"
 import { RefreshCw } from "lucide-react"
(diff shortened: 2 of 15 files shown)


kristofferremback:
### Code review

Found 3 issues:

1. `apps/backoffice/src/pages/workspace-detail-members.tsx:262-269` — Dead-letter banner branch fires before the `isPolling` guard. If one event dead-letters early while others are still pending, the operator sees a terminal-looking "fan-out failed for X of N events" message even though polling is ongoing, and the count can grow as more events dead-letter. Two reasonable fixes: gate the error branch on `!isPolling` so we wait for the final count, or change the copy to acknowledge ongoing polling (e.g. "X events failed so far — still checking the rest"). (design: count is understated during active polling)
   https://github.com/threahq/threa/blob/6ea9f76ad3db1c6ccd5b4be720ba8dd323e96a5b/apps/backoffice/src/pages/workspace-detail-members.tsx#L260-L270

2. `apps/control-plane/tests/e2e/backoffice.test.ts:394` — Raw `'control-plane'` string literal in the test SQL instead of importing `CONTROL_PLANE_LISTENER_ID`. The constant was added in this PR specifically to centralise the listener name; using it here keeps the e2e test in lock-step with production if the id ever changes. (CLAUDE.md [INV-33, Code hygiene]: "Avoid magic strings by centralizing constants at the source of truth.")
   https://github.com/threahq/threa/blob/6ea9f76ad3db1c6ccd5b4be720ba8dd323e96a5b/apps/control-plane/tests/e2e/backoffice.test.ts#L392-L397

3. `apps/control-plane/src/db/migrations/002_outbox.sql:40` — `outbox_dead_letters` has only a single-column index on `listener_id`. The new `getEventStatuses` query filters on `listener_id = $1 AND outbox_event_id = ANY($2::bigint[])`. With one control-plane listener, the index doesn't narrow, so the planner heap-scans every DLQ row for that listener after the index lookup. A composite `(listener_id, outbox_event_id)` index makes it an exact match. DLQs usually stay small, so this is borderline — flag is mostly forward-looking. (performance: missing composite index for new query pattern)
   https://github.com/threahq/threa/blob/6ea9f76ad3db1c6ccd5b4be720ba8dd323e96a5b/apps/control-plane/src/db/migrations/002_outbox.sql#L32-L41

---

All pieces match the PR description: `runForOrganization` → outbox ids → handler → API client types → frontend polling → 5-state banner. No missing corresponding changes (e.g. types are exported from `packages/backend-common` barrels).
</details>

Agent 2 looked at: bigint comparison semantics, listener-row-missing handling (returns all pending), polling stop conditions, useEffect cleanup, state branch reachability, BigInt() reachability vs Zod regex, order preservation, race between mutation success and first poll. The only concrete bug ideas were sub-80 (e.g. a one-extra-fetch closure-staleness in `refetchInterval`).
</details>

<details><summary>📋 CLAUDE.md Compliance [1 issue]</summary>

INV-33 magic string — see item #2 above.

Checked against INV-20 race-safe writes (queries are read-only, no concurrency hazard), INV-30 pool-vs-withClient (single-query path uses `pool` directly), INV-32 HttpError/Zod (Zod schema with `.regex(/^\d+$/).max(200)`, `BackofficeServiceError` → 400 in service), INV-34 thin handlers, INV-51/52 colocation (status endpoint sits in `features/backoffice`), INV-55 Zod input validation (done), INV-46 structured backend responses (status enum, no display strings), INV-59 multi-view URL (not applicable — overlay UI on existing page).
</details>

- Dead-letter banner ordering (item #1 above).
- Missing composite DLQ index for new query pattern (item #3 above).

Otherwise: status endpoint is single-purpose with sensible scope; cursor + sliding-window logic stays inside `OutboxRepository` (one source of truth); listener id constant lives in a sensible `lib/` location; polling reuses TanStack `refetchInterval` rather than reinventing.
</details>

Endpoint is gated by `auth` + `requirePlatformAdmin`. Zod regex `^\d+$` strips anything non-numeric before BigInt parsing — no SQL injection. Listener id is hardcoded server-side, not user-controllable. No data exposure beyond what a platform admin already sees (status is processed/pending/dead-lettered booleans, not payload content).
</details>

---

PR #551, opened 2026-05-16
docs: add pi remote v2 adapter reference

kristofferremback:
## Problem

The Pi adapter needs to move from ad-hoc message polling to the runtime protocol.

## Solution

Adds a checked-in reference Pi extension for `/remote-control`, runtime heartbeat, targeted claim polling while idle, and completion/failure through the public runtime API.

## Test plan

- [x] Pre-commit lint/typecheck/OpenAPI check

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/551"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/bot-runtimes/repository.ts
+++ b/apps/backend/src/features/bot-runtimes/repository.ts
@@ -347,6 +347,22 @@ export const BotRuntimeSessionLinkRepository = {
     )
     return result.rows[0] ? mapSessionLink(result.rows[0]) : null
   },
+
+  async findActiveByRuntimeSession(
+    db: Querier,
+    params: {
+      workspaceId: string
+      botId: string
+      runtimeKind: BotRuntimeKind
+      instanceId: string
+      runtimeSessionId: string
+    }
+  ): Promise<BotRuntimeSessionLink | null> {
+    const result = await db.query<BotRuntimeSessionLinkRow>(
+      sql`SELECT * FROM bot_runtime_session_links WHERE workspace_id = ${params.workspaceId} AND bot_id = ${params.botId} AND runtime_kind = ${params.runtimeKind} AND instance_id = ${params.instanceId} AND runtime_session_id = ${params.runtimeSessionId} AND status = 'active'`
+    )
+    return result.rows[0] ? mapSessionLink(result.rows[0]) : null
+  },
 }
 
 export const BotInvocationRepository = {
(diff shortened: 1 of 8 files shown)


kristofferremback:
### Code review
**Confidence: 3/7** — Needs work
**Review models:** Orchestrator: Pi coding agent | Reviewers: manual static review

Found 3 issues:

1. `docs/examples/pi-remote/threa-remote-v2.ts:99-110` — The reference adapter calls `POST /api/v1/workspaces/:workspaceId/bot-runtime/sessions`, but Phase 2 did not add that route or handler. `/remote-control` will fail before it can create/link the dedicated scratchpad, so the checked-in adapter cannot actually be used against the stacked backend.
   https://github.com/threahq/threa/blob/1d3981809d3e26e0c392f1b76ae9c271cf0e5db3/docs/examples/pi-remote/threa-remote-v2.ts#L99-L110

2. `docs/examples/pi-remote/threa-remote-v2.ts:130-135` — The adapter claims with a 120s TTL but never renews the claim while the agent is working. The backend completion path only accepts non-expired claims, so any normal coding task that runs longer than two minutes will fail to complete and may later be reclaimed by another poller. Either add/use a claim heartbeat endpoint or choose a completion protocol that does not expire an actively running local turn without renewal.
   https://github.com/threahq/threa/blob/1d3981809d3e26e0c392f1b76ae9c271cf0e5db3/docs/examples/pi-remote/threa-remote-v2.ts#L130-L135

3. `docs/examples/pi-remote/threa-remote-v2.ts:154-160` — `textFromAgentMessages()` stringifies structured assistant `content` values, which are normally arrays/objects, so the posted Threa result can become `[object Object]` instead of the assistant text. The current global plugin handles `content: string` and `content: [{ type: "text", text }]`; the reference adapter should keep that extraction logic.
   https://github.com/threahq/threa/blob/1d3981809d3e26e0c392f1b76ae9c271cf0e5db3/docs/examples/pi-remote/threa-remote-v2.ts#L154-L160

kristofferremback:
Accepted/fixed. Added the runtime session API used by the reference adapter, added claim renewal and adapter-side renewal while a Pi turn is running, and preserved structured assistant text extraction so object/array content does not post as .

kristofferremback:
Correction: Accepted/fixed. Added the runtime session API used by the reference adapter, added claim renewal and adapter-side renewal while a Pi turn is running, and preserved structured assistant text extraction so object/array content does not post as `[object Object]`.

kristofferremback:
@coderabbitai review

PR #586, opened 2026-05-21
feat(conversations): multi-membership and reassignment

kristofferremback:
## Summary

Stops the boundary extractor from leaving messages stranded in the wrong conversation. The classifier can now (a) place a message in more than one conversation, (b) move an earlier message into the freshly created conversation when later context makes the earlier classification look wrong.

The conversation row no longer stores `message_ids` / `participant_ids`. Both are derived at read time from a new join table, `conversation_message_assignments`, with `is_primary` distinguishing the canonical membership from secondaries. A partial unique index on `(message_id) WHERE is_primary` enforces "at most one primary per message"; a full unique index on `(conversation_id, message_id)` enforces "no duplicate membership"; and a deferred constraint trigger added in a follow-up migration enforces "exactly one primary at commit" (see Hardening below).

### Schema

- New table `conversation_message_assignments` with the two indexes above.
- Follow-up migration `20260521120000_assignments_primary_constraint.sql` adds a `DEFERRABLE INITIALLY DEFERRED` constraint trigger (`cma_exactly_one_primary_per_message`) so the partial index's "at most one" is upgraded to a true "exactly one" invariant — checked at COMMIT, which lets the two-step DELETE-then-UPSERT in `assignPrimary` and the "secondary-first, primary-second" loop pass through their transient mid-transaction states.
- Append-only migrations only (INV-17); no edits to existing files.

### Extractor

- `extract()` now returns `assignments: MessageAssignment[]` (one or more, each with `isPrimary`) instead of a single `conversationId`, plus an optional `reassignments[]` list of prior messages to move.
- Internal extractor types (`MessageAssignment`, `Reassignment`, `CompletenessUpdate`, `ExtractionResult`) are now `z.infer<>` from internal Zod schemas in `boundary-extraction/types.ts` (INV-31) so schema and consumer types can't drift.
- Thread streams no longer take the deterministic "join the parent conversation" shortcut — the LLM sees both the parent's conversation and the thread's active conversation in the candidate set and decides. The shortcut was wrong when threads branched topically.
- Parent-thread context messages are now in the reassignment candidate set, so the LLM can also pull a parent-channel message into a freshly created thread topic.
- Reassignment targets are validated twice: the prompt is hard-coded with the candidate message IDs, and the service double-checks each `r.messageId` is in the Phase 1 surrounding/thread window before writing.
- Completeness updates and reassignment targets are restricted to `validUpdateTargets` (the conversations the extractor actually saw), so the LLM cannot touch unrelated conversations.

### Service

- Three-phase pattern preserved (INV-41): read with `withClient`, AI with no connection, write in one transaction.
- `assignPrimary` UPSERTs on `(conversation_id, message_id)` and then deletes any other primary for the same message. Concurrent `assignSecondary` between the two steps either lands first (and gets promoted in place) or no-ops on its `ON CONFLICT DO NOTHING` — both indexes stay satisfied under concurrency (INV-20).
- Phase 3 touched-conversation loops are batched (INV-56): one `UPDATE ... WHERE id = ANY(...)` to bump `last_activity_at`, one `findByIds` to fetch them for outbox events.
- Fails loudly (INV-11) if the extractor ever returns zero or multiple primaries — better an aborted transaction than an orphaned message.
- Outbox emits `conversation:created` for the new conversation, `conversation:updated` for everything else touched, plus per-assignment `conversation:message_assigned` and per-move `conversation:message_reassigned`.
  - `conversation:message_assigned` now carries `parentStreamId` so the parent channel's room receives the membership update for thread messages (the broadcast handler fans out generically when the field is set).
  - `conversation:message_reassigned` routes to the reassigned message's own stream (`existingPrimary.streamId`), not the triggering message's stream — context-window messages reassigned from another stream were previously emitted to the wrong room.

### Repository

- `ConversationRepository` now joins a `LEFT JOIN LATERAL` against `conversation_message_assignments` to derive `messageIds`, `participantIds`, and `secondaryMessageIds` in a single pass. The join is produced by `assignmentAggregateJoin(alias)` and consumed from both the read paths (`alias = 'c'`) and the UPDATE...RETURNING CTE (`alias = 'u'`) so the two stay in lockstep instead of drifting via byte-copy.
- All cross-tenant-capable methods (`findByIds`, `findByMessageId`, `findByMessageIds`, `bumpActivityForIds`, `update`) now take `workspaceId` and filter by it at the SQL level (INV-8). The `update` no-op short-circuit is also workspace-scoped so a misrouted ID from another tenant returns `null` instead of leaking the row.
- Added `findByIds` and `bumpActivityForIds` for batch use from the service.
- `Conversation` shape gains `secondaryMessageIds` (messages that live here but have their primary elsewhere).

### Audit script

- `scripts/analyze-conversation-boundaries.ts` was reading the dropped array columns; rewrote its three queries to go through `conversation_message_assignments`, ordered by `m.sequence`.

### Hardening / review follow-ups

Two rounds of CodeRabbit review (`pullrequestreview-4336XXXXXXX` and `pullrequestreview-4337273037`) drove the following on top of the original implementation:

- Workspace scoping (INV-8) on every repository method that takes a conversation ID directly, plus matching integration tests covering cross-tenant read/update isolation.
- Zod-derived internal types in `boundary-extraction/types.ts` (INV-31).
- `parentStreamId` on `conversation:message_assigned`; `existingPrimary.streamId` routing on `conversation:message_reassigned`.
- Eval suite tightened to `filter() + length === 1` so a multi-primary regression in the extractor also fails loudly instead of silently using the first primary.
- `assignmentAggregateJoin(alias)` helper to keep the read and update CTEs in sync.
- Integration tests for the new constraint trigger: a transient "secondary then primary" sequence inside one transaction commits OK; a "secondary-only" sequence aborts at COMMIT.

## Test plan

- [x] `bun run typecheck` clean across all packages.
- [x] `bun run lint` clean.
- [x] `apps/backend/src/features/conversations/boundary-extraction/llm-extractor.test.ts` — 21 passing (covers multi-assignment, reassignment validation, completeness updates, candidate-set enforcement, parent-thread reassignment).
- [x] Pre-commit hook (lint + typecheck + Dockerfile workspace check + OpenAPI spec drift) green on every push to the branch.
- [ ] `apps/backend/tests/integration/conversation-repository.test.ts` workspace-scoping + constraint-trigger tests — require Postgres on `:5454`; not runnable in this sandbox. CI runs these.
- [ ] E2E suite requires the docker-compose postgres on port 5454; not run in this sandbox.
- [ ] Verify in staging that an existing stream's conversations still render after deploy (the LATERAL aggregate is now the only source of truth for `messageIds`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EZqdYwwVfTCvjaeQMw2ePc)_


kristofferremback:
_This review has been superseded by [a newer one](https://github.com/threahq/threa/pull/586#issuecomment-4510615686)._

<details><summary>Original review (click to expand)</summary>


### Code review
**Confidence: 2/7** — Significant Concerns

Found 8 issues — one blocker that prevents the reassignment path from ever committing.

1. `apps/backend/src/features/conversations/assignment-repository.ts:81-104` — **BLOCKER.** `assignPrimary` aborts the transaction whenever it's invoked to *reassign* a message that already has a primary in a different conversation (i.e. the entire reassignment code path). `ON CONFLICT (conversation_id, message_id)` only arbitrates on the full unique index. When reassigning `M` from `A` to `B`, the proposed `(B, M, is_primary=TRUE)` row has no conflict on `(conv, message)` (no `(B, M)` row exists), so the INSERT proceeds and Postgres then enforces the partial unique index `(message_id) WHERE is_primary` — which collides with the still-present `(A, M, is_primary=TRUE)`. Raises `duplicate key value violates unique constraint "conversation_message_assignments_message_primary"` and aborts. The step-2 DELETE never executes because step 1 already errored. The doc comment at lines 65-78 describes the intended invariant but the implementation gets the steps in the wrong order. Fix: DELETE any existing primary for `message_id` in a *different* conversation BEFORE the upsert, not after.
   https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/features/conversations/assignment-repository.ts#L80-L114

2. `apps/backend/src/features/conversations/boundary-extraction-service.ts:312-319` — **bug.** Cross-stream reassignments persist the wrong `stream_id`. The denorm is bound to the outer `streamId` (the triggering message's stream), not the reassigned message's original stream. For a thread message that triggers reassignment of a parent-channel message (the parent's conv is in `validUpdateTargets` via the `parentMessageConversations` fetch at `boundary-extraction-service.ts:99-102`), the new assignment row stores the thread's `stream_id`. `existingPrimary.streamId` is already in scope at line 297 and holds the correct value. Breaks the stream-scoped reads at `conversation_message_assignments_stream` for those rows.
   `https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/features/conversations/boundary-extraction-service.ts#L310-L320`

3. `apps/backend/src/db/migrations/20260521103505_conversation_message_assignments.sql:73` — **CLAUDE.md violation.** Backfill generates IDs as `'cma_' || replace(gen_random_uuid()::text, '-', '')` — a UUID, not a ULID. INV-2 (Data Model and Persistence Safety): "All entities use prefixed ULIDs like `stream_xxx`." The runtime `conversationMessageAssignmentId()` uses `ulid()`; only the migration backfill bypasses it. Backfilled rows will sort lexicographically against ULIDs incorrectly and have no embedded timestamp.
   `https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/db/migrations/20260521103505_conversation_message_assignments.sql#L62-L86`

4. `apps/backend/src/features/conversations/boundary-extraction-service.ts:297-300` — **N+1 inside Phase 3 transaction.** `findPrimaryByMessageId` is called per reassignment. All candidate `messageId`s are already known (`validReassignmentMessageIds`). One `findPrimariesByMessageIds(ids)` before the loop collapses this to a single query. Cost matters because Phase 3 is the only place we hold a transactional connection.
   `https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/features/conversations/boundary-extraction-service.ts#L296-L329`

5. `apps/frontend/src/hooks/use-conversations.ts` (unchanged) — **plan gap, INV-53.** Plan §"Outbox / Frontend Impact" requires: "Frontend conv stores subscribe to these and patch their local `messageIds` arrays on receipt." The two new socket events (`conversation:message_assigned`, `conversation:message_reassigned`) are emitted but no frontend handler exists. INV-53 (Frontend and UX behavior): "Socket subscriptions must always pair with bootstrap fetches, and bootstrap must be invalidated on reconnect/resubscribe to close event gaps." The PR description acknowledges this as a follow-up but the plan order has frontend as step 5 within the single PR.
   https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/docs/plans/conversation-multi-membership-and-reassignment.md#L155-L163

6. `apps/backend/src/features/conversations/repository.ts:304-326` — **wasted round-trip on every conversation write.** `insert` does a bare `INSERT` and then a full `findById` (which runs the `ASSIGNMENT_AGGREGATE_JOIN` LATERAL). At insert time the join table has zero rows for this conversation, so the aggregate always returns the empty-array shape — the post-insert read is fully deterministic from the params plus PG-side defaults. `update` has the same shape: bare `UPDATE` then `findById`. In Phase 3 this doubles the round-trips on every conversation write while still inside the transaction.
   https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/features/conversations/repository.ts#L299-L369

7. `apps/backend/src/features/conversations/boundary-extraction-service.ts:423-445` — **secondaries silently excluded from extractor candidate set.** `buildConversationSummaries` filters `c.messageIds` (primary-only per the LATERAL aggregate) against `contextMessageIds`. Messages that are secondary-assigned to a conversation aren't surfaced to the LLM as reassignment candidates. Plan D1 scopes the candidate set to "the messages already in `ExtractionContext` (recentMessages plus the tails of `activeConversations` that the prompt surfaces)" — secondaries within a surfaced conv should qualify, but the code excludes them. Either widen the filter to include `secondaryMessageIds`, or document the exclusion in the plan.
   `https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/features/conversations/boundary-extraction-service.ts#L423-L446`

8. `apps/backend/src/features/conversations/assignment-repository.ts:80-114` — **API safety.** `assignPrimary` is correct only inside a transaction (two-step write where step 1 alone leaves two primaries in the table). The signature accepts `Querier`, which includes `Pool`. A future caller passing the pool directly silently violates the partial unique index invariant. Either narrow the type to `PoolClient` or assert it at runtime; either way, document the transaction requirement in the doc comment.
   https://github.com/threahq/threa/blob/99012be6ff881b48ca0447c9d4069034cd6dd20e/apps/backend/src/features/conversations/assignment-repository.ts#L62-L114

---

<details><summary>📐 Plan Adherence — 2 issues</summary>

- Frontend store + socket handlers (plan step 5) absent — see issue 5.
- Secondaries silently excluded from reassignment candidate set vs. plan D1 — see issue 7.

</details>

<details><summary>🔍 Bugs — 2 issues</summary>

- `assignPrimary` always aborts on reassignment due to partial-unique violation — see issue 1.
- Reassignment stream_id denorm uses triggering stream not reassigned message's stream — see issue 2.

</details>

<details><summary>📋 CLAUDE.md Compliance — 2 violations</summary>

- INV-2 violation: migration backfill uses UUID-hex not ULID — see issue 3.
- INV-53 violation: new socket events without paired frontend bootstrap/subscription — see issue 5.

</details>

<details><summary>🏗️ Design — 3 concerns</summary>

- N+1 `findPrimaryByMessageId` calls in reassignment loop — see issue 4.
- `insert`/`update` do bare write + full LATERAL read (wasted round-trip) — see issue 6.
- `assignPrimary` accepts `Querier` but requires transactional caller — see issue 8.

</details>

<details><summary>🔒 Security</summary>

CLEAN — no high-confidence vulnerabilities identified. SQL paths use parameterized `sql\`\`` templates throughout. The LLM-controlled `reassignments[].messageId` and `toConversationId` fields are validated against Phase 1 sets (`validReassignmentMessageIds`, `validUpdateTargets`) before any writes.

</details>

---

</details>

kristofferremback:
_This review has been superseded by [a newer one](https://github.com/threahq/threa/pull/586#issuecomment-4510615686)._

<details><summary>Original review (click to expand)</summary>


### Code review

Supersedes the earlier 2/7 review on this PR — the `assignPrimary` ordering, cross-stream `streamId`, Phase 3 N+1, and `insert`/`update` round-trip blockers are all resolved on `058fb20`. Four issues remain.

1. `apps/frontend/src/hooks/use-conversations.ts:90-91` — The hook subscribes to `conversation:created` and `conversation:updated` but not the two new event types this PR introduces (`conversation:message_assigned`, `conversation:message_reassigned`). Both are emitted from `boundary-extraction-service.ts:393, 405`. Per INV-53 ("Socket subscriptions must always pair with bootstrap fetches, and bootstrap must be invalidated on reconnect/resubscribe to close event gaps"), without these handlers the conversation `messageIds` cache stays stale across every reassignment until the next page reload. The plan's "Outbox / Frontend Impact" section and step 5 of "Implementation as a Single PR" explicitly required this wiring.
   https://github.com/threahq/threa/blob/058fb20d933f3d229f0c173778f436ada9d78c10/apps/frontend/src/hooks/use-conversations.ts#L88-L97

2. `apps/backend/src/lib/outbox/broadcast-handler.ts:232` — `conversation:message_assigned` and `conversation:message_reassigned` are stream-scoped but fall through to the generic `isStreamScopedEvent` branch, which only routes to the message's own stream room. The sibling branch at line 178-185 handles the asymmetry for `conversation:created`/`conversation:updated` by also fanning out to `parentStreamId` when the conversation belongs to a thread. The new events miss that fan-out, so clients viewing a parent channel never see assignment/reassignment signals from child thread conversations even though `boundary-extraction-service.ts:367-368` already resolves `parentStreamId` and could be threaded through the payload.
   https://github.com/threahq/threa/blob/058fb20d933f3d229f0c173778f436ada9d78c10/apps/backend/src/lib/outbox/broadcast-handler.ts#L175-L237

3. `apps/backend/src/features/conversations/assignment-repository.ts:87-121` — The two-step `assignPrimary` (DELETE other-conv primaries, then UPSERT) closes the race against concurrent `assignSecondary` (correctly proven in the doc comment) but **not** against two concurrent `assignPrimary(convX, M)` / `assignPrimary(convY, M)` in different transactions. Under READ COMMITTED both DELETEs evaluate their predicates against their own snapshots before either commits; the second INSERT then proposes `(convY, M, is_primary=TRUE)` which violates the partial unique index `(message_id) WHERE is_primary` against `convX`'s freshly-committed row — and `ON CONFLICT (conversation_id, message_id)` does not arbitrate cross-conversation index conflicts, so the constraint raises and the Phase 3 transaction aborts. Concrete trigger paths: job-queue retry on the same `messageId`, or worker A processing message N while worker B processes message M when N triggers a reassignment of M. Options: `SELECT pg_advisory_xact_lock(hashtext(messageId))` at Phase 3 start, or wrap `assignPrimary` in a retry-on-23505 loop.
   https://github.com/threahq/threa/blob/058fb20d933f3d229f0c173778f436ada9d78c10/apps/backend/src/features/conversations/assignment-repository.ts#L86-L121

4. `apps/backend/src/features/conversations/repository.ts:381-398` — The LATERAL aggregate is duplicated verbatim inside `update()`'s CTE (byte-for-byte identical to `ASSIGNMENT_AGGREGATE_JOIN` at lines 104-123 except for `c.id` → `u.id`). The constant can't be reused because the alias differs, and there's no comment marking the two copies as paired. A future column addition (e.g. the planned `tertiary_message_ids` or a `confidence` aggregate) applied to one copy will silently diverge from the other. A small inline-comment cross-reference or a parameterized helper would prevent that.
   https://github.com/threahq/threa/blob/058fb20d933f3d229f0c173778f436ada9d78c10/apps/backend/src/features/conversations/repository.ts#L100-L130

---

Issue #1 — frontend subscriptions for the two new outbox event types are missing. Plan section "Outbox / Frontend Impact" and step 5 of "Implementation as a Single PR" both called for these handlers paired with bootstrap invalidation.

</details>

Issue #3 — concurrent `assignPrimary` against the partial unique index. Unlikely under steady-state single-worker-per-message processing but a real failure mode on retry or cross-message reassignment overlap.

</details>

INV-1 (no FKs), INV-2 (`cma_` prefix — backfill UUID-suffix matches established pattern in 5 prior migrations), INV-3 (TEXT for `reason`, validated in code), INV-8 (workspace_id denorm on every join row), INV-17 (single new migration, no edits to prior files), INV-20 (race-safe partial-index upsert except for the concurrent-primary case in #3), INV-41 (three-phase preserved — withClient → no-conn AI → withTransaction), INV-51/52 (feature colocation + barrel export of `ConversationMessageAssignmentRepository`), INV-56 (Phase 3 uses `bumpActivityForIds` + `findByIds`) all hold.

</details>

<details><summary>🏗️ Design [1 issue]</summary>

Issue #4 — LATERAL aggregate verbatim duplication in `update()` without a cross-reference. The CTE-with-alias approach is fine; the lack of comment binding the two copies is the gap.

</details>

Workspace isolation: every `conversation_message_assignments` row carries `workspace_id` and writes pass it explicitly. LLM-controlled `toConversationId` is filtered through `validUpdateTargets` (line 298) and `messageId` through `validReassignmentMessageIds` (line 268). The dynamic SQL in `update()` interpolates only hardcoded column names (`topic_summary`, `completeness_score`, ...) with parameterized values — no injection path.

</details>

---

</details>

kristofferremback:
_This review has been superseded by [a newer one](https://github.com/threahq/threa/pull/586#issuecomment-4510914424)._

<details><summary>Original review (click to expand)</summary>


### Code review

Found 3 issues:

1. `apps/backend/src/features/conversations/repository.ts:462` — `bumpActivity(db, id)` updates `conversations` without a `workspace_id` filter, while its sibling `bumpActivityForIds` correctly scopes. Test-only caller today, but exported on the repo's public surface. CLAUDE.md [INV-8, "Data Model and Persistence Safety"]: "All product/domain data persisted in PostgreSQL must be workspace-scoped via `workspace_id` and must be queried/mutated with a workspace filter."
   https://github.com/threahq/threa/blob/4771c47e67e31f9b5bd1e3949f9e6ea4362f2763/apps/backend/src/features/conversations/repository.ts#L461-L470

2. **Plan step 5 (frontend) is missing.** The committed plan (`docs/plans/conversation-multi-membership-and-reassignment.md:175`) explicitly lists "Frontend. Conv store reads new shape (`messageIds` now derived from primary assignments; `secondaryMessageIds` optional). Subscribe to `conversation:message_assigned` and `conversation:message_reassigned`." and earlier states "this lands as one PR" (line 167). HEAD has zero frontend files in the diff — `grep -r "message_assigned\|message_reassigned\|secondaryMessageIds" apps/frontend/src/` returns no results. Backend now emits `conversation:message_assigned` and `conversation:message_reassigned` outbox events with no consumer; clients will not reflect membership changes until a full re-bootstrap. INV-53 ("Socket subscriptions must always pair with bootstrap fetches") applies — the bootstrap shape includes `secondaryMessageIds` but nothing renders it.
   https://github.com/threahq/threa/blob/4771c47e67e31f9b5bd1e3949f9e6ea4362f2763/docs/plans/conversation-multi-membership-and-reassignment.md#L165-L178

3. `apps/backend/src/features/conversations/repository.ts:247-266` — `findPrimariesByMessageIds` does the per-id lookup application-side: nested `for (id of messageIds) if (conv.messageIds.includes(id))`. With M conversations × N input IDs × K avg primary-array length, this is O(M·N·K). At today's reassignment-window scale (handful of IDs, modest array sizes) the cost is trivial; the design concern is that the SQL `&&` overlap pulls the right rows in one shot but the JS loop immediately re-scans arrays. Cleaner: `CROSS JOIN LATERAL unnest(message_ids) WITH ORDINALITY` so the DB returns `(message_id, conversation_row)` pairs and the Map is built directly.
   https://github.com/threahq/threa/blob/4771c47e67e31f9b5bd1e3949f9e6ea4362f2763/apps/backend/src/features/conversations/repository.ts#L243-L266

---

Frontend step (5/7 in the plan's "Implementation as a Single PR" ordering) is missing. New outbox events emitted without any consumer; `secondaryMessageIds` on the `Conversation` type is unread on the frontend.

</details>

Correctness reviewer found nothing scoring ≥80. Locking via `SELECT ... FOR UPDATE` on `messages.id = ANY(...)` correctly serializes concurrent reassignments. Array-mutation helpers are idempotent (CASE guards on `array_append`). Reassignment ordering (remove-from-old before add-to-new) prevents transient double-membership. Participant-id staleness after `removePrimaryMessage` is documented intentional behavior.

</details>

INV-8 (workspace scoping) violation on the new `bumpActivity` repo method. Test-only caller currently, but the method is exported and will attract production callers. Fix: drop the method (callers can use `bumpActivityForIds` with one ID), or add a `workspaceId` parameter to match the rest of the post-CodeRabbit repo surface.

</details>

`findPrimariesByMessageIds` rebuilds the result Map by re-scanning each returned conversation's `messageIds` array in JS. SQL-side `unnest WITH ORDINALITY` would return pre-paired `(message_id, conversation)` rows. Not a real hot path at today's scale (small reassignment windows), but the join-shaped data is being assembled twice.

</details>

No high-confidence vulnerabilities. Workspace scoping is consistent across the new array-mutation methods (`addPrimaryMessage`, `addSecondaryMessage`, `removePrimaryMessage` all filter by `workspace_id`). The lock query `SELECT id FROM messages WHERE id = ANY(...) FOR UPDATE` is scoped only by id — intentional, since the id set comes from `validReassignmentMessageIds` which is built from messages already queried in the current stream + thread streams; no cross-tenant exposure path.

</details>

</details>

kristofferremback:
### Code review
**Confidence: 5/7** — Good

Last round's three issues all resolved in `bea38c4` — verified: `bumpActivity` dropped, `findPrimariesByMessageIds` rewritten with `CROSS JOIN LATERAL unnest`, frontend step 5 wired with new socket handlers + reconnect invalidation. Two new non-blocking observations below.

Found 2 issues:

1. `apps/frontend/src/hooks/use-conversations.ts:70-81` — The list `useQuery` doesn't override TanStack's `refetchOnReconnect: true` default, so when `navigator.onLine` flips back on TanStack will refetch — and the new INV-53 effect at line 78 ALSO invalidates on socket-reconnect. Network restoration and socket reconnect are different signals but frequently coincident (laptop wake, brief network loss). When they happen serially TanStack won't coalesce them — you get one HTTP request per signal. Cheap fix: pass `refetchOnReconnect: false` on the `useQuery` since the socket-driven invalidation is the authoritative INV-53 path (socket can reconnect without `navigator.onLine` flipping, but not vice versa for our model).
   https://github.com/threahq/threa/blob/bea38c464b9ec1876ec47672a7ff6d4c61733e2a/apps/frontend/src/hooks/use-conversations.ts#L70-L81

2. **Plan step 6 (tests) — partial coverage.** The plan (`docs/plans/conversation-multi-membership-and-reassignment.md:176`) calls for service E2E covering "cross-topic split, late reveal triggering reassignment, thread root secondary on parent." Reassignment is well covered in `boundary-extraction-service.test.ts:451-688` (3 tests). The other two scenarios — a single extraction producing both a primary and a secondary assignment (multi-assignment / cross-topic split), and a thread-root message being added as secondary to a parent-channel conv — are exercised at the extractor unit level but never end-to-end through the service + DB layer. That leaves the `addSecondaryMessage` write path without integration coverage.
   https://github.com/threahq/threa/blob/bea38c464b9ec1876ec47672a7ff6d4c61733e2a/docs/plans/conversation-multi-membership-and-reassignment.md#L174-L178

---

Step 6 E2E coverage is partial: reassignment is covered, but multi-assignment (cross-topic split) and thread-root-as-parent-secondary scenarios are not exercised through the full service path. Extractor unit tests cover the LLM contract for these cases; the missing integration scope is the `addSecondaryMessage` write + outbox loop.

</details>

Correctness reviewer found nothing scoring ≥80. Verified:
- `findPrimariesByMessageIds` CROSS JOIN LATERAL has no column ambiguity (`SELECT_FIELDS` uses bare column names that only exist in `conversations c`; unnest exposes only `m.message_id`). Returns one entry per input message ID; empty input short-circuits.
- All 4 socket `.on()` calls paired with `.off()` cleanups. No orphan callers of `bumpActivity`.
- INV-53 reconnect effect correctly guards on `reconnectCount === 0` to skip mount.
- `conversation:message_assigned` filter accepts both thread-stream and parent-stream subscribers via `payload.streamId !== streamId && payload.parentStreamId !== streamId`.
- `conversation:message_reassigned` correctly excluded from the parent-stream fan-out branch and falls through to generic stream-scoped routing.
- `FOR UPDATE` lock scope (id-only) has no cross-tenant exposure — id set comes from `validReassignmentMessageIds` already filtered to this workspace's stream + thread streams.

</details>

No new INV violations. The `bumpActivity` INV-8 gap from the prior review is gone; the workspace-scoped `bumpActivityForIds` is the only remaining surface.

</details>
<details><summary>🏗️ Design [1 minor concern, see issue 1]</summary>

The TanStack `refetchOnReconnect` overlap with the manual socket-reconnect invalidate (issue 1) is the only design concern surviving the threshold. The broadcast-handler conditional now matches 3 conversation:* event types — currently coherent under one routing rule ("stream + optional parent stream"), so not flagged. `conversationKeys.messages` is built as a raw literal rather than spreading `conversationKeys.all`, but since `"conversations"` is the literal first element of both, TanStack prefix matching against `conversationKeys.all` still catches it — no functional gap.

</details>

No high-confidence vulnerabilities. Workspace scoping consistent across the new repo surface. No injection, no data exposure path. The narrow `FOR UPDATE` scope on `messages.id = ANY(...)` is intentional and documented.

</details>

PR #612, opened 2026-05-24
feat(gallery): preview markdown and HTML attachments inline

kristofferremback:
## Summary

- Markdown and HTML attachments now open in the existing image-gallery overlay instead of triggering a download. Markdown renders through `MarkdownContent`; HTML is sandboxed inside a `srcDoc` iframe (`sandbox=""`, null origin, no scripts) which works around the strict-sandbox + cross-origin S3 whitescreen we hit when pointing the iframe directly at the presigned URL. Both viewers get a raw-source toggle and a copy-to-clipboard action.
- Sidebar thumbnails now reuse the URL already loaded by `ImageAttachment` (mirrors the `VideoAttachment` callback pattern) so the strip doesn't show spinners while the gallery is open. Long filenames middle-truncate (with a shadcn `Tooltip` for the full name on desktop hover) so text never pushes thumbnails out of view, and thumbnails use `object-contain` over a letterbox instead of `object-cover` so neither axis crops.
- Fixed the swipe-to-quote regression: every gallery touch handler now `stopPropagation` up front, before any early returns, and bails when the target is inside `[data-gallery-text-viewer]`. React events propagate through the React tree, not the DOM tree, so Radix Portal'd gallery content was still bubbling into the underlying `MessageEvent`.
- Native `overflow-auto overscroll-contain` replaces Radix `ScrollArea` on the text panel because the latter intercepts iOS horizontal touch and broke pan-to-scroll inside wide code blocks.
- Desktop panel max width is now `800px` (matches the stream message column) and drops horizontal padding so narrow windows aren't wasted.

## Test plan

- [ ] Open a stream with an image, a markdown file, and an HTML file attached. Click each:
  - [ ] Markdown renders with the app's markdown styling; toggle raw-source; copy works.
  - [ ] HTML renders inside the sandboxed iframe; toggle raw-source; copy works.
  - [ ] Image still opens normally and the sidebar shows real thumbnails (no spinner column).
- [ ] On desktop, hover a long-filename sidebar entry — the shadcn tooltip shows the full name.
- [ ] On mobile (iOS Safari), swipe horizontally inside a wide markdown code block — page scrolls horizontally without triggering the quote/swipe gesture on the underlying message.
- [ ] On mobile, swipe-to-quote still works when starting the gesture outside the viewer panel.
- [ ] Resize the window narrow on desktop — the markdown panel uses the full available width up to `800px` with no leftover horizontal padding.

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/612"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782240242&installation_id=129946197&pr_number=612&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F612&signature=4b4ab4264f243345671fc5296fbe39fac2e2a33636e646a96317261cd93d0e2c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/gallery/html-viewer.tsx
+++ b/apps/frontend/src/components/gallery/html-viewer.tsx
@@ -1,0 +1,71 @@
+import { Loader2 } from "lucide-react"
+import { useTextContent } from "./use-text-content"
+
+interface HtmlViewerProps {
+  url: string
+  filename: string
+  rawMode?: boolean
+}
+
+/**
+ * Renders an HTML attachment inside a hard-sandboxed iframe. We always fetch
+ * the source ourselves and pass it via `srcDoc` rather than pointing the
+ * iframe at the presigned S3 URL — that earlier setup produced a permanent
+ * white iframe in real browsers because the strict sandbox + cross-origin S3
+ * response combination tripped the browser's render path. With srcDoc the
+ * iframe gets a literal HTML string under a null origin, no network hop, no
+ * Content-Type guessing.
+ *
+ * `sandbox=""` means: render HTML + CSS, but no scripts, no same-origin, no
+ * top-frame navigation. Safe for arbitrary user-uploaded HTML.
+ *
+ * In rawMode the same source is shown as a `<pre>` block, sharing the panel
+ * surface with the markdown viewer.
+ */
+export function HtmlViewer({ url, filename, rawMode = false }: HtmlViewerProps) {
+  const { content, error } = useTextContent(url || null)
+
+  if (!url || (content === null && !error)) {
+    return (
+      <div className="absolute inset-0 flex items-center justify-center">
+        <Loader2 className="h-6 w-6 animate-spin text-white/50" />
+      </div>
+    )
+  }
+
+  if (error) {
+    return (
+      <div className="absolute inset-0 flex items-center justify-center">
+        <p className="text-sm text-white/70">Couldn't load {filename}.</p>
+      </div>
+    )
+  }
+
+  if (rawMode) {
+    return (
+      <div className="absolute inset-0 pb-16 pt-14 sm:py-16">
+        <div
+          data-gallery-text-viewer="true"
+          className="mx-auto h-full w-full max-w-[800px] overflow-auto overscroll-contain rounded-lg bg-card text-foreground sm:border sm:border-border sm:shadow-2xl"
+        >
+          <pre className="px-4 py-6 sm:px-8 sm:py-8 font-mono text-xs leading-relaxed whitespace-pre text-foreground">
+            {content ?? ""}
+          </pre>
+        </div>
+      </div>
+    )
+  }
+
+  return (
+    <div className="absolute inset-0 pb-16 pt-14 sm:py-16">
+      <iframe
+        data-gallery-text-viewer="true"
+        srcDoc={content ?? ""}
+        title={filename}
+        sandbox=""
+        referrerPolicy="no-referrer"
+        className="mx-auto block h-full w-full max-w-[800px] rounded-lg bg-white sm:border sm:border-border sm:shadow-2xl"
+      />
+    </div>
+  )
+}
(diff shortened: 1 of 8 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (a few non-blocking polish items)

Found 3 issues:

1. `apps/frontend/src/components/timeline/attachment-list.tsx:609-628` — Three identical `useCallback` wrappers (`handleImageClick`, `handleVideoClick`, `handleTextOpen`) all just call `openMedia(attachmentId)` with the same dep array. Collapse to a single `handleMediaOpen = useCallback((id) => openMedia(id), [openMedia])` shared across all three attachment surfaces. (CLAUDE.md [INV-35, Architecture and Dependency Flow]: "Reuse existing helpers and abstractions instead of parallel implementations" — and the dead-clone aspect of [INV-38, Maintainability and Scope Control]: "Delete dead code immediately.")
   https://github.com/threahq/threa/blob/ea040f2e01f163db721f67d8cbfbd35db8235b0c/apps/frontend/src/components/timeline/attachment-list.tsx#L609-L628

2. `apps/frontend/src/components/image-gallery.tsx:49-56` — `truncateMiddle` lives inside `image-gallery.tsx` (a UI component file), but `OpenableFileChip` over in `attachment-list.tsx` independently end-truncates the same filename data with a different strategy. Not a duplicate today, but the helper is invisible to the chip — if either label width is ever retuned, the two surfaces will silently diverge. Move to `apps/frontend/src/lib/` (or a shared `string-utils`) so the chip can adopt it.
   https://github.com/threahq/threa/blob/ea040f2e01f163db721f67d8cbfbd35db8235b0c/apps/frontend/src/components/image-gallery.tsx#L46-L56

3. `apps/frontend/src/components/gallery/use-text-content.ts:7` — Module-level `textCache` Map of presigned-S3-URL → fetched text has no eviction policy and lives for the tab's lifetime. Safe today because each `getDownloadUrl` call returns a fresh URL (so the cache key effectively rotates), but that's a load-bearing assumption with no test or comment. Either drop a `// safe because URLs are request-scoped` note explaining the invariant, or add a soft TTL/size cap. The failure-path delete is correct; only successful entries linger.
   https://github.com/threahq/threa/blob/ea040f2e01f163db721f67d8cbfbd35db8235b0c/apps/frontend/src/components/gallery/use-text-content.ts#L1-L70

---

No formal plan; PR description covers all visible changes. Markdown/HTML viewers, sidebar thumbnail forwarding, swipe-bail, native scroll, and 800px desktop column all present.

</details>

Two reviewer flags dropped as not high-impact:
- `truncateMiddle` `slice(-0)` edge case — unreachable today (only call site passes `maxLength=18` → `tail=8`).
- `loadedTextUrls` in `useEffect` deps causes one redundant re-run after fetch — guarded; benign.

</details>

- INV-35/INV-38: three identical `useCallback` wrappers (see issue 1).

</details>

- `truncateMiddle` placement (see issue 2).
- `textCache` lifetime (see issue 3).

Gallery item type expansion (image/video/markdown/html) is cumulative but matches the trajectory of #485/#544/#414 — acceptable.

</details>

HtmlViewer correctly uses `sandbox=""` (no `allow-scripts`, no `allow-same-origin`) on a `srcDoc` iframe — null origin, no script execution. MarkdownContent renders via `react-markdown` without `dangerouslySetInnerHTML`. No XSS surface introduced.

</details>

PR #622, opened 2026-05-25
feat(bot-runtime): WebSocket transport for bot scratchpad invocations

kristofferremback:
## Why

Cloudflare flagged 50% daily free-tier usage and it traces to the new bot scratchpad invocation framework polling on a 2s cadence (claim + presence). Every poll is one worker request, even when nothing's happening.

WebSocket connections **bypass the worker entirely** after the upgrade (`docs/system-overview.md:54-57`). A bot on WS drops from ~1800 worker requests/hour to ~1.

## What's in the branch

Docs (the plan + impl guide that opened this PR):
- `docs/bot-runtime-websocket-plan.md` — architecture, security model, failure modes
- `docs/bot-runtime-websocket-implementation.md` — 14-section build order

Backend code — the actual transport, on top of the existing HTTP path:
- **Outbox events:** 6 new bot-scoped types (`bot_invocation:{available,claimed,cancelled}`, `bot_session_link:invalidated`, `bot:active_actor_changed`, `bot:resync`) in `apps/backend/src/lib/outbox/repository.ts` + `isBotScopedEvent` type guard
- **Service emission:** every mutation that bots care about now writes its row + the outbox event in the same `withTransaction` — `BotRuntimeService.createInvocation` (fires only when `xmax = 0`, so retried HTTP calls don't double-push), `claimNextInvocation`, `setActiveActorInTransaction` (single mutation path, computed `affectedBotIds` covers both displaced + new bot), `requestResyncInTransaction`
- **Routing:** `BroadcastHandler.dispatchBotEvent` picks narrowest-room-wins (session > instance > bot > workspace) and fans out on `io.of("/bot")`
- **Auth:** `createBotSocketAuthMiddleware` reuses `BotApiKeyService.validateKey`, requires `BOT_RUNTIME_WRITE`, fails closed on throw, supports both `auth.token` and `Authorization: Bearer`
- **Namespace handler:** `attachBotNamespace` validates `bot:hello` with Zod, upserts presence on hello, joins `bot:{ws}` / `:bot:{botId}` / `:instance:` / `:session:` rooms, returns bootstrap `{serverGeneratedAt, availableInvocations, ownedClaims}` so reconnects don't lose in-flight claims
- **Registry:** `BotSocketRegistry` keyed by `(workspaceId, botId, instanceId)` for admin force-disconnect + diagnostics
- **Discovery:** `POST /bot-runtime/presence` now returns `ws: {url, path, namespace}` derived from `x-forwarded-proto` so runtimes upgrade with zero out-of-band config
- **Security:** per-socket key revalidation ticker (default 60s) — closes the revocation gap that would otherwise wait for the engine ping timeout

## Tests

- `service.test.ts` (9) — outbox emission for createInvocation (fresh vs idempotent), claimNextInvocation (claimed vs no-row), setActiveActorInTransaction (displaced bot fan-out, persona→bot, unchanged), requestResyncInTransaction (instanceId-without-botId guard + null normalization)
- `socket-auth.test.ts` (6) — happy path, missing token, Bearer fallback, invalid key, missing scope, throw failure
- `bot-socket-registry.test.ts` (5) — register/unregister/get, isolation across keys, disconnectInstance count, unknown-key no-op
- `broadcast-handler.test.ts` (+5) — namespace routing for available (all 4 target variants), active_actor_changed fan-out to both bots, resync narrowing

Total: 20 new bot-runtime tests + 5 dispatch routing tests. All pass via `bun run --cwd apps/backend test:unit`.

## Notable correctness decisions

- `bot_invocation:available` fires **only on first insert** — `RETURNING (xmax = 0)` distinguishes a fresh row from an idempotent ON CONFLICT update, so retried HTTP calls don't re-push
- `bot_invocation:claimed` payload omits `claimedByInstanceId` — siblings only need "stop racing", not the winner's identity
- `stream_active_actors` has exactly one mutation entry point (`setActiveActorInTransaction`) — repo method now carries a warning comment, and the existing pi-local session linker routes through the wrapper
- Bootstrap SELECT filters `claimed_by_instance_id = $thisInstance` for owned claims and applies a 24h `sinceCursor` floor on available
- `instanceId` / `runtimeSessionId` come from a Zod-validated hello payload before being used in room names

## Deferred (intentional, not in this PR)

- `markOffline` callback on `BotSocketRegistry` disconnect — HTTP presence heartbeats already handle this and the WS layer skipping it isn't a correctness gap, just a minor latency one
- Passing the bot namespace as a `BroadcastHandler` constructor dep — we look it up via `io.of("/bot")` inside `dispatchBotEvent` instead, which is fine because the namespace is a stable string
- Smoke test script `scripts/smoke-bot-ws.ts` — manual verification only for now; the unit-level routing + emission tests cover the same paths

## Backward compatibility

- HTTP claim/complete/fail endpoints are unchanged
- Old bots that haven't adopted WS keep polling and keep working
- WS is opt-in per client — runtimes that read the `ws` hint on presence response and upgrade, drop the poll loop and stop hitting the worker

## Test plan

- [x] `bun run --cwd apps/backend test:unit src/features/bot-runtimes/`
- [x] `bun run --cwd apps/backend test:unit src/lib/outbox/broadcast-handler.test.ts`
- [x] `bun run --cwd apps/backend typecheck`
- [x] `bun run --cwd apps/backend lint`
- [ ] Manual smoke: connect with `socket.io-client`, send `bot:hello`, verify ack shape + bootstrap; trigger a mention in another tab, verify push arrives without polling

https://claude.ai/code/session_013uS3yWPxEy4BxJdaGYpKVj


kristofferremback:
@coderabbitai review

---

kristofferremback:
> _**Superseded** by the newer review at https://github.com/threahq/threa/pull/622#issuecomment-4546289165. Original content preserved below for history._

<details><summary>Original review (click to expand)</summary>

### Code review

**Confidence: 3/7** — Needs Work
**Review models:** Orchestrator: Opus 4.7 | Reviewers: Sonnet x3

Found 3 issues — all in `apps/backend/src/features/bot-runtimes/socket-handler.ts`. Two were converged on by multiple reviewers independently.

1. **Bootstrap SELECT runs before `socket.join()` — race window for missed pushes.**
   `socket-handler.ts:131-158`. The hello handler does presence upsert → `getBootstrapForRuntime` → then the three `socket.join(...)` calls. An invocation inserted between the bootstrap SELECT and the join emits `bot_invocation:available` to a room the socket has not joined yet — silently dropped (not in the snapshot, not delivered live). Bot recovers via the 30s backstop poll, so this isn't permanent loss, but it's exactly the gap INV-53 ("Socket subscriptions must always pair with bootstrap fetches… to close event gaps") and the plan doc (`docs/bot-runtime-websocket-plan.md` §4.7, §6) said to close by joining first. The comment at lines 126-130 justifies the inverted order as "clean rollback on error", but the rollback-safety argument is recoverable a different way: join first, do the read, and on error leave the rooms (Socket.IO cleans them up on disconnect anyway, and the registry / `joined` flag can be set after the read succeeds).
   https://github.com/threahq/threa/blob/1082f20a9ebd024d4c330546380bc58746dd6d7f/apps/backend/src/features/bot-runtimes/socket-handler.ts#L126-L165

2. **`instanceId` / `runtimeSessionId` lack the room-fragment regex the plan requires.**
   `socket-handler.ts:11-24`. `helloSchema` constrains both to `.min(1).max(128)` only — no character class. They are interpolated directly into room names at lines 154-158 (`bot:${workspaceId}:bot:${botId}:instance:${data.instanceId}`). A connected bot can craft `instanceId: "foo:session:bar"` to land its socket in a room overlapping the session-targeted namespace, or use `:` / `/` to collide with sibling rooms inside the same `(workspaceId, botId)`. Cross-workspace / cross-bot leakage is blocked because both come from the verified API key, so impact is bounded to within one bot's room set — but the plan doc (§4.3, §6.1) called this out explicitly and prescribed `^[A-Za-z0-9_-]{1,64}$`. The shipped schema also allows length 128 vs the planned 64.
   https://github.com/threahq/threa/blob/1082f20a9ebd024d4c330546380bc58746dd6d7f/apps/backend/src/features/bot-runtimes/socket-handler.ts#L11-L24

3. **Bootstrap response missing `activeActorByStream` and `activeSessionLinks`.**
   `socket-handler.ts:216-227` (`BotHelloResponse`) and the producer in `service.ts` only return `{ serverGeneratedAt, availableInvocations, ownedClaims }`. The plan specifies two more bootstrap fields (`docs/bot-runtime-websocket-plan.md:144-147`, also `docs/bot-runtime-websocket-implementation.md:356-357`): `activeActorByStream` so a reconnecting bot knows which streams it's the active actor on, and `activeSessionLinks` so it knows which Pi-local sessions are still linked. Without these in bootstrap, a reconnect needs follow-up HTTP calls to reconstruct that state — re-opening the gap the bootstrap was meant to close, and partially reintroducing the poll cost the PR is trying to eliminate. Not in the PR's "Deferred" list, so this reads as a missed scope rather than an intentional cut.
   https://github.com/threahq/threa/blob/1082f20a9ebd024d4c330546380bc58746dd6d7f/apps/backend/src/features/bot-runtimes/socket-handler.ts#L216-L227

---

<details><summary>📐 Plan Adherence — 3 issues</summary>

- `socket-handler.ts:11-24` — `helloSchema` missing `^[A-Za-z0-9_-]{1,64}$` on `instanceId` / `runtimeSessionId` (plan §4.3, §6.1).
- `socket-handler.ts:131-158` — bootstrap runs before `socket.join()`, inverting the plan's "join first, read second" ordering (plan §4.7, §6 / INV-53).
- `socket-handler.ts:216-227` — bootstrap omits `activeActorByStream` and `activeSessionLinks` (plan lines 144-147, impl guide lines 356-357).

</details>

<details><summary>🔍 Bugs — 1 issue</summary>

- `socket-handler.ts:131-158` — bootstrap-then-join race window for `bot_invocation:available` events arriving between the SELECT and the room join. Not silent loss in practice (backstop poll covers it), but it's the exact gap the join-first ordering exists to close.

</details>

No invariant violations that weren't already covered by the PR body's "Deferred" list (broadcaster `io.of("/bot")` lookup vs INV-12 / INV-13, and the missing `markOffline`-on-disconnect path, were both explicitly deferred).

</details>

No leaky-abstraction, parallel-implementation, N+1, or config-sprawl concerns introduced by this PR. The two design-flavor items reviewers raised (broadcaster's inline `io.of("/bot")`, missing disconnect-grace timer + `onInstanceOffline` callback) are explicitly deferred in the PR body.

</details>

<details><summary>🔒 Security — 1 issue</summary>

- `socket-handler.ts:11-24` — same finding as plan adherence #1: room-name fragment regex missing on user-controlled `instanceId` / `runtimeSessionId`. Impact is bounded (no cross-workspace / cross-bot leak), but a malicious instance can collide with sibling rooms within its own `(workspace, bot)` scope.

</details>

---

</details>

kristofferremback:
> _**Superseded** by the newer review at https://github.com/threahq/threa/pull/622#issuecomment-4546289165. Original content preserved below for history._

<details><summary>Original review (click to expand)</summary>

### Code review

**Confidence: 3/7** — Needs Work
**Review models:** Orchestrator: Opus 4.7 | Reviewers: Sonnet x3

Found 5 issues — 1 race-safety bug (independently flagged by two reviewers), 1 transaction-snapshot gap, 1 INV-20 violation, and 2 plan-adherence misses. Nothing blocking, but the two race-safety items are real.

1. `apps/backend/src/features/bot-runtimes/socket-handler.ts:128-179` — `bug`: concurrent `bot:hello` guard bypass. `if (joined)` is checked synchronously at L128 but `joined` is only assigned at L179, after three awaits (presence upsert → room joins → bootstrap SELECT). Two `bot:hello` frames arriving on the same socket before the first chain settles will both observe `joined === null`, both pass the guard, both run to completion — emitting two acks, double-calling `upsertPresenceFromBotKey`, and double-`register`-ing in the registry. The guard's stated intent ("one hello per connection") is silently violated. Fix: set a synchronous `helloInFlight` sentinel before the first await, clear it on both success and catch.
   https://github.com/threahq/threa/blob/a1d7e863197a85bc82cd009cec33adc43a69187b/apps/backend/src/features/bot-runtimes/socket-handler.ts#L127-L180

2. `apps/backend/src/features/bot-runtimes/service.ts:362-389` — `bug`: bootstrap snapshot not transactionally consistent. `getBootstrapForRuntime` fires three queries in parallel via `Promise.all`, each acquiring its own pool connection and therefore its own MVCC snapshot. A claim that commits between the `findBootstrapInvocations` sub-queries can leave an invocation in neither `available` nor `ownedClaims`. The join-rooms-first fix closes the *post*-bootstrap window for missed events, but the *within*-bootstrap gap is still open. Wrap the three reads in a single `withClient` call so they share a snapshot.
   https://github.com/threahq/threa/blob/a1d7e863197a85bc82cd009cec33adc43a69187b/apps/backend/src/features/bot-runtimes/service.ts#L361-L390

3. `apps/backend/src/features/bot-runtimes/service.ts:128-129` — `claude-md` (INV-20, *Data Model and Persistence Safety*): "Never do select-then-update without locking/concurrency control". `setActiveActorInTransaction` does `findByRootStream` (a plain `SELECT`) and then `upsert` inside the same READ COMMITTED tx. Two concurrent calls on the same `rootStreamId` can both read the same `previousActorId` and emit `bot:active_actor_changed` events with stale/duplicate `affectedBotIds` payloads. Either `SELECT ... FOR UPDATE` on `findByRootStream`, or fold the previous-actor lookup into the UPSERT's `RETURNING` so the read and the write are one statement.
   https://github.com/threahq/threa/blob/a1d7e863197a85bc82cd009cec33adc43a69187b/apps/backend/src/features/bot-runtimes/service.ts#L127-L130

4. `apps/backend/src/features/bot-runtimes/socket-handler.test.ts` — `plan`: implementation guide §12 (test table, `docs/bot-runtime-websocket-implementation.md:574`) explicitly maps this file to "Bootstrap" coverage — `serverGeneratedAt` monotonicity across reconnects, `sinceCursor` filtering of pending invocations, stale-cursor (>24h) fall-back, and "claimed-by-me invocations always included regardless of cursor". The file isn't in the PR. `service.test.ts` covers the new bootstrap field shape but not these cursor-filter behaviors. Not blocking but the test plan claim is currently overstated.

5. `apps/backend/src/features/bot-runtimes/service.ts:362-363` — `plan`: `since = sinceCursor > lookbackFloor ? sinceCursor : lookbackFloor` accepts any future date a clock-skewed bot sends, while `docs/bot-runtime-websocket-implementation.md:407` ("ignores values that are in the future, malformed, or older than 24h") and plan §4.7 both require future cursors to be treated as malformed. A bot whose clock is ahead by an hour gets an empty `available` list on every bootstrap and silently loses pending work until the cursor self-corrects. Clamp `sinceCursor` against `serverGeneratedAt` (or `now`) as the upper bound.
   https://github.com/threahq/threa/blob/a1d7e863197a85bc82cd009cec33adc43a69187b/apps/backend/src/features/bot-runtimes/service.ts#L361-L364

---

<details><summary>📐 Plan Adherence [2 issues]</summary>

- Missing `socket-handler.test.ts` — issue #4 above. Plan's Step 12 test table calls this out explicitly for bootstrap-cursor coverage.
- `sinceCursor` future-value handling — issue #5 above. Plan §4.7 + implementation guide §6 both spell out the "ignore future" rule.

</details>

- Concurrent `bot:hello` race — issue #1 above (flagged independently by the Correctness and Design reviewers).
- Cross-connection MVCC gap in `getBootstrapForRuntime` — issue #2 above.

</details>

- INV-20 `setActiveActorInTransaction` — issue #3 above.

</details>

No issues scoring ≥80 in this category beyond the bootstrap-snapshot item already counted under Bugs.

</details>

No high-confidence vulnerabilities found. Workspace isolation comes from the API-key-validated `workspaceId` (not from the hello payload), all SQL goes through parameterized templates, identifiers are now Zod-regex-constrained before becoming room names, and the WS URL hint derived from `x-forwarded-proto` is already noted in code as dev-only.

</details>

---

</details>

kristofferremback:
> _**Superseded** by the newer review at https://github.com/threahq/threa/pull/622#issuecomment-4551843599. Original content preserved below for history._

<details><summary>Original review (click to expand)</summary>


### Code review
**Review models:** Orchestrator: opus | Reviewers: sonnet x3

Second pass after the fixes in `15a9204`. The five blockers from the prior review (https://github.com/threahq/threa/pull/622#issuecomment-4545168188) are all resolved: `helloInFlight` sentinel closes the concurrent-hello race; `getBootstrapForRuntime` opens an explicit `REPEATABLE READ READ ONLY` snapshot; `findByRootStreamForUpdate` + `SELECT ... FOR UPDATE` makes the active-actor swap race-safe (INV-20); `sinceCursor` is clamped to `(lookbackFloor, now]`; the new `socket-handler.test.ts` covers the previously-uncovered surface.

Correctness and security are clean on this pass. One real coverage gap remains, plus one forward-compat advisory.

Found 1 issue:

1. `apps/backend/src/features/public-api/handlers.test.ts` — the new `ws` field on the `POST /bot-runtime/presence` response (`handlers.ts:754` — `ws: buildRuntimeWsHint(req, botRuntimeWsUrl)`) has no test. The plan's test table (Step 10) explicitly asks for an assertion that the response body includes the WS hint derived from the injected `botRuntimeWsUrl`. `grep -n "ws:\|wsUrl\|botRuntimeWsUrl" handlers.test.ts` returns zero matches — the discovery contract that runtimes rely on to auto-upgrade is untested.
   https://github.com/threahq/threa/blob/15a920489da38642e7ad784549fe503d1f73ef33/apps/backend/src/features/public-api/handlers.ts#L750-L760

---

- `handlers.test.ts` — missing test for the new `ws` discovery field on the presence response (see main finding).

The `markOffline` / 30s grace timer from plan §4.8 is **not** flagged here: the PR body explicitly defers it ("`markOffline` callback on `BotSocketRegistry` disconnect — HTTP presence heartbeats already handle this and the WS layer skipping it isn't a correctness gap, just a minor latency one"). Worth a follow-up issue so it doesn't get lost, but not a blocker for this PR.

</details>

No new bugs found. Re-verified the fixes:

- `socket-handler.ts:101-143` — `helloInFlight` is set synchronously before the first `await` and reset in both success and catch paths; the dedicated test (`socket-handler.test.ts:178-207`) covers the concurrent-frames race with a deferred-resolve pattern.
- `service.ts` `getBootstrapForRuntime` — `BEGIN ISOLATION LEVEL REPEATABLE READ READ ONLY` wraps the 3-way `Promise.all` of SELECTs, with `COMMIT` on success and `ROLLBACK().catch(() => {})` on error to avoid masking the original throw.
- `setActiveActorInTransaction` — calls `findByRootStreamForUpdate`, so the read of `previousActorId` and the subsequent UPSERT lock the same row, preventing two concurrent swaps from emitting events with stale displaced-actor IDs.
- `sinceCursor` clamp — `cursor && cursor > lookbackFloor && cursor <= now` correctly handles future-dated clocks and the 24h floor.

Bootstrap error-path inconsistency (presence row written but rooms left on `getBootstrapForRuntime` throw) is acceptable — the client retries hello and the next attempt overwrites presence.

</details>

No INV violations introduced. The bootstrap snapshot now respects INV-20 (no select-then-update without locking) via the new `FOR UPDATE` repo method, and INV-30 is satisfied because the multi-SELECT bootstrap legitimately needs `withClient` for MVCC consistency.

</details>

<details><summary>🏗️ Design [1 advisory, non-blocking]</summary>

- `socket-handler.ts:302-330` `serializeInvocation` includes `promptMarkdown` in both `availableInvocations` and `ownedClaims` in the bootstrap ack. The plan §4 (`bot-runtime-websocket-plan.md:204`) says the `bot_invocation:available` **push** is "plaintext metadata only ... The bot fetches the source message over HTTP and decrypts client-side." The outbox event payload (`BotInvocationAvailableOutboxPayload`) correctly omits content. The bootstrap is the WS replacement for HTTP polling, so it inherits the same shape as the HTTP path which does include the prompt — that's defensible today, but when E2E lands the bootstrap will need to switch to metadata-only and have bots HTTP-fetch the prompt. Worth a comment near `serializeInvocation` flagging this as a known asymmetry, or a follow-up issue tied to the E2E work.
- `broadcast-handler.ts` `dispatchBotEvent` resolves the bot namespace via `this.io.of("/bot")` on each call rather than receiving it as a constructor dep (plan suggested the dep injection). `io.of()` is O(1) so no perf concern, but a `botNamespace` field would make the routing easier to test in isolation. Non-blocking.

</details>

Traced data flow from handshake auth and `bot:hello` payload through to room names. `instanceId` and `runtimeSessionId` are Zod-validated against `^[A-Za-z0-9_-]+$` (1-64 chars) before any room name is constructed, so no room-name injection is possible. The scope check (`BOT_RUNTIME_WRITE`) is enforced in `socket-auth.ts` and fails closed on throw. The 60s revalidation ticker bounds the post-revocation window. `availableInvocations` carry an expired `claimToken` for stale claims, but the HTTP claim endpoint validates `claim_expires_at > NOW()` so an expired token can't be used — no exploit.

</details>

> Note: GitHub's API doesn't expose a "patch comment" operation through the MCP tools available here, so the two prior unified-review comments (https://github.com/threahq/threa/pull/622#issuecomment-4542864803 and https://github.com/threahq/threa/pull/622#issuecomment-4545168188) couldn't be marked superseded in-place. This comment supersedes both.

---


</details>

kristofferremback:
> **⚠️ Superseded by [#issuecomment-4552305627](https://github.com/threahq/threa/pull/622#issuecomment-4552305627).** All four findings below were addressed in `8f56238`; the new review is clean.

<details><summary>Original review (click to expand)</summary>


### Code review
**Confidence: 4/7** — Acceptable (one real exploit path + one INV-36 violation; two acknowledged deferrals worth re-examining)
**Review models:** Orchestrator: claude-opus-4-7[1m] | Reviewers: sonnet x3

Found 4 issues:

1. `apps/backend/src/features/bot-runtimes/socket-handler.ts:302-330` — **Bootstrap leaks `claimToken` in `ownedClaims`**. `serializeInvocation` returns the full row including `claim_token`, and the `ownedClaims` SELECT filters only on attacker-supplied `instanceId` (`repository.ts:629-630`: `claimed_by_instance_id = ${params.instanceId}`). Any client with a valid bot API key + scope `BOT_RUNTIME_WRITE` can send `bot:hello` with `instanceId = <another live instance's id>` and lift that instance's claim tokens out of the ack — and `claimToken` is the credential the complete/renew/fail HTTP endpoints authorize on (`repository.ts:526,535,553,564,582`: `WHERE … AND claim_token = ${params.claimToken}`). Within a shared-bot-key fleet this is a meaningful escalation from "can race-claim new invocations" to "can mark in-flight claims of sibling instances complete/failed". Fix is local: drop `claimToken` (and arguably `claimedByInstanceId`) from the bootstrap serialization — runtimes already hold their own tokens, they have no need to receive them via bootstrap.
   https://github.com/threahq/threa/blob/f1479a8942f98b9dbcffe077274c058a2080f396/apps/backend/src/features/bot-runtimes/socket-handler.ts#L302-L330

2. `apps/backend/src/lib/outbox/repository.ts:81-82,665-666,787-796` + `apps/backend/src/lib/outbox/broadcast-handler.ts:299-321` — **`bot_invocation:cancelled` and `bot_session_link:invalidated` registered with zero production emitters** (CLAUDE.md [INV-36, *Maintainability and Scope Control*]: "Do not add speculative features, configuration, or comments for imagined requirements"). Both event types appear in `OutboxEventType`, `OutboxEventPayloadMap`, `BOT_SCOPED_EVENTS`, and `dispatchBotEvent` routing branches, but `grep -rn 'bot_invocation:cancelled\|bot_session_link:invalidated' apps/backend/src --include='*.ts'` returns hits only in the outbox plumbing and tests — no service call ever writes these rows. Either remove them and reintroduce alongside emitters, or wire the emitter as part of this PR.
   https://github.com/threahq/threa/blob/f1479a8942f98b9dbcffe077274c058a2080f396/apps/backend/src/lib/outbox/repository.ts#L79-L84

3. `apps/backend/src/features/bot-runtimes/bot-socket-registry.ts:1-82` — **Grace-timer `onInstanceOffline` + `BotRuntimeInstanceRepository.markOffline` not implemented**, despite the plan explicitly requiring them (`docs/bot-runtime-websocket-plan.md:182`: "After a 30 s grace window with no reconnect… the registry calls `BotRuntimeInstanceRepository.markOffline(...)` which sets `status='offline'`, `accepting_invocations=false`"; impl-doc step 5 line 243 and step 8 line 482). The PR body defers this with rationale "HTTP presence heartbeats already handle this", but the same plan doc states (`docs/bot-runtime-websocket-implementation.md:424`): "**No periodic `last_seen_at` ticker.** We deliberately do NOT write `last_seen_at` on a schedule". If the deferral rationale is correct, the plan doc contradicts itself; if the plan is correct, instances that crash without explicit `'busy'/'paused'` transitions stay `status='available', accepting_invocations=true` indefinitely and `target_instance_id`-pinned invocations get routed to a dead runtime. Worth confirming which is true (and updating either the code or the plan) before merging.
   https://github.com/threahq/threa/blob/f1479a8942f98b9dbcffe077274c058a2080f396/apps/backend/src/features/bot-runtimes/bot-socket-registry.ts#L1-L82

4. `apps/backend/src/lib/outbox/broadcast-handler.ts:273` — **`dispatchBotEvent` hard-codes the `/bot` namespace string in `lib/`** by calling `this.io.of("/bot")` on every dispatch instead of taking the namespace as a constructor dep (CLAUDE.md [INV-13, *Architecture Boundaries and Dependency Flow*]: "Construct long-lived collaborators once; avoid per-call dependency assembly"). Plan impl-doc step 3 explicitly required: "Inject the bot namespace as a dependency on `BroadcastHandler`… add `botIo: Namespace`". The PR body defers this and the lookup is functionally OK (`io.of(name)` is a map lookup), but the feature name lives in cross-cutting infrastructure and `BroadcastHandler` tests that mock `this.io` could silently miss `/bot` routing regressions. Cheap to fix (thread `botNs: Namespace` through `server.ts`).
   https://github.com/threahq/threa/blob/f1479a8942f98b9dbcffe077274c058a2080f396/apps/backend/src/lib/outbox/broadcast-handler.ts#L271-L275

---

<details><summary>📐 Plan Adherence — 2 issues</summary>

- Grace-timer + `markOffline` deferred but plan-doc rationale contradicts itself (issue 3 above).
- `BroadcastHandler` namespace dep deferred per PR body, but plan impl-doc step 3 required injection (issue 4 above).
- Everything else lines up — outbox events, service emission, bootstrap shape, room scheme, narrowest-room-wins dispatch, per-socket revalidation ticker all match the plan.

</details>

Correctness scan went through: `setActiveActorInTransaction` (advisory_lock → find → upsert order is correct, fan-out covers both displaced and new actor), `createInvocation` (xmax detection is solid, won't double-emit on retry), `claimNextInvocation` (transactional with outbox row), `requestResyncInTransaction` (instanceId-without-botId guard works), socket revalidation ticker (cleared on disconnect, double-disconnect is a no-op), narrowest-room-wins dispatch (session > instance > bot > workspace, can't fan out wider than intended), bootstrap snapshot isolation (REPEATABLE READ READ ONLY, sinceCursor floor, dedup is the runtime's job).

The IPv6 bracketed-host fix in `f1479a8` was the right call.

</details>

- INV-36 violation: speculative event types (issue 2 above).

Verified compliance:
- INV-4, INV-6, INV-7: outbox row written in same `withTransaction` as state mutation ✓
- INV-5: data access through `BotInvocationRepository` / `BotRuntimeInstanceRepository` ✓
- INV-20: `setActiveActorInTransaction` uses `pg_advisory_xact_lock(hashtextextended(...))` keyed by `(workspaceId, rootStreamId)` before find/upsert ✓
- INV-32: `HttpError` with `status`/`code` for `BOT_RUNTIME_WS_URL_NOT_CONFIGURED` ✓
- INV-51, 52: bot-runtimes feature stays colocated; new exports go through `index.ts` ✓
- INV-55: `bot:hello` payload validated with Zod before room names are constructed ✓

</details>

- `BroadcastHandler` namespace coupling (issue 4 above).

`BotSocketRegistry` keyed by `(workspaceId, botId, instanceId)` with `runtimeSessionId` tracked separately is fine — sessions are sub-instance and don't need their own registry tier. Bootstrap queries are bounded (LIMIT 200) so no N+1.

</details>

<details><summary>🔒 Security — 1 issue</summary>

- Claim token leak in bootstrap (issue 1 above).

Other surfaces checked clean:
- WS URL hint construction fails closed for non-loopback hosts (bracketed IPv6 fix in `f1479a8`).
- Socket auth middleware requires `BOT_RUNTIME_WRITE` scope, fails closed on `validateKey` throw, accepts both `auth.token` and `Authorization: Bearer`.
- Room name construction uses Zod-validated `instanceId` / `runtimeSessionId` so a crafted hello can't reach across workspaces.

</details>

</details>

kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-7[1m] | Reviewers: sonnet x3

All four findings from the previous review at HEAD `f1479a8` were addressed in `8f56238`:
- `claimToken` and `claimedByInstanceId` removed from bootstrap serialization (security)
- `bot_invocation:cancelled` and `bot_session_link:invalidated` removed entirely (INV-36)
- `BotSocketRegistry` now wires a 30s grace timer that calls `BotRuntimeInstanceRepository.markOffline` (plan §4.8)
- `BroadcastHandler` takes `botNamespace: Namespace` as a constructor dep (INV-13)

PR #633, opened 2026-05-26
fix(accounts): propagate WorkOS-rotated sealed sessions in account switcher

kristofferremback:
## Summary

Switching back to a parked account 401'd, and logging out always took every signed-in account on the browser with it. Two related fixes:

### 1. The switcher bug: `resolveAlts` discarded WorkOS-rotated sealed sessions

WorkOS access tokens are short-lived. `authService.authenticateSession()` transparently calls `session.refresh()` when the access token expires and returns `{ refreshed: true, sealedSession: <rotated> }` — at which point the pre-refresh sealed value is **revoked** at WorkOS. The control-plane `resolveAlts` ignored this and kept handing the dead pre-refresh sealed back to callers. So:

- `switch` promoted the dead sealed to the active cookie — the next `/api/auth/me` 401'd, browser saw "no session" and looked logged out.
- `list` left dead sealeds in the alt cookies — parked sessions died silently the moment the switcher view triggered a refresh.
- `addAndParkActive` parked the dead pre-refresh sealed of the previous active mid-OAuth-callback.

Fix: `resolveAlts` now writes the rotated sealed back to the alt cookie via `setAltSessionCookie` and returns it; the callers use the returned (post-refresh) sealed instead of the cookie the browser sent. `addAndParkActive` does the same for the previous-active sealed. `createAuthMiddleware` stamps `req.sealedSession` with the post-refresh value so the switch/remove handlers re-park the live token.

The e2e suite missed this because `StubAuthService.authenticateSession` always returns `refreshed: false`. Added focused unit tests with a `ProgrammableAuthService` that can return `refreshed: true` — three tests lock in the refreshed-sealed propagation through `switch`, `list`, and `addAndParkActive`.

### 2. Logout scope choice for multi-account browsers

When more than one account is signed in, clicking "Log out" now opens a confirm dialog with two options:

- **Sign out of \<active email\>** — revokes just the active WorkOS session and promotes a parked alt (the user stays signed in to the next account).
- **Sign out of all accounts** — the existing scope=all behavior.

Mechanics:
- Sidebar's logout action fetches `accountsApi.list()` (cached, reused by the switcher); if `liveCount > 1` it opens the dialog via a `?logout-confirm` search param, otherwise it calls `logout()` directly. Cold-call failure falls through to all-accounts so the user is never trapped.
- New `LogoutScopeDialog` mounted globally alongside `AccountSwitcherDialog`. `useAuth().logout` now takes `{ scope?: "current" | "all" }` and hits `/api/auth/logout?scope=current` or the default endpoint.
- Backend `logout` handler intercepts `scope=current`: re-uses `accountsService.remove(activeUser.id)` (which revokes the active session and promotes the lowest parked alt), then redirects to the same-app origin. Falls through to the full-logout path if there's no parked alt or the promote step throws.

## Files

- `packages/backend-common/src/auth/middleware.ts` — adds `req.sealedSession` with the post-refresh value
- `apps/control-plane/src/features/accounts/service.ts` — `resolveAlts` writes/returns rotated sealed; `addAndParkActive` parks rotated previous-active
- `apps/control-plane/src/features/accounts/handlers.ts` — switch/remove read `req.sealedSession`; list/resolve pass `res` through
- `apps/control-plane/src/features/auth/handlers.ts` — `scope=current` branch
- `apps/control-plane/src/features/accounts/service.test.ts` — new unit tests with refresh-aware stub
- `apps/frontend/src/auth/context.tsx` — `logout({ scope })` signature
- `apps/frontend/src/components/account-switcher/logout-scope-dialog.tsx` — new dialog
- `apps/frontend/src/components/layout/sidebar/sidebar-footer.tsx` — fetch-count + branch
- `apps/frontend/src/pages/workspace-layout.tsx` — mount `LogoutScopeDialog`

## Test plan

- [x] `bun test apps/control-plane/src/features/accounts/service.test.ts` — 3 pass (new refresh-propagation cases)
- [x] `bun run --cwd apps/backend test:unit` — 1423 pass, 0 fail
- [x] `bun run --cwd apps/frontend test` — 1824 pass, 0 fail
- [x] `bun run --cwd packages/backend-common typecheck`
- [x] `bun run --cwd apps/control-plane typecheck`
- [x] `bun run --cwd apps/backend typecheck`
- [x] `bun run --cwd apps/frontend typecheck`
- [ ] Manual: sign in to account A, add account B, switch back to A — A stays signed in (was 401'ing before).
- [ ] Manual: signed in to A and B, click "Log out" — dialog appears; "Sign out of A" lands you signed in as B; "Sign out of all accounts" empties the jar.
- [ ] Manual: signed in to A only, click "Log out" — no dialog, direct full logout.
- [ ] Backend e2e (`apps/control-plane/tests/e2e/accounts.test.ts`) — needs Postgres on :5454, not available in remote env; re-run locally before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_011vVdzv2dAEVK9hUHce6HcQ)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/633"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782421126&installation_id=129946197&pr_number=633&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F633&signature=e0f5960dbf4aa79b9124b02f21197ab39f9db9f4d93607150f5602b6b4daa3d4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/control-plane/src/features/accounts/handlers.ts
+++ b/apps/control-plane/src/features/accounts/handlers.ts
@@ -1,6 +1,6 @@
 import type { Request, Response } from "express"
 import { z } from "zod/v4"
-import { HttpError, SESSION_COOKIE_NAME } from "@threa/backend-common"
+import { HttpError } from "@threa/backend-common"
 import type { AccountsService } from "./service"
 
 interface Dependencies {
@@ -24,7 +24,7 @@ export function createAccountsHandlers({ accountsService }: Dependencies) {
       if (!req.workosUserId || !req.authUser) {
         throw new HttpError("Not authenticated", { status: 401, code: "NOT_AUTHENTICATED" })
       }
-      res.json(await accountsService.list(req.cookies, req.authUser))
+      res.json(await accountsService.list(res, req.cookies, req.authUser))
     },
 
     async resolve(req: Request, res: Response) {
@@ -35,31 +35,33 @@ export function createAccountsHandlers({ accountsService }: Dependencies) {
       if (!parsed.success) {
         throw new HttpError("Invalid query", { status: 400, code: "VALIDATION_ERROR" })
       }
-      res.json(await accountsService.resolve(req.cookies, req.authUser, parsed.data))
+      res.json(await accountsService.resolve(res, req.cookies, req.authUser, parsed.data))
     },
 
     async switch(req: Request, res: Response) {
-      if (!req.workosUserId || !req.authUser) {
+      if (!req.workosUserId || !req.authUser || !req.sealedSession) {
         throw new HttpError("Not authenticated", { status: 401, code: "NOT_AUTHENTICATED" })
       }
       const parsed = targetSchema.safeParse(req.body)
       if (!parsed.success) {
         throw new HttpError("Invalid request body", { status: 400, code: "VALIDATION_ERROR" })
       }
-      const activeSealed = req.cookies[SESSION_COOKIE_NAME]
-      res.json(await accountsService.switch(res, req.cookies, activeSealed, req.authUser, parsed.data.targetUserId))
+      res.json(
+        await accountsService.switch(res, req.cookies, req.sealedSession, req.authUser, parsed.data.targetUserId)
+      )
     },
 
     async remove(req: Request, res: Response) {
-      if (!req.workosUserId || !req.authUser) {
+      if (!req.workosUserId || !req.authUser || !req.sealedSession) {
         throw new HttpError("Not authenticated", { status: 401, code: "NOT_AUTHENTICATED" })
       }
       const parsed = targetSchema.safeParse(req.body)
       if (!parsed.success) {
         throw new HttpError("Invalid request body", { status: 400, code: "VALIDATION_ERROR" })
       }
-      const activeSealed = req.cookies[SESSION_COOKIE_NAME]
-      res.json(await accountsService.remove(res, req.cookies, activeSealed, req.authUser, parsed.data.targetUserId))
+      res.json(
+        await accountsService.remove(res, req.cookies, req.sealedSession, req.authUser, parsed.data.targetUserId)
+      )
     },
   }
 }
--- a/apps/control-plane/src/features/accounts/service.test.ts
+++ b/apps/control-plane/src/features/accounts/service.test.ts
@@ -45,6 +45,7 @@ type AuthReply = AuthResult | ((sealed: string) => AuthResult)
 
 class ProgrammableAuthService implements AuthService {
   private replies = new Map<string, AuthReply>()
+  readonly revoked: string[] = []
 
   on(sealed: string, reply: AuthReply): void {
     this.replies.set(sealed, reply)
@@ -71,8 +72,9 @@ class ProgrammableAuthService implements AuthService {
   async getLogoutUrl(): Promise<string | null> {
     return null
   }
-  async revokeSession(): Promise<boolean> {
-    return false
+  async revokeSession(sealed: string): Promise<boolean> {
+    this.revoked.push(sealed)
+    return true
   }
 }
 
@@ -165,5 +167,65 @@ describe("AccountsService — refreshed sealed propagation", () => {
     expect(result).toEqual({ ok: true })
     expect(res.__cookies.get(SESSION_COOKIE_NAME)).toBe("sealed_b_new")
     expect(res.__cookies.get(altCookieName(0))).toBe("sealed_a_rotated")
+  })
+
+  test("removeCurrentAndPromote revokes the rotated sealed and promotes the alt", async () => {
+    // The "sign out of just this account" path: WorkOS may rotate the active
+    // session's refresh token during the authenticate() pre-check. If we then
+    // pass the pre-refresh sealed to revokeSession, WorkOS's session.refresh
+    // fails on a stale token and revoke silently returns false — the live
+    // WorkOS session stays alive. The rotated value must reach revoke.
+    const auth = new ProgrammableAuthService()
+    auth.on("sealed_a_active", {
+      success: true,
+      refreshed: true,
+      sealedSession: "sealed_a_rotated",
+      user: makeUser("user_a"),
+    })
+    auth.on("sealed_b_parked", { success: true, refreshed: false, user: makeUser("user_b") })
+
+    const service = new AccountsService({ authService: auth, membership: NEVER_MEMBER })
+    const res = makeRes()
+    const cookies = { [altCookieName(0)]: "sealed_b_parked" }
+
+    const promoted = await service.removeCurrentAndPromote(res, cookies, "sealed_a_active")
+
+    expect(promoted).toBe(true)
+    // Active cookie now holds the promoted parked sealed; the parked slot was
+    // cleared on promotion.
+    expect(res.__cookies.get(SESSION_COOKIE_NAME)).toBe("sealed_b_parked")
+    expect(res.__cookies.get(altCookieName(0))).toBeNull()
+    // The rotated sealed is what reached revokeSession — without the fix this
+    // would be "sealed_a_active" and the active WorkOS session would survive.
+    expect(auth.revoked).toEqual(["sealed_a_rotated"])
+  })
+
+  test("removeCurrentAndPromote returns false when there is no parked alt to promote", async () => {
+    // No alts means a "sign out current" would empty the cookie jar — that's
+    // indistinguishable from a full logout, so the caller falls through to
+    // the WorkOS SSO logout flow instead of going through revoke + promote.
+    const auth = new ProgrammableAuthService()
+    auth.on("sealed_a_active", { success: true, refreshed: false, user: makeUser("user_a") })
+
+    const service = new AccountsService({ authService: auth, membership: NEVER_MEMBER })
+    const res = makeRes()
+
+    const promoted = await service.removeCurrentAndPromote(res, {}, "sealed_a_active")
+
+    expect(promoted).toBe(false)
+    // No cookie writes, no revoke calls — caller is responsible for full logout.
+    expect(res.__cookies.size).toBe(0)
+    expect(auth.revoked).toEqual([])
+  })
+
+  test("removeCurrentAndPromote returns false when there is no active sealed", async () => {
+    const auth = new ProgrammableAuthService()
+    const service = new AccountsService({ authService: auth, membership: NEVER_MEMBER })
+    const res = makeRes()
+
+    const promoted = await service.removeCurrentAndPromote(res, {}, undefined)
+
+    expect(promoted).toBe(false)
+    expect(res.__cookies.size).toBe(0)
   })
 })
--- a/apps/control-plane/src/features/accounts/service.test.ts
+++ b/apps/control-plane/src/features/accounts/service.test.ts
@@ -218,6 +218,45 @@ describe("AccountsService — refreshed sealed propagation", () => {
     expect(auth.revoked).toEqual([])
   })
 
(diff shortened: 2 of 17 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-7[1m] | Reviewers: sonnet x3

Found 4 issues (1 real bug, 3 compliance):

1. `apps/control-plane/src/features/auth/handlers.ts:208-215` — **The `scope=current` logout path leaks the live WorkOS session when refresh rotates mid-handler.** `authenticateSession(session)` on line 209 may rotate the refresh token and return a new `auth.sealedSession`, at which point the original `session` (raw pre-refresh cookie) is dead at WorkOS. Line 214 then passes that dead `session` to `accountsService.remove()` as `activeSealed`. Inside `remove`, `revokeSession(activeSealed)` calls `session.authenticate()` on the stale sealed, fails (`!authenticated`), and returns `false` silently — the rotated (live) sealed never reaches `revokeSession`, so the user's active WorkOS session stays alive after "Sign out of current account". This is exactly the rotation-aware pattern this PR already established for `addAndParkActive` (`prevSealedToPark = prevAuth.refreshed && prevAuth.sealedSession ? prevAuth.sealedSession : prevActiveSealed`); it just wasn't applied in the new scope=current branch. Fix: derive `const sealedToRevoke = auth.refreshed && auth.sealedSession ? auth.sealedSession : session` and pass `sealedToRevoke` into `accountsService.remove(...)`.
   https://github.com/threahq/threa/blob/c40e7078b47a7e3fd69c78e61135a796183a29d8/apps/control-plane/src/features/auth/handlers.ts#L207-L221

2. `apps/control-plane/src/features/auth/handlers.ts:191-193` — CLAUDE.md [INV-55, _API Contracts, Validation, and Types_]: _"Inputs (`body`, `query`, `params`) are validated with Zod schemas, not manual `typeof` checks."_ The new `scope` query parameter uses `typeof req.query.scope === "string" ? req.query.scope : undefined` while every other input in this handler (`callbackSchema`, `magicSendSchema`, `magicVerifySchema`) goes through `safeParse`. It also gates the most security-sensitive new branch (the revoke-and-promote path that acts on `auth.user.id`). Add a `z.object({ scope: z.enum(["current"]).optional() })` parse.
   https://github.com/threahq/threa/blob/c40e7078b47a7e3fd69c78e61135a796183a29d8/apps/control-plane/src/features/auth/handlers.ts#L190-L194

3. `apps/frontend/src/components/layout/sidebar/sidebar-footer.tsx:128-131` — CLAUDE.md [INV-33, _API Contracts, Validation, and Types_]: _"Avoid magic strings by centralizing constants at the source of truth."_ The string `"logout-confirm"` is written inline here while the canonical definition exists as `const SEARCH_PARAM = "logout-confirm"` inside `logout-scope-dialog.tsx` (unexported). The dialog owns this param — export `LOGOUT_CONFIRM_PARAM` from the dialog (or a shared module) and import it here so a rename can't silently desync the setter and reader.
   https://github.com/threahq/threa/blob/c40e7078b47a7e3fd69c78e61135a796183a29d8/apps/frontend/src/components/layout/sidebar/sidebar-footer.tsx#L127-L133

4. `apps/frontend/src/components/layout/sidebar/sidebar-footer.tsx:120` — CLAUDE.md [INV-33, _API Contracts, Validation, and Types_]: same rule. `["accounts", "list"]` is now in three places — inline here, and as file-local `ACCOUNTS_LIST_KEY` consts in both `account-switcher-dialog.tsx` and `logout-scope-dialog.tsx`. `sidebar-footer.tsx` uses `fetchQuery` so TanStack won't warn on a shape mismatch. Export once (e.g. from `@/api/accounts.ts`) and import in all three consumers.
   https://github.com/threahq/threa/blob/c40e7078b47a7e3fd69c78e61135a796183a29d8/apps/frontend/src/components/layout/sidebar/sidebar-footer.tsx#L117-L141

---

`docs/plans/multi-account-login-split.md` lists "The auth middleware does not change" as a security invariant. This PR adds `req.sealedSession` to the middleware, but the invariant's three load-bearing properties — _"still reads exactly `req.cookies[SESSION_COOKIE_NAME]`, runs one `authenticateSession`, and sets `req.authUser`. One request → one validated `workosUserId`"_ — are all preserved. The new field is a read-only propagation of the post-refresh sealed for handlers that re-park, not a new auth surface. Spirit of the invariant intact.

The `scope=current` / `LogoutScopeDialog` feature is net-new (user-requested) and not specified in the plan, but the plan's cookie model (env-scoped, parked alts) is unchanged.
</details>

<details><summary>🔍 Bugs [1 issue — see #1 above]</summary>

The scope=current revoke bug (item 1) is the only correctness finding ≥80. Both correctness and design reviewers independently surfaced it, scoring 85 and 88.
</details>

<details><summary>📋 CLAUDE.md Compliance [3 violations — see #2-4 above]</summary>

INV-55 (Zod for inputs) on the new `scope` query param; INV-33 (centralize constants) on `"logout-confirm"` and `["accounts", "list"]`.

Other rules checked: INV-9, INV-11 (the `catch { /* fall through */ }` in scope=current is a documented safe-downgrade to all-accounts logout, not a hidden default), INV-18, INV-23/24/26, INV-32, INV-34, INV-35/37, INV-38, INV-40, INV-25 (test "without the fix" comment is borderline regression-prevention rather than change history), INV-36.
</details>

`resolveAlts` now mutating cookies on GET paths (`list`, `resolve`) is justified — without it, viewing the switcher kills parked sessions silently. The comment that GET stays Set-Cookie-free was load-bearing only until WorkOS rotation was modeled. Both reviewers flagged this as defensible.

Query keys for the accounts list could be shared between dialog and footer (covered as INV-33 finding #4).

Test gap that hid the original bug: `StubAuthService` always returns `refreshed: false`. The new `ProgrammableAuthService` in this PR closes the gap at unit level, but extending the stub (or adding a refresh toggle) would catch future regressions at the e2e level too. Not blocking.
</details>

The scope=current revoke leak (item 1) is correctness, not exploitable from outside — a refresh-window race on the user's own session, not cross-account.

LogoutScopeDialog renders `user?.email || user?.name` inside React text nodes (escaped by default, no `dangerouslySetInnerHTML`). `auth.user.id` in the scope=current path is server-derived from the validated session, not from request input.
</details>

---

kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: claude-opus-4-7[1m] | Reviewers: sonnet x3

Four findings from the prior review are all resolved (scope=current rotated-sealed revoke; `logoutQuerySchema` Zod gate; `LOGOUT_CONFIRM_PARAM` / `ACCOUNTS_LIST_KEY` extracted). Two architecture-level concerns surfaced this pass, both non-blocking:

1. `apps/control-plane/src/features/auth/handlers.ts:213-231` — CLAUDE.md [INV-34, _Architecture Boundaries and Dependency Flow_]: _"Handlers and workers stay thin and infrastructure-focused, delegating business logic to services and domain modules."_ The new `scope=current` branch performs multi-step orchestration directly in the handler — `authService.authenticateSession()`, then `readAltSessionCookies()`, then derives `sealedToRevoke`, then delegates to `accountsService.remove()`. The symmetric "park previous active" sequence already lives inside `AccountsService.addAndParkActive`. Consider extracting `AccountsService.removeCurrentAndPromote(res, cookies, session)` so the handler shrinks back to a thin pass-through and the rotation invariant lives next to the other cookie writes.
   https://github.com/threahq/threa/blob/2f65bbd92314ebef2d2c2b1ad54c40099e16b637/apps/control-plane/src/features/auth/handlers.ts#L213-L231

2. `apps/control-plane/src/features/accounts/service.ts:90`, `apps/control-plane/src/features/auth/handlers.ts:221`, `packages/backend-common/src/auth/middleware.ts:62` — Triple-duplicated rotation idiom. The line `r.refreshed && r.sealedSession ? r.sealedSession : original` appears in three files for the same invariant. `resolveAlts` was previously a pure read helper and now takes `Response` solely to do the cookie write as a side effect — but the same pattern still had to be re-inlined in the `logout` handler because that handler calls `remove` directly rather than going through `resolveAlts`. A one-liner `pickSealed(auth: AuthResult, original: string): string` in `@threa/backend-common` would let middleware/service/handler share one source of truth, and let `resolveAlts` stay effect-free by returning rotated sealeds as data for the caller to commit (mirroring how `createAuthMiddleware` already stamps `req.sealedSession` as data rather than writing the cookie).
   https://github.com/threahq/threa/blob/2f65bbd92314ebef2d2c2b1ad54c40099e16b637/apps/control-plane/src/features/accounts/service.ts#L88-L95

---

No standalone plan file for this PR — the PR description IS the spec. Both halves are present with the matching frontend wiring (`LogoutScopeDialog`, sidebar fetchQuery+branch, `useAuth().logout({ scope })`, `req.sealedSession` propagation through middleware → handlers → service).
</details>

The previous review's correctness finding (scope=current passing pre-refresh sealed to revoke) is now resolved — `auth.refreshed && auth.sealedSession ? auth.sealedSession : session` in `auth/handlers.ts:221` matches the rotation-aware pattern used by `addAndParkActive` and `resolveAlts`. No new correctness issues ≥80.

scope=current skips `getLogoutUrl`, so the WorkOS SSO session at the IdP isn't terminated for the active account. Same behavior as the existing `/api/accounts/remove` endpoint — not introduced here, flagged for awareness only.
</details>

<details><summary>📋 CLAUDE.md Compliance [1 violation — see #1 above]</summary>

Four prior findings resolved: scope=current rotated-sealed revoke, `logoutQuerySchema` (INV-55), `LOGOUT_CONFIRM_PARAM` exported (INV-33), shared `ACCOUNTS_LIST_KEY` (INV-33).

Remaining: INV-34 on the logout handler (item #1). Other rules checked: INV-9, INV-11, INV-12/13, INV-15, INV-18, INV-23/24/26, INV-32, INV-36, INV-38, INV-40, INV-55.
</details>

<details><summary>🏗️ Design [1 concern — see #2 above]</summary>

The fix shape is good: data flow through `req.sealedSession` mirrors the existing `req.authUser` pattern, and `resolveAlts` performing cookie writes is a justified workaround for the fact that the validate→use loop straddles requests in time. The remaining concern is duplication (#2) — same idiom in three files with no shared helper.

Test gap noted: `StubAuthService` doesn't model rotation; `ProgrammableAuthService` in this PR closes the gap at unit level. Extending the stub would catch future regressions at e2e level too.
</details>

`scope=current` restricted by Zod enum (`z.enum(["current"])`); unknown values fall through to default all-accounts logout. The `forwardedHost`-derived `sameAppOrigin` reuses the existing `isTrustedHost` allow-list — no new open-redirect surface. `req.sealedSession` is server-side only and not echoed in any handler response.

`LogoutScopeDialog` renders `user?.email || user?.name` inside React text nodes (escaped by default, no `dangerouslySetInnerHTML`).
</details>

---

PR #637, opened 2026-05-27
feat(e2e): Phase 3.5 — keep ciphertext at rest, decrypt on render

kristofferremback:
## Summary

Phase 3.5 of E2E-encrypted scratchpads. Up to now we decrypted on sync (when an event arrived), which meant plaintext landed in IDB alongside ciphertext. That trades the lock guarantee for convenience: if the user locks the session, the plaintext is still sitting in IndexedDB on disk.

This change moves decryption out of the sync layer entirely. Events arrive ciphertext-only and stay that way in IDB (`db.events.payload` keeps the wire `{ciphertext, envelope}` shape). Decryption happens only in memory at render time, on demand, behind an LRU-bounded in-memory cache. When the user locks (or switches accounts) the cache is dropped — plaintext never outlives the unlocked session.

### What changed

**New: render-time decryption layer**
- `lib/crypto/decrypt-cache.ts` — module-level cache with insertion-order LRU (cap 500), per-event listener pattern, coalesces concurrent decrypts via an inflight Promise map. `clearDecryptCache()` drops everything and notifies subscribers.
- `hooks/use-decrypted-message-content.ts` — `useSyncExternalStore` reads the cache, kicks off decrypts on miss when unlocked, returns a discriminated union (`plaintext | locked | pending | decrypted | failed`) so the UI can render distinct states.

**Removed: sync-time decryption**
- `sync/stream-sync.ts` no longer decrypts in `applyStreamBootstrap` or `handleMessageCreated` (deleted `decryptEventPayloadIfNeeded`, `decryptBootstrapEventsIfNeeded`, the `_userId` param).
- `sync/workspace-sync.ts` no longer takes a `userId` param or runs `decryptBootstrapEventsIfNeeded`. Field renamed to `streamBootstraps`.
- `sync/sync-engine.ts`, `pages/workspace-layout.tsx` — dead `currentWorkspaceUserId` plumbing removed (its only consumer was the sync-time decrypt path).

**UI integration**
- `components/timeline/message-event.tsx` — runs each E2E message through `useDecryptedMessageContent` and shows `🔒 Locked — unlock to view` / `🔒 Decrypting…` / `🔒 Decryption failed` placeholders during non-decrypted states.
- `components/layout/sidebar/{stream-item,scratchpad-item}.tsx` — sidebar previews show `🔒 Encrypted message` rather than leaking a wire placeholder.

**Lock-time invalidation**
- `stores/e2e-session-store.ts` — `lock()` and `resetE2eSessionStoreCache()` now call `clearDecryptCache()` so locking really does drop plaintext.

### Self-review fix (second commit)

A race surfaced during self-review: a `requestDecryption` that was in flight when `clearDecryptCache()` ran would still write its plaintext back into the cleared cache on resolve, defeating the lock guarantee. Fixed by snapshotting a module-level `generation` counter before the decrypt's `await`; if the generation has changed (cache was cleared) by the time the decrypt resolves, the result is dropped rather than touched into the cache. Test added.

## Test plan

- [x] `bun run test` — 1838/1838 passing (added one new test for the lock-race fix)
- [x] `bun run typecheck` clean
- [x] `bun run lint` clean
- [ ] Manual: lock session mid-decrypt; verify cached plaintext is dropped and no leak in DevTools after relock
- [ ] Manual: scroll through a long E2E scratchpad; verify per-message decrypt latency is unnoticeable on repeat visits (cache hits)
- [ ] Manual: hard-refresh while locked; verify `🔒 Locked` placeholder shows and unlocking populates content without a re-fetch

---
_Generated by [Claude Code](https://claude.ai/code/session_01QtinEwrFfpJEdkENwAKCj8)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/637"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782468127&installation_id=129946197&pr_number=637&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F637&signature=4a8f21e96429f2f706ba69c0d9e5375bcaa83d4f196942870c8b1cc41cc42c63"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/layout/sidebar/scratchpad-item.tsx
+++ b/apps/frontend/src/components/layout/sidebar/scratchpad-item.tsx
@@ -158,6 +158,7 @@ export function ScratchpadItem({
                 compact={compact}
                 showPreviewOnHover={showPreviewOnHover}
                 isMobile={isMobile}
+                e2eEnabled={streamWithPreview.e2eEnabled}
               />
             </div>
           </div>
--- a/apps/frontend/src/components/layout/sidebar/stream-item.tsx
+++ b/apps/frontend/src/components/layout/sidebar/stream-item.tsx
@@ -99,6 +99,13 @@ interface StreamItemPreviewProps {
   compact: boolean
   showPreviewOnHover: boolean
   isMobile: boolean
+  /**
+   * E2E streams: sidebar previews are public surfaces, so we never decrypt
+   * here. Phase 3.5 keeps ciphertext at rest and routes decryption through
+   * the per-message render hook — show a sentinel in the sidebar so the
+   * placeholder zero-width-space wire content never leaks.
+   */
+  e2eEnabled?: boolean
 }
 
 export function StreamItemPreview({
@@ -108,6 +115,7 @@ export function StreamItemPreview({
   compact,
   showPreviewOnHover,
   isMobile,
+  e2eEnabled,
 }: StreamItemPreviewProps) {
   if (!preview?.content) return null
 
@@ -123,7 +131,9 @@ export function StreamItemPreview({
       aria-hidden={hoverPreview ? "true" : undefined}
     >
       <span className="truncate flex-1">
-        {getActorName(preview.authorId, preview.authorType)}: {truncateContent(preview.content, 50, toEmoji)}
+        {e2eEnabled
+          ? `${getActorName(preview.authorId, preview.authorType)}: 🔒 Encrypted message`
+          : `${getActorName(preview.authorId, preview.authorType)}: ${truncateContent(preview.content, 50, toEmoji)}`}
       </span>
       <RelativeTime date={preview.createdAt} className="flex-shrink-0" />
     </div>
@@ -330,6 +340,7 @@ export function StreamItem({
                 compact={compact}
                 showPreviewOnHover={showPreviewOnHover}
                 isMobile={isMobile}
+                e2eEnabled={stream.e2eEnabled}
               />
             </div>
           </div>
--- a/apps/frontend/src/components/timeline/message-event.tsx
+++ b/apps/frontend/src/components/timeline/message-event.tsx
@@ -20,6 +20,7 @@ import { usePendingMessages, usePanel, createDraftPanelId, useTrace, useMessageS
 import { useUserProfile } from "@/components/user-profile"
 import { useFormattedDate } from "@/hooks/use-formatted-date"
 import { useMessageMarkdownCopy } from "@/hooks/use-message-markdown-copy"
+import { useDecryptedMessageContent, type DecryptedMessageContent } from "@/hooks/use-decrypted-message-content"
 import { useEditLastMessage } from "./edit-last-message-context"
 import {
   useActors,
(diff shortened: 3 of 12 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: Opus | Reviewers: Sonnet x3

Found 1 issue (non-blocking):

1. `apps/frontend/src/stores/draft-store.ts` (missing change) — **Composer drafts in `db.draftMessages` for E2E streams still persist plaintext.** Plan line 338 (`.claude/plans/ok-nice-let-s-plan-rosy-goblet.md`) explicitly lists this as a Phase 3.5 requirement alongside ciphertext-at-rest for events: *"Composer drafts in `db.draftMessages` for E2E streams get the same treatment."* The PR closes the events leak but leaves the drafts surface open — a user who locks the session can still have their unsent draft sitting plaintext in IDB. Could land as a follow-up if you're scoping this PR strictly to the event payload path; flagging because it's named in the same phase.

---

The events ciphertext-at-rest work matches the plan precisely. The draft-store gap above is the only adherence item that scored at threshold — it's explicitly enumerated in Phase 3.5's scope but no `db.draftMessages` write path or table definition is touched in the diff.

</details>

Bug scan focused on the generation-counter race fix, LRU eviction logic, listener cleanup, reactivity, sync-layer removal call sites, the `` zero-width-space wire placeholder, and IDB persistence. Nothing reached the threshold. Sub-threshold observations only: a theoretical "≥501 concurrent in-flight decrypts" eviction edge that affects UI flash but not safety; an unstable `subscribe` arrow function (pre-existing pattern). Generation guard correctly closes the post-lock decrypt window.

</details>

No invariant violations introduced. INV-15 (UI components stay UI-focused), INV-18 (no nested components), INV-25 (comments explain why), INV-36 (no speculative features), INV-38 (delete dead code — note the PR actively removes dead `currentWorkspaceUserId` plumbing), INV-48 (no `mock.module`/`vi.mock` — tests use scoped `spyOn` against namespace imports), INV-58 (`contentJson` canonical), INV-60 (preview surfaces strip markdown — sidebar uses the `🔒 Encrypted message` sentinel) all hold.

</details>

The cache-as-module-singleton + per-key listener + inflight-coalesce + generation-counter shape is clean. Sub-threshold observations worth noting but not blocking: (a) the `decrypted` discriminated-union return creates a fresh object every render, but no downstream consumer is `React.memo`-wrapped so no bailout is lost; (b) `ThreadCard` (`apps/frontend/src/components/timeline/thread-card.tsx`) and the mobile long-press drawer in `scratchpad-item.tsx` render thread/preview content from `contentMarkdown` without an `e2eEnabled` sentinel substitution — they'll show invisible (ZWS) text for E2E thread replies. Same surface family as the sidebar fix this PR shipped; consider extending the sentinel pattern when the draft work lands.

</details>

Specifically verified: (a) cache truly drops plaintext on lock (`clearDecryptCache` clears entries + inflight + bumps generation), (b) post-lock in-flight decrypts cannot leak plaintext back into the cache (generation snapshot + check), (c) wire ciphertext never enters the DOM (only `contentMarkdown` and rendered placeholder strings are passed downstream; `payload.ciphertext` is read in the hook but never returned), (d) zero-width-space placeholder is replaced with `🔒 Locked — unlock to view` for non-`plaintext` statuses in `applyDecryptedContent`. No `dangerouslySetInnerHTML` introduced.

</details>

---

kristofferremback:
Re the plan-adherence finding above (composer drafts): tracking as follow-up in #638. Scoped to ship separately because the lock-time composer behavior and existing-draft handling are open design calls the plan flagged for "resolve when the phase starts" — keeping #637 focused on the events-at-rest slice that's already complete.

---

PR #639, opened 2026-05-27
fix: package pi remote extension dependencies

kristofferremback:
## Problem

#636 made the Pi remote extension require `socket.io-client` for the `/bot` WebSocket transport, but the extension was still documented/stored as a single-file example under `docs/examples/pi-remote/`. Copying only that file into `~/.pi/agent/extensions/` makes `/reload` fail:

```text
Failed to load extension: Cannot find module 'socket.io-client'
```

## Solution

Move the remote-control adapter into a real Pi package at `extensions/pi-remote/`:

- adds `package.json` with `socket.io-client` as a runtime dependency
- declares the Pi extension entrypoint via `pi.extensions`
- moves the extension and tests under `extensions/pi-remote/src/`
- adds package install instructions for local Pi usage
- updates the `update-pi-remote-plugin` skill to point at the new canonical location

This directory is intentionally **not** under `packages/` so it is not part of the Bun workspace graph and does not get pulled into cloud Docker images.

## Local install shape

```bash
rm -f ~/.pi/agent/extensions/threa-remote.ts
rm -rf ~/.pi/agent/extensions/threa-remote
mkdir -p ~/.pi/agent/extensions/threa-remote
cp -R extensions/pi-remote/. ~/.pi/agent/extensions/threa-remote/
cd ~/.pi/agent/extensions/threa-remote
npm install
```

Then run `/reload`.

## Test plan

- [x] `bun test ./extensions/pi-remote/src/threa-remote.test.ts`
- [x] `bun --check extensions/pi-remote/src/threa-remote.ts`
- [x] `bun run check:dockerfiles`
- [x] commit hook full lint/typecheck/API-doc check passed
- [x] installed locally under `~/.pi/agent/extensions/threa-remote/` with `socket.io-client`

---


kristofferremback:
Don't think this is necessary

PR #646, opened 2026-05-27
feat(e2e): Phase 5a.2 — enclave service, EIK registry, invite UI

kristofferremback:
## Summary

Phase 5a.2 of the Ariadne enclave plan. Stands up the end-to-end loop without tools so the deliverable from the plan is reachable:

> *"Kris and his wife can journal against Ariadne in an E2E scratchpad. No tools. Token cost shows up in usage."*

Builds on 5a.1 (`@threa/crypto` + `@threa/agent-runtime` extraction, #642). 5a.3 (encrypted trace steps) and 5a.4 (`web_search` + `read_url`) follow.

## What's in this PR

**New service: `apps/enclave/`** — Bun + Express, mirrors the regional backend stack. HPKE keystore generates an X25519 EIK at boot; `/healthz`, `/pubkey`, `/attestation` (stub), `/invoke` (bearer-auth). Registers with the backend at boot, heartbeats every 30s, gracefully revokes on shutdown. Falls out of the live set at 2 minutes of staleness.

**Backend `enclave-runtimes` feature** — `enclave_runtimes` table (elr_ prefix), repository/service/handlers, internal endpoints for register-key / heartbeat / revoke. New workspace-scoped `GET /enclave/active-keys` so the frontend can fetch the current live EIK fleet per send.

**`enclave-dispatcher` + worker** — Loads ciphertext history + the owner UIK + the live EIK set, dispatches to a randomly picked instance, writes the sealed reply via `EventService`, records token usage via `AICostService` with the sidecar. Falls back to a second live runtime on primary failure. INV-30 / INV-41: the dispatcher releases the pool before the slow LLM HTTP round-trip.

**Companion outbox routing** — branches on `e2eRow.invitedAgentKind === "enclave"` to dispatch `ENCLAVE_PERSONA_AGENT` (always-on companion in E2E since `@`-mention can't see ciphertext). The plaintext path is unchanged.

**Ariadne persona** — gains `e2eCapable: true` and `e2eEnabledTools: [SEND_MESSAGE]` (web_search + read_url ship in 5a.4).

**Frontend invite UI + multi-recipient encryption** — `encryptMessage` now takes `recipients[]`; enclave-invited streams fan envelopes to `[UIK, ...liveEIKs]`. New `useInviteEnclave` hook, `InviteEnclaveButton` (also surfaces the "Ariadne is here" badge), `api/enclave.ts` client. `VirtualStream` carries `e2eInvitedAgentKind` through so the send path can branch.

**Bun X25519 unblock** — swapped the KEM in `@threa/crypto` from `@hpke/core`'s `DhkemX25519` to `@hpke/dhkem-x25519` (noble-curves backed). Bun 1.3's WebCrypto doesn't implement X25519 seal/open, so the enclave would crash on first invoke without this. The browser side is unaffected because the noble suite works there too.

## Architectural choices worth highlighting

- **No `keyId` pinned on the stream.** Recipients are recomputed per-message from the live EIK set so HA isn't tied to "invite the enclave once". Verified by `getActiveKeys` in the send path.
- **Multi-instance from day one.** No "single active" constraint on `enclave_runtimes`; uniqueness is on `key_id` only. Heartbeat-based liveness (2-minute window) instead of explicit revocation for normal churn.
- **Plaintext never leaves the backend.** Dispatcher reads only `ciphertext` / `envelope` / `e2eVersion` from history rows; a dedicated test asserts the serialized invoke request contains no plaintext when the history has plaintext columns alongside the ciphertext.
- **No `PersonaAgent` reuse.** The orchestrator is ~200 LOC and DI-friendly (`AIFactory` seam for tests); the 700-LOC plaintext orchestrator is untouched.

## Self-review fixes pushed in the follow-up commit

- `functionId` style: `enclave_persona_agent` → `enclave-persona-agent` (kebab, matching `agent-loop` / `summary-update`).
- `enclaveApi.invite` return type: `Stream` (the wire actually returns `Stream`, not `StreamWithPreview`); the bootstrap mapper preserves `lastMessagePreview` via spread.
- Enclave bearer auth: `timingSafeEqual` with explicit length check instead of `!==`. Defense in depth even though the enclave isn't public-facing in 5a.

## Test plan

- [x] `bun run typecheck` clean across all workspaces
- [x] `bun run --cwd apps/backend lint` clean (pre-commit caught two INV-52 internals imports — fixed before push)
- [x] `bun test apps/backend/src/features/agents` — 271 pass (incl. 5 new dispatcher tests)
- [x] `bun test apps/backend/src/features/enclave-runtimes` — 7 new tests (register/heartbeat/listLive/revoke SQL shape + idempotency)
- [x] `bun test apps/enclave/src/orchestrator.test.ts` — 3 tests (real HPKE round-trip via noble X25519 under Bun, allowlist rejection, empty-history guard)
- [x] `bun test apps/backend/src/features/e2e-streams apps/backend/src/features/streams` — 60 pass
- [x] `bun test packages/crypto` — 16 pass (HPKE round-trip under Bun confirmed)
- [ ] Manual end-to-end on a single laptop with one enclave instance: create encrypted scratchpad → invite Ariadne → send → reply renders, `messages.ciphertext IS NOT NULL`, `ai_usage_records` row tagged with the session.
- [ ] HA smoke: two enclave instances running, kill the primary mid-conversation, send another message — dispatcher fallback succeeds against the surviving instance.

> Frontend test suite has pre-existing DOM-setup failures on `main` (`document[isPrepared]` in `@testing-library/user-event`) that are unrelated to this change — verified by stash + re-run.

## Out of scope (queued for follow-ups)

- 5a.3 — encrypted `agent_session_steps` (ciphertext + envelope columns, new step endpoint, `EncryptedTraceObserver`, frontend per-step decrypt cache).
- 5a.4 — `web_search` + `read_url` wired in the enclave's tool array. No new backend endpoints needed once 5a.3 ships.
- 5b — workspace-coupled tools (`list_streams`, `search_*`, `describe_memo`, attachments, GitHub, Linear) gated on a callback-channel subsystem.
- Real TEE attestation. `GET /attestation` stub returns commit SHA + build hash for docs only.

https://claude.ai/code/session_01QtinEwrFfpJEdkENwAKCj8

---
_Generated by [Claude Code](https://claude.ai/code/session_01QtinEwrFfpJEdkENwAKCj8)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/646"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782503946&installation_id=129946197&pr_number=646&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F646&signature=11025b1cb2dd74cabe270f8b49ab6791e5133d0cba5823a9d855c7eba5746d95"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review
**Review models:** Orchestrator: Opus 4.7 | Reviewers: sonnet x3

Found 4 issues:

1. `apps/backend/src/features/agents/enclave-dispatcher.ts:80` — `e2eRow.invitedAgentKind !== "enclave"` uses a raw string literal. The companion-outbox handler in this same PR routes through `E2eInvitedAgentKinds.ENCLAVE` (`companion-outbox-handler.ts:121`); the dispatcher should follow suit. CLAUDE.md [INV-33, API and backend contracts]: *"Avoid magic strings by centralizing constants at the source of truth (INV-33)."*
   https://github.com/threahq/threa/blob/572e666fe5feaac0514825fc0e89643e60a991e3/apps/backend/src/features/agents/enclave-dispatcher.ts#L79-L86

2. `apps/backend/src/features/agents/enclave-dispatcher.ts:178` — `authorType: "persona"` literal. `AuthorTypes.PERSONA` is exported from `@threa/types` (`packages/types/src/constants.ts:41`) and used elsewhere in this PR. CLAUDE.md [INV-33, API and backend contracts]: *"Avoid magic strings by centralizing constants at the source of truth (INV-33)."*
   https://github.com/threahq/threa/blob/572e666fe5feaac0514825fc0e89643e60a991e3/apps/backend/src/features/agents/enclave-dispatcher.ts#L174-L184

3. `apps/enclave/src/config.ts:31` — `allowedPersonaIds: ["persona_system_ariadne"]` hardcodes the same string that backend defines as `ARIADNE_AGENT_ID` in `built-in-agents.ts:6`. Both apps already depend on `@threa/types`; lifting the constant there (alongside the existing `ARIADNE_PERSONA_SLUG`) would eliminate the silent-drift risk on rename.
   https://github.com/threahq/threa/blob/572e666fe5feaac0514825fc0e89643e60a991e3/apps/enclave/src/config.ts#L29-L33

4. `apps/backend/src/features/agents/enclave-dispatcher.ts:15` + `apps/enclave/src/orchestrator.ts:29` — `HISTORY_LIMIT = 50` and `MAX_HISTORY_MESSAGES = 50` are the same semantic value (how many history rows the backend sends / the enclave consumes) defined independently in two apps. The orchestrator comment explicitly says *"Mirrors the backend's HISTORY_LIMIT"* — acknowledging the duplication. A single exported constant in `@threa/types` would prevent silent drift on the wire shape.
   https://github.com/threahq/threa/blob/572e666fe5feaac0514825fc0e89643e60a991e3/apps/enclave/src/orchestrator.ts#L25-L32

---

All items specified for 5a.2 (enclave service, EIK registry, invite UI, `tools: []`) are present. 5a.3 (encrypted trace steps) and 5a.4 (`web_search` + `read_url`) correctly remain out of scope. Multi-recipient envelope, per-message live-EIK fan-out, no pinned `keyId` on streams — all match the plan.

</details>

No high-confidence bugs surfaced. Spot-checks that came back clean: bearer auth uses `timingSafeEqual` with length check; pool is not held across the enclave HTTP call (INV-30/41); `pickInstance` has a length-guard so `Math.random() * 0` can't fire; UIK is always first in the recipient list so the user is never locked out; `contentMarkdown` on the persona reply row is always the placeholder; AAD parts (streamId + senderId + replyMessageId) align between dispatcher and enclave-side `buildMessageAad`; the empty-history filter still admits the triggering message because it was committed before the outbox event fires.

</details>
<details><summary>📋 CLAUDE.md Compliance [2 violations]</summary>

Both INV-33 (magic strings) — see findings 1 and 2 above.

</details>

Both config-sprawl — see findings 3 and 4 above. No leaky abstractions, no parallel implementations of existing helpers, no N+1, no DI hygiene issues. Feature folder shape for `enclave-runtimes/` matches the established pattern; barrel exports follow INV-51/52. Orchestrator deliberately does not reuse `PersonaAgent` per the plan's "no shared base class" decision (INV-29).

</details>

No high-confidence vulnerabilities. Enclave `/invoke` bearer is constant-time; `GET /enclave/active-keys` is workspace-scoped through the existing middleware; the multi-recipient encryption fans `[UIK, ...liveEIKs]` so the user can always decrypt; enclave never writes decrypted text into the row's `contentMarkdown`. The HPKE KEM swap (`@hpke/dhkem-x25519`) is the noble-curves binding — same primitive, different implementation.

</details>

kristofferremback:
### Security review

**Reviewers:** sonnet x3 (crypto / E2E, trust boundaries, data exposure) — independent fan-out, findings cross-checked.

**Verdict:** 1 HIGH (fixed in `77a7a500`) + 5 MEDIUM hardenings (3 fixed in `77a7a500`, 2 deferred to a follow-up) + 1 INFO (operational doc note) + 1 known design limitation (matches the 5a plan).

---

#### HIGH — fixed in `77a7a500`

**SSRF / bearer-leak via unvalidated `instanceUrl` in `register-key`** — `apps/backend/src/features/enclave-runtimes/handlers.ts:11`

Independently flagged by two reviewers. `z.string().url()` accepted any scheme/host, so a caller with `INTERNAL_API_KEY` could register `instanceUrl: "http://attacker.example.com"` (or AWS IMDS at `http://169.254.169.254`). On the next dispatch, the backend POSTs `Authorization: Bearer <INTERNAL_API_KEY>` plus the sealed envelope to that host. Bearer-harvest is the real damage because the same secret authenticates control-plane↔backend, so SSRF here is lateral.

**Fix shipped:** scheme allow-list (http/https only), explicit block on `169.254.169.254` / `metadata.google.internal` / `169.254.170.2` (ECS), and an optional `ENCLAVE_INSTANCE_URL_ALLOWED_PREFIXES` env var that production should set to the enclave service's hostname prefix. https://github.com/threahq/threa/blob/77a7a500/apps/backend/src/features/enclave-runtimes/handlers.ts#L7-L41

---

#### MEDIUM — fixed in `77a7a500`

1. **Verbose 500 response body** — `apps/enclave/src/index.ts:51` — flagged by all three reviewers. Catch-all returned `err.message` to the caller, which `EnclaveClient.invoke` then logged into the backend's structured log stream. Current throw sites are benign, but the surface grows with 5a.3/5a.4. **Fix:** opaque `"Internal Server Error"` on the wire; full `err` stays in `logger.error({ err })`. https://github.com/threahq/threa/blob/77a7a500/apps/enclave/src/index.ts#L49-L52

2. **Unbounded `history[]` / `recipients[]` arrays in `/invoke`** — `apps/enclave/src/invoke-handler.ts:43-52`. Only gate was `express.json({ limit: "4mb" })`; a caller could send hundreds of entries that each pass schema validation and force the orchestrator to HPKE-decrypt-loop through them. The slice-to-50 in the orchestrator happens *after* the full parse. **Fix:** `.max(ENCLAVE_HISTORY_LIMIT)` on history, `.max(32)` on recipients. https://github.com/threahq/threa/blob/77a7a500/apps/enclave/src/invoke-handler.ts#L49-L52

---

#### MEDIUM — deferred to a follow-up PR

3. **`registerKey` `ON CONFLICT DO UPDATE` resurrects a revoked row by clearing `revoked_at`** — `apps/backend/src/features/enclave-runtimes/repository.ts:63-72`. An attacker with both `INTERNAL_API_KEY` and a known revoked `keyId` (observable in any persisted envelope) can revive the row. Suggested fix: guard the upsert with `WHERE revoked_at IS NULL`, or fail loudly + log when the upsert clears it.

4. **`persona.model` accepted as arbitrary string** — `apps/enclave/src/invoke-handler.ts:47`, `apps/enclave/src/orchestrator.ts:137`. Only exploitable by direct `INTERNAL_API_KEY` holders; backend always sends a server-side constant. Suggested fix: derive `EnclaveConfig.allowedModels` from `allowedPersonaIds` (or hardcode `[ARIADNE_MODEL]` for now) and reject anything else in the orchestrator.

5. **`/revoke` accepts any `keyId` from any caller** — `apps/backend/src/features/enclave-runtimes/handlers.ts:87-93`. Any holder of `INTERNAL_API_KEY` (i.e., any enclave instance) can revoke every peer. Heartbeat-driven re-registration recovers in ≤30s, so the impact is bounded availability. Suggested fix: bind revocation to the caller's own `instance_id` (mirror the body shape that heartbeat already uses) or move revoke behind a separate operator credential.

---

#### INFO — known design limitation

**A compromised backend can spoof Ariadne replies.** The dispatcher controls `aadParts.senderId` and `replyMessageId`, so a hostile operator can pre-encrypt forged "Ariadne said X" to the user's UIK with a valid AAD before calling `EventService.createMessage`. The client cannot distinguish.

This is the explicit Phase 5a trust trade-off — *"operator runs the published binary"* per the plan's "trust shifts from cryptographic verification to operational trust" section. Closed properly in a future phase via Nitro / Confidential-VM attestation with client-side verification. Worth a `SECURITY.md` line; no code change in 5a.2.

---

#### INFO — operational

**`GET /pubkey` is unauthenticated.** Same data is served via the workspace-scoped `/enclave/active-keys`, so the only risk is that an exposed enclave port lets an unauthenticated probe enumerate EIK fingerprints. Document in `apps/enclave/README.md` that the listen port must be ingress-restricted to the backend's IP range.

---

🤖 Generated with security-review automation

kristofferremback:
We're rebuilding this separately

PR #663, opened 2026-05-28
fix(timeline): re-anchor firstItemIndex on window slides to stop cold-load jump

kristofferremback:
## Summary

Fixes the regression where the timeline jumps ~half a screen a frame after load, **only on the first visit to a long stream** (not when navigating back).

On a cold first visit the list mounts off stale IndexedDB data, then the bootstrap response **slides the window forward** — leading rows drop as the floor ratchets up, newer rows append, so the item count is roughly unchanged. The post-#582 `firstItemIndex` compensation used a count-growth heuristic that did nothing in that case, leaving `firstItemIndex` stale so every surviving row shifted virtual index and Virtuoso yanked the viewport. Navigating back had bootstrap cached at mount, so no slide and no jump — which is exactly the reported asymmetry.

## Changes (`apps/frontend/src/hooks/use-virtuoso-scroll.ts`)

- **Anchor-based delta instead of count.** Find the first row that survived from the previous render and shift `firstItemIndex` by the negation of its index delta. This handles prepend, leading removal, and a forward-sliding window uniformly (live bottom-append yields delta 0, so it's a no-op there).
- **Synchronous baseline reset on stream change.** Moved the ref-backed virtual-index reset out of the `useLayoutEffect` (which ran *after* the render that already recorded the new stream's baseline, clobbering it and skipping the first slide) into a render-time `resetKeyChanged` guard. The layout effect now only catches up the React state that drives re-renders.

This restores the anchor logic from #579 for this hook only — it does **not** reintroduce #579's cold-boot blank-gap behavior (the #582 fix in `stream-content.tsx`/`event-list.tsx` is untouched), nor its physical-distance jump detection or scroll listener.

Bottom-landing is preserved; unread handling is intentionally left alone (to be revisited separately).

## Test plan

- [x] `bunx vitest run src/hooks/use-virtuoso-scroll.test.tsx` — added 2 tests: forward window-slide (`firstItemIndex` +10) and prepend guard (−10)
- [x] `use-unread-divider` + `use-events` (25) and timeline components (219) green
- [x] frontend typecheck + lint clean (also ran via pre-commit hook across the monorepo)
- [ ] Manual browser check on a long stream's first cold visit (not runnable in the remote container — needs a live backend + primed IndexedDB)

---
_Generated by [Claude Code](https://claude.ai/code/session_019ch9gUWKsmBqomdTEUiPYN)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/663"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782581107&installation_id=129946197&pr_number=663&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F663&signature=349ad5280a370c0ce9a3f73e1b562d9f36e5139b3fd7ebd376f81158031c4533"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/hooks/use-virtuoso-scroll.test.tsx
+++ b/apps/frontend/src/hooks/use-virtuoso-scroll.test.tsx
@@ -51,6 +51,31 @@ function makeScrollableDiv(initial: { clientHeight: number; scrollTop?: number }
 
 type HookApi = ReturnType<typeof useVirtuosoScroll>
 
+type ScrollOptions = Parameters<typeof useVirtuosoScroll>[0]
+
+function renderScrollHook(initialOptions: ScrollOptions): {
+  current: HookApi
+  rerender: (options: ScrollOptions) => void
+} {
+  const ref: { current: HookApi | undefined } = { current: undefined }
+  function Probe({ options }: { options: ScrollOptions }) {
+    ref.current = useVirtuosoScroll(options)
+    return null
+  }
+  const utils = render(<Probe options={initialOptions} />)
+  return {
+    get current(): HookApi {
+      if (!ref.current) throw new Error("Probe did not capture the hook return value")
+      return ref.current
+    },
+    rerender: (options: ScrollOptions) => act(() => utils.rerender(<Probe options={options} />)),
+  }
+}
+
+function makeKeys(prefixes: number[]): string[] {
+  return prefixes.map((n) => `e${n}`)
+}
+
 function renderHookWithScroller(
   options: Parameters<typeof useVirtuosoScroll>[0],
   scrollerEl: HTMLDivElement,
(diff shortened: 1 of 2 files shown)


kristofferremback:
@coderabbitai review

PR #665, opened 2026-05-28
feat(e2e): SSK foundation — per-stream key storage + v2 message path

kristofferremback:
## What & why

Migrates E2E scratchpad messages off the per-message recipient **fan-out (v1)** onto a **per-stream symmetric key (SSK)**, the unifying message-crypto path. New messages seal once under the stream's SSK as envelope **v2**; v1 fan-out stays **read-compatible** so nothing already written breaks.

Why move off fan-out: v1 re-runs one HPKE seal *per recipient* and bakes the recipient set into every message, so adding/removing an actor would mean rewriting history. The SSK decouples the message payload from the recipient set — one AEAD seal per message, and recipient-set changes roll a key *generation* forward without re-encrypting history. This is the foundation rotation and multi-actor invites build on. For this PR the generation is always `0` (owner-only; rotation is a later phase).

## Storage

- New `stream_e2e_key_wraps` table — HPKE-wrapped SSK, one row per `(stream, key_generation, recipient)`. The server only ever holds the SSK wrapped to a recipient's long-term public key (**INV-E7**), never in plaintext. Unique slot index backs the race-safe upsert.
- New `e2e_streams.current_key_generation` column — the generation a new send seals under.
- `StreamE2eKeyWrapsRepository`: set-based UNNEST `insertMany`, idempotent per slot (INV-20/56), and `listForStream`. BYTEA crosses the boundary as base64.

## Message path

- Envelope **v2** (`keyGeneration` + `iv` + `aad`); the schema accepts v2 and v1 as a structurally-disjoint union, so it discriminates without a version literal. Send seals under the current generation.
- **Two-phase E2E create.** The wrap AAD binds to the server-minted `streamId`, which the client doesn't know until create returns — so it can't wrap inline. The client creates the stream (E2E flag set atomically), learns the id, wraps the SSK, then POSTs it to `…/e2e/key-wraps`. `storeOwnerKeyWrap` validates the caller owns the stream. The brief no-wrap window is benign: the stream is empty and the send path refuses until the SSK resolves; the store is idempotent (ON CONFLICT DO NOTHING).
- Frontend in-memory **SSK cache**: fetch-and-unwrap, de-duplicated per slot, dropped on lock/account-switch (same boundary as the decrypt cache). Create seeds it with the freshly minted SSK to skip an immediate round-trip.

## New invariants

- **INV-E7** — server never sees an SSK or any private key; wraps are forwarded verbatim; key-wrap writes are race-safe.
- **INV-E8** — a v2 message's `keyGeneration` must be `<= current_key_generation`; read any historical generation, write only the current one.

## Testing

- Backend: `StreamE2eKeyWrapsRepository` unit tests (UNNEST batch insert / idempotency / scoped read), mocked-Querier.
- Frontend: v2 SSK loopback round-trip, "no wrap → null", tampered-ciphertext → null, v1 fan-out read-compat, unknown-version → null; SSK cache fetch/dedup/clear behaviour.
- Full-workspace `typecheck` clean. DB-dependent paths run in CI (no local migrate).

Plan doc (`docs/plans/e2e-encrypted-scratchpads.md`) synced with the SSK design, storage, INV-E7/E8, and a Phase 1.5 delivery note.

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/665"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782589398&installation_id=129946197&pr_number=665&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F665&signature=3d2e60a8c609a1d20700156e4af46a84bb105e1ef3984e1e2ba79255fe2d04f5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review
**Confidence: 5/7** — Good (one real robustness gap, otherwise clean)
**Review models:** Orchestrator: Opus 4.8 | Reviewers: sonnet x3

Found 1 issue:

1. `apps/frontend/src/hooks/use-create-encrypted-scratchpad.ts:46-56` — **Orphaned, permanently-unwritable scratchpad if the wrap store fails.** The stream is created server-side first (`streamService.create`, L34), but the SSK is generated as a local (`ssk`, L46) and only persisted by `e2eKeyWrapsApi.store` (L52). If that POST throws (network blip / 5xx / 4xx), the SSK is lost, `putStreamKey` (L56) never runs, and the local cache writes (L73-101) are skipped — yet the E2E stream already exists on the server with no `generation 0` wrap. On next bootstrap it reappears in the sidebar, `resolveCurrentStreamKey` returns `null`, and every send throws "You don't have access to this encrypted scratchpad's key" with no recovery path (the SSK can't be regenerated to match the AAD-bound stream id). Consider wrapping create+store so a failed store tears down / hides the half-provisioned stream, or making the owner wrap retriable from the locked-state UI.
   https://github.com/threahq/threa/blob/36a452ceb844ff88e6300fa7a1e10350eed47d96/apps/frontend/src/hooks/use-create-encrypted-scratchpad.ts#L46-L56

---

Matches "Phase 1.5 — SSK foundation" in `docs/plans/e2e-encrypted-scratchpads.md`: `stream_e2e_key_wraps` table + `current_key_generation` column, set-based repo, v2 envelope as a disjoint union with v1 read-compat, two-phase create, and the in-memory SSK cache are all present with corresponding frontend/backend wiring. No missing corresponding changes found.
</details>
<details><summary>🔍 Bugs — 1 issue</summary>

See finding #1 above (store-failure orphan).
</details>

No INV violations introduced. `enabled_at` and `current_key_generation` both `DEFAULT` in the migration so the omission in `markStreamE2e` is fine; repo uses set-based UNNEST (INV-56) with `ON CONFLICT DO NOTHING` (INV-20); endpoints are workspace-scoped (INV-8); barrel imports respected (INV-52).
</details>
<details><summary>🏗️ Design — 1 minor (non-blocking)</summary>

`apps/frontend/src/lib/crypto/stream-key-cache.ts:93-104` — `resolveCurrentStreamKey` doesn't use the `inflight` dedup map that `resolveStreamKey` uses, so two concurrent first-sends on an unseeded stream can issue two parallel `GET /e2e/key-wraps` calls. Consequence is benign (a redundant idempotent read; both unwrap the same SSK and cache it), and the create path seeds the cache so the window is narrow — hence non-blocking — but reusing `inflight` keyed on the stream slot would make it consistent with its sibling.

(Considered and dropped: `storeOwnerKeyWrap` using `withClient` rather than `withTransaction` — `e2e_streams.owner_user_id` is immutable with no mutation path, and the insert is idempotent per slot, so the guard-then-write is not racy today.)
</details>

Key-wrap endpoints are protected at both the membership (`validateStreamAccess`) and owner (`e2e.ownerUserId !== userId`) levels, with `userId`/`workspaceId` taken from the auth session rather than the request body. The server only ever stores/forwards HPKE-wrapped bytes — no plaintext SSK or private key crosses the boundary (INV-E7). No injection / auth-bypass / data-exposure paths found.
</details>

---

PR #671, opened 2026-05-29
feat(e2e): Pi-remote key plumbing — BIK registration, SSK key roll on invite, pinned actor identity

kristofferremback:
## What this is

Phase 2 (Pi remote E2E) groundwork for encrypted scratchpads. It lets an external runtime register an encryption key, rolls the per-stream symmetric key (SSK) forward when an actor is invited, and pins each invited actor to a concrete identity. The server never sees the SSK or any private key — only opaque HPKE wraps.

Three stacked commits, each self-contained:

### 1. Register Bot Identity Key (BIK) on runtime presence (`66400dd`)
- Migration adds nullable `public_key` + `public_key_id` to `bot_runtime_instances` (opaque base64 TEXT — server never inspects it, so the hot `SELECT *`/`RETURNING *` presence reads need no BYTEA encode/decode).
- `upsertPresence` writes the BIK columns with `COALESCE(EXCLUDED, existing)` on conflict (INV-20), so a presence upsert that omits the key (invocation touch, session-link path) can't clobber a key an earlier session registered, while a session start that carries one rotates it.
- The WS `bot:hello` and HTTP presence schemas share one validated key fragment (32-byte X25519, base64) hoisted into `lib/schemas` (a zero-feature-dep leaf) so the two registration paths can't drift (INV-31) and no value crosses the cyclic public-api ↔ bot-runtimes import edge.

### 2. Roll the stream key and wrap it to invited actors (`5fda3d3`)
- `bumpKeyGeneration`: race-safe roll guarded on the prior generation (INV-20).
- `inviteActor` resolves the live actor recipient set and returns it as a `keyRoll`; `rollStreamKey` + `POST …/e2e/key-generations` stores the client's wrap batch **and** bumps `current_key_generation` in one transaction (INV-7) so a send never seals under a generation with no wraps. Owner-only; validates generation == current+1 and that the owner's own wrap is present (no self-lockout).
- Frontend `rekeyStream` mints a fresh SSK, wraps it to the owner's UIK plus every actor recipient (each AAD-bound to stream + generation + recipient), POSTs the batch, and seeds the cache.

### 3. Pin invited actors to a concrete identity; allow multiple bots (`5cc4640`)
- Addresses a security smell: previously the bot was *inferred* from whoever was the active actor, so the underlying bot could be swapped as long as some key was wrapped to it.
- Migration adds `e2e_stream_actors.actor_id`; PK becomes `(workspace_id, stream_id, kind, actor_id)`. Enclave rows use a fixed sentinel (one Ariadne); bot rows carry the real `bot_id`. Legacy rows backfill `actor_id = kind`.
- `inviteActor` now pins a validated principal (bot must exist + be non-archived; the enclave uses its sentinel) and wraps to *that* bot's live BIKs directly — the active-actor inference is gone. A scratchpad can now hold multiple bots; a bot's key may rotate under its id, but the underlying bot can't be silently swapped.

## Deploy safety

Self-contained and backward-compatible:
- Blast radius is the E2E-scratchpad feature. The one shared touch (`bot_runtime_instances.upsertPresence`) is additive + `COALESCE` — a no-op for every current caller. Non-E2E chat/bots/scratchpads are unchanged.
- The new invite→roll machinery is effectively inert in prod until a live BIK (Pi adapter, not yet shipped) or EIK (enclave, not yet shipped) exists: with no live actor key, `resolveActorRecipients` returns empty → `keyRoll: null` → invite behaves exactly as before.
- Migrations are additive (one PK change on the small, new `e2e_stream_actors` table). They are verified in CI, not locally (read-only prod DB).

### Known follow-ups / caveats
- **Rolling-deploy window:** the `e2e_stream_actors` PK change means old backend pods doing an actor-invite during the deploy window would hit an `ON CONFLICT (workspace_id, stream_id, kind)` mismatch. Affects only E2E actor-invites in that window (a brand-new, ~zero-traffic path; only the enclave button is wired today).
- **No repair UI yet** if the roll POST fails after the invite commits (the actor is recorded but unkeyed; the stream still works at the old generation for the owner). Same shape as the create-flow repair affordance — a re-key trigger is future work.
- **Bot-picker UI is the next slice:** choosing which bot to invite, listing/removing multiple bots, and making `canInviteActor`/`isActorInvited` bot-id-aware. This PR lands the data model, identity pinning, and resolution.
- **Identity is a server-table mapping**, not yet cryptographically verified: pinning `bot_id` makes the bot a named principal and makes removal meaningful, but full E2E identity needs out-of-band key-fingerprint verification (the same thing the UIK story needs in Phase 4). This PR is the foundation for that.

## Tests
- Backend: BIK presence round-trip + COALESCE-preserve; hello/HTTP schema accept/reject (half-key, wrong length); `bumpKeyGeneration` guard; `rollStreamKey` (atomic store+bump, owner-only, stale-generation, owner-wrap-missing, lost-race); `inviteActor` recipient resolution (enclave EIKs, bot-by-id, multiple bots, bot-not-found, missing-bot-id); actor repo with `actor_id`.
- Frontend: `rekeyStream` proves both owner and actor wraps decrypt to the same fresh SSK under their own AAD, plus cache seeding.
- Full frontend suite green (1921); backend feature suites green. DB-integration runs in CI.

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/671"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782629899&installation_id=129946197&pr_number=671&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F671&signature=75215da6ae3918cc9398340acd1e8de280a0b0bf0896a4ead5e88c917a5e1c41"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
Thanks @coderabbitai — both findings were valid and are fixed in `89bb734`:

1. **BIK retention** — agreed this was a real correctness hole (a reconnecting runtime reusing `instanceId` could be handed SSK wraps for a key it rotated away). `upsertPresence` now **overwrites** the BIK by default, so a session that registers no key clears any stale one. Retention is opt-in via `retainBik` for the only two server-internal writes that legitimately don't carry the key: the invocation-side presence touch and the session-link path.

2. **`publicKey` schema** — tightened the Zod source (it generates the OpenAPI) to an exact 32-byte base64: `length 44` + `^[A-Za-z0-9+/]{43}=$`, replacing `z.base64()` + a non-serializing refine. The published contract now matches runtime validation (no empty/any-length key). Spec regenerated.

One partial: the `publicKey`/`publicKeyId` all-or-nothing pairing stays enforced at runtime (a shared object-level `.refine`) but isn't expressed in the generated JSON schema — our zod→OpenAPI generator doesn't serialize object refinements, and I didn't want to restructure the wire shape (nest under a `bik` object) just for the doc. Open to that later if the doc-level guarantee matters.

---

PR #673, opened 2026-05-29
fix(link-previews): capture metadata streamed into body after </head>

kristofferremback:
Streaming-SSR frameworks (Next.js App Router and friends) emit a
metadata-less <head> and then stream the real <title>/<meta og:*> tags
into the <body> after </head> closes. The generic metadata fetcher
stopped reading at the first </head>, so it never saw those tags and
fell back to a URL-slug title with no description or image — e.g. an
anthropic.com/news link unfurled as just "claude opus 4 8".

Stop reading at </head> only after a metadata signal has been observed;
pages that never provide one stay bounded by MAX_HTML_BYTES. The signal
is scanned per-chunk rather than on the trimmed tail window so it isn't
lost when a large chunk overflows the window.

https://claude.ai/code/session_01DSvSfn7xgx4UX6p7ZQ5hYE

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/673"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782632047&installation_id=129946197&pr_number=673&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F673&signature=f488797d66a12557edf60638e67a141eea54f411c838d8b8a2bba578b6c2bb5b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/link-previews/worker.test.ts
+++ b/apps/backend/src/features/link-previews/worker.test.ts
@@ -395,6 +395,83 @@ describe("createLinkPreviewWorker", () => {
     )
   })
 
+  test('ignores body attributes like name="descriptionField" and still captures real metadata', async () => {
+    // A body attribute whose value merely starts with "description" must not flip the early-stop
+    // gate: if it did, the loop would stop at </head> and drop the real og:* tags streamed after it.
+    const headChunk =
+      `<html><head><meta name="theme-color" content="#000"/></head>` +
+      `<body><form><input name="descriptionField" value="not metadata"/></form>`
+    const metaChunk =
+      `<title>Real Page Title</title>` +
+      `<meta property="og:title" content="Real Page Title">` +
+      `<meta property="og:description" content="The genuine description.">` +
+      `</body></html>`
+
+    const fetchMock = mock(async () => streamingHtmlResponse([headChunk, metaChunk]))
+    globalThis.fetch = fetchMock as unknown as typeof fetch
+
+    const completePreviewsAndPublish = mock(async () => {})
+    const worker = createLinkPreviewWorker({
+      linkPreviewService: {
+        extractAndCreatePending: mock(async () => [{ id: "lp_901", url: "https://example.com/article" }]),
+        getPreviewById: mock(async () => ({
+          id: "lp_901",
+          workspaceId: "ws_123",
+          url: "https://example.com/article",
+          normalizedUrl: "https://example.com/article",
+          title: null,
+          description: null,
+          imageUrl: null,
+          faviconUrl: null,
+          siteName: null,
+          contentType: "website",
+          status: "pending",
+          previewType: null,
+          previewData: null,
+          targetWorkspaceId: null,
+          targetStreamId: null,
+          targetMessageId: null,
+          fetchedAt: null,
+          expiresAt: null,
+          createdAt: new Date("2026-05-29T10:00:00.000Z"),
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Code review (self-review)
**Confidence: 6/7** — Very Good (one issue found and fixed in this pass)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet ×3

Ran a 3-perspective review (spec/CLAUDE.md, correctness/security, design). Three candidate findings were traced; two dissolved on inspection, one was a real foot-gun and is now fixed in `ed7cb76`.

**Fixed — premature early-stop via over-matched signal**
The metadata-signal gate matched `name=…description` without anchoring the value's end, so body attributes like `name="descriptionField"` or `data-name="description-toggle"` could flip `seenMeta` early. While `</head>` was still in the rolling window, that re-triggered the exact early-stop bug this PR fixes — dropping `og:*` tags streamed after `</head>`. Anchored the match (`name=["']?description["'\s>]`), added `twitter:image` for symmetry with `og:image`, and added a regression test.

<details><summary>🔍 Bugs [1 fixed]</summary>

- `worker.ts` — `name="description"` over-match → premature `seenMeta`. **Fixed** in `ed7cb76` + regression test (`ignores body attributes like name="descriptionField"…`).
</details>

<details><summary>🏗️ Design [accepted-by-design]</summary>

- *`META_SIGNAL` token list could drift from `parseHtmlMeta`'s keys.* Accepted: the gate is a **best-effort early-stop optimization, not a correctness gate** — `parseHtmlMeta` always runs on the full accumulated buffer, so a missed/stale signal only costs extra bytes (bounded by `MAX_HTML_BYTES`), never lost metadata. Added a comment documenting this. A shared-derivation refactor would be over-abstraction for a loose heuristic (INV-36).
</details>

No violations. `META_SIGNAL` is an HTML structural-token match, not a language/semantic decision (INV-54 N/A). Tests assert event content, not counts (INV-23/24).
</details>

SSRF remains handled by `isBlockedUrl` (pre/post-redirect); no new vectors. Reading more bytes is bounded by `MAX_HTML_BYTES`; DoS out of scope.
</details>

<details><summary>Dismissed after tracing</summary>

- *"Test doesn't exercise chunk boundaries"* — false: `ReadableStream` delivers one chunk per `read()` (verified), and the pre-fix code genuinely failed the streaming test, proving chunks are processed separately.
- *"Split-boundary signal could be missed silently"* — not a bug: when `seenMeta` never fires, the loop reads **more** (bounded), and `parseHtmlMeta` runs on the full buffer, so metadata is never lost — only the early-stop optimization is skipped.
</details>

---

PR #685, opened 2026-05-29
feat(e2e): enclave /invoke — Ariadne decrypts, calls the LLM, seals the reply

kristofferremback:
## What this is

Phase 5 (Ariadne enclave) groundwork. The enclave can now take an end-to-end-encrypted scratchpad turn, decrypt it **in isolation**, call the LLM, and return an encrypted reply — Threa's main backend and DB never see plaintext. This PR is the **enclave side + the backend-side plumbing to call it**; the dispatch/reply-write wiring (what makes a live user turn flow through it) is the immediate follow-up.

For our scope this is the **Railway / operational-isolation** model (per the product call), not a hardware TEE: trust is "the operator runs the published binary, which holds its key in memory, never persists/logs plaintext, and only egresses to the LLM provider." Hardware attestation (Nitro/PCR) stays out of scope.

### `43e6ea9` — enclave `/invoke` + Node runtime
- **`/invoke`**: unwrap the forwarded SSK with the in-memory EIK → open the prompt + decryptable history → call OpenRouter → seal the reply back under the SSK. Pure `handleInvoke` core + an Express adapter that answers status-only on failure (error bodies could echo content). Plaintext lives only in-process, for the request.
- **OpenRouter client** pinned to zero-retention providers (`provider.data_collection: "deny"`), injectable so the handler is loopback-tested with no network. Intentionally **not** the backend `createAI` wrapper (documented INV-28 exemption: the enclave must emit no telemetry carrying content and keep egress to one host).
- **Runtime: Node, not Bun.** The enclave must do X25519 HPKE *decap* at runtime, and Bun 1.3.x's WebCrypto lacks X25519 `deriveBits` (it has `generateKey` but not the ECDH primitive) — this would fail in production, not just tests. Bun stays the *bundler* (`bun build --target=node` → one self-contained ESM file); the Docker runtime stage is plain `node:22-slim`. Tests move to **vitest** so HPKE runs on Node. The shared `@threa/crypto` package is untouched, so the live frontend E2E path is unaffected.

### `47f0383` — shared contract + forwarder + capability flag
- **`@threa/types`**: the canonical `/invoke` contract (`EnclaveInvokeRequest`/`Response`, sealed-message + SSK-wrap shapes). The enclave validates against it with zod; the backend forwarder builds it — one definition, no drift.
- **`EnclaveForwarder`**: backend → enclave `POST /invoke` client, authenticated with the internal-api-key header, generous timeout for LLM round-trips. Errors carry status only — never the request.
- **Ariadne `e2eCapable: true`** on its built-in config (Empty: false) + `isE2eCapablePersona` gate. Config-only — the enclave runs the built-in Ariadne, so no persona-table migration.

### `03528e9` — authenticate `/invoke` (self-review fix)
The forwarder sent the internal-api-key but the enclave wasn't checking it. The EIK pubkey is public, so without this gate anyone reaching the enclave could wrap their own SSK to it and use `/invoke` as an LLM/decryption oracle. Now only a caller presenting the shared secret is accepted (constant-time compare); `/pubkey`, `/healthz`, `/attestation` stay public.

## Tests
- Enclave (vitest, Node): loopback proof — wrap an SSK to a generated EIK, seal a prompt, run `handleInvoke` with a stub LLM, confirm the reply ciphertext opens back under the SSK; plus history roles, skipping generations with no wrap, the no-wrap error, and the `/invoke` auth rejection. Registration tests ported to vitest.
- Backend (bun): `EnclaveForwarder` (URL/header/body, non-ok status, network error) and `isE2eCapablePersona`.
- Full monorepo typecheck/lint/dockerfile-check/OpenAPI green.

## Deploy safety
Self-contained: the enclave is a separate service, and `@threa/crypto` is unchanged, so nothing here alters existing backend/frontend behavior. The `EnclaveForwarder` and `e2eCapable` flag have no caller yet (the dispatch follow-up wires them), so they're inert in prod until then.

## Follow-up (not in this PR)
The dispatch vertical: a `message:created` outbox handler that, on an E2E stream where the enclave actor is invited + Ariadne is e2e-capable, enqueues an enclave-invoke job; a worker that builds the request (resolve live EIK + wraps + encrypted history), forwards, and writes the sealed reply back as an Ariadne-authored E2E message via the existing `EventService`. That's the piece that makes a real user turn reach Ariadne.

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/685"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782643516&installation_id=129946197&pr_number=685&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F685&signature=4fe9d758b6d1a0c91c4116604532e3a7d5a140942156460ab3ab30ff163fa064"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: Opus 4.8 | Reviewers: sonnet x3

Solid groundwork. Crypto correctness, the `/invoke` auth gate (the decryption-oracle fix in `03528e9`), and plaintext-leakage paths all checked out clean. Found 3 non-blocking issues, all forward-looking/maintainability:

1. `packages/types/src/api.ts:445-450` — `EnclaveInvokeResponse` is a single-sealed-message shape (`{ ciphertext, envelope, model, usage? }`), but the plan's HR#2 specifies the enclave→backend response as `{ messages: [...], sidecar }` to support multi-message output. Since this `@threa/types` type is the canonical contract, the dispatch follow-up will need a breaking change here. Worth either aligning the shape now or noting it as provisional. (plan: `e2e-enclave-ssk.md` §3 HR#2)
   https://github.com/threahq/threa/blob/03528e9a0c2cdec43c3d9d7ea4dccdb5ee5c81d9/packages/types/src/api.ts#L445-L450

2. `packages/types/src/api.ts:405-410` — `EnclaveStreamEnvelope` is a field-for-field duplicate of `StreamEnvelope` in `@threa/crypto` (`{ v, keyGeneration, iv, aad }`), and `invoke.ts` passes `EnclaveStreamEnvelope` instances into `openMessageAsString`/`sealMessage` (which expect `StreamEnvelope`) under structural typing with no bridging. If `StreamEnvelope` gains a required field the two silently diverge and the error surfaces only in `invoke.ts`. Consider re-exporting the crypto type as the single source of truth. (design)
   https://github.com/threahq/threa/blob/03528e9a0c2cdec43c3d9d7ea4dccdb5ee5c81d9/packages/types/src/api.ts#L405-L410

3. `apps/enclave/src/invoke.ts:54-57` — the history-window limit `200` is a bare magic number inline in the Zod schema. The plan §7 explicitly lists "INV-33 (centralize constants — generation start, staleness window, **history limit**)" as an invariant to respect for this stack, naming the history limit specifically. Suggest a named constant. (CLAUDE.md [INV-33, *Maintainability and Scope Control*]: "Avoid magic strings by centralizing constants at the source of truth")
   https://github.com/threahq/threa/blob/03528e9a0c2cdec43c3d9d7ea4dccdb5ee5c81d9/apps/enclave/src/invoke.ts#L54-L57

---
<details><summary>📐 Plan Adherence [1 issue]</summary>Response shape diverges from the plan's multi-message `{ messages, sidecar }` contract (#1). Scope deferral (no dispatcher/reply-write wiring) matches the PR body and the plan's PR-4 boundary — not flagged.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>HPKE unwrap AAD binding, per-call IV freshness, 32-byte SSK length check, AES-GCM tag binding via stored AAD, and the no-wrap error path all verified correct. History loop correctly skips generations with no wrap.</details>
<details><summary>📋 CLAUDE.md Compliance [1 violation]</summary>INV-33 history-limit magic number (#3). The INV-28 deviation (no `createAI` wrapper) is documented and justified (no content-bearing telemetry, single egress host) — not a violation.</details>
<details><summary>🏗️ Design [1 concern]</summary>Envelope type duplication across `@threa/types` and `@threa/crypto` (#2). Contract is otherwise a clean single-source: enclave validates with zod, backend forwarder builds against the same type.</details>
<details><summary>🔒 Security [CLEAN]</summary>`/invoke` auth precedes body parse and `handleInvoke` (no decryption oracle); missing header short-circuits to 401; `timingSafeEqual` after length check is the standard pattern. Error bodies are status/structural only — no plaintext echo. Access log emits `{id, method, url}` only. OpenRouter pinned to `data_collection: "deny"`.</details>

kristofferremback:
Addressed the review in `46b0327`:

- **#3 (history-limit magic number, INV-33)** — fixed. Extracted the `/invoke` schema bounds (history window, wrap count, system/model size, max tokens) into named constants.
- **#2 (envelope duplication)** — kept the duplicate but documented it as a deliberate mirror. Re-exporting `@threa/crypto`'s `StreamEnvelope` from `@threa/types` would pull `@hpke/core` into every `@threa/types` consumer (frontend, control-plane, …) to share four fields; instead, the structural handoff in `invoke.ts` fails to compile (CI-caught) if the crypto type gains a field, so it can't diverge silently.
- **#1 (multi-message `{messages, sidecar}` response shape)** — no change. The cited `e2e-enclave-ssk.md §3 HR#2` doesn't exist (the plan is `docs/plans/e2e-encrypted-scratchpads.md`), and there's no `sidecar`/multi-message shape anywhere in it — Flow C specifies *"Compose reply envelope"* (a single reply to `[UIK, EIK]`), which is exactly `EnclaveInvokeResponse`. A multi-message agent-loop output would be a deliberate future extension, not built speculatively now (INV-36). Flagging that this finding's plan citation was a hallucination.

Full typecheck/lint/dockerfile/OpenAPI green.

---

kristofferremback:
Addressed all of CodeRabbit's findings in `4af75a3`:

- **`/invoke` parsed the body before the auth gate** — fixed. Extracted `requireInternalKey` as a middleware mounted *before* a route-scoped `express.json` (and dropped the app-wide parser), so an unauthorized caller is rejected without us parsing their (up to 4 MB) JSON.
- **Forwarder trusted the 200 payload** — fixed. `EnclaveForwarder.invoke` now validates the response shape (ciphertext/model strings + envelope present) and throws `EnclaveForwardError` on a malformed 200; added a test.
- **OpenRouter fetch had no timeout** — fixed. Added `AbortSignal.timeout`, kept under the forwarder's timeout so the enclave aborts cleanly first.
- **Dockerfile ran as root** — fixed. Runtime stage now copies `--chown=node:node` and runs `USER node`.
- **`invoke.ts` cast hid drift** — fixed. Typed assignment (`const request: EnclaveInvokeRequest = parsed.data`) instead of `as`, so schema/contract drift is a compile error.
- **README** — `PORT` now lists `/invoke`; the access-log note corrected to reflect that the serializer emits only `{id, method, url}` (no headers logged, so the api key is never written; `redact` is belt-and-braces).

Enclave 9 tests + forwarder 4 tests green; full typecheck/lint/dockerfile/OpenAPI green.

---

PR #690, opened 2026-05-29
feat(e2e): dispatch a scratchpad turn to the enclave and write Ariadne's reply

kristofferremback:
## What this is

Phase 5, the dispatch vertical — the piece that makes **Ariadne actually reply in an E2E scratchpad**. Builds on the merged enclave `/invoke` (#685). A user turn in an E2E scratchpad where the enclave actor is invited now gets forwarded to the enclave, which decrypts in isolation, calls the LLM, and returns a sealed reply that's stored and fanned out — **Threa's backend and DB never see plaintext** (only ciphertext + the SSK wraps addressed to a live EIK cross the boundary).

## How it works
- **`EnclaveDispatchHandler`** — a `message:created` outbox handler, the mirror of `CompanionHandler` but *for* E2E streams (companion mode is forced off there): on a **user** turn in an E2E stream with the enclave actor invited, enqueue an `ENCLAVE_INVOKE` job. Persona-authored replies never re-trigger (user-only gate).
- **`buildEnclaveInvokeRequest`** (pure, unit-tested) — picks a live EIK that can decrypt the **current** generation (has a wrap for it), ships its SSK wraps + the prompt + decryptable history (roles mapped, chronological), strips the `openrouter:` model prefix, and binds the reply descriptor. Returns `null` when no live enclave can serve the stream (→ worker no-ops).
- **enclave-invoke worker** — load → build → `EnclaveForwarder.invoke` → write the sealed reply via the existing `EventService` as an Ariadne-authored E2E message (INV-E1 placeholder content), under the **server-minted id the reply's AAD was bound to**. Idempotent per trigger (`clientMessageId`).
- **Plumbing** — `JobQueues.ENCLAVE_INVOKE` + payload; `EventService.createMessage` gains an optional `id` (so the reply row id matches its sealed AAD); `server.ts` wires the forwarder + worker + handler, all **gated on a configured internal-api-key**.

## Tests
- `buildEnclaveInvokeRequest` unit-tested (EIK selection, current-generation matching, no-actor / no-live-enclave → null, history role mapping, model-prefix stripping) — this is where the logic risk lives.
- Backend typecheck clean (DI wiring boots), messaging 84 + enclave-runtimes 18 tests pass, full lint/dockerfile/OpenAPI green.
- The worker/handler **runtime** behavior is DB/outbox-coupled — validated by CI integration like the other outbox handlers/workers.

## Deploy safety
- The whole path is gated on `internalApiKey` being configured; absent it, the handler/worker aren't even registered (no behavior change).
- Reuses the existing E2E message-create path (no new write semantics) and only triggers on E2E streams with the enclave actor invited — non-E2E chat/companion flows are untouched.
- The `EventService.createMessage` `id` param is additive (`params.id ?? messageId()`); all existing callers are unaffected.

## Known limitations / follow-ups (by design, not blockers)
- **Single final reply, no progress/step traces yet.** The enclave does one LLM call and returns one sealed message. Encrypted trace-step streaming (seal each step under the SSK → POST back over the authenticated channel → encrypted `agent_session_steps` → existing trace UI decrypts) is the planned next slice — the storage, broadcast channel, crypto, and trace UI all already support it.
- **No turn coalescing.** Each user message dispatches one job (like the bot path); rapid messages → multiple concurrent forwards. Session semantics come with the in-enclave agent loop.
- **In-enclave agent loop** (run Ariadne "as close to normal as possible," with tools) is the larger follow-up; the `EnclaveInvokeRequest/Response` contract is the additive seam for multi-message + steps. The open design fork there: which of Ariadne's tools survive E2E (GAM memory needs the backend to see content — web/GitHub egress from the enclave is fine).

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/690"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782662552&installation_id=129946197&pr_number=690&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F690&signature=b8286a9cf5b19264191eda5a4c5341c6d6d953a0d6ec2ab3f65c5d5434c620d8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
Addressed both in `da4bf89`:

**Trigger-generation decryption (Major) — fixed.** Good catch. The chosen EIK now must have a wrap for **both** the current generation (to seal the reply) and `trigger.envelope.keyGeneration` (to open the prompt), so a stream that rotated between the user's turn and dispatch can't pick a key that seals but can't decrypt. Added a rotated-key regression test (gen-0 prompt under a gen-1 current: dispatch only succeeds when an EIK covers both).

**Invoke idempotency (Critical) — partially fixed now, full fix deferred with rationale.** Pushing back gently on the severity: the *message write* was already idempotent via `clientMessageId`, so **no duplicate reply ever reaches the user** — the only residual is wasted LLM spend if a retry lands in the narrow window after `invoke()` but before `createMessage()`. I added a cheap pre-check (`findByClientMessageId('enclave-reply:<id>')`) before invoking, so the **common** at-least-once case (job redelivered after a prior *full* success) now skips the enclave call entirely.

The remaining window (invoke succeeded but the reply never committed) can still re-invoke. The proper fix is durable invoke-tracking (a tracking table per INV-57, committed before the external call) — but that lifecycle belongs with the in-enclave **agent-loop/session** follow-up, which reworks this path and brings session semantics anyway. Building a standalone tracking table now would be subsumed by it. I'd rather not add that surface in this slice; flagged as the explicit follow-up.

Backend typecheck + enclave tests (19) + full lint/dockerfile/OpenAPI green.

---

PR #695, opened 2026-05-29
feat(agents): GitHub tools read public repos without an integration

kristofferremback:
## What

Ariadne — and any persona with GitHub tools enabled — can now read **public / open-source** GitHub repositories even when the workspace has **not** connected a GitHub App integration. Private repositories still require credentials.

GitHub's REST API serves public repos anonymously, so an unauthenticated Octokit client works for them; private resources answer `404`, which the tool layer already surfaces as `GITHUB_NOT_FOUND`.

## Why

Previously, the GitHub tools were registered for Ariadne but every call returned `GITHUB_NOT_CONNECTED` until a workspace admin connected the GitHub App. That blocked the common case of "look at this open-source repo" — which needs no credentials at all.

## How

- **`GitHubClient`** accepts `null` `record`/`credentials`/`metadata` and uses an unauthenticated `Octokit` in that mode. In anonymous mode it skips token refresh (no credentials to renew) and skips rate-limit persistence (anonymous limits are per-IP, not workspace state). `GitHubClient.anonymous()` is the named factory.
- **`getGithubClient`** gains an `allowUnauthenticatedFallback` option. The *no-usable-installation* paths (no app configured, no active record, undecryptable or unrefreshable credentials) return an anonymous client when the flag is set. The default (`false`) preserves the existing null-on-missing behavior, so **link-preview enrichment is unchanged**.
- The **near-rate-limit branch deliberately does *not* fall back**. It stays a back-off circuit-breaker (keeps returning `null` → `GITHUB_NOT_CONNECTED`) so a throttled installation isn't silently downgraded onto the shared per-IP anonymous quota, and the rate-limit state stays visible (INV-11: fail loudly).
- **Agent tools** opt in via `createMemoizedGithubClient`, so `GITHUB_NOT_CONNECTED` no longer blocks public-repo reads.

## Design decisions

- **Option flag rather than always-on fallback.** Link previews intentionally keep authenticated-only semantics; only the agent-tool path opts into anonymous reads.
- **Rate-limit branch is not a fallback path.** This was raised in self-review (two independent reviewers, INV-11) — handing back an anonymous client when an authenticated installation is throttled would mask the throttle and burn the shared anonymous IP quota.

## Tests

- `service.test.ts` (new): no-app path returns `null` without the flag and an anonymous `GitHubClient` with it; an exploding pool asserts the no-app path never touches the DB; `GitHubClient.anonymous()` builds without installation state.
- `client-accessor.test.ts`: asserts the agent-tool accessor opts into `{ allowUnauthenticatedFallback: true }` (regression guard so Ariadne doesn't silently lose public-repo access).
- All GitHub tool + link-preview tests pass (32/32). Full-monorepo lint + typecheck pass via pre-commit hooks.

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/695"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782679444&installation_id=129946197&pr_number=695&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F695&signature=ac240ad2e15e1f8e4ff99f98a4a4832867d487dfb3090361029a778bb5a53a7e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/tools/github/client-accessor.test.ts
+++ b/apps/backend/src/features/agents/tools/github/client-accessor.test.ts
@@ -6,6 +6,21 @@ function makeService(fn: () => Promise<GitHubClient | null>): WorkspaceIntegrati
   return { getGithubClient: fn } as unknown as WorkspaceIntegrationService
 }
 
+describe("createMemoizedGithubClient options", () => {
+  it("requests the unauthenticated fallback so agent tools can read public repos without an integration", async () => {
+    let receivedOptions: unknown
+    const service = {
+      getGithubClient: async (_workspaceId: string, options?: unknown) => {
+        receivedOptions = options
+        return null
+      },
+    } as unknown as WorkspaceIntegrationService
+    const getClient = createMemoizedGithubClient(service, "ws_1")
+    await getClient()
+    expect(receivedOptions).toEqual({ allowUnauthenticatedFallback: true })
+  })
+})
+
 describe("createMemoizedGithubClient", () => {
   it("fetches once and reuses the resolved client across calls", async () => {
     let calls = 0
--- a/apps/backend/src/features/agents/tools/github/client-accessor.ts
+++ b/apps/backend/src/features/agents/tools/github/client-accessor.ts
@@ -37,6 +37,10 @@ export async function withGithubClient<T>(
  * agent turn so chained GitHub tool calls share one DB fetch and (possibly) one token
  * refresh. A rejection clears the cache so a later call in the same turn gets a fresh
  * attempt instead of inheriting a transient failure.
+ *
+ * Agent tools opt into the unauthenticated fallback: when the workspace has no GitHub
+ * integration, the tools still work against public/open-source repositories instead of
+ * reporting GITHUB_NOT_CONNECTED.
  */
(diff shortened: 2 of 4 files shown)


kristofferremback:
Addressed CodeRabbit's nitpick: added a `service.test.ts` case asserting `getGithubClient(..., { allowUnauthenticatedFallback: true })` still returns `null` when an active integration is near its rate limit — locking in the deliberate no-fallback back-off contract. (91e05da)

PR #721, opened 2026-05-31
Fix infinite "Updating Threa" loop on stale-deploy recovery

kristofferremback:
## Problem

A tab could get permanently trapped on the chunk-load auto-recovery screen — **"Updating Threa / Fetching the latest version…"** — with no way out (no buttons, just a spinner). Reported from a mobile/PWA session.

That screen is the stale-deploy recovery state in `error-boundary.tsx`: when an old tab requests a lazy JS chunk whose content-hashed filename was replaced by a newer build, the import 404s and the app tries to self-heal (unregister SW, clear caches, hard-reload). The escape hatch — the error card with **Try Again / Back to Workspaces / Clear cache & reload** — only appears once `runSwRecovery()` reaches its per-session cap (`MAX_ATTEMPTS = 2`), returns `false`, and flips `recoveryDeclined = true`.

## Root cause

The cap could never be reached. The CSS watchdog in `index.html` cleared the shared `sw-recovery-attempts` counter on **every** load where stylesheets were present:

```js
if (hasStyles) {
  sessionStorage.removeItem("sw-recovery-attempts")  // ← the bug
  return
}
```

A broken **JS** chunk loads **CSS** just fine. So each reload cycle went: load → watchdog sees styles, zeroes the counter → router re-enters the dead route → chunk 404 → `runSwRecovery()` (counter 0→1) → reload → repeat. The counter ping-ponged 0↔1 forever, the cap was never hit, the escape buttons never rendered. Permanent spinner.

## Fix

Stamp each recovery attempt with a timestamp (`sw-recovery-last`), and only clear the counter on a healthy load when that stamp is older than a 60s cooldown:

- In a tight reload loop, attempts are seconds apart → the counter accumulates to the cap → the escape UI surfaces (within ~2 cycles).
- A genuine one-off recovery (app heals and stays up) still clears the counter later, so future unrelated breaks can recover.

The cooldown/keys are duplicated across `index.html` (plain JS, can't import TS) and `sw-recovery.ts`, mirroring the existing `MAX_ATTEMPTS` duplication; both sides document the sync requirement.

## Testing

- `apps/frontend/src/lib/sw-recovery.test.ts` — added coverage that each auto-recovery attempt stamps `sw-recovery-last`, and that `force: true` does not. 10/10 green.
- Existing cap/increment/force tests still pass.
- Full pre-commit suite (monorepo lint + typecheck) passed on push.

## Follow-up (not in this PR)

This guarantees escape from the loop but doesn't address *why* reloads weren't curing the underlying 404 — that points at stale assets being served post-deploy (CDN/edge cache on `index.html`, or the SW re-serving old content). Worth a separate look at the deploy/caching layer.

https://claude.ai/code/session_019CUa7RRYsV54b29dX6i5qs

---
_Generated by [Claude Code](https://claude.ai/code/session_019CUa7RRYsV54b29dX6i5qs)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/721"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782845138&installation_id=129946197&pr_number=721&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F721&signature=47daf9d2c9616bd864971da34075dc72935cf830ee131eaff357f0231fd7e451"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/index.html
+++ b/apps/frontend/index.html
@@ -153,11 +153,21 @@
             }
           }
           if (hasStyles) {
-            // Styles loaded successfully — reset counter so future breaks can recover
-            sessionStorage.removeItem("sw-recovery-attempts")
+            // Styles loaded — but a broken JS chunk loads CSS fine too, so
+            // "hasStyles" is NOT proof the last recovery worked. Only clear the
+            // counter when the last attempt is old enough that the app has
+            // clearly been healthy since; otherwise a chunk-error reload loop
+            // would keep zeroing this and never reach the cap that surfaces the
+            // escape UI. Cooldown mirrors the note in src/lib/sw-recovery.ts.
+            var last = parseInt(sessionStorage.getItem("sw-recovery-last") || "0", 10)
+            if (!last || Date.now() - last > 60000) {
+              sessionStorage.removeItem("sw-recovery-attempts")
+              sessionStorage.removeItem("sw-recovery-last")
+            }
             return
           }
           sessionStorage.setItem("sw-recovery-attempts", String(attempts + 1))
+          sessionStorage.setItem("sw-recovery-last", String(Date.now()))
           var tasks = []
           if (navigator.serviceWorker) {
             tasks.push(
--- a/apps/frontend/src/lib/sw-recovery.test.ts
+++ b/apps/frontend/src/lib/sw-recovery.test.ts
@@ -47,6 +47,7 @@ describe("runSwRecovery", () => {
   afterEach(() => {
     Object.defineProperty(window, "location", { configurable: true, value: originalLocation })
     sessionStorage.clear()
+    vi.restoreAllMocks()
   })
 
   it("returns false without reloading once the per-session cap is reached", async () => {
--- a/apps/frontend/src/lib/sw-recovery.test.ts
+++ b/apps/frontend/src/lib/sw-recovery.test.ts
@@ -63,6 +63,22 @@ describe("runSwRecovery", () => {
     expect(sessionStorage.getItem("sw-recovery-attempts")).toBe("2")
   })
 
+  it("stamps the last-attempt time so the watchdog's reset can be cooldown-gated", async () => {
+    const now = 1_700_000_000_000
+    vi.spyOn(Date, "now").mockReturnValue(now)
+    await runSwRecovery()
+    // The index.html CSS watchdog reads this stamp and refuses to clear the
+    // shared counter while it's recent — that's what stops a broken-chunk
+    // reload loop from zeroing the counter and never reaching the cap.
+    expect(sessionStorage.getItem("sw-recovery-last")).toBe(String(now))
+  })
+
+  it("force: true does not stamp the last-attempt time", async () => {
+    const result = await runSwRecovery({ force: true })
+    expect(result).toBe(true)
+    expect(sessionStorage.getItem("sw-recovery-last")).toBeNull()
+  })
+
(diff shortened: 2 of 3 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-8[1m] | Reviewers: sonnet x3

One minor design finding, already addressed in [`ffa1041d1869ff79b2627d797ce937d467bba164`](https://github.com/threahq/threa/commit/ffa1041d1869ff79b2627d797ce937d467bba164); correctness, security, CLAUDE.md, and plan adherence all clean.

1. `apps/frontend/src/lib/sw-recovery.ts` — The 60s reset cooldown is enforced only in `index.html` (the sole reader of the `sw-recovery-last` stamp), while `sw-recovery.ts` writes the stamp but didn't name the canonical value. Adding a `COOLDOWN_MS` constant to the TS module would be **unused** there (no TS code reads the cooldown) and would conflict with INV-36/INV-38, so instead the header comment now names the value and its owner for discoverability. `index.html` genuinely cannot import the TS module (inline plain-JS script), so the cross-file coordination via sessionStorage keys is intrinsic, not avoidable sprawl.
   https://github.com/threahq/threa/blob/ffa1041d1869ff79b2627d797ce937d467bba164/apps/frontend/src/lib/sw-recovery.ts#L8-L21

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Diff implements exactly what the PR describes: stamp each recovery attempt (`sw-recovery.ts`), gate the watchdog's counter-clear on a 60s cooldown (`index.html`), plus two tests. No missing corresponding changes.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Loop termination verified across timing orderings (chunk error before vs after the 3s watchdog) — attempts accumulate to the cap so the escape UI always surfaces. `parseInt` on a missing/empty/corrupt `sw-recovery-last` yields `0`/`NaN`, and `!last` treats both as "no stamp" (clears the counter — the safe direction). Clock skew (negative `now - last`) falls to "don't reset," the loop-safe direction. `force: true` correctly skips the stamp.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>Comments explain why, not change history (INV-25). Test uses scoped `vi.spyOn` (INV-48) and now restores mocks in `afterEach`. The duplicated `sw-recovery-*` keys / cap across `index.html` and the TS module are pre-existing and intrinsic to the ES-module boundary, not a new INV-33 violation.</details>
<details><summary>🏗️ Design [1 concern, addressed]</summary>See finding 1 above — cooldown value now documented in the typed module without introducing dead code.</details>
<details><summary>🔒 Security [CLEAN]</summary>No user-controlled inputs; sessionStorage-only state, no injection/XSS/auth surface.</details>

PR #731, opened 2026-06-02
feat(sidebar): user-defined custom sections

kristofferremback:
## What

Adds hand-curated **custom sections** to the sidebar — the lightweight cousin of labels. You name a section in the sidebar editor, then file streams into it. No shared catalog, no per-stream attribution: membership is per-user and lives inline in the sidebar config (the same place quick-links and label sections already live), so there's no new DB table to sync.

## Three ways to file a stream

All three enforce **one section per stream** (filing into one removes it from any other):

- **Desktop drag-and-drop** — drag a stream row onto a custom section. A pill follows the cursor, the source row dims, and the target section highlights. Disabled on mobile so touch scroll / long-press stay untouched.
- **"Add to section…"** action — in the desktop hover menu and the mobile action drawer. Opens a picker (centered dialog on desktop, bottom drawer on mobile via `ResponsiveDialog`) to pick/unpick a section or create a new one and file in one step.
- **The editor** — create sections by name, rename them inline, reorder and remove them like any other section.

## "Custom trumps order"

A stream filed into a custom section shows **only** there — even when a smart/label/type section ordered *above* it would also match. Custom membership is collected up front and withheld from every non-custom section regardless of position, rather than competing for the topmost claim. (The headline example from the request — a stream that fits the first labeled section but is put in the second custom section shows only in the custom section — is the first `resolve-sections` test.)

## Implementation

- **Types** (`packages/types/src/sidebar.ts`): new `{ kind: "custom"; sectionId; name; streamIds }` section spec. `normalizeSidebarConfig` sanitizes `streamIds` (strings only, in-section dedupe) and enforces single-membership across sections on every read/write boundary — first listing wins — and drops blank-named sections.
- **Backend** (`sidebar-config/handlers.ts`): the `custom` branch added to the Zod discriminated union (trimmed name, length cap, `streamIds` capped + defaulted).
- **Resolution** (`resolve-sections.ts`): two-pass — gather custom membership, then exclude it from every non-custom section regardless of order. Zero overhead for users without custom sections (no extra Set allocation when none exist).
- **Config helpers** (`sidebar-config.ts`): pure, tested `createCustomSection` / `renameCustomSection` / `setStreamCustomSection` / `getStreamCustomSectionId` / `customSections`, plus the shared `customSectionId()` id helper.
- **DnD** (`sidebar-dnd.tsx`, `sections.tsx`, `sidebar-stream-list.tsx`): stream rows become drag sources (desktop only — bare rows on mobile, no dead draggables); custom sections become drop targets. Pointer sensor with an 8px activation distance so a plain click still navigates. Keyboard users file via the accessible "Add to section…" action rather than keyboard drag.
- **Surfaces**: `SectionPicker` (new), wired into both `stream-item.tsx` and `scratchpad-item.tsx`.

## Design notes

- **Membership in per-user JSONB config, not a DB table.** Matches the "without the hassle of labels" intent and reuses the existing optimistic + cross-device sync path that quick-links/label sections already use. Stale ids (archived/deleted streams) are benign — `resolveCustomSection` filters against visible streams, so a lingering id simply resolves to nothing, exactly like a lingering label assignment.
- The new `custom` kind is handled at every `spec.kind` exhaustion site (presentation, id derivation, resolver, editor row, normalize, Zod schema).

## Testing

`bun run typecheck` (whole monorepo), lint, and unit suites green. **+103 sidebar tests pass**, with new coverage for:
- the resolver (trumps-order, label/custom interaction, activity sort, single-membership)
- config helpers (create/rename/file/move/clear, exclusivity)
- `normalizeSidebarConfig` sanitization (blank name, non-array/dup streamIds, cross-section dedupe)
- the backend Zod schema (accepts custom, defaults `streamIds`, rejects blank name)
- the editor (inline create + rename) and `SectionPicker` (pick / unpick / create-and-file)

## Review

Self-reviewed with a 3-perspective pass (spec-compliance, correctness/security, design). Spec and correctness came back clean; two design findings (a dead-draggable registration on mobile, and a duplicated id-prefix convention) were fixed in the follow-up commit.

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_018PdtjFBBX1cy6mW18na17F

---
_Generated by [Claude Code](https://claude.ai/code/session_018PdtjFBBX1cy6mW18na17F)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/731"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783007818&installation_id=129946197&pr_number=731&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F731&signature=556287ed6d6632d843334af96bf27d761ec9a22d99254b3a797bc9f1d44397a8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/sidebar-config/handlers.ts
+++ b/apps/backend/src/features/sidebar-config/handlers.ts
@@ -9,6 +9,7 @@ import {
   SIDEBAR_QUICK_LINKS,
   SIDEBAR_QUICK_LINK_VISIBILITIES,
   MAX_CUSTOM_SECTION_NAME_LENGTH,
+  MAX_CUSTOM_SECTION_STREAM_IDS,
 } from "@threa/types"
 
 const sidebarSectionSpecSchema = z.discriminatedUnion("kind", [
@@ -22,7 +23,7 @@ const sidebarSectionSpecSchema = z.discriminatedUnion("kind", [
     kind: z.literal("custom"),
     sectionId: z.string().min(1).max(64),
     name: z.string().trim().min(1).max(MAX_CUSTOM_SECTION_NAME_LENGTH),
-    streamIds: z.array(z.string().min(1).max(64)).max(500).optional().default([]),
+    streamIds: z.array(z.string().min(1).max(64)).max(MAX_CUSTOM_SECTION_STREAM_IDS).optional().default([]),
   }),
   z.object({ kind: z.literal("quicklinks") }),
 ])
--- a/apps/backend/src/features/sidebar-config/handlers.ts
+++ b/apps/backend/src/features/sidebar-config/handlers.ts
@@ -8,12 +8,22 @@ import {
   SIDEBAR_BASE_PRESETS,
   SIDEBAR_QUICK_LINKS,
   SIDEBAR_QUICK_LINK_VISIBILITIES,
+  MAX_CUSTOM_SECTION_NAME_LENGTH,
 } from "@threa/types"
 
 const sidebarSectionSpecSchema = z.discriminatedUnion("kind", [
   z.object({ kind: z.literal("smart"), bucket: z.enum(SIDEBAR_SECTION_KEYS) }),
   z.object({ kind: z.literal("type"), streamType: z.enum(SIDEBAR_TYPE_SECTIONS) }),
   z.object({ kind: z.literal("label"), labelId: z.string().min(1).max(64) }),
+  // Custom section: a hand-curated membership the service normalizes (sanitizes
+  // streamIds, enforces single membership) on write. Cap streamIds so the
+  // per-user config document can't grow unbounded.
+  z.object({
+    kind: z.literal("custom"),
+    sectionId: z.string().min(1).max(64),
+    name: z.string().trim().min(1).max(MAX_CUSTOM_SECTION_NAME_LENGTH),
+    streamIds: z.array(z.string().min(1).max(64)).max(500).optional().default([]),
+  }),
   z.object({ kind: z.literal("quicklinks") }),
 ])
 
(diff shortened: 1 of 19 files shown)


kristofferremback:
Addressed the custom-section id-integrity finding in `27a84b4`, with a slightly different approach than the suggested Zod `refine`:

`normalizeSidebarConfig` now dedupes custom sections by `spec.sectionId` (in addition to the existing dedup by `id`). The real integrity risk is two custom sections sharing a `sectionId` — that's what would desync the filing/rename helpers (they key off `sectionId`) and collide the drop-zone id. A drifted-but-unique `id` on its own is only a cosmetic collapse-state key.

I kept this in `normalizeSidebarConfig` rather than a Zod refine because (a) the service normalizes on both read and write, so it also heals any already-persisted bad row, consistent with how normalize already sanitizes stray data; and (b) it keeps the `custom:` id-derivation convention in the frontend presentation layer (`sectionIdForSpec`) rather than duplicating that string into the types/backend validation layer. Covered by a new test in `packages/types/src/sidebar.test.ts`.

---

PR #741, opened 2026-06-03
feat(sidebar): support pinning channels into the Pinned section

kristofferremback:
## What

Wires channel pinning end-to-end so the sidebar's **Pinned** section works, composing correctly with labeled streams and custom sections.

The backend was already complete — `stream_members.pinned`/`pinned_at`, `POST /api/workspaces/:id/streams/:id/pin`, and `resolveSections` already resolved the `"pinned"` smart bucket. The gap was entirely the frontend: nothing set a stream's section to `"pinned"`, and there was no way to pin one. `categorizeStream` even carried a `// TODO: Add pinned support when backend implements it`.

## Changes (frontend only)

- **`categorizeStream`** now takes an `isPinned` flag; a pinned stream always lands in the Pinned section regardless of read state.
- **`sidebar.tsx`** derives the pinned set from IDB memberships (pin state is per-membership, not on the stream row) and threads `isPinned` into categorization and onto `StreamItemData`.
- **`streamsApi.pin` + `usePinStream`** — optimistically writes the pin to IDB (reverting on error) so the Pinned section updates instantly, then reconciles the server's authoritative `pinnedAt` and the workspace bootstrap cache. Mirrors the existing `markAsRead` / `setNotificationLevel` patterns.
- **`StreamItem`** gains a **Pin / Unpin** action.

## Composing with labels & custom sections

This was the key constraint, and it falls out of the existing `resolveSections` precedence — pinning just feeds the same `"pinned"` bucket those rules already handle:

- **Custom sections** stay exclusive and trump pinning: a stream filed into a custom section shows only there (custom members are withheld from every smart bucket, including Pinned).
- **Labels** stay topmost-wins: a pinned + labeled stream is claimed by whichever of its Pinned / label section sits higher in the user's order.

## Behavior note

Pins are **sticky** — a pinned stream stays in Pinned even with an unread mention (the urgency strip still flags it in place), rather than being promoted to Important. This keeps the section stable and predictable. Easy to switch to promote-on-unread if preferred.

## Tests

- `categorizeStream` pinned cases.
- Pin / Unpin action fires the mutation with the right args.
- `usePinStream` optimistic IDB write + error revert + cache reconcile.
- All 110 sidebar tests pass; frontend typecheck clean. The pre-commit hook ran the full monorepo lint + typecheck — all green.

https://claude.ai/code/session_01BzeaUxkzYkyShM7UspT7qG

---
_Generated by [Claude Code](https://claude.ai/code/session_01BzeaUxkzYkyShM7UspT7qG)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/741"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783060011&installation_id=129946197&pr_number=741&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F741&signature=a0957171dc11aeba5f9e3e739c965f5f83ad86a3d37234947abefe2ed29fac0a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/api/streams.ts
+++ b/apps/frontend/src/api/streams.ts
@@ -152,6 +152,13 @@ export const streamsApi = {
     return res.membership
   },
 
+  async pin(workspaceId: string, streamId: string, pinned: boolean): Promise<StreamMember> {
+    const res = await api.post<{ membership: StreamMember }>(`/api/workspaces/${workspaceId}/streams/${streamId}/pin`, {
+      pinned,
+    })
+    return res.membership
+  },
+
   async join(workspaceId: string, streamId: string): Promise<StreamMember> {
     const res = await api.post<{ data: { membership: StreamMember } }>(
       `/api/workspaces/${workspaceId}/streams/${streamId}/join`
--- a/apps/frontend/src/components/layout/sidebar/sidebar.tsx
+++ b/apps/frontend/src/components/layout/sidebar/sidebar.tsx
@@ -104,6 +104,14 @@ export function Sidebar({ workspaceId }: SidebarProps) {
 
   // Build set of muted streams (for suppressing unread badges)
   const mutedStreamIdSet = useMemo(() => new Set(unreadState?.mutedStreamIds ?? []), [unreadState?.mutedStreamIds])
+
+  // Streams the viewer pinned — drives the smart "Pinned" section. Pinned state
+  // is per-membership (not on the stream row), so derive it from memberships.
+  const pinnedStreamIdSet = useMemo(() => {
+    const ids = new Set<string>()
+    for (const m of idbStreamMemberships) if (m.pinned) ids.add(m.streamId)
+    return ids
+  }, [idbStreamMemberships])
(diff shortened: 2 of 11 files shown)


kristofferremback:
Not sure I think we need pinning now that we have custom sections.

PR #755, opened 2026-06-03
fix(push): stop the service worker silencing notifications you can't actually see

kristofferremback:
## Problem

Push notifications were intermittently not showing on mobile. I checked production first to rule out the server side: the user's Android subscription was fresh (re-registered ~2.7h prior), the `push-notifications` outbox listener was fully caught up (cursor = max id, 0 retries, 0 dead-letters), and the notification level was `all`. So events were flowing and delivery was being attempted without error — the gap was on the device.

Two service-worker behaviors were dropping notifications:

1. **Blanket focus suppression.** The SW suppressed *every* notification whenever any Threa window was focused, regardless of which stream you were looking at. A message in another stream produced nothing while the app was open. And because the backend deliberately routes a push to the focused/attended device expecting the SW to drop it, each of those was a `userVisibleOnly` silent push — which mobile browsers penalize against the push-display budget, degrading later delivery.

2. **Focus treated as presence.** Suppression keyed on `client.focused` alone, with no check that you were actually *there*. A tab left focused and walked away from kept eating notifications for the stream on screen — even though the backend's attended-device routing already requires a real interaction within 2 minutes. The two layers disagreed on what "present" means.

## Change

- **Stream-aware suppression** (`3d3f02a`): suppress only when a focused window is viewing *that* stream (read from the client URL, `/w/{ws}/s/{stream}`; an open thread `?m=…` still resolves to its stream). A push for any other stream now shows even with the app focused, and only the on-screen stream is ever skipped — which also drastically cuts the silent-push budget burn.

- **Interaction-gated presence** (`c6d802e`): suppress only when the focused window viewing that stream was also interacted with (pointer/key/touch) in the last 2 minutes. The page records its last interaction into a shared Cache entry (throttled); the SW reads it in the push handler. Cache Storage is shared across the origin and survives the SW's frequent restarts, so this needs no message round-trip — and it fails open to *showing* when presence is unknown.

- **Self-review hardening** (`6abd2c0`): guard Cache Storage access so an error can't reject the push handler and silently drop a notification; register the presence cache in the SW activate-cleanup allowlist; hoist the 2-minute window to `@threa/types` (`PRESENCE_INTERACTION_WINDOW_MS`) so the backend routing and the SW share one source of truth; extract the presence writer into a `useSwPresence` hook.

## Notes

- Behavior change confirmed with the user: when you're actively at your laptop, notifications surface on the laptop and the phone stays quiet (no desk double-buzz); the phone fans back in automatically once the laptop isn't actively attended.
- "Interaction" matches the backend exactly (`pointerdown`/`keydown`/`touchstart`). Passive reading via mouse-wheel for >2 min then a message *in that same stream* can now notify — the intended consequence of "interacted, not just focused."
- Takes effect per device once the new SW installs and claims (next app load).

## Testing

- New unit tests: `isViewingStream` (stream/workspace match, query strings, non-stream routes, missing ids, malformed URLs) and `isWithinPresenceWindow` (window boundaries, never-recorded).
- `bun run test` for the touched suites (format, presence, contexts) — green. Full monorepo lint + typecheck green via pre-commit.

🤖 Self-reviewed with the multi-perspective code-review skill (3× Sonnet) before opening; findings addressed in `6abd2c0`.

---
_Generated by [Claude Code](https://claude.ai/code/session_012hboNRpoVYgcMg3Fg3fQz3)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/755"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783107014&installation_id=129946197&pr_number=755&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F755&signature=6a4fe9df0dd2cc011ab1a546812ff57140cbc68126d5aaff7af7b62880c63ffe"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/push/service.ts
+++ b/apps/backend/src/features/push/service.ts
@@ -7,6 +7,7 @@ import {
   PrefNotificationLevels,
   ActivityTypes,
   StreamTypes,
+  PRESENCE_INTERACTION_WINDOW_MS,
   type PrefNotificationLevel,
   type StreamType,
 } from "@threa/types"
@@ -36,7 +37,9 @@ const CURRENTLY_FOCUSED_WINDOW_MS = 60_000
  * fall through to fanout so the user gets notified on whichever device they
  * pick up next.
  */
-const RECENT_INTERACTION_WINDOW_MS = 2 * 60 * 1_000 // 2 minutes
+// Shared with the frontend SW's push-suppression check so the two layers agree
+// on what "present" means (@threa/types).
+const RECENT_INTERACTION_WINDOW_MS = PRESENCE_INTERACTION_WINDOW_MS
 
 /**
  * Per-device session expiry window. If a specific device has not sent a heartbeat
--- a/apps/frontend/src/contexts/socket-context.tsx
+++ b/apps/frontend/src/contexts/socket-context.tsx
@@ -5,21 +5,13 @@ import { api } from "@/api/client"
 import { getCachedWsConfig, setCachedWsConfig } from "@/lib/cached-ws-config"
 import { usePageActivity } from "@/hooks/use-page-activity"
 import { usePageInteraction } from "@/hooks/use-page-interaction"
-import { recordInteractionPresence } from "@/lib/sw-presence"
+import { useSwPresence } from "@/hooks/use-sw-presence"
 
 /** Periodic heartbeat tick for session liveness — must be < ACTIVE_SESSION_WINDOW_MS on the backend (60s). */
 const PERIODIC_HEARTBEAT_INTERVAL_MS = 30_000
 
 /** Throttle for focus-change-driven heartbeats so a flicker of blur/focus events doesn't flood the socket. */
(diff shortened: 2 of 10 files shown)


kristofferremback:
@coderabbitai review

PR #710, opened 2026-05-30
feat(enclave): web tools + bounded research inside the E2E enclave (Slice PR 6)

kristofferremback:
## Problem

The Ariadne enclave runs the real `AgentRuntime` loop next to decrypted E2E
plaintext, but with `tools: []` — she could reason and reply, but not search the
web or read a URL inside an E2E scratchpad. Phase 5's PR 6 was scoped as "wire
`web_search` + `read_url`", but that predates the **general researcher** (#689),
which is now the persona's default research surface. Shipping only the raw
primitives would leave the enclave persona materially behind the non-E2E one.

## Solution

Give the enclave persona the same web research surface the backend persona has —
`web_search`, `read_url`, and a **web-only `general_research`** loop — with every
tool step sealed under the stream SSK and streamed to the trace, so the server
still never sees plaintext (INV-E7).

The general researcher's core loop was backend-only. Rather than reimplement it
in the enclave (a second copy that would drift — INV-35/37), the runtime-agnostic
core is **extracted to `@threa/agent-runtime`** and both hosts drive it.

### How it works

```
backend persona ─┐                          ┌─ resolves model+config via ConfigResolver/AI
                 ├─► runGeneralResearch ◄────┤
enclave turn ────┘   (shared package loop)   └─ supplies AgentRuntimeAI + assignment model
                          │
                          └─ drives primitive tools → synthesised cited brief
```

The loop now takes **resolved primitives** (`AgentRuntimeAI` + `model` +
`modelString` + numeric config) instead of the backend's full `AI` +
`ConfigResolver`, so it imports cleanly on the enclave's OTEL-free barrel. The
backend `GeneralResearcher` becomes a thin adapter that resolves config + model
and delegates; every backend importer keeps its path via re-export shims, and
behavior is byte-identical (existing suites unchanged).

In the enclave, `buildEnclaveTools` assembles only the enclave-safe subset:
`web_search` (when a Tavily key is present), `read_url` (always), and
`general_research` driving that web subset **in-process** — no backend callback,
so plaintext never leaves the enclave. The turn constructs the enclave AI once so
the turn loop and the research sub-loop share one usage accumulator (INV-19).

### Key design decisions

**1. Extract the researcher core to the shared package (not reimplement).**
One implementation, two callers (INV-35/37). The only backend coupling was
`AI`/`ConfigResolver`, which only supply `modelId`/`temperature`/`maxIterations`
— pushed out to explicit caller params. Prompt + budgets are re-exported from the
[...]

PR #778, opened 2026-06-04
fix(e2e): deliver attachment ciphertext to the enclave + load_attachment tool

kristofferremback:
## The bug (why Ariadne couldn't read any encrypted attachment)

The enclave attachment slice (#765) **never worked in integration**: `sessionAssignmentSchema` in `apps/enclave/src/sessions.ts` was missing `attachmentCiphertexts`, and **Zod strips unknown keys** — the backend shipped the bytes, the enclave's request parse silently deleted them, and every file reached the model as `[Attachment "…" is unavailable]`. (The schema's own doc comments warn about exactly this failure mode for two other fields.)

The slice's 45 tests were green because they called `runEnclaveTurn` directly, bypassing the HTTP parse that broke it. Confirmed live with debug taps in dev: backend logged `loaded … bytes: 17668`, enclave logged `shipped: []`.

A second finding from Kris's testing ("it's not PDFs, it's all files"): browsers report **no MIME type** for `.md`/code/CSV files, so their refs say `application/octet-stream` — which no model accepts as a `file` part even once the bytes arrive.

## What this PR does

**Delivery (the root-cause fix)**
- Declare `attachmentCiphertexts` in the enclave schema + a regression test that round-trips the assignment through `sessionAssignmentSchema.parse` so the next undeclared field can't ship silently.
- `/sessions` body limit 4mb → 48mb, with matching worker-side budgets: 16MB/file, ~32MB base64 + 64-file total, dropped files always `logger.warn`'d (no silent caps) and surfaced to the model as "unavailable" notes.

**Trigger files feed eagerly** (as before, plus):
- images → image parts, PDFs → file parts (unchanged)
- **new:** anything valid UTF-8 (strict `TextDecoder`) is inlined as text contents — covers the no-MIME reality of markdown/code/CSV; true binary degrades to an honest note instead of a request the provider would reject.

**History files load on demand — `load_attachment` (Kris's call)**
- The worker now ships ciphertext for the trigger **plus recent history** (newest-first priority within the budget).
- History messages carry `[Attached: "name" (attachment id)]` notes instead of auto-fed bytes (token cost proportional to use), and a new **in-enclave `load_attachment` tool** decrypts the inline-shipped ciphertext next to the model call — never S3, never a backend callback. Same tool name as the main app's, added to `ENCLAVE_ENABLED_TOOLS` for prompt parity.
- Deliberately **outside the stream's tool-privacy gate**: the files are the conversation's own content (their refs already ride the messages the model reads), not workspace reads or web egress — so a web-restricted scratchpad can still read its own files.
- `AgentToolResult.multimodal` gains a `file` variant (runtime injection maps it to an AI-SDK file part) so a tool can hand the model a PDF, not just images. The main app's image-only `load_attachment` is unaffected.

**Test-infra fix**
- `scripts/dev-test.ts` never passed `INTERNAL_API_KEY` to the wrangler workspace-router, so `resolveRegion` was skipped and every uncached workspace 404'd ("Workspace not found"). One `--var` fixes dev:test for everyone.

## Verification

Beyond units (enclave 48, backend enclave-runtimes 56, agent-runtime 141, all typechecks):

**Live end-to-end via dev:test + Playwright + real OpenRouter inference** —
1. Fresh user → workspace → **New Encrypted Scratchpad** → passphrase setup.
2. Uploaded a `.md` file (ref mime: `application/octet-stream`) + asked for specifics → Ariadne replied: *"The launch codename is **VELVET-OTTER-99**, and the most important rule is: 'never deploy on a Friday.'"* — exact file contents, sealed end-to-end.
3. Follow-up turn (**no attachment on the trigger**): "What were the two key dates listed in that file?" → *"Code freeze: June 12, Launch: June 19"* — facts that appear in no message text, and the session trace shows the `tool_call` step where she loaded the history file.

Debug taps used for diagnosis were removed before commit.

## Notes for review

- The enclave's `load_attachment` being ungated by tool privacy is a deliberate judgment call (reasoning above + in code comments) — flag if you disagree.
- Known follow-ups (unchanged scope): very large files fall back to "unavailable" notes rather than a backend-proxy fetch; mid-turn reconsideration; PR9 workspace tools.

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/778"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783203981&installation_id=129946197&pr_number=778&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F778&signature=8fa09c638508533532b1518760e50b44a23a69123c2b036d8d753c7f3a0660e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/enclave-system-prompt.ts
+++ b/apps/backend/src/features/agents/enclave-system-prompt.ts
@@ -23,6 +23,9 @@ const ENCLAVE_ENABLED_TOOLS: string[] = [
   AgentToolNames.WEB_SEARCH,
   AgentToolNames.READ_URL,
   AgentToolNames.GENERAL_RESEARCH,
+  // Conversation-local file reads: the enclave's in-process variant decrypts
+  // inline-shipped ciphertext (no S3, no backend callback).
+  AgentToolNames.LOAD_ATTACHMENT,
 ]
 
 export async function buildEnclaveSystemPrompt(params: {
--- a/apps/backend/src/features/enclave-runtimes/dispatch/enclave-invoke-worker.test.ts
+++ b/apps/backend/src/features/enclave-runtimes/dispatch/enclave-invoke-worker.test.ts
@@ -76,7 +76,7 @@ function arrangeDispatch() {
   spyOn(EnclaveRuntimesRepository, "listLive").mockResolvedValue([EIK])
   spyOn(StreamE2eKeyWrapsRepository, "listForStream").mockResolvedValue([WRAP])
   spyOn(MessageRepository, "findSurrounding").mockResolvedValue([])
-  spyOn(AttachmentRepository, "findByMessageId").mockResolvedValue([])
+  spyOn(AttachmentRepository, "findByMessageIds").mockResolvedValue(new Map())
   spyOn(StreamRepository, "findById").mockResolvedValue({ id: "stream_1", workspaceId: "ws_1" } as never)
   spyOn(UserPreferencesService.prototype, "getPreferences").mockResolvedValue({} as never)
   spyOn(UserRepository, "findByIds").mockResolvedValue([{ name: "Kris" }] as never)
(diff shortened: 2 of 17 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: opus-4.8 | Reviewers: sonnet x3

Found 2 issues (both non-blocking design concerns):

1. `apps/backend/src/features/enclave-runtimes/dispatch/enclave-invoke-worker.ts:44` / `apps/enclave/src/sessions.ts:38` — The attachment count cap (64) is duplicated as two separate constants in two separate packages, coordinated only by a comment ("mirrored by the enclave schema's `MAX_INLINE_ATTACHMENTS`"). If the worker's cap is ever raised above the enclave schema's `.max(64)`, valid assignments get rejected with a 400 at the enclave parse. The failure would be loud, but the coupling is invisible to whoever edits one side. The enclave already imports `@threa/types` (e.g. `AgentStepTypes` in `attachment-tool.ts`), so the constant could live once there, next to the assignment contract. (abstraction: config sprawl)
   https://github.com/threahq/threa/blob/e717a5bfa133384cecda040a73c910aa707adfac/apps/backend/src/features/enclave-runtimes/dispatch/enclave-invoke-worker.ts#L42-L44
   https://github.com/threahq/threa/blob/e717a5bfa133384cecda040a73c910aa707adfac/apps/enclave/src/sessions.ts#L37-L38

2. `apps/enclave/src/agent/attachment-tool.ts:44-51` — `load_attachment`'s not-found branch conflates two distinguishable cases: an id the enclave genuinely doesn't know (`!ref`) vs a budget-dropped file (`ref` present, `ciphertext` absent — the file's `[Attached: …]` note still rides the message, so the model will plausibly ask for it). The message does hedge ("unknown id, or the file was too large to ship"), but since `ref && !ciphertext` identifies the drop case exactly, a precise "this file was too large to ship into the enclave" would stop the model from second-guessing an id it correctly copied from the note. (abstraction: imprecise error for a reachable case)
   https://github.com/threahq/threa/blob/e717a5bfa133384cecda040a73c910aa707adfac/apps/enclave/src/agent/attachment-tool.ts#L41-L51

---

Checked the diff against the PR description and `.claude/plans/e2e-attachments.md` (Slice C). All stated changes are present: `attachmentCiphertexts` declared in `sessionAssignmentSchema` + a regression test that round-trips the assignment through `sessionAssignmentSchema.parse`; `/sessions` body limit raised to 48mb with matching worker-side budgets (16MB/file, ~32MB base64 total, 64-file cap) and `logger.warn` on every drop; UTF-8 text inlining via strict `TextDecoder`; in-enclave `load_attachment` tool with inline ciphertext (no S3, no backend callback); `file` multimodal variant plumbed through agent-runtime; `INTERNAL_API_KEY` passed to the wrangler workspace-router in `scripts/dev-test.ts`. No missing corresponding changes found.
</details>

Traced: trigger-first ordering in `attachmentCiphertextIds` (correct — trigger is hardcoded first, history is filtered + reversed to newest-first); the budget loop's `continue` (correct — warns per dropped file rather than breaking silently); base64 budget (~32MB) vs the 48mb body limit (fits with history + system prompt headroom); `load_attachment` always advertised even with empty maps (graceful "not available" response, intentional); decryption failure handled with an explicit error result. No significant bugs found.
</details>

Verified against the invariant playbook: INV-30/41 (ciphertext loads use `pool` directly; S3 I/O happens after the DB query, no connection held during network work), INV-11 (budget overruns warn loudly, no silent fallback), INV-55 (sessions handler validates via Zod), INV-5/27 (reuses the composable `findByMessageIds` repo method), INV-36/38 (no speculative features, no dead code left). The enclave not using `createAI` (INV-28) is pre-existing and documented as deliberate in `llm.ts`. No violations introduced.
</details>

The two issues above: the duplicated 64-count cap across `enclave-invoke-worker.ts` / `sessions.ts`, and the conflated unknown-id/budget-dropped error in `load_attachment`. Otherwise: the `file` multimodal variant is plumbed consistently through `agent-tool.ts` → `agent-runtime.ts`, and the new tool matches the established enclave patterns from #710/#711 (local tool construction in `buildEnclaveTools`, trace formatting, honest degradation notes).
</details>

Data-flow traced with the E2E threat model in mind ("no key material or plaintext on the server, ever"): refs (key/iv/filename) are decrypted only inside the enclave from sealed payloads; the backend ships opaque ciphertext only; drop warnings log attachment id + sizes, never filenames or content; the decrypt happens next to the model call. The tool-privacy-gate bypass for `load_attachment` is the PR's stated deliberate judgment call (conversation-own content, no workspace read or web egress) and the reasoning holds. No high-confidence vulnerabilities.
</details>

PR #797, opened 2026-06-05
Fix "keep me unlocked" not persisting across app restarts

kristofferremback:
"Keep me unlocked on this device" stored the UIK's X25519 private key
directly as a non-extractable CryptoKey in IndexedDB. Some browsers
(observed on Android Chrome) resolve the IndexedDB write — so the unlock
reports success — but silently drop newer X25519/Ed25519 CryptoKeys when
the database is flushed to disk. A reload within the same session still
found it, but a true cold start read the row back without the key and fell
through to the passphrase prompt, so the device was never actually kept
unlocked across app restarts.

Seal the private bytes under a freshly generated, non-extractable AES-GCM
device key instead, and store that AES key plus the sealed bytes. AES-GCM
keys round-trip through on-disk IndexedDB reliably on every engine. The
security posture is unchanged: the AES key is non-extractable, so the IDB
row alone still cannot yield the raw private bytes — its presence remains
the "this device is trusted" signal.

The in-memory session key is now always extractable (it already was after
setup/rotation); at-rest protection comes entirely from wrapping, which
lets the settings "keep me unlocked" toggle re-seal without re-deriving.

Cold start unwraps the sealed bytes to resume unlocked; a corrupt device
key is forgotten and falls back to the passphrase prompt. Existing trusted
devices that did persist (e.g. desktop) re-unlock once, after which the new
format takes over.

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/797"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783275395&installation_id=129946197&pr_number=797&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F797&signature=b319c320f010993cbaf38bf243bdf93e7245d09e757d03760f777220e0e439c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/db/database.ts
+++ b/apps/frontend/src/db/database.ts
@@ -593,10 +593,10 @@ export interface CachedLabelAssignment {
  * sign-out delete it. `keyId` is stored so a server-side key rotation can be
  * detected and the now-stale device key discarded.
  *
- * We previously stored the X25519 private key directly as a `CryptoKey`, but
- * some browsers (Android Chrome) drop newer X25519 keys when IndexedDB is
- * flushed to disk, so it never survived a cold start. The AES-GCM device key
- * round-trips reliably; see `device-wrap-key.ts` for the full rationale.
+ * The X25519 private key is sealed under the AES key rather than stored as a
+ * `CryptoKey` because some browsers (Android Chrome) drop newer X25519 keys
+ * when IndexedDB is flushed to disk; AES-GCM keys round-trip reliably. See
+ * `device-wrap-key.ts` for the full rationale.
  */
 export interface CachedE2eDeviceKey {
   id: string // `${workspaceId}:${userId}` — deterministic, one trusted key per user
--- a/apps/frontend/src/db/database.ts
+++ b/apps/frontend/src/db/database.ts
@@ -581,17 +581,22 @@ export interface CachedLabelAssignment {
 }
 
 /**
- * "Keep me unlocked on this device" record. Persists the UNWRAPPED X25519
- * private key as a NON-EXTRACTABLE `CryptoKey` (structured-clone-serializable
- * native key) so a page reload can resume the unlocked session without the
- * passphrase / Argon2id step.
+ * "Keep me unlocked on this device" record. Persists the UIK private key sealed
+ * for at-rest storage so a cold start can resume the unlocked session without
+ * the passphrase / Argon2id step.
  *
- * Security note: the key is non-extractable — even an attacker with this IDB
- * record cannot read the raw private bytes back out (`crypto.subtle.exportKey`
- * rejects); it stays usable only for in-process HPKE decryption. The presence
- * of a row here IS the "this device is trusted" state. `lock()` / sign-out
- * delete it. `keyId` is stored so a server-side key rotation can be detected
- * and the now-stale device key discarded.
+ * Security note: the private key is never stored in the clear. For the plain
+ * (no PIN/biometric) case it's sealed under a NON-EXTRACTABLE AES-GCM device
+ * key (see `device-wrap-key.ts`) — even an attacker with this IDB record cannot
+ * recover the raw private bytes, because that AES key can't be exported. The
+ * presence of a row here IS the "this device is trusted" state. `lock()` /
+ * sign-out delete it. `keyId` is stored so a server-side key rotation can be
+ * detected and the now-stale device key discarded.
+ *
+ * We previously stored the X25519 private key directly as a `CryptoKey`, but
+ * some browsers (Android Chrome) drop newer X25519 keys when IndexedDB is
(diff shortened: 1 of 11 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (issues are non-blocking)
**Review models:** Orchestrator: Opus | Reviewers: Sonnet x3

Found 2 issues (neither is a correctness bug):

1. `apps/frontend/src/lib/crypto/device-wrap-key.ts:38-61` — Parallel crypto implementation. `wrapPrivateKeyForDevice`/`unwrapPrivateKeyFromDevice` re-do `exportPrivateKey` → `crypto.subtle.encrypt/decrypt` → manual `iv||ciphertext` layout inline, instead of delegating to `wrapPrivate`/`unwrapPrivate` in `keys.ts`. The two sibling device-unlock modules established the opposite pattern — bring your own KEK, delegate the AES-GCM layer: `pin-device-key.ts` calls `wrapPrivate(privateKey, kek)` / `unwrapPrivate(...)`, and `webauthn-device-key.ts` does the same. This PR breaks that pattern, yielding a second AES-GCM-over-private-key path with a divergent wire format (`[iv|ct]` vs `keys.ts`'s versioned `[version|iv|ct]`) and a duplicated `IV_LENGTH = 12` constant (also at `keys.ts:13`). The non-extractable AES key could be generated and passed straight to `wrapPrivate`/`unwrapPrivate` as the KEK; the only delta is the harmless 1-byte version prefix. (CLAUDE.md [INV-35/37, *Architecture Boundaries*]: "Reuse existing helpers and abstractions instead of parallel implementations.")
   https://github.com/threahq/threa/blob/b9cd1d338c746c4f55ca257bc9547e2611438c15/apps/frontend/src/lib/crypto/device-wrap-key.ts#L38-L61

2. `apps/frontend/src/db/database.ts:596-599` — Comment narrates change history. The paragraph beginning "We previously stored the X25519 private key directly as a `CryptoKey`, but…" describes what the code used to do rather than why the current design works, and the full rationale already lives in `device-wrap-key.ts` (which the comment points to). (CLAUDE.md [INV-25, *Maintainability and Scope Control*]: "Comments explain why the current implementation works; they do not narrate the change history.") Minor.
   https://github.com/threahq/threa/blob/b9cd1d338c746c4f55ca257bc9547e2611438c15/apps/frontend/src/db/database.ts#L596-L599

---

No committed plan; checked against the PR description. One candidate finding was raised and **dismissed**: old-format `e2eDeviceKeys` rows (carrying the removed `privateKey` field) are not migrated. This is intentional and documented in the PR body ("Existing trusted devices that did persist re-unlock once, after which the new format takes over"). On the affected browser (Android Chrome) no such rows exist on disk; where they do (desktop), the row is replaced by the `db.e2eDeviceKeys.put` upsert on the next trust-enable, and migrating in place is infeasible because the old key is non-extractable (its bytes can't be re-exported to re-wrap). The proposed remediation would be a no-op or impossible.
</details>

Resume path: the `loadGeneration` guard is correctly re-checked after the new `await unwrapPrivateKeyFromDevice`; corrupt-key fallback deletes the row and lands on `locked` (covered by a new test). AES-GCM IV length/slicing is correct (12-byte IV, `[iv|ct]`). `setDeviceTrust(true)` requiring an extractable in-memory key is satisfied because all unlock paths now return extractable keys.
</details>
<details><summary>📋 CLAUDE.md Compliance [1 minor]</summary>

INV-25 (issue #2 above). INV-38/49 satisfied — `toNonExtractable` and `unwrapPrivate`'s `opts` removed cleanly, no deprecated aliases left.
</details>

Parallel implementation / reuse (issue #1 above). The `{deviceWrapKey, deviceWrappedPrivate}` field pair is otherwise coherent with the existing `pin*`/`webauthn*` fields and correctly documented as mutually exclusive.
</details>

The AES device key is generated non-extractable and verified unexportable by test; raw private bytes are never written to disk in the clear, so the "IDB row alone can't yield the key" posture holds. Making in-memory session keys extractable is a real but non-exploitable delta — an attacker in the JS context can already use (and intercept) the key regardless of the flag — and it matches the pre-existing post-setup/post-rotation behavior.
</details>

---

PR #798, opened 2026-06-05
fix(e2e): seal the first thread reply under an encrypted scratchpad

kristofferremback:
## What

Sending a message in a thread under an **encrypted** scratchpad was broken: the first reply never went through. This fixes it.

## Why it broke

#793 made threads under an E2E scratchpad inherit the root stream's symmetric key (SSK) server-side — a thread is sealed the instant it's created, and the backend's INV-E1 gate then *requires* ciphertext on every message POST to it.

But the **first reply** to a message is what creates the thread, and it goes through a different frontend path than normal sends: the draft-promotion path (`queueDraftMessage`), which never encrypted. So the sequence was:

1. User replies → message queued as **plaintext** with thread-creation metadata.
2. Drain loop promotes the draft → backend creates the thread and **seals it**.
3. Drain sends the queued plaintext to the now-sealed thread → **INV-E1 → 400**, retried forever.

Diagnosed from the prod backend logs (repeated `POST /api/workspaces/.../messages 400`). Subsequent replies to an already-created thread use the normal (encrypting) send path, which is why only the first one failed.

## The fix

All frontend (keys only live on the client):

- **`queueDraftMessage` now seals** when the draft lives under an E2E root: resolves the root's current SSK (held in memory after unlock), seals the body + attachment refs, and stashes `ciphertext/envelope/e2eVersion` on the pending row. The drain's existing E2E branch forwards it after promotion. The AAD binds the root id, but the promoted thread carries a copy of the root's wraps, so it still opens under the thread id (decryption resolves the key by the wraps and uses `envelope.aad` verbatim — verified across the frontend and enclave decrypt paths).
- **The draft-thread composer** reads `e2eEnabled` off the parent stream instead of the always-`undefined` panel stream. Without this it uploaded **attachments as plaintext** under an encrypted root (a leak) and persisted plaintext drafts to IDB.
- **Carry `contentJson`** on the optimistic event so the post-promotion server echo can seed the decrypt cache (stream-sync needs both `contentMarkdown` + `contentJson`) — otherwise the author saw the opaque placeholder for their own reply.

## Self-review hardening (2nd commit)

- **Extracted `sealOutgoingMessage`** (`lib/crypto/seal-send`) and routed both the live send path and the draft path through it, so the encrypt-before-enqueue rule lives on one path (INV-35) instead of two copies.
- **Fail closed in the composer**: don't queue a reply until the parent stream resolves. A cold/slow load could otherwise leave `parentStream` undefined at submit, queue plaintext, and jam the outbox.
- **Best-effort heal-on-send**: when the encrypted root has invited actors, revive stale actor wraps on the root at queue time so the promoted thread copies live wraps — closing a parity gap with the live send path (an enclave whose key rotated would otherwise park the first thread turn; it self-heals on the next reply).

## Tests

- New `use-queue-draft-message` suite: seals under the root, heals when the root has actors, stays plaintext with no root, refuses when the session is locked.
- Frontend hooks / sync / crypto / encryption / timeline suites (734) + full-monorepo typecheck green.

> Note: I couldn't exercise the full browser flow from this environment — worth a manual smoke test (reply in an encrypted scratchpad, confirm the thread's first message lands and renders decrypted, with and without an attachment).

https://claude.ai/code/session_01Tuv1cX1PEHCNDbHWFBHTWe

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tuv1cX1PEHCNDbHWFBHTWe)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/798"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783275946&installation_id=129946197&pr_number=798&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F798&signature=767978f842a00e0e2e330235d494625de2dce548894652b1079dd0282266df6c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/e2e-streams/key-wrap-repository.ts
+++ b/apps/backend/src/features/e2e-streams/key-wrap-repository.ts
@@ -85,34 +85,6 @@ export const StreamE2eKeyWrapsRepository = {
   },
 
   /**
-   * Copy every wrap from `fromStreamId` onto `toStreamId` (same workspace),
-   * minting fresh row ids. Used when a thread inherits its root scratchpad's
-   * SSK: the thread reuses the same key, so it carries the same per-recipient
-   * wraps (owner + enclave + any bots) under its own stream id. Idempotent via
-   * the slot conflict, so a re-run (or a racing thread create) is a no-op.
-   * Wrap counts are tiny (a handful of recipients), so a list + batch insert is
-   * fine — no hot path.
-   */
-  async copyToStream(
-    db: Querier,
-    params: { workspaceId: string; fromStreamId: string; toStreamId: string }
-  ): Promise<void> {
-    const wraps = await this.listForStream(db, params.workspaceId, params.fromStreamId)
-    await this.insertMany(
-      db,
-      wraps.map((w) => ({
-        workspaceId: params.workspaceId,
-        streamId: params.toStreamId,
-        keyGeneration: w.keyGeneration,
-        recipientKeyId: w.recipientKeyId,
-        recipientKind: w.recipientKind,
-        wrapEnc: w.wrapEnc,
-        wrapCt: w.wrapCt,
-      }))
-    )
-  },
-
-  /**
    * All wraps for a stream, across recipients and generations. Wrap bytes are
    * HPKE ciphertext (decryptable only by the matching private key), so it is
    * safe to return the full set to any stream member; the caller selects its
--- a/apps/backend/src/features/enclave-runtimes/dispatch/enclave-invoke-worker.test.ts
+++ b/apps/backend/src/features/enclave-runtimes/dispatch/enclave-invoke-worker.test.ts
@@ -124,6 +124,44 @@ describe("createEnclaveInvokeWorker", () => {
     expect(assignSession).toHaveBeenCalledTimes(1)
   })
 
+  it("resolves the SSK + wraps from the root for a thread message (reply still lands in the thread)", async () => {
+    arrangeDispatch()
+    // The trigger lives in a THREAD whose root is stream_root; the thread shares
+    // the root's SSK and carries no wraps of its own.
+    spyOn(StreamRepository, "findById").mockImplementation((async (_p: unknown, id: string) =>
+      id === "stream_1"
+        ? { id: "stream_1", workspaceId: "ws_1", rootStreamId: "stream_root" }
+        : { id: "stream_root", workspaceId: "ws_1", rootStreamId: null }) as never)
+    const getE2e = spyOn(E2eStreamsRepository, "getByStreamId").mockResolvedValue({ ...E2E, streamId: "stream_root" })
+    const listWraps = spyOn(StreamE2eKeyWrapsRepository, "listForStream").mockResolvedValue([WRAP])
+    const insertSession = spyOn(AgentSessionRepository, "insertRunningOrSkip").mockResolvedValue({
+      id: "session_1",
+      createdAt: new Date(),
+    } as never)
+    spyOn(StreamEventRepository, "insert").mockResolvedValue({ id: "evt_1" } as never)
+    spyOn(OutboxRepository, "insert").mockResolvedValue(undefined as never)
+    const assignSession = mock(async (_instanceUrl: string, _assignment: unknown) => {})
+    const { io } = fakeIo()
+
+    const worker = createEnclaveInvokeWorker({
+      pool,
+      io,
+      enclaveForwarder: { assignSession } as unknown as EnclaveForwarder,
+      storage: FAKE_STORAGE,
+    })
+    await worker(JOB)
+
+    // Key material (e2e row + wraps) is fetched against the root...
+    expect(getE2e).toHaveBeenCalledWith(pool, "ws_1", "stream_root")
+    expect(listWraps).toHaveBeenCalledWith(pool, "ws_1", "stream_root")
+    // ...the assignment carries the root id, so the enclave unwraps under the
+    // AAD the wraps were sealed with...
+    const assignment = assignSession.mock.calls[0]![1] as { streamId: string }
+    expect(assignment.streamId).toBe("stream_root")
+    // ...but the session (and therefore Ariadne's reply) lands in the thread.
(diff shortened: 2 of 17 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (diff is clean; non-blocking notes below). **But see the out-of-diff blocker — this PR alone does not make encrypted thread messaging work end-to-end.**
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3

The diff itself is sound: it correctly seals draft-thread first replies under the encrypted root's SSK via the shared `sealOutgoingMessage` helper, adds `contentJson` to the optimistic event so the sender's post-promotion decrypt-cache seed works, fails closed when `parentStream` is unresolved, and preserves the live-send path through the refactor. 3 minor, non-blocking items.

---

### ⚠️ Out-of-diff blocker (inherited from base #793 — not introduced here, but it gates this PR's goal)

#793's `StreamE2eKeyWrapsRepository.copyToStream` copies each SSK wrap's bytes **verbatim** from the root onto the thread's rows. But an SSK wrap is HPKE-sealed with `buildWrapAad({ streamId, keyGeneration, recipientKeyId })` — bound to the **root's** `streamId`. Every reader resolves the thread's SSK using the **thread's** `streamId` in the wrap AAD (frontend [`fetchAndUnwrap`](https://github.com/threahq/threa/blob/43c715d7fc1eb78475e502f48d83e3d99f2a80bd/apps/frontend/src/lib/crypto/stream-key-cache.ts#L326-L331); enclave [`unwrap`](https://github.com/threahq/threa/blob/43c715d7fc1eb78475e502f48d83e3d99f2a80bd/apps/enclave/src/agent/run-turn.ts#L324-L331)). AAD mismatch ⇒ HPKE open throws ⇒ the thread SSK is unresolvable.

The project's own crypto suite already asserts exactly this relocation throws: [`stream-key.test.ts#L213-L220`](https://github.com/threahq/threa/blob/43c715d7fc1eb78475e502f48d83e3d99f2a80bd/packages/crypto/src/__tests__/stream-key.test.ts#L213-L220) ("rejects a wrap relocated to a different stream").

Consequence: with this PR, sending now succeeds, but **the enclave can't decrypt the user's thread message (no Ariadne reply), and the owner can't read thread history after a reload** — only the live `seedDecryption` cache masks it for the sender in-session. The server can't fix this (it never holds the SSK plaintext); the SSK must be re-wrapped to each recipient under the **thread's** id client-side, or thread SSK resolution must target the root's wraps. Filing/handling separately — flagged here because it determines whether this PR meets its stated purpose.

---

Found 3 non-blocking issues in the diff:

1. `apps/frontend/src/hooks/use-queue-draft-message.test.ts:134-156` — the new test asserts the plaintext path leaves `ciphertext`/`envelope`/`e2eVersion` unset, but nothing asserts the optimistic event payload carries `contentJson` — the regression this PR explicitly fixes ("flash the opaque placeholder for its own author"). Worth one `mockEventsAdd` assertion. (plan/test-coverage)
   https://github.com/threahq/threa/blob/43c715d7fc1eb78475e502f48d83e3d99f2a80bd/apps/frontend/src/hooks/use-queue-draft-message.test.ts#L134-L156
2. `apps/frontend/src/lib/crypto/seal-send.ts:24-30` — `SealOutgoingMessageResult.owner` returns a raw `CryptoKey` purely so callers can pass it into `reviveStaleActorWraps`. Minor seam concern: the helper's public return type now advertises key material. Optional — internalizing the revive (or a callback) would keep keys off the result type. (design)
   https://github.com/threahq/threa/blob/43c715d7fc1eb78475e502f48d83e3d99f2a80bd/apps/frontend/src/lib/crypto/seal-send.ts#L24-L30
3. `apps/frontend/src/components/thread/stream-panel.tsx:365-369` — `hasActors` is derived from `parentStream.e2eActors`, which is only persisted to IDB via the actor-invite path (not bootstrap sync), so it can be a false-negative and silently skip the best-effort heal. Same gating the live send path already uses, and the heal is best-effort (self-heals on the next reply), so low impact. (design)
   https://github.com/threahq/threa/blob/43c715d7fc1eb78475e502f48d83e3d99f2a80bd/apps/frontend/src/components/thread/stream-panel.tsx#L365-L369

---
<details><summary>📐 Plan Adherence [1 issue]</summary>Diff implements the stated send-path fix fully. Gap: no test asserts the `contentJson`-on-optimistic-event behavior the PR calls out (#1 above). Separately, the PR's end-to-end goal is blocked by the inherited wrap-AAD issue (see callout).</details>
<details><summary>🔍 Bugs [CLEAN in-diff]</summary>No bugs in the changed lines. AAD-vs-wrap-AAD asymmetry is a base (#793) characteristic, not introduced here — see callout.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No invariant violations. Earlier INV-23/24 test-assertion nits were already addressed; mocking uses scoped `spyOn` (INV-48); shared `sealOutgoingMessage` satisfies INV-35.</details>
<details><summary>🏗️ Design [2 notes]</summary>`owner` return shape (#2) and `hasActors` IDB reliability (#3). The `sealOutgoingMessage` unification is a clear improvement over the prior duplication.</details>
<details><summary>🔒 Security [CLEAN in-diff]</summary>No high-confidence vulnerabilities in the diff. The composer now reads `e2eEnabled` off the parent so attachments encrypt before upload (closes a plaintext-attachment leak). Note: the out-of-diff wrap-AAD issue is a correctness/availability problem (undecryptable), not a confidentiality leak.</details>

PR #799, opened 2026-06-05
Retry Giphy GIF loads on transient CDN errors

kristofferremback:
## Problem

A Giphy GIF on a sent message could fail to load after a transient CDN/network blip and stay broken until a full page refresh (see the bottom message in the reported screenshot — a broken-image placeholder while other GIFs rendered fine).

The shared renderer `apps/frontend/src/components/giphy/giphy-image.tsx` rendered a bare ``&lt;img src={url} loading="lazy"&gt;`` with no `onError` handling. The browser does not retry a failed image load on its own, so a single failed request left that GIF permanently broken until a fresh page load created a new `` element.

## Fix

`GiphyImage` is the single component both the composer node view and the timeline preview use, so the fix covers every GIF surface. It now:

- **Retries on error with exponential backoff** — up to 5 attempts (500ms → 8s).
- **Cache-busts each retry** (`_threaRetry=N` query param) so the browser issues a genuinely fresh request instead of reusing the failed/negatively-cached response. The first load still uses the pristine Giphy URL, so the normal path is unchanged.
- **Recovers when connectivity returns** — listens for the `online` event and re-attempts an exhausted GIF.
- **Falls back to a "tap to retry" affordance** only after automatic retries are exhausted, instead of silently showing a broken image.

This follows the existing `onError` fallback pattern in `link-preview-card.tsx` / `attachment-list.tsx`, but goes further since GIFs are content we can't afford to drop — it actively recovers rather than just swapping in a placeholder.

## Tests

Added `giphy-image.test.tsx` (4 tests):
- First load uses the clean URL (no cache-bust param).
- A transient error triggers a cache-busted retry after backoff.
- Exhausting retries shows the retry button, which recovers the image on tap.
- Changing the GIF URL resets retry state.

All giphy tests pass (11) and `tsc --noEmit` on the frontend is clean.

## Note / follow-up

This is the client-side render path; the GIF URL is stored inline pointing at `media.giphy.com`, so reliability still ultimately depends on Giphy's CDN being reachable. Full independence from Giphy uptime would mean proxying/caching the GIF bytes through our own storage at send time — out of scope here, can be scoped separately if it matters.

https://claude.ai/code/session_01UbZGxUHofZJMRBuHeDmghw

---
_Generated by [Claude Code](https://claude.ai/code/session_01UbZGxUHofZJMRBuHeDmghw)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/799"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783286851&installation_id=129946197&pr_number=799&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F799&signature=74d198fa090f08eb9945623ba4c97c071a769078be16c4fd937668ca9a99de35"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/giphy/giphy-image.test.tsx
+++ b/apps/frontend/src/components/giphy/giphy-image.test.tsx
@@ -5,9 +5,19 @@ import { GiphyImage } from "./giphy-image"
 const GIF_URL = "https://media.giphy.com/media/abc123/giphy.gif"
 
 function getImg(): HTMLImageElement {
-  const img = document.querySelector('img[data-type], [data-type="giphy-embed"] img') as HTMLImageElement | null
+  const img = document.querySelector('[data-type="giphy-embed"] img') as HTMLImageElement | null
   if (!img) throw new Error("GIF image not rendered")
   return img
+}
+
+/** Drive every automatic retry until the manual-retry fallback appears. */
+function exhaustRetries() {
+  for (let i = 0; i <= 5; i++) {
+    act(() => {
+      fireEvent.error(getImg())
+      vi.advanceTimersByTime(500 * 2 ** i)
+    })
+  }
 }
 
 describe("GiphyImage — load reliability", () => {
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Self-review (multi-perspective)
**Review models:** Orchestrator: claude-opus-4-8[1m] | Reviewers: sonnet x3 (spec compliance, correctness, design)

Ran a 3-perspective review on the diff. Two findings surfaced (both ≥80), both fixed in `9eaa607`:

1. `apps/frontend/src/components/giphy/giphy-image.tsx` — **INV-14**: the failed-state "tap to retry" affordance was a raw `<button>` instead of the Shadcn `Button` primitive (missing focus-visible rings / variant handling). → switched to `<Button variant="outline" size="sm">`.
2. `apps/frontend/src/components/giphy/giphy-image.tsx` — **bug**: a single counter was used for both the backoff cap and the cache-bust token, so tapping retry / coming back online bumped it *past* `MAX_RETRIES` and gave one instant re-fail with no backoff. → split into a monotonic `reloadToken` (drives `key` + cache-bust) and a per-cycle `retries` count that resets on manual/online recovery, so each starts a fresh backoff cycle. Added a regression test.

<details><summary>📐 Plan Adherence — CLEAN</summary>Diff matches the stated purpose: backoff retry + cache-bust + online recovery + manual fallback. No missing corresponding changes.</details>
<details><summary>🔍 Bugs — 1 fixed</summary>Retry-cycle reset bug (finding #2) fixed in 9eaa607.</details>
<details><summary>📋 CLAUDE.md — 1 fixed</summary>INV-14 raw button (finding #1) fixed in 9eaa607. INV-39 (mount real component, observable behavior), INV-25 (comments explain why), INV-18 — all clean.</details>
<details><summary>🏗️ Design — CLEAN</summary>Retry logic appropriately localized in the shared `GiphyImage`; no premature shared abstraction warranted (link-preview / attachment use static fallbacks; active retry is justified for the external Giphy CDN). `online` listener cleaned up correctly.</details>
<details><summary>🔒 Security — CLEAN</summary>URL is host-validated upstream; `new URL().searchParams.set` is injection-safe; React JSX is XSS-safe here (no `dangerouslySetInnerHTML`).</details>

Frontend tests pass (giphy suite 12/12, `GiphyImage` 5/5) and `tsc --noEmit` is clean.

---

PR #820, opened 2026-06-10
fix(sync): close the reconnect catch-up gap that dropped live messages

kristofferremback:
## Problem

A message sent while a client was briefly disconnected (deploy bounce, phone sleep) could stay missing from the timeline until a full reload, while newer messages kept arriving fine.

Verified against prod for the reported incident (DM `stream_01KJMS776MNP2Q382MJ639Y2JD`):

- seq **3501** created 18:55:11Z while the viewer's device was suspended (its `stream:read` events stop at 18:53:31 and resume 18:55:36) → broadcast missed
- seq **3502** arrived live at 18:55:34, exactly mid-resume
- 3501 never backfilled — visible in the screenshots that triggered this investigation

## Root cause

`SyncEngine.bootstrapWorkspace`'s reconnect path re-joined stream rooms (re-registering socket handlers) **before** reading the per-stream catch-up cursor from `getLatestPersistedSequence`. A live message landing in that window (3502) is written to IndexedDB and advances the cursor past the disconnect gap, so the `bootstrap?after=3502` fetch returns nothing and 3501 is permanently skipped. Since stream bootstrap queries run with `staleTime: Infinity` and all refetch triggers disabled, nothing ever heals the hole short of a reload. `performStreamRefresh` had the same ordering flaw.

INV-53's subscribe-then-fetch makes *overlap* safe (writes dedupe by event id) — but this ordering created *gaps*, which nothing tolerated.

## Fix (two layers)

1. **Root cause:** read each visible stream's catch-up cursor before re-joining its room, in both the reconnect path and `performStreamRefresh`. The regression test fails against the old ordering with the exact production symptom (`after: "3"` instead of `after: "1"`).
2. **Safety net:** tail sequence-gap detection on the live socket write path. When an incoming event's sequence skips past the latest persisted one, the handler reports the pre-write tail via a new `onSequenceGap` callback and `SyncEngine.backfillStreamGap` runs a single-flighted `bootstrap?after=` fetch. This heals *any* missed-event scenario (zombie socket, server bounce, failed catch-up) instead of leaving the hole until the next full-replace bootstrap. A distinct gap reported mid-flight is queued and drained into one follow-up fetch; duplicate reports of the same gap (both handler registration sites observe it) collapse onto the in-flight fetch.

Note: other users' command events legitimately occupy sequence slots that are never delivered to this viewer, so a reported gap can come back empty — the backfill is idempotent and self-quiets because the next live event sits contiguously on the new tail.

## Design decisions

- `backfillStreamGap` deliberately does not coordinate with `activeStreamRefreshes`: concurrent applies are idempotent by design (append-mode merges, writes dedupe by id), so the worst case of an overlapping navigation refresh is one redundant fetch — not worth coupling the two flight guards.
- Gap detection runs inside the same Dexie transaction as the dedupe + put in both `handleMessageCreated` and `handleAppendEvent`, so a concurrent write can't make the reported cursor lie; the backfill itself fires only after commit.

## Testing

- Gap detection: contiguous events, cold cache, older-than-tail events, pending optimistic placeholder rows, duplicate redelivery, append-style (membership/session) events
- Pre-join cursor regression tests for both the reconnect path and navigation refresh (verified to fail with the old ordering)
- Backfill single-flighting + queued mid-flight cursor drain
- Full frontend suite: 2338 tests green; monorepo typecheck + lint green via pre-commit hooks

https://claude.ai/code/session_01A7PKnoGy94kset59mDPyYw

---
_Generated by [Claude Code](https://claude.ai/code/session_01A7PKnoGy94kset59mDPyYw)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/820"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783713556&installation_id=129946197&pr_number=820&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F820&signature=1a658ccd69df36b9062d8e1e6fded1e8bb5500b2a0c197da1fbfe3626e1fc01d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/hooks/use-stream-socket.ts
+++ b/apps/frontend/src/hooks/use-stream-socket.ts
@@ -3,6 +3,7 @@ import { useQueryClient } from "@tanstack/react-query"
 import { useSocket } from "@/contexts"
 import { joinRoomFireAndForget } from "@/lib/socket-room"
 import { registerStreamSocketHandlers } from "@/sync/stream-sync"
+import { useOptionalSyncEngine } from "@/sync/sync-engine"
 
 /**
  * Hook to handle real-time message/reaction events for a specific stream.
@@ -16,6 +17,8 @@ export function useStreamSocket(workspaceId: string, streamId: string, options?:
   const shouldSubscribe = options?.enabled ?? true
   const queryClient = useQueryClient()
   const socket = useSocket()
+  // Optional: draft panels can mount outside the workspace SyncEngine provider.
+  const syncEngine = useOptionalSyncEngine()
 
   useEffect(() => {
     if (!socket || !workspaceId || !streamId || !shouldSubscribe) return
@@ -29,7 +32,14 @@ export function useStreamSocket(workspaceId: string, streamId: string, options?:
     // Register all stream-level socket handlers — they write to IDB only.
     // queryClient is passed for transitional workspace bootstrap preview updates
     // (will be removed in Phase 3).
-    const cleanupHandlers = registerStreamSocketHandlers(socket, workspaceId, streamId, queryClient)
+    const cleanupHandlers = registerStreamSocketHandlers(socket, workspaceId, streamId, queryClient, {
+      // A live event that skips past the cached tail means events were missed
+      // (zombie socket, server bounce). Route to the engine's single-flighted
+      // backfill so the hole is fetched instead of persisting until reload.
+      onSequenceGap: syncEngine
+        ? ({ streamId: gapStreamId, afterSequence }) => void syncEngine.backfillStreamGap(gapStreamId, afterSequence)
+        : undefined,
+    })
 
     return () => {
       abortController.abort()
@@ -38,5 +48,5 @@ export function useStreamSocket(workspaceId: string, streamId: string, options?:
       // a single leave undoes ALL joins. The SyncEngine also joins this room
       // for stream:activity delivery — leaving here would break sidebar updates.
     }
-  }, [socket, workspaceId, streamId, shouldSubscribe, queryClient])
+  }, [socket, workspaceId, streamId, shouldSubscribe, queryClient, syncEngine])
 }
--- a/apps/frontend/src/sync/stream-sync.test.ts
+++ b/apps/frontend/src/sync/stream-sync.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach } from "vitest"
+import { describe, it, expect, beforeEach, vi } from "vitest"
 import { QueryClient } from "@tanstack/react-query"
 import type { Socket } from "socket.io-client"
 import { db } from "@/db"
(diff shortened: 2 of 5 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: sonnet x3

Pre-PR self-review of the branch diff. 4 findings scored ≥80; 2 were fixed in `51991d7` before this PR was opened, 1 was declined with rationale, 1 was clean.

**Fixed before opening:**

1. `stream-sync.ts` — `handleAppendEvent` ran dedupe → gap-read → put as three separate awaits with no enclosing Dexie transaction, unlike `handleMessageCreated`. A concurrent append landing between the gap read and the write could make the reported cursor lie. Now wrapped in `db.transaction("rw", [db.events], ...)`. (design, 85)
   https://github.com/threahq/threa/blob/51991d760035149c734041ac91e969540d0d10cd/apps/frontend/src/sync/stream-sync.ts#L805-L832
2. `sync-engine.ts` — `backfillStreamGap` single-flighted on `streamId` only, so a *distinct* gap reported mid-flight was silently dropped if the in-flight fetch didn't happen to cover it. Mid-flight reports with a different cursor are now queued (lowest wins) and drained into one follow-up fetch; duplicate reports of the same gap still collapse. (bug, 82)
   https://github.com/threahq/threa/blob/51991d760035149c734041ac91e969540d0d10cd/apps/frontend/src/sync/sync-engine.ts#L240-L300

**Declined:**

3. `backfillStreamGap` doesn't coordinate with `activeStreamRefreshes` (design, 80) — concurrent applies are idempotent by design (INV-53 overlap-safety: append merges, writes dedupe by event id), so an overlapping navigation refresh costs at most one redundant fetch. Coupling the two flight guards adds complexity without a correctness payoff. Documented in the PR description.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>All four claimed fix pieces present and tested; no plan file by design (intentionally not committed).</details>
<details><summary>🔍 Bugs [1 issue — fixed]</summary>Mid-flight gap report dropped by single-flight (see #2 above).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No INV violations introduced; gap callback fires post-commit (no network inside Dexie transactions), no hidden singletons, INV-53 is the invariant being repaired.</details>
<details><summary>🏗️ Design [2 concerns — 1 fixed, 1 declined]</summary>Transaction-shape inconsistency in handleAppendEvent (fixed); flight-guard coordination (declined, see above).</details>

---

Issue #849, opened 2026-06-11
Conversation overlay follow-ups (from #842 self-review)


kristofferremback:
Items deliberately left out of #842 (conversation overlay + corrections), noted in the merge-time self-review. None were blocking; only the first has a data footprint.

- [ ] **Empty conversations after corrections.** Moving a conversation's only message out leaves an empty conversation. It disappears from the in-view panel (no rows on screen) but stays in the conversations list and the correction picker — which conveniently makes it an undo target. If corrections get heavy use, decide on cleanup/archival semantics (e.g. archive on empty, or let staleness age it out).
- [ ] **Pending spinner tracks only the latest in-flight correction.** `pendingMessageId` comes from the single TanStack mutation's `variables`; two rapid corrections on different messages show the pending affordance only on the second. Cosmetic — consider a per-message pending set.
- [ ] **History asymmetry on overlay toggle.** The header split button pushes a history entry (`setSearchParams` default) while the panel's X uses `replace: true`. Back-button behavior differs depending on how the overlay was closed; pick one.
- [ ] **Split-button chevron is a 20px touch target.** `h-8 w-5` works with a mouse but is fiddly on touch. Consider widening on mobile or moving the conversations-list entry elsewhere for touch.
- [ ] **Narrow desktop (~640–900px): panel can cover top rows' topic chips.** The top-right panel overlaps the centered column's chip corner at these widths. Solid panel background keeps it clean and it resolves on scroll/collapse, but a smarter anchor (or hiding chips under the panel) would polish it.

Larger follow-ups already listed in #842's description: conversation merge/split corrections, secondary-membership visualization, and the pre-existing live-ordering bug (two rapid sends render newest-first until reload — reproduced on `main`).

PR #852, opened 2026-06-11
fix(conversations): resolve a conversation emptied by a correction, reactivate on undo

kristofferremback:
First of a five-PR stack for the conversation-overlay follow-ups in #849. Closes item 1 of #849.

## Problem

`ConversationService.reassignMessage` could leave the previous conversation with `message_ids = {}` when a correction moved its only message out. The empty conversation lingered as `active` forever — still surfacing in the conversations list and the correction picker.

## Change

- **`ConversationRepository.resolveIfEmpty`**: sets `status = 'resolved'` when the conversation owns no primary messages. The emptiness check lives inside the `UPDATE` (INV-20), so a concurrent extraction appending to `message_ids` makes the condition fail under the row lock instead of resolving a conversation that just gained a message.
- **`ConversationRepository.reactivateIfResolved`**: flips a resolved conversation back to `active`; also conditional in SQL, no-op for active/stalled.
- **`reassignMessage`** calls `resolveIfEmpty` on the previous conversation right after `removePrimaryMessage`, and `reactivateIfResolved` on the target right after `addPrimaryMessage` — the undo path (moving the message back) reactivates the emptied conversation. Both run inside the existing `withTransaction` block and before the final `findByIds` re-read, so the `conversation:updated` outbox payloads carry the new status (INV-4, INV-7).

Statuses use `ConversationStatuses` from `@threa/types` (`active | stalled | resolved` — no separate "archived" state; INV-33).

## Tests

Extended `apps/backend/tests/integration/conversation-reassign.test.ts`:
- moving the only message out → previous conversation returned with `status: 'resolved'`, DB row resolved, and the `conversation:updated` outbox payload reflects it
- moving the message back into the resolved-empty conversation → back to `'active'`

Ran: backend integration (480 pass / 0 fail), backend unit (1677 pass / 0 fail), repo-wide `bun run typecheck` — all green.

https://claude.ai/code/session_017CdLNS7gX8Za142XJruk4x

---
_Generated by [Claude Code](https://claude.ai/code/session_017CdLNS7gX8Za142XJruk4x)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/852"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783804138&installation_id=129946197&pr_number=852&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F852&signature=d573ea19c998d2aef8c2a2965285cc4dd7845b1f26139a5214e244f12a982535"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/conversations/repository.ts
+++ b/apps/backend/src/features/conversations/repository.ts
@@ -1,5 +1,5 @@
 import { sql, type Querier } from "../../db"
-import type { ConversationStatus } from "@threa/types"
+import { ConversationStatuses, type ConversationStatus } from "@threa/types"
 
 interface ConversationRow {
   id: string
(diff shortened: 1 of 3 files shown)


kristofferremback:
@coderabbitai review

---

kristofferremback:
Thanks — addressed in df14175. Point-by-point:

**1. `resolveIfEmpty` and `stalled` — intended behavior, no status predicate added.** `stalled` is only ever assigned by the boundary-extraction model (`boundary-extraction/config.ts`); there is no separate human-decision lifecycle behind it. An emptied conversation has no content to stall on, and `idx_conversations_stalled` treats `active`+`stalled` as the "unresolved" set for notifications — leaving an empty conversation `stalled` would keep it surfacing there, which is exactly the lingering-empty-shell problem this PR removes. Resolving any emptied conversation regardless of prior status is deliberate.

**2. Unconditional `reactivateIfResolved` — intentional, docs updated.** A user moving a message into *any* resolved conversation is declaring it has activity again, so it returns to the active lifecycle; this isn't undo-specific. The inline comment in `service.ts` and the `repository.ts` JSDoc now say so explicitly.

**3. Undo-test coverage — assertion added, but note the premise:** after the first move conversation A owns `[msg1, msg2, msg0]`, so the undo leaves it `[msg1, msg2]` — *not* empty. The valuable check is the negative one, which the test now asserts: resolve-on-empty must NOT fire for a previous conversation that still has messages (`previousConversation` comes back `active` with `[msg1, msg2]`).

**4. Row counts as return values:** skipping per INV-36 — no current caller needs them, and adding plumbing for a hypothetical metric is speculative.

Backend integration (480) + unit (1677) suites and repo typecheck re-run green after the changes.

---

PR #887, opened 2026-06-12
feat: discoverable add-filter menu on both search surfaces

kristofferremback:
## Problem

The search split in #883 carried over the typed filter autocomplete (`from:@…`, `in:#…`) and the removable chips, but dropped the quick switcher's discoverable filter UI (the "Add filter" dropdown + value pickers from `use-search-items.tsx`). There is now no way to find out *what* you can query for without already knowing the syntax — and on mobile, typing query operators on a phone keyboard is impractical even when you do know it.

## Solution

A new `SearchFilterMenu` component renders an always-visible **"Add filter"** trigger on both search surfaces, opening a two-step kind → value flow that commits by rewriting the query string via the existing `addFilterToQuery`. The query string stays the single source of truth: a filter added through the menu is identical to one typed by hand and renders as the same removable chip.

### How it works

- **Desktop sidebar panel:** a compact pill under the search input opens a popover.
- **Mobile `/search` page:** the same trigger opens a vaul bottom drawer (same responsive split as `SearchableSelect`).
- **Step 1 — filter kind:** From user, With user, In channel, In DM with, Stream type, Status, After/Before date. Each entry shows its typed syntax (`from:@user`, `in:#channel`, `is:thread`, …) as a hint, so the menu doubles as documentation of the query language. The kind list is itself searchable.
- **Step 2 — value:** user picker (with the `me` shortcut), channel picker (labels via `streamLabel`), fixed option lists with descriptions, and date presets (Today, Yesterday, Last week, …) plus a calendar for arbitrary dates (with a back button to the presets).

### Key design decisions

**1. Query string remains the single source of truth.** The menu never holds filter state; it appends `from:@slug ` etc. to the query and closes. Chips, parsing, slug→id resolution, and debounce all flow through the existing `useMessageSearch` path unchanged.

**2. `in:` split into "In channel" and "In DM with".** The syntax distinguishes `in:#slug` from `in:@slug` and the two need different value pickers, so the menu mirrors that rather than inventing a combined picker.

**3. User pickers are users-only.** `useMessageSearch` resolves `from:`/`with:`/`in:@` slugs against workspace users exclusively — offering personas/bots would commit a chip whose filter silently never applies (self-review finding).

**4. Value sources reused, not duplicated.** Stream-type and status options import from the typed autocomplete extensions (`FILTER_TYPE_OPTIONS`, `STATUS_FILTER_OPTIONS`), date presets from `lib/dates`, channel labels from `lib/streams` — both input paths stay in lockstep (INV-33, INV-35).

**5. Keyboard isolation.** Keystrokes inside the menu stop propagating so arrows/Escape drive the menu, not the panel's result navigation; pickers without a text input focus the cmdk root on mount so arrow/Enter navigation works.

### Bugs fixed along the way (manual testing on staging)

- **`RichInput` collapsed trailing whitespace** when syncing external values: the menu commits `status:active ` with a trailing space to park the cursor outside the filter token, but the HTML parse dropped it, so the typed suggestion popover re-opened on focus. Fixed with `preserveWhitespace: "full"` on both parse paths.
- **The search input didn't grow with wrapped queries** (fixed `h-8` → `h-auto min-h-8` on both surfaces).
- **Date filters were broken end-to-end (prod too):** the API validates `before:`/`after:` with `z.string().datetime()` but the syntax carries bare `YYYY-MM-DD`, so every date filter failed validation. `useMessageSearch` now widens date-only values to the ISO instant of local midnight (`localStartOfDayISO` in `lib/dates`), covering the typed path and the menu alike.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/search/search-filter-menu.tsx` | The add-filter trigger + popover/drawer with kind list and value pickers |

## Modified files

| File | Change |
| ---- | ------- |
| `apps/frontend/src/components/search/sidebar-search-panel.tsx` | Always-visible chips + menu row under the input; input grows with content |
| `apps/frontend/src/pages/search.tsx` | Same row + growth on the mobile/full-page surface |
| `apps/frontend/src/components/search/search-filter-chips.tsx` | Root is `display: contents` so chips and trigger wrap in one caller-owned row |
| `apps/frontend/src/components/search/use-message-search.ts` | Date-only `before:`/`after:` values widened to local-midnight ISO datetimes |
| `apps/frontend/src/components/quick-switcher/rich-input.tsx` | Preserve trailing whitespace when parsing/syncing content |
| `apps/frontend/src/lib/dates.ts` | New `localStartOfDayISO` helper (+ unit tests) |
| `apps/frontend/src/components/search/index.ts` | Export `SearchFilterMenu` |
| `apps/frontend/src/components/search/sidebar-search-panel.integration.test.tsx` | 9 new tests for the menu and the date-filter API contract |

## Test plan

- [x] 9 new integration tests: kind list contents/syntax hints, from/DM/channel/status/date-preset flows, back navigation, keyboard isolation, trailing-space round-trip, date widening in the API payload
- [x] 3 unit tests for `localStartOfDayISO`
- [x] Full frontend suite green; monorepo typecheck + lint clean
- [x] Self-review (3-perspective): 3 findings ≥80 fixed in `3af1ec8`; CodeRabbit review addressed in `7a03ab4` (4 fixed, 2 disputed and withdrawn)
- [x] Manual verification on staging (caught the trailing-space, input-growth, and date-filter bugs above)

<details>

**Goal:** Restore the discoverable search filter options lost in the #883 search split, working well on both desktop and mobile.

**What Was Built:**
- `SearchFilterMenu`: responsive (Popover desktop / vaul Drawer mobile) two-step filter builder rendered next to the existing filter chips on both search surfaces, always visible.
- Kind list teaches the typed syntax per entry; value pickers reuse the same data sources as the typed autocomplete triggers (mentionables, slugged streams, canonical option lists, date presets + calendar).
- Commit path: `addFilterToQuery(query, type, value) + " "` → existing parse/resolve/debounce pipeline; chip removal unchanged.

**Design Evolution:**
- Initial version duplicated stream-type/status option lists and offered personas/bots in `from:`/`with:`; the self-review pass (3 parallel reviewers, ≥80 score threshold) caught the silent-filter bug and the INV-33 duplication (`3af1ec8`).
- CodeRabbit review added the calendar back button, the picker truncation hint, the `in:@user` test, and computed example years (`7a03ab4`); its `before:`-preset and divider concerns were disputed (typed-path parity / always-populated row) and withdrawn.
- Manual staging testing surfaced three real bugs fixed in follow-ups: trailing-whitespace collapse in `RichInput` (`1d8955d`), fixed-height input clipping wrapped queries (`49dff60`), and date-only filter values failing API datetime validation — a pre-existing prod bug (`965e1c1`).

**What's NOT Included:** Fixing the same persona/bot silent-drop in the *typed* `from:@`/`with:@` autocomplete (pre-existing); search-page-specific test file (surface shares all components with the tested sidebar panel).

**Status:** Complete; all suites green; CI green on every commit.

</details>

---


kristofferremback:
@coderabbitai review

---

kristofferremback:
@coderabbitai Thanks — addressed in 7a03ab4. Disposition per finding:

**1. Calendar back navigation — fixed.** The calendar view now has a "Date presets" back button (`ChevronLeft`) that returns to the preset list without losing the kind context.

**2. `before:` future presets — intentional, keeping as-is.** The menu deliberately reuses the exact preset set the typed `before:` autocomplete shows (`getDateFilterOptions` → `getFutureDatePresets`, which already mixes in Today/Yesterday/Last week as past-cap options). Forking the menu's presets would drift the two input paths apart (INV-43: shared behavior on one path). If the preset semantics should change, that's a change to `getFutureDatePresets` in `lib/dates` affecting both paths — happy to do that as a follow-up if desired.

**3. Always-rendered `border-t` — intentional, keeping as-is.** The row beneath it is no longer conditionally rendered: it always contains the "Add filter" trigger (the discoverability point of this PR), so the divider separates populated rows rather than floating over empty space. A conditional border would toggle a 1px layout shift the moment the first chip appears (INV-21 spirit).

**4. Silent 20-result cap — fixed.** Both searchable pickers now render a "Showing first 20 — keep typing to narrow" hint row whenever results are truncated (`PICKER_RESULT_CAP`).

**5. Missing `in:@user` test — fixed.** Added an integration test covering In DM with → user pick → `in:@martin` in the query and `{ in: ["member_1"] }` in the API call.

**6. Hardcoded year — fixed.** Syntax hints compute the example date from the current year.

---

PR #908, opened 2026-06-13
feat: sweep legacy unread-counter entries from sync_log

kristofferremback:
## What

A one-time, content-based data migration that deletes the legacy unread-counter
entries from `sync_log` — the rows the client guard
`isLegacyUnreadCounterEntry` (`apps/frontend/src/sync/unread-counters.ts`)
currently skips on catch-up.

This is **PR 1 of 2**. It unblocks deleting the guard itself, which follows as a
separate PR once this sweep has deployed to prod.

## Why

The four unread-counter events (`stream:activity`, `stream:read`,
`stream:read_all`, `activity:created`) carry absolute fields since the phase-2c
backend deploy (#872). Entries written before it lack those fields; catch-up
replay of them is unsafe, so the client guard skips them and leans on the
workspace bootstrap as their authority (INV-53). The guard can only die once no
legacy entry can reach catch-up.

**Pure time+count retention (#907) never reliably gets us there:** a quiet
workspace with fewer than `minKeep` (2,000) lifetime sync events never trips the
count floor (`head − minKeep` is negative), so its legacy rows would persist
*indefinitely* and the guard could never be deleted by waiting.

## How

`20260613083239_prune_legacy_counter_entries.sql` deletes exactly the rows the
guard skips, using the **same per-type absent-field predicate, mirrored in
SQL** — keyed on payload **content**, not a hardcoded deploy timestamp, so it
can never touch a valid post-#872 row.

## Design decisions

- **Content-based, not timestamp-based.** No hardcoded boundary; the predicate
  mirrors `isLegacyUnreadCounterEntry` exactly (`jsonb_typeof(...) IS DISTINCT
  FROM 'number'/'array'`), so it deletes precisely what the guard skips.
- **A migration, not a worker + marker table.** Run-once-then-inert is free from
  the migrator; no throwaway-table/second-deploy-removal machinery needed.
  Matches existing one-time data migrations (`backfill_system_streams`,
  `fix_persona_model`).
- **Does NOT advance `retained_from`.** The legacy set is *sparse* (counter rows
  interleaved among non-counter rows). Advancing the floor to the max deleted id
  would falsely mark surviving lower-id rows as pruned and force needless
  bootstraps. Retention's floor is for its contiguous time-prefix; a sparse
  removal owes no floor signal.

## Safety

The guard is **client-side** (`sync-engine.ts:1113`) and the workspace sync
cursor advances to `head` with **no per-sync_id contiguity check** (INV-61
contiguity is a separate per-stream `broadcastSequence` mechanism). So deleting a
legacy row server-side is byte-for-byte equivalent to the client skipping it:
surviving entries still apply, the cursor still climbs to `head`, and the deleted
row opens no gap. Idempotent; runs before the server accepts connections.

## Sequencing

PR 2 (deleting `isLegacyUnreadCounterEntry` + its sync-engine call sites) is safe
only **after this sweep deploys to prod** — the migration runs at backend startup
before serving, so once deployed every backend is legacy-free. Removing the guard
before then would let a pre-sweep backend feed field-less payloads to the
appliers.

## Testing

`apps/backend/tests/integration/prune-legacy-counter-entries.test.ts` runs the
**actual shipped migration SQL** (read from the file — no duplication) against
seeded legacy / survivor / non-counter rows, asserting only legacy rows are
deleted, survivors + non-counter rows remain, and `retained_from` is untouched.
Plus an idempotency case. (Live-DB integration suite is CI-only; verified here by
typecheck + reasoning.)

<details>

See `docs/plans/sync-v2-log-retention.md` ("What this unblocks") and
`docs/plans/sync-v2-healing-deletion-inventory.md` (ground rules), both updated
in this PR.

Approach: one-time content-based sweep migration removes the legacy entries the
guard skips (by content, because pure retention leaves a quiet workspace's legacy
rows in place forever), then a follow-up PR deletes the guard once the sweep has
deployed. No floor advance (sparse set). Tested against the real migration SQL in
the live-DB integration harness.

</details>

---
_Generated by [Claude Code](https://claude.ai/code/session_01J7rV7XXFhxzS7QAZMCaByJ)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/908"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783932268&installation_id=129946197&pr_number=908&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F908&signature=d6dfa8dd7d5d858b94144dacf61fe4de105f857d3a1e5635c210eaad582e0ab5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/db/migrations/20260613083239_prune_legacy_counter_entries.sql
+++ b/apps/backend/src/db/migrations/20260613083239_prune_legacy_counter_entries.sql
@@ -1,0 +1,43 @@
+-- One-time sweep of legacy unread-counter entries from sync_log.
+-- (sync engine v2; docs/plans/sync-v2-log-retention.md, "What this unblocks".)
+--
+-- The four unread-counter events (stream:activity, stream:read, stream:read_all,
+-- activity:created) carry ABSOLUTE fields since the phase-2c backend deploy
+-- (#872). Entries written before that deploy lack those fields; catch-up replay
+-- of them is unsafe, so the client guard `isLegacyUnreadCounterEntry`
+-- (apps/frontend/src/sync/unread-counters.ts) skips them on apply and leans on
+-- the workspace bootstrap as their authority (INV-53). That guard can only be
+-- deleted once no legacy entry can reach catch-up.
+--
+-- Pure time-based retention does not get us there reliably: a quiet workspace
+-- with fewer than minKeep (2,000) lifetime sync events never trips the count
+-- floor, so its legacy rows would persist indefinitely. This sweep removes them
+-- directly, by CONTENT rather than a hardcoded deploy timestamp: it deletes
+-- exactly the rows the guard already skips (the same per-type absent-field
+-- predicate, mirrored in SQL), so it can never touch a valid post-#872 row.
+--
+-- Safety: the guard is client-side and the workspace sync cursor advances to
+-- head with no per-sync_id contiguity check (contiguity is a separate
+-- per-stream broadcastSequence mechanism, INV-61). Deleting a legacy row is
+-- therefore identical to the client skipping it — surviving entries still apply
+-- and the cursor still climbs to head. The deleted set is SPARSE (legacy
+-- counter rows interleaved among non-counter rows), so `retained_from` is
+-- deliberately NOT advanced: advancing the floor past surviving lower-id rows
+-- would falsely mark them pruned and force needless bootstraps. A sparse
+-- removal of skipped-anyway rows opens no gap, so no floor signal is owed.
+--
+-- Idempotent: re-running finds no legacy rows. No FKs (INV-1); workspace-scoped
+-- by table (INV-8). Runs before the server accepts connections, so no outbox
+-- events are emitted.
+
+DELETE FROM sync_log
+WHERE
+  (event_type = 'stream:activity'
+    AND jsonb_typeof(payload -> 'messageOrdinal') IS DISTINCT FROM 'number')
+  OR (event_type = 'stream:read'
+    AND jsonb_typeof(payload -> 'lastReadOrdinal') IS DISTINCT FROM 'number')
+  OR (event_type = 'stream:read_all'
+    AND jsonb_typeof(payload -> 'reads') IS DISTINCT FROM 'array')
+  OR (event_type = 'activity:created'
+    AND (jsonb_typeof(payload -> 'counts' -> 'mentionCount') IS DISTINCT FROM 'number'
+      OR jsonb_typeof(payload -> 'counts' -> 'activityCount') IS DISTINCT FROM 'number'));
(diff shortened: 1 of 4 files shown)


kristofferremback:
### Code review
**Review models:** Reviewers: Sonnet ×3 (parallel, inline self-scoring ≥80)

No blocking issues found. Three reviewers (spec/CLAUDE.md compliance, correctness/security, design) independently cleared the diff.

Key things verified:
- **SQL predicate ≡ `isLegacyUnreadCounterEntry`** across all four branches, including the edge cases: missing `counts` (`NULL IS DISTINCT FROM 'number'` → TRUE, matches `typeof undefined !== "number"`), empty `reads: []` (kept as non-legacy on both sides), and `activity:created` two-field OR logic.
- **Safety of not advancing `retained_from`** re-derived from the client: `cursorStore.advance(entry.syncId)` runs for *every* entry (the guard only gates `gate.dispatch`), and the workspace cursor has no per-sync_id contiguity check (INV-61 contiguity is per-stream `broadcastSequence`), so a sparse server-side delete opens no gap.
- **INV-17** (new migration file, not an edit), **INV-20** (startup DELETE, no select-then-act), and the predicate-vs-guard duplication (forced — SQL can't call TS — and mitigated by the test running the real migration artifact).
- **Reusing `pruneExpiredEntries` would be wrong** here: it time-gates and floor-advances, both inappropriate for a sparse content sweep.

<details><summary>📐 Plan Adherence [CLEAN]</summary>Migration + integration test + both plan-doc updates all present; no missing corresponding change. Guard deletion correctly deferred to a follow-up PR sequenced after this deploys.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No legacy row missed, no non-legacy/non-counter row at risk. Test ids and assertions verified correct (outboxId→syncId mapping, string comparison against DB rows).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>One sub-threshold note (~55): the idempotency test's `expect(afterSecond.size).toBe(1)` is a count assertion (mild INV-23) that's redundant with the preceding content equality — left as-is since it documents the idempotency invariant and the content check already covers identity.</details>
<details><summary>🏗️ Design [CLEAN]</summary>Unbatched DELETE acceptable: legacy window is ~1 day (sync_log created 2026-06-11, absolute fields shipped 2026-06-12) and the migration runs before the server accepts connections. Test approach (execute the real migration file) is sound — `pg` simple-query handles it and the migration is a single statement.</details>
<details><summary>🔒 Security [CLEAN]</summary>No injection/exposure surface — static migration SQL, no user input.</details>

---

PR #919, opened 2026-06-13
feat(drafts): stage 1 backend — centralized drafts table, feature, outbox + routes

kristofferremback:
**Plan:** `docs/plans/centralized-draft-system.md` (merged in #914). This is **Stage 1 of 4**.

## Problem

Drafts in Threa are local-only. The composer auto-saves to IndexedDB (`db.draftMessages`, debounced 500ms) and keeps manual stash entries (`db.stashedDrafts`, Cmd+S), but none of it leaves the device. A reply you start on your phone is invisible on your laptop. There's also no safe story for the same draft being touched on two devices — the cardinal sin is losing a phone edit because a stale laptop tab overwrote it.

The plan makes drafts **centralized but local-first**: one first-class entity per composer payload, mirrored to the backend so it roams, with concurrency resolved by **local-wins, split-on-drift** rather than overwrite. Duplicated drafts are acceptable; lost drafts are not.

## Solution

Stage 1 is **backend-only** — the table and endpoints exist and are tested, but nothing calls them yet (no frontend wiring, no producers). It mirrors `apps/backend/src/features/scheduled-messages/` almost exactly: the same integer-`version` optimistic CAS, the same user-private outbox lane, the same handler/service/repository/view/barrel layout (INV-51/52). The one structural departure from scheduled-messages is the **split**: on a `version` mismatch the service does not 409 — it keeps the existing row and inserts a fresh draft for the incoming content.

### How it works

- **One `drafts` table** (`20260613130739_drafts.sql`), workspace + user scoped (INV-8/50), prefixed-ULID `draft_` id (INV-2), no FKs (INV-1), no enums — `scope` is `TEXT` validated in code (INV-3). Content is one of two shapes: plaintext (`content_json` / `content_markdown`) or the E2E triple (`ciphertext` / `envelope` / `e2e_version`) with the plaintext columns null. `deleted_at` is the cross-device tombstone for resolve-on-send and explicit discard.
- **`last_client_write_id`** is the per-push idempotency key — a deliberate addition to the plan's SQL. A lost ack makes the client retry the same upsert; matching this column short-circuits to the existing row so the retry doesn't read as drift and spuriously split.
- **The split/idempotent upsert** (`service.ts` `upsert`, one `withTransaction`): (a) `insertIfAbsent` (`ON CONFLICT (id) DO NOTHING`) — a brand-new id lands at version 1; (b) else `findByIdForUpdate` (`SELECT … FOR UPDATE`) locks the row — if its `last_client_write_id` equals the incoming `writeId`, it's a lost-ack retry, return as-is; (c) else `casUpdate` on `expectedVersion` — match → `version + 1`; (d) else (version drift, or the original id is a resolved tombstone) → **split**: leave the existing row untouched, insert a freshly minted `draftId()` at version 1, return `originalId` so the client migrates its local state.
- **Resolve-on-send is CAS** (`resolve`): soft-delete only when `version` still matches, so a copy that drifted since the send started survives as a stash entry instead of being collaterally deleted. `delete` is the unconditional explicit-discard path (no CAS).
- **Outbox → sync pipeline:** `draft:upserted` / `draft:deleted` are written in the same transaction as the row mutation (INV-4/7), as **user-scoped** events (`targetUserId`). `resolveDeliveryGroups` already routes user-scoped events to `user:{userId}`, and `SyncLogRepository.listEntriesForUser` already admits `user:{userId}` log entries on reconnect catch-up (`l.groups && ARRAY['workspace', 'user:'||userId]`) — so live delivery and catch-up both work with no changes to the sync feature. They are **not** in `TIMELINE_BROADCAST_EVENT_TYPES`: drafts are private, not timeline rows, so the contiguity/`broadcastSequence` machinery (INV-61) correctly does not apply.

### Key design decisions

**1. Mirror scheduled-messages, diverge only at the conflict.** Same CAS primitive, same user-private outbox, same colocation. The only new behavior is split-instead-of-409. Reusing a proven user-scoped lane is why drafts slot into the recent sync-v2 catch-up pipeline for free — the catch-up filter is group-based, not a per-event-type allowlist.

**2. `last_client_write_id` idempotency key (addition to the plan's SQL).** Without it, a lost ack on an accepted write makes the client retry with the same `expectedVersion`; the server would see the now-incremented `version`, read drift, and split — a spurious duplicate on every dropped ack. Matching the write id collapses the retry to a no-op. The alternative (dedup on content hash) is fragile against whitespace/ordering and can't distinguish a retry from a genuine identical re-edit.

**3. Soft delete, not hard.** Resolve and discard set `deleted_at` so the removal propagates to other devices as a `draft:deleted` tombstone (INV-53). A hard delete would leave peer devices unable to learn the draft is gone.

**4. No stream-access check on upsert (deliberate).** A draft is private composer state that only ever returns to its own author (user-scoped outbox, every query filters `user_id`), so a draft against an inaccessible stream is harmless scratch. Access is enforced at send time, not draft time. Flagged in self-review (L2) and left as Stage-4 (send-path) scope.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260613130739_drafts.sql` | `drafts` table + two partial indexes (recency, per-scope) |
| `apps/backend/src/features/drafts/repository.ts` | `insertIfAbsent` / `findByIdForUpdate` / `casUpdate` / `softDeleteCas` / `softDelete` / `listByUser`, snake↔camel |
| `apps/backend/src/features/drafts/service.ts` | Split-on-drift `upsert`, CAS `resolve`, `delete`, `list`; outbox-in-transaction |
| `apps/backend/src/features/drafts/handlers.ts` | Zod-validated `list` / `upsert` / `resolve` / `delete`; markdown derived at the boundary (INV-58) |
| `apps/backend/src/features/drafts/view.ts` | Row → wire `Draft`; drops internal-only `lastClientWriteId` / `deletedAt` |
| `apps/backend/src/features/drafts/index.ts` | Feature barrel (INV-52) |
| `apps/backend/src/features/drafts/repository.test.ts` | Repository SQL-shape unit tests |
| `apps/backend/src/features/drafts/service.test.ts` | Split-algorithm + resolve-CAS service tests |

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/api.ts` | `Draft`, `DraftScope`, `DraftCommand`, upsert/resolve/list I-O, outbox payloads (Stage-1 groundwork commit) |
| `packages/types/src/index.ts` | Export the new draft types |
| `packages/backend-common/src/id.ts` + `index.ts` | `draftId()` helper + curated barrel re-export |
| `apps/backend/src/lib/id.ts` | Re-export `draftId` for backend use |
| `apps/backend/src/lib/outbox/repository.ts` | `draft:upserted` / `draft:deleted` in `OutboxEventType`, payload interfaces, `OutboxEventPayloadMap`, `UserScopedEventType` union **and** the `USER_SCOPED_EVENTS` array |
| `apps/backend/src/routes.ts` | Mount `GET` / `PUT :id` / `POST :id/resolve` / `DELETE :id` under `/drafts` |
| `apps/backend/src/server.ts` | Construct `DraftsService` and inject into the routes factory |

## Out of scope (deliberate)

- **All frontend wiring** — Dexie store, composer rewrite, push/pull through the offline queue, socket apply, bootstrap pairing. Stages 2–3.
- **Resolve-on-send wiring, thread re-pointing, E2E seal-before-push.** Stage 4. The columns (`ciphertext`/`envelope`/`e2e_version`) and the resolve CAS exist; nothing produces or consumes them yet.
- **Stream-access validation on upsert** (decision 4 above; self-review L2).
- **E2E triple co-presence validation** — `handlers.ts` rejects `contentJson` + `ciphertext` together but doesn't yet require `envelope`/`e2e_version` to travel with `ciphertext`. Deferred to Stage 4 when the seal path defines that contract (self-review L1).
- **Tombstone retention/expiry** — `deleted_at` rows accumulate; cleanup is post-v1.

## Test plan

- [x] `bun test src/features/drafts` — 14 pass (47 assertions): insert / CAS-update / split-on-drift / lost-ack idempotent retry / soft-delete CAS / resolve decline-on-drift
- [x] `bun test src/lib/outbox` — 42 pass (broadcast routing unaffected)
- [x] `tsc --noEmit` clean across `@threa/types` and `apps/backend`; eslint clean on the new + modified files
- [x] Pre-commit hook (prettier / tsc all packages / astro check / OpenAPI `--check`) passed on commit
- [x] Self-review: 1 reviewer agent over the diff — M1 fixed (recency index now carries the `id DESC` tiebreak to fully cover `listByUser`); L1/L2 documented as deliberate Stage-4 scope above; race-safety / INV-8 / INV-4-7 / outbox union+array verified clean
- [ ] No live-DB harness in this environment, so repository tests assert SQL shape against a mocked `Querier` (the established pattern for repo tests here) — not integration-tested against Postgres

<details>

**Goal:** Centralized, local-first drafts that roam across a user's devices, with local-wins/split-on-drift concurrency. Full spec: `docs/plans/centralized-draft-system.md`.

**Stage sequence:**
1. **(this PR)** Types + backend feature + migration. Table and endpoints exist and are tested; nothing calls them.
2. Frontend unified local store — fold `draftMessages` + `stashedDrafts` into one Dexie `drafts` store + device-local `composerLoaded` pointer; rewrite `use-draft-message.ts` / `draft-store.ts` / `use-stash-composer.ts`. Local-first, no backend calls yet.
3. Push/pull sync + client-side drift — `operation-queue` `upsert_draft` / `delete_draft` ops (silent retry), socket apply with client-side split, `GET /drafts` bootstrap + reconnect invalidation (INV-53).
4. Resolve-on-send + thread re-pointing + E2E — send carries `draftId + version` → CAS resolve; `moveMessagesToThread` re-scopes `thread:{messageId}` drafts → `stream:{threadStreamId}`; seal-before-push for E2E streams.

**What this PR built:** migration, `features/drafts/` (repository/service/handlers/view/barrel + tests), outbox event types + user-scoped routing, routes + server wiring, shared types + `draftId()`.

**Design decisions:** see "Key design decisions" above. Notable deviation from the plan's SQL: added `last_client_write_id` for per-push idempotency.

**Schema changes:** new `drafts` table, append-only migration, no changes to existing tables.

**What's NOT included:** all frontend, resolve/thread/E2E wiring, stream-access checks, tombstone GC — see "Out of scope".

**Status:** backend complete, tested, typechecked. Ready for review; Stage 2 follows in its own PR.

</details>

---

---
_Generated by [Claude Code](https://claude.ai/code/session_018ytvNevQzhKZ5QEXo86eUc)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/919"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783952970&installation_id=129946197&pr_number=919&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F919&signature=3542ae783d8c581f6fedf789cab204697add39965e05d827910d355635a75704"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review — superseded

This review has been superseded by a newer pass: https://github.com/threahq/threa/pull/919#issuecomment-4698886703

<details><summary>Original review (superseded)</summary>


### Code review
**Confidence: 5/7** — Good (a few non-blocking)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3

Found 3 issues (none blocking; all are Stage-1 scope/hygiene, no correctness or security defects in the live paths):

1. `handlers.ts:40-43` — `resolveSchema` requires `writeId` (`z.string().min(1)`) and `ResolveDraftInput` declares it, but it is never forwarded to `draftsService.resolve()` (`ResolveDraftParams` has no `writeId`) — the server enforces the field on every caller and discards it. On `upsert`, `writeId` drives lost-ack dedup via `last_client_write_id`; on `resolve` the equivalent idempotency is absent, so a retry after a lost ack returns `{ resolved: false }` (row already tombstoned), indistinguishable from "version drifted, draft kept". Flagged independently by two reviewers. CLAUDE.md §7 — INV-38 "Delete dead code immediately" / INV-36 "Do not add speculative features, configuration … for imagined requirements". Either wire it through for resolve idempotency, or drop it from the schema + `ResolveDraftInput` until Stage 4 wires resolve-on-send.
   https://github.com/threahq/threa/blob/951f9a71873df9624e7d6346e0e32ba3a110fb82/apps/backend/src/features/drafts/handlers.ts#L40-L43

2. `repository.ts:268-276` — `listByUser` has no `LIMIT`. Both peer features cap their bootstrap reads (`saved-messages` and `scheduled-messages` use `LIMIT` + cursor). Live drafts are sparse in practice, but the query is structurally unbounded — a defensive cap (the personal stash is a bounded set, so a high `LIMIT` guardrail rather than full pagination) keeps it consistent with the established pattern.
   https://github.com/threahq/threa/blob/951f9a71873df9624e7d6346e0e32ba3a110fb82/apps/backend/src/features/drafts/repository.ts#L268-L276

3. `service.ts:145-147` — the split insert uses `newRow!` on `insertIfAbsent`, which returns `null` on `ON CONFLICT (id) DO NOTHING`. The id is a freshly minted ULID so a collision is astronomically unlikely, but if it ever fired, `null` flows into `toDraftView` and throws an opaque `TypeError` at the call site rather than failing loudly. A one-line `if (!newRow) throw new HttpError(...)` makes the impossible case legible. Low-likelihood hardening, not a live defect.
   https://github.com/threahq/threa/blob/951f9a71873df9624e7d6346e0e32ba3a110fb82/apps/backend/src/features/drafts/service.ts#L145-L147

---

Matches `docs/plans/centralized-draft-system.md` Stage 1. The planned per-feature `outbox-handler.ts` is intentionally absent — user-scoped routing is handled generically by `USER_SCOPED_EVENTS` + `resolveDeliveryGroups`, an implementation-choice difference, not a gap. Frontend / resolve-on-send / thread re-pointing / E2E are correctly deferred to Stages 2–4.
</details>
<details><summary>🔍 Bugs [1 hardening nit — issue 3]</summary>

Upsert concurrency traced and sound: `insertIfAbsent` (the insert-conflict block serializes the brand-new-id window) + `findByIdForUpdate` `FOR UPDATE` (serializes the already-exists window) + version CAS + fresh-id split — no TOCTOU gap, no lost write. One reviewer raised the foreign-PK-collision case (an id that exists under a different user → `findByIdForUpdate` returns null → split); **dismissed** — the caller's content is preserved under the fresh split id and the client migrates via `split:true`, so there is no data loss and no cross-user read/write. `jsonbOrNull` + `::jsonb` casts are correct (JSON-string params, not raw objects). `expectedVersion` is unbounded JS int vs PG `INTEGER` — a deliberately bad value 500s cleanly, not a security issue.
</details>
<details><summary>📋 CLAUDE.md Compliance [1 violation — issue 1]</summary>

INV-1 (no FKs), INV-3 (no enums; `scope` is TEXT), INV-2 (prefixed ULID), INV-8/50 (every query filters `workspace_id` + `user_id`), INV-20 (set-based CAS in the WHERE), INV-4/7 (outbox written in the same `withTransaction` as the mutation; no-op branches correctly skip publishing), INV-51/52 (colocated feature + barrel), INV-55 (Zod), INV-58 (markdown derived at the boundary) all hold. Only the dead `writeId` (issue 1) violates INV-38/INV-36.
</details>
<details><summary>🏗️ Design [1 concern — issue 2]</summary>

Deliberately mirrors `scheduled-messages`/`saved-messages`: same Querier-first repo, snake↔camel, user-private outbox lane. Outbox wiring adds the events to the `OutboxEventType` union, `OutboxEventPayloadMap`, the `UserScopedEventType` union **and** the `USER_SCOPED_EVENTS` runtime array (the recurring "type-but-not-array" footgun is avoided). Dependencies constructed once (INV-13). Only `listByUser`'s missing cap (issue 2) diverges from the peer pattern.
</details>

No injection (parameterized SQL throughout), no auth bypass (every read/write scoped to `req.user`/`req.workspaceId`), no cross-workspace/cross-user draft exposure, no secrets. Drafts are user-private and never broadcast.
</details>

</details>

kristofferremback:
### Code review — superseded

Superseded by a newer pass: https://github.com/threahq/threa/pull/919#issuecomment-4699282603

<details><summary>Earlier review (superseded)</summary>


### Code review (re-review @ `0d1c66f`)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3

The three findings from the [prior review](https://github.com/threahq/threa/pull/919#issuecomment-4698860898) are verified fixed and clean: `writeId` removed from the resolve path (schema + `ResolveDraftInput`, no stale refs), `listByUser` capped with `MAX_DRAFTS_PER_USER = 500`, and the split insert now fails loud instead of `newRow!`. Two new issues, both non-blocking:

1. `service.ts:116` — the lost-ack idempotency branch (b) short-circuits on `existing.lastClientWriteId === writeId` **without excluding tombstones**. `findByIdForUpdate` deliberately does not filter `deleted_at IS NULL`, so if a write is accepted (setting `last_client_write_id = W`), the draft is then resolved (soft-deleted, `draft:deleted` emitted), and the client retries the original push with the same `writeId = W` (lost ack), branch (b) returns `toDraftView(existing)` with `split: false` — and `toDraftView` strips `deletedAt`, so the client gets a **live-looking response for a deleted draft**, contradicting the `draft:deleted` event already on the wire. Narrow timing window and no consumer in Stage 1, but the response is contract-wrong. Fix: gate branch (b) on a live row (`existing && !existing.deletedAt && existing.lastClientWriteId === ...`); a tombstoned-but-matching row then falls through to the split, which is consistent with the documented "duplicates acceptable, never resurrect-as-live" rule.
   https://github.com/threahq/threa/blob/0d1c66fc74b0fa36591c71572088ea6f04c2cce8/apps/backend/src/features/drafts/service.ts#L115-L118

2. `repository.test.ts:203-216` — the `listByUser` test asserts the filter predicates and `ORDER BY` but not the new `LIMIT`. `MAX_DRAFTS_PER_USER` was added as last commit's fix; without an assertion the cap can be silently removed and the suite still passes. Flagged by two reviewers. Add `expect(captured.text).toContain("LIMIT")` (and `expect(captured.values).toContain(500)`).
   https://github.com/threahq/threa/blob/0d1c66fc74b0fa36591c71572088ea6f04c2cce8/apps/backend/src/features/drafts/repository.test.ts#L203-L216

---

Still matches `docs/plans/centralized-draft-system.md` Stage 1. The fix commit changed nothing structural. Frontend / resolve-on-send / thread / E2E remain correctly deferred.
</details>
<details><summary>🔍 Bugs [1 narrow — issue 1]</summary>

Upsert concurrency re-traced and still sound (insert-conflict block + `FOR UPDATE` + version CAS + fresh-id split). The new `if (!newRow) throw` guard is correct. Only the tombstone/lost-ack interaction in branch (b) (issue 1) is wrong, and only in a narrow resolve-races-a-retry window with no Stage-1 consumer. `jsonbOrNull`/`::jsonb` casts, CAS WHERE clauses (version + deleted_at + workspace + user), and the new `LIMIT` are all correct.
</details>

INV-1/2/3/8/50/20/4/7/11/13/36/38/51/52/55/58 all hold. The prior INV-38/INV-36 dead-`writeId` violation is resolved.
</details>

Outbox events present in the `OutboxEventType` union, `OutboxEventPayloadMap`, the `UserScopedEventType` union **and** the `USER_SCOPED_EVENTS` runtime array. `DraftsService` constructed once (INV-13). `MAX_DRAFTS_PER_USER` is a named const (no magic number). Lost-ack branch correctly avoids double-publishing.
</details>

Parameterized SQL throughout, every read/write scoped to `req.user`/`req.workspaceId`, no cross-user/workspace exposure, no secrets.
</details>

</details>

kristofferremback:
### Code review (pass 3 @ `d0455e9`)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3

No issues found. All five findings from the two prior passes are verified fixed, and three independent reviewers (spec / correctness / design) each returned clean.

Verified this pass:
- **Branch (b)/(c) tombstone gating** — the lost-ack idempotency branch and the CAS branch are both gated on `!existing.deletedAt`; a tombstoned row with a matching `writeId` falls cleanly to the split instead of being returned as live. The two guards are semantically distinct (return-as-is vs attempt-CAS), not collapsible duplication.
- **Concurrency** — `insertIfAbsent` (insert-conflict block) + `findByIdForUpdate` `FOR UPDATE` + version CAS + fresh-id split serialize correctly; no TOCTOU, no lost write; the fail-loud split guard is correct.
- **INV compliance** — INV-1/2/3/8/50 (no FKs/enums, prefixed ULID, every query scoped to `workspace_id` + `user_id`), INV-4/7/6 (outbox in the same transaction; services own boundaries), INV-11 (fail loud), INV-13 (service constructed once), INV-20 (set-based CAS), INV-30 (single-query `list` on `pool`, writes via `withTransaction`), INV-51/52 (colocated + barrel), INV-55/32 (Zod + `HttpError`), INV-58 (markdown derived at the boundary).
- **Outbox/sync** — `draft:upserted`/`draft:deleted` present in the `OutboxEventType` union, `OutboxEventPayloadMap`, the `UserScopedEventType` union **and** the `USER_SCOPED_EVENTS` array; routed to `user:{targetUserId}`, written to the sync log for catch-up, and correctly absent from `TIMELINE_BROADCAST_EVENT_TYPES` (no `broadcastSequence` slot, INV-61 N/A).
- **Security** — parameterized SQL throughout, no cross-user/workspace exposure, no secrets.

PR #920, opened 2026-06-13
refactor: sync-v2 unread-counter cleanup + scope invitation events to members:write

kristofferremback:
Two sync-v2 healing follow-ups on top of #917, plus the inventory doc that tracks this line of work. They are independent; grouped here because they share the inventory entry and were done in one sitting.

---

### Problem

The three unread-counter socket payloads — `stream:read`, `stream:read_all`, `stream:activity` — have carried absolute fields (`lastReadOrdinal`, `reads[]`, `messageOrdinal`) since sync-v2 phase 2c. The backend emits them unconditionally (`messaging/event-service.ts` for `messageOrdinal`; `streams/service.ts` for `lastReadOrdinal` and `reads[]`). The frontend handlers in `apps/frontend/src/sync/workspace-sync.ts` still carried the pre-2c increment/zero fallback for absent fields as a defensive guard — dead code now that every live and replayed payload has the fields.

### Solution

Make the three fields required on the payload interfaces and delete the absent-field branches in `handleStreamRead` / `handleStreamReadAll` / `handleStreamActivity` (both the TanStack query-cache update and the IndexedDB write), so each handler routes straight through the absolute appliers (`applyStreamReadOrdinal` / `applyStreamsReadAllOrdinals` / `applyStreamActivityOrdinal`).

The absolute path is byte-for-byte unchanged — the deleted code only ran when a field was missing, which the backend never does. The obsolete `"falls back to the legacy increment when stream:activity lacks an ordinal"` test is removed (it covered a path that no longer exists).

---

### Problem

Invitation outbox events (`invitation:sent` / `accepted` / `revoked` / `link-created` / `link-claimed`) matched no scoping list in `lib/outbox/repository.ts`, so `resolveDeliveryGroups` routed them through its final `WORKSPACE_GROUP` fallthrough. That meant they were appended to `sync_log` and broadcast to **every** workspace member's socket — even though the payloads carry invitee `email`, invite-link `tokenHash`, and accepted-user identity (`InvitationSentOutboxPayload` etc.). Inert today only because no client reads invitation events, but a leak waiting for the first `invitation:*` listener. The workspace bootstrap already gates the `invitations` field on `members:write` (`workspaces/handlers.ts`), so the broadcast was strictly broader than the REST surface.

### Solution

Introduce a permission-scoped delivery group and route the five invitation events to it, applied congruently to both the live socket path and catch-up replay.

### How it works

1. **`delivery-groups.ts`** — `permissionGroup(slug)` → `permission:<slug>`; `DELIVERED_PERMISSION_SCOPES` (currently just `MEMBERS_WRITE`) is the set of scopes that get their own group; `permissionGroupsForRole(role)` derives a member's permission groups via `permissionsForRole` from `@threa/types`. `resolveDeliveryGroups` gains a branch returning `[permissionGroup(MEMBERS_WRITE)]` for the five invitation events, before the workspace fallthrough.
2. **Live (`socket.ts`)** — on workspace join, after the existing user-room join, the socket auto-joins the role-derived permission rooms (`permissionGroupsForRole(workspaceUser.role).map(groupToRoom)`) and leaves them on workspace leave. The `userRooms` bookkeeping map now tracks `permissionRooms` alongside `userRoom`. A re-join on the same socket re-derives the set and leaves rooms the member no longer qualifies for.
3. **Catch-up (`sync/handlers.ts` → `service.ts` → `repository.ts`)** — the catch-up handler derives `permissionGroups` from `req.user.role` (the same role source the socket join uses) and threads it into `listEntriesForUser`, whose group-overlap test becomes `l.groups && (ARRAY['workspace', 'user:' || $userId]::text[] || $permissionGroups::text[])`. So replay admits exactly what live delivery reached.

### Key design decisions

**1. Role-derived permissions, not the JWT claim.** Socket auth carries only `workosUserId` (no permissions claim), and using one source in both paths keeps live and catch-up congruent by construction — no gap or duplicate for any custom-permission edge. `requireWorkspacePermission` honors a JWT `permissions` claim when present and falls back to `permissionsForRole(role)`; this change uses the role fallback in both delivery paths. Both `workspaceUser.role` (socket) and `req.user.role` (catch-up) come from the same mirror-aware `mapRowToUser` read, so they agree. Consequence: a hypothetical user with a custom JWT grant of `members:write` but `member` role would see `invitations` in bootstrap (JWT) but not get live/catch-up invitation updates (role) — strictly better than today (nobody gets live updates) and never a leak to a non-privileged member.

**2. Scope is `members:write`, mirroring the bootstrap gate.** `workspaces/handlers.ts` already includes the bootstrap `invitations` field only for `members:write` holders, and the invitation routes gate on `requireWorkspacePermission(members:write)`. The delivery scope now matches the REST surface exactly.

**3. All five invitation events, not just sent/accepted/revoked.** `link-created` and `link-claimed` carry `tokenHash` / `email` too, so scoping anything less would leave a leak.

### Self-review (2 Sonnet reviewer agents)

- **Re-join leak (fixed before push).** First version joined the new permission rooms on every workspace `join` but never left previously-joined ones. Since the frontend re-joins the workspace room on the same socket (`sync-engine.ts` joins at two sites), a role change mid-connection would have left a demoted admin in the stale `members:write` room. Fixed in `socket.ts`: on re-join the handler leaves any tracked permission room not in the freshly-derived set.
- **Live-path demotion window (accepted limitation).** After the re-join fix, a demoted admin whose socket stays connected and never re-joins still receives live invitation events until reconnect. This matches the system's existing permission model — `requireWorkspacePermission` itself accepts "a demoted member loses access on next session refresh (≤ access-token TTL)." Catch-up is always current (role derived per request). Reacting to role-change events on the socket to close the window fully is a possible follow-up, not done here.
- Remaining reviewer notes (permission rooms not metered in `wsConnectionsActive`, `userRooms` not cleared on disconnect) are pre-existing and match the existing user-room treatment; socket.io drops rooms on disconnect, so no live leak. Not changed.

## Out of scope (deliberate)

- **The frontend `invitation:*` client handler.** Wiring `sync/workspace-sync.ts` to invalidate `["invitations", workspaceId]` so the settings invitations list heals live remains the separate follow-up logged in the inventory. This PR only closes the broadcast leak so that wiring is safe to add later.
- **Backfilling historical `sync_log` rows.** Pre-change invitation entries keep their old `['workspace']` groups until retention (30-day horizon) ages them out, so catch-up still returns them to all members until then. They are inert — no client handler reads invitation events — so no rewrite here; if the client handler lands before those rows expire, add a one-time migration to rewrite their `groups` (INV-17 allows a new append-only migration).
- **The unconsumed `lastReadSequence?` / `sequence?` wire fields** on the unread payload interfaces — declared but never read, dead before this PR. Left untouched; a separate dead-code question, not part of the fallback cleanup.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/lib/outbox/delivery-groups.test.ts` | Unit tests: invitation events route to `members:write`; `permissionGroupsForRole` (admin/owner→group, member→none) |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/sync/workspace-sync.ts` | Make `lastReadOrdinal`/`reads`/`messageOrdinal` required; delete legacy fallback branches in the three handlers (cache + IDB); correct the `handleStreamActivity` comment |
| `apps/frontend/src/sync/workspace-sync.test.ts` | Remove the obsolete legacy-increment test |
| `apps/backend/src/lib/outbox/delivery-groups.ts` | Add `permissionGroup`, `DELIVERED_PERMISSION_SCOPES`, `permissionGroupsForRole`; route invitation events to `permission:members:write` |
| `apps/backend/src/lib/outbox/index.ts` | Export the three new delivery-group symbols |
| `apps/backend/src/socket.ts` | Auto-join/leave role-derived permission rooms on workspace join/leave; drop stale rooms on re-join |
| `apps/backend/src/features/sync/handlers.ts` | Derive `permissionGroups` from `req.user.role`, pass to catch-up |
| `apps/backend/src/features/sync/service.ts` | Thread `permissionGroups` through `catchUp` |
| `apps/backend/src/features/sync/repository.ts` | Admit the viewer's permission groups in the `listEntriesForUser` overlap test |
| `apps/backend/src/features/sync/{handlers,service}.test.ts` | Supply/assert `permissionGroups` (incl. admin→members:write derivation) |
| `apps/backend/tests/integration/sync-catch-up.test.ts` | Integration case: holders get permission-group entries, members don't |
| `docs/plans/sync-v2-healing-deletion-inventory.md` | Log the `invitation:*` entry, mark it scoped (done) and `usePageResumeRefresh` done (#917) |

## Test plan

- [x] Frontend `bunx vitest run src/sync` — 193 pass
- [x] Backend `bun test src/lib/outbox src/features/sync` — 64 pass
- [x] `tsc --noEmit` clean across frontend and backend; eslint + prettier clean on changed files
- [x] Self-review: 2 Sonnet reviewer agents (backend security/correctness, frontend correctness) — 1 correctness fix applied (re-join room leak), 1 limitation documented (demotion window), 1 comment nit fixed; remaining notes dismissed as pre-existing/by-design
- [ ] `tests/integration/sync-catch-up.test.ts` permission-group case — not run locally (live-DB harness is CI-only, `ECONNREFUSED 127.0.0.1`); runs in CI

---

### DIFF
--- a/apps/backend/src/features/sync/handlers.test.ts
+++ b/apps/backend/src/features/sync/handlers.test.ts
@@ -4,9 +4,9 @@ import { createSyncHandlers } from "./handlers"
 import { HttpError } from "../../lib/errors"
 import type { SyncService } from "./service"
 
-function makeReqRes(query: Record<string, string>) {
+function makeReqRes(query: Record<string, string>, role: string = "member") {
   const req = {
-    user: { id: "usr_alice" },
+    user: { id: "usr_alice", role },
     workspaceId: "ws_1",
     query,
   } as unknown as Request
@@ -33,7 +33,13 @@ describe("createSyncHandlers.catchUp", () => {
     const { req, res, json } = makeReqRes({ after: "5", limit: "50" })
     await handlers.catchUp(req, res)
 
-    expect(catchUp.mock.calls[0][0]).toEqual({ workspaceId: "ws_1", userId: "usr_alice", after: 5n, limit: 50 })
+    expect(catchUp.mock.calls[0][0]).toEqual({
+      workspaceId: "ws_1",
+      userId: "usr_alice",
+      permissionGroups: [],
+      after: 5n,
+      limit: 50,
+    })
     expect(json.mock.calls[0][0]).toEqual({
       entries: [
         {
@@ -68,7 +74,29 @@ describe("createSyncHandlers.catchUp", () => {
     const { req, res } = makeReqRes({})
     await handlers.catchUp(req, res)
 
-    expect(catchUp.mock.calls[0][0]).toEqual({ workspaceId: "ws_1", userId: "usr_alice", after: 0n, limit: 200 })
+    expect(catchUp.mock.calls[0][0]).toEqual({
+      workspaceId: "ws_1",
+      userId: "usr_alice",
+      permissionGroups: [],
+      after: 0n,
+      limit: 200,
+    })
+  })
+
+  it("derives permission delivery groups from the member role (admin → members:write)", async () => {
+    const catchUp = mock(async (_params: Parameters<SyncService["catchUp"]>[0]) => ({ entries: [], head: 0n }))
+    const handlers = createSyncHandlers({ syncService: { catchUp } as unknown as SyncService })
+
+    const { req, res } = makeReqRes({}, "admin")
+    await handlers.catchUp(req, res)
+
+    expect(catchUp.mock.calls[0][0]).toEqual({
+      workspaceId: "ws_1",
+      userId: "usr_alice",
+      permissionGroups: ["permission:members:write"],
+      after: 0n,
+      limit: 200,
+    })
   })
 
   it("rejects a non-numeric cursor with a 400", async () => {
--- a/apps/backend/src/features/sync/handlers.test.ts
+++ b/apps/backend/src/features/sync/handlers.test.ts
@@ -1,7 +1,9 @@
 import { describe, expect, it, mock } from "bun:test"
 import type { Request, Response } from "express"
+import { WORKSPACE_PERMISSION_SCOPES } from "@threa/types"
 import { createSyncHandlers } from "./handlers"
 import { HttpError } from "../../lib/errors"
+import { permissionGroup } from "../../lib/outbox"
 import type { SyncService } from "./service"
 
 function makeReqRes(query: Record<string, string>, role: string = "member") {
@@ -93,7 +95,7 @@ describe("createSyncHandlers.catchUp", () => {
     expect(catchUp.mock.calls[0][0]).toEqual({
       workspaceId: "ws_1",
       userId: "usr_alice",
-      permissionGroups: ["permission:members:write"],
+      permissionGroups: [permissionGroup(WORKSPACE_PERMISSION_SCOPES.MEMBERS_WRITE)],
       after: 0n,
       limit: 200,
     })
--- a/apps/backend/src/features/sync/handlers.ts
+++ b/apps/backend/src/features/sync/handlers.ts
@@ -2,6 +2,7 @@ import type { Request, Response } from "express"
 import { z } from "zod"
 import type { SyncCatchUpResponse } from "@threa/types"
 import { HttpError } from "../../lib/errors"
+import { permissionGroupsForRole } from "../../lib/outbox"
 import type { SyncService } from "./service"
 
 const catchUpQuerySchema = z.object({
@@ -18,6 +19,10 @@ export function createSyncHandlers({ syncService }: Dependencies) {
     async catchUp(req: Request, res: Response) {
       const userId = req.user!.id
       const workspaceId = req.workspaceId!
+      // Permission-scoped log entries (e.g. invitation lifecycle) replay only to
+      // holders. Derived from role to stay congruent with the socket join, which
+      // joins the same permission rooms off the same role.
+      const permissionGroups = permissionGroupsForRole(req.user!.role)
 
       const parsed = catchUpQuerySchema.safeParse(req.query)
       if (!parsed.success) {
@@ -27,6 +32,7 @@ export function createSyncHandlers({ syncService }: Dependencies) {
       const { entries, head, requiresBootstrap } = await syncService.catchUp({
         workspaceId,
         userId,
+        permissionGroups,
         after: BigInt(parsed.data.after),
         limit: parsed.data.limit,
       })
--- a/apps/backend/src/features/sync/repository.ts
+++ b/apps/backend/src/features/sync/repository.ts
@@ -119,8 +119,12 @@ export const SyncLogRepository = {
 
   /**
    * Lists log entries after a cursor, filtered to what the requesting user is
-   * allowed to see: workspace-group entries, their own user-group entries, and
-   * stream-group entries for streams they can read.
+   * allowed to see: workspace-group entries, their own user-group entries,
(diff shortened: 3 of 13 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (no bugs or security issues; two non-blocking notes)
**Review models:** Orchestrator: Claude Code | Reviewers: sonnet ×3

Correctness and security came back clean: the invitee-data leak is closed on both paths (live socket room + catch-up SQL), the `permissionGroups` SQL overlap is correct for the empty-array (member) case, the socket join/leave/re-join room bookkeeping has no stale-room leak, and the deleted frontend legacy branches are behaviour-preserving on the absolute path (own-message / viewing cases still resolve through `applyStreamActivityOrdinal`).

Found 2 non-blocking issues:

1. `apps/backend/src/lib/outbox/delivery-groups.ts:74-82,253-261` — **design (future footgun):** `DELIVERED_PERMISSION_SCOPES` (which rooms a socket joins and which groups catch-up admits, via `permissionGroupsForRole`) and the invitation routing branch in `resolveDeliveryGroups` are two independent lists with only a comment binding them. No current bug — `MEMBERS_WRITE` is in both. But if a future event is routed to `permissionGroup(X)` where `X` is not in `DELIVERED_PERMISSION_SCOPES`, live delivery still works (the room exists) while catch-up silently drops those entries (the SQL only admits groups `permissionGroupsForRole` returns) — a live-vs-replay asymmetry. Consider deriving the routing from the same source, or an exhaustiveness-checked scope→event-types map.
   https://github.com/threahq/threa/blob/61eb25e8e8b3b07aa52cec2199461bf9c213de05/apps/backend/src/lib/outbox/delivery-groups.ts#L74-L82

2. `apps/backend/src/features/sync/handlers.test.ts:96` (and `apps/backend/tests/integration/sync-catch-up.test.ts:109,115`) — **minor consistency (INV-33, debatable):** these assertions hardcode `"permission:members:write"` while `delivery-groups.test.ts` derives it via `permissionGroup(...)`. Pinning the literal wire value in a test is defensible (it catches a format change independently), so this is a consistency nit, not a real drift risk — flagging only because the sibling test does it the other way.
   https://github.com/threahq/threa/blob/61eb25e8e8b3b07aa52cec2199461bf9c213de05/apps/backend/src/features/sync/handlers.test.ts#L95-L97

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Both halves land their corresponding changes — the new delivery group is honoured in live (socket) and replay (catch-up) paths; required-field changes match the unconditional backend emit; exported symbols are consumed.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No logic errors, null hazards, or races found in the changed lines. SQL empty-array case, socket re-join cleanup, and frontend own/viewing unread handling all verified correct.</details>
<details><summary>📋 CLAUDE.md Compliance [1 — debatable]</summary>INV-33 test magic-string (issue 2). No INV-8 / INV-11 / INV-20 / INV-55 violations introduced.</details>
<details><summary>🏗️ Design [1 concern]</summary>The `DELIVERED_PERMISSION_SCOPES` ↔ `resolveDeliveryGroups` manual-sync footgun (issue 1).</details>
<details><summary>🔒 Security [CLEAN]</summary>The leak (invitee email / invite-link tokenHash broadcast workspace-wide) is closed; no residual exposure path introduced. Non-`members:write` members never join the room and never match the catch-up group filter.</details>

PR #933, opened 2026-06-14
fix(workspace-authz): broadcast workspace_user:updated on role change

kristofferremback:
## Problem

Role changes fanned out from the control plane land in the regional `workspace_user_permissions` mirror through a single `WorkspaceUserPermissionsRepository.upsert` in `WorkspaceAuthzService.applyMembershipChange` (`apps/backend/src/features/workspace-authz/service.ts`) — the landing point for the CP fan-back at `POST /internal/authz/memberships` (`handlers.ts:syncMembership`). That upsert emitted **no outbox event**.

The regional backend re-derives a user's role from the mirror on every request (`UserRepository`'s `JOIN_AUTHZ_MIRROR` + `pickMirroredRole`), so the access boundary was already correct. But connected clients cache the role in their workspace bootstrap and IndexedDB and only refresh it on reconnect/reload. So a just-demoted admin kept `role: "admin"` — and the admin-only UI it gates — in every open tab until they reconnected. This is a client-freshness gap, not privilege escalation.

This re-scopes the old "demotion window" follow-up (#920/#924): the real fix is the missing event, not dropping stale rooms on a role-change socket event — there was no role-change event to begin with.

## Solution

Emit the `workspace_user:updated` event the frontend already consumes, in the same transaction as the mirror upsert. No new event type, no frontend change.

### How it works

`applyMembershipChange` now wraps its work in `withTransaction`:

1. Upsert the mirror on the transaction `client` (was `this.pool`). The repository's `last_event_at < EXCLUDED.last_event_at` guard still rejects stale/duplicate fan-out, returning `null` when nothing changed.
2. On a real change, load the affected user with `UserRepository.findByWorkosUserIdInWorkspace(client, …)`. Role is derived from the mirror at read time, so reading **after** the upsert **in the same transaction** yields the new role.
3. Emit `OutboxRepository.insert(client, "workspace_user:updated", { workspaceId, user: serializeBigInt(user) })` — the same payload shape `updateUserProfile`/`setUserStatus` already produce.

The frontend handler `handleWorkspaceUserUpdated` (`apps/frontend/src/sync/workspace-sync.ts:762`) already writes this payload into the bootstrap cache, IndexedDB (`db.workspaceUsers`), and the in-memory user cache, so the cached role (and its gated UI) updates live. Its `WorkspaceUserPayload` carries `role`, confirming the field flows through.

### Key design decisions

**1. Read the user inside the same transaction, after the upsert.** Role is mirror-derived, so an in-transaction read post-upsert is what makes the broadcast carry the *new* role without a second round trip or a re-derivation at the call site. Outbox-in-transaction also keeps the mirror write and its broadcast atomic (INV-4/INV-7).

**2. Only emit on an applied change.** A stale upsert returns `null` and emits nothing, so duplicated/out-of-order fan-out doesn't generate spurious broadcasts. Preserves the existing "ignored as stale" debug log.

**3. Skip the broadcast when the user row hasn't landed yet.** The mirror can be written before the regional `users` row exists (e.g. between invite acceptance and the next WorkOS poll). In that window the load returns `null`; there's nothing to broadcast, and the user's first read derives the role from the mirror anyway.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/workspace-authz/service.ts` | Wrap `applyMembershipChange` in `withTransaction`; load the user post-upsert and emit `workspace_user:updated` on a real change. |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/features/workspace-authz/service.test.ts` | Covers the three paths: change emits with the derived role, stale upsert emits nothing, user-not-yet-present skips the broadcast. |

## Out of scope (deliberate)

- **`applyMembershipRemoval`** (`service.ts`) also emits nothing, but full workspace removal emits `workspace_user:removed` from a different path. Left untouched — not symmetric enough to fold in here.
- **FIX #2 (public-root leg in catch-up `visible_streams`)** and **FIX #3 (notification-level membership event)** from the same sync-gap audit ship as their own PRs.

## Test plan

- [x] `bun test src/features/workspace-authz` — 8 pass (3 new + existing handler tests)
- [x] `bun test src/features/workspaces/service.test.ts src/features/workspace-authz` together — 19 pass (no cross-file `spyOn(db, "withTransaction")` interference)
- [x] `bunx tsc --noEmit` (backend) clean
- [x] Pre-commit hook (full monorepo lint + per-package tsc + tests) passed on commit
- [ ] No live-DB harness in this suite, so the in-transaction read-after-upsert is covered via scoped `spyOn` stubs (INV-48) like its `workspaces/service.test.ts` siblings — not integration-tested against Postgres

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01NnHLHsPdKGtPd4YVrM7RKm)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/933"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784028796&installation_id=129946197&pr_number=933&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F933&signature=8facf5064ed2125a907145f319813701e0d1eb52414c2ab279b56da7ef2cb4bf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/workspace-authz/service.test.ts
+++ b/apps/backend/src/features/workspace-authz/service.test.ts
@@ -74,9 +74,9 @@ describe("WorkspaceAuthzService.applyMembershipChange", () => {
     const service = new WorkspaceAuthzService({ pool: {} as never })
     await service.applyMembershipChange(input)
 
-    expect(upsertSpy).toHaveBeenCalledTimes(1)
+    expect(upsertSpy).toHaveBeenCalled()
     expect(userSpy).toHaveBeenCalledWith({} as never, "ws_1", "workos_user_1")
-    expect(insertSpy).toHaveBeenCalledTimes(1)
+    expect(insertSpy).toHaveBeenCalled()
     const [, eventType, payload] = insertSpy.mock.calls[0]!
     expect(eventType).toBe("workspace_user:updated")
     expect(payload).toMatchObject({
@@ -98,6 +98,26 @@ describe("WorkspaceAuthzService.applyMembershipChange", () => {
     await service.applyMembershipChange(input)
 
     expect(upsertSpy).toHaveBeenCalledTimes(1)
+    expect(userSpy).not.toHaveBeenCalled()
+    expect(insertSpy).not.toHaveBeenCalled()
+
+    upsertSpy.mockRestore()
+    userSpy.mockRestore()
+    insertSpy.mockRestore()
+  })
+
+  test("emits nothing when the upsert applied a non-active membership", async () => {
+    // An inactive/pending mirror derives no role from the active-only read
+    // join, so loading the user would broadcast the stale users.role fallback.
+    const upsertSpy = spyOn(WorkspaceUserPermissionsRepository, "upsert").mockResolvedValue(
+      buildMirror({ status: "inactive" })
+    )
+    const userSpy = spyOn(UserRepository, "findByWorkosUserIdInWorkspace")
+    const insertSpy = spyOn(OutboxRepository, "insert")
+
+    const service = new WorkspaceAuthzService({ pool: {} as never })
+    await service.applyMembershipChange({ ...input, status: "inactive" })
+
     expect(userSpy).not.toHaveBeenCalled()
     expect(insertSpy).not.toHaveBeenCalled()
 
--- a/apps/backend/src/features/workspace-authz/service.test.ts
+++ b/apps/backend/src/features/workspace-authz/service.test.ts
@@ -1,0 +1,124 @@
+import { afterEach, describe, expect, spyOn, test } from "bun:test"
+import { WORKSPACE_ROLE_SLUGS } from "@threa/types"
+import * as db from "../../db"
+import { OutboxRepository } from "../../lib/outbox"
+import { UserRepository, type User } from "../workspaces"
+import { WorkspaceAuthzService } from "./service"
+import { WorkspaceUserPermissionsRepository, type WorkspaceUserPermissions } from "./repository"
+
+function buildUser(overrides: Partial<User> = {}): User {
+  return {
+    id: "usr_1",
+    workspaceId: "ws_1",
+    workosUserId: "workos_user_1",
+    email: "user@example.com",
+    role: WORKSPACE_ROLE_SLUGS.MEMBER,
+    slug: "user",
+    name: "User",
+    description: null,
+    avatarUrl: null,
+    timezone: null,
+    locale: null,
+    pronouns: null,
+    phone: null,
+    githubUsername: null,
+    statusEmoji: null,
+    statusText: null,
+    statusExpiresAt: null,
+    statusPausesNotifications: false,
+    notificationsPausedUntil: null,
+    notificationsPausedIndefinitely: false,
+    setupCompleted: true,
+    joinedAt: new Date("2026-01-01T00:00:00.000Z"),
+    ...overrides,
+  }
+}
+
+function buildMirror(overrides: Partial<WorkspaceUserPermissions> = {}): WorkspaceUserPermissions {
+  return {
+    workspaceId: "ws_1",
+    workosUserId: "workos_user_1",
+    roleSlugs: [WORKSPACE_ROLE_SLUGS.MEMBER],
+    status: "active",
+    lastEventAt: new Date("2026-01-02T00:00:00.000Z"),
+    ...overrides,
+  }
+}
+
+describe("WorkspaceAuthzService.applyMembershipChange", () => {
+  // Run the transaction callback inline against a stub client so the upsert,
+  // user read, and outbox insert exercise the real control flow without a DB.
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (one substantive issue found and fixed pre-merge)
**Review models:** Orchestrator: Opus 4.8 | Reviewers: sonnet x6 (spec, correctness, data-flow, design, UX, mobile)

Found 2 issues, both fixed in `97e560f`. 1 dismissed.

1. `apps/backend/src/features/workspace-authz/service.ts` — **Stale role broadcast on non-active membership** (bug / data-flow — flagged independently by the correctness and data-flow lenses, 90 & 88). `WorkosMembershipStatus` is `"active" | "inactive" | "pending"`. For an `inactive`/`pending` fan-out the mirror upsert still returns a row (its staleness guard is status-agnostic), but `findByWorkosUserIdInWorkspace` derives `role` through a JOIN gated on `wup.status = 'active'` (`JOIN_AUTHZ_MIRROR`). The just-written non-active row misses the join, `mirror_role_slugs` is `null`, and `pickMirroredRole` falls back to the stale `users.role` — so the broadcast shipped the **pre-change** role, the opposite of the fix's intent.
   **Fixed:** gate the broadcast on the upserted status being `active`, mirroring the `status !== 'active'` gate in `resolveActivePermissions`; non-active transitions skip the broadcast (deactivation is surfaced through the removal/credential paths). New test covers the inactive case.
   https://github.com/threahq/threa/blob/97e560f74ae58224c7157bf158464c72a0a021ab/apps/backend/src/features/workspace-authz/service.ts#L84-L92

2. `apps/backend/src/features/workspace-authz/service.test.ts` — **INV-23 count assertion** (claude-md, 83). The happy-path test asserted `toHaveBeenCalledTimes(1)` on the outbox spy; INV-23 (Testing) says "Do not assert event counts; assert presence/content of specific events." Content was already asserted alongside it.
   **Fixed:** relaxed to `toHaveBeenCalled()` + the existing payload/content assertion.

---
<details><summary>📐 Plan Adherence — CLEAN</summary>Matches FIX #1 intent: wrap the upsert in a transaction, load the user post-upsert, emit the existing `workspace_user:updated` event. No frontend change needed (handler already consumes it). `applyMembershipRemoval` left untouched per the deliberate out-of-scope note.</details>
<details><summary>🔍 Bugs — 1 fixed</summary>See item 1 (non-active stale-role broadcast).</details>
<details><summary>🔁 Data Flow — 1 fixed (same as item 1)</summary>Payload round-trip to the frontend `WorkspaceUserPayload` confirmed: `role` and all profile fields present; Dates serialize at the wire boundary as the other `workspace_user:updated` emitters already rely on. Self-echo (a demoted user receiving their own update) merges correctly via the in-place cache patch.</details>
<details><summary>📋 CLAUDE.md Compliance — 1 fixed</summary>See item 2 (INV-23). INV-4/INV-7 (outbox-in-transaction) and INV-52 (barrel import of `UserRepository`) satisfied.</details>
<details><summary>🏗️ Design — CLEAN</summary>`UserRepository` used as a static namespace import via the `../workspaces` barrel, consistent with the sibling `WorkspaceUserPermissionsRepository` usage and the `updateUserProfile` template. Read-your-own-writes within the transaction makes the post-upsert read see the new role. Per-event user load is not a hot path.</details>
<details><summary>🔒 Security — CLEAN</summary>No injection/auth/exposure surface; the access boundary was already correct (backend re-derives role per request). This change is client-cache freshness only.</details>
<details><summary>🎛️ UX — CLEAN</summary>The frontend handler patches cache/IDB in place (no forced navigation or modal close). A demoted admin's gated UI updating live is the intended behavior, not a regression.</details>
<details><summary>📱 Mobile — CLEAN</summary>Backend-only change; no layout/touch/viewport surface.</details>

**Dismissed:** "wrapping the upsert in `withTransaction` rolls back the mirror write if the outbox insert throws" (82) — this is the mandated outbox-in-transaction pattern (INV-4/INV-7) used by every write path in the codebase; atomic domain-write + event is the intended behavior, not a regression. A permanent outbox-insert failure is not a realistic mode for a plain insert into the outbox table.

---

PR #934, opened 2026-06-14
feat(agents): C-2a — episode-aware context hydration policy for DMs

kristofferremback:
**Design doc:** `docs/plans/agent-runtimes-unification-redesign.md` — agent-runtimes §2.5 (C-2 / Hydrate) and §2.8 Q7+Q8. No Linear ticket.

## Problem

A companion agent's context window is a flat `MAX_CONTEXT_MESSAGES = 20` applied identically by every stream-type builder (`context-builder.ts`), and prior turn digests (C-1) are injected unconditionally for the stream/persona (`companion/context.ts` → `loadTurnDigestPromptBlock`). For a DM — the one unbounded, never-ending agent surface — that means the agent always carries forward the last sessions' tool-work digest chain, even across a multi-week gap that a human reads as a brand-new conversation. §2.8 Q8 ruled the DM episode boundary; nothing implemented it. There was also no seam to resolve "which window / does continuity carry" per turn — the §2.5 `Hydrate` step the redesign introduces.

## Solution

Introduce the per-turn `ContextWindowPolicy` and resolve it once at the dispatch (`Hydrate`) seam, then use it to gate digest carry on the DM episode boundary. The boundary is anchored on the **sequential window**, not the async `conversations` segmenter: a DM continues its prior episode only while the prior completed session's cursor is still inside the window about to be built.

### How it works

1. **`resolveContextWindowPolicy(db, { stream, maxMessages? })`** (`context-window-policy.ts`) returns `{ episode, maxMessages, carryDigests }`. Bounded surfaces (scratchpad / thread / channel) → `{ kind: "stream" }`, `carryDigests: true` — one continuous episode (Q7).
2. **DM episode (Q8):** read the prior **completed** session via `AgentSessionRepository.findLatestCompletedByStream` — which skips the in-flight RUNNING session the turn inserts before it builds context, so the policy reads the *prior* episode's `lastSeenSequence`, not the current one. Compare it to `MessageRepository.findWindowFloorSequence(stream, maxMessages)` (the oldest sequence inside a budgeted newest-first window, mirroring `list()`'s `deleted_at IS NULL` filter). `lastSeenSequence >= floor` → continue + carry digests; a gap that clears the whole window → fresh, no carry. A `null` floor (stream smaller than the budget) → the window covers everything → continue. No prior completed session → fresh (no digests exist to carry).
3. **`persona-agent.ts`** resolves the policy just before `buildAgentContext` and passes it through. `buildAgentContext` threads `policy.maxMessages` into `buildStreamContext` (replacing the hardcoded constant) and gates the digest block: `policy.carryDigests ? loadTurnDigestPromptBlock(...) : null`.

### Key design decisions

**1. Anchor on the sequence, not on conversation identity.** The DM episode could have keyed on the `conversations` segmenter (Q8's original framing). It doesn't: extraction is async/debounced and membership is sparse (a topic can own messages 4 and 6 but not 5), so consuming `message_ids` directly would block the turn and produce a discontiguous context. The sequential window is the reliable, non-blocking base; the conversation is demoted to a (deferred) highlight. Anchoring on the cursor also means a new message that merely references a weeks-old message can never revive that old session.

**2. Bias toward continue (Q8 error asymmetry).** A false *continue* injects a stale digest chain — cheap and self-correcting (the reader block re-verifies, the live turn outweighs system-context, the carry ages out). A false *fresh* drops context the user expected the agent to hold — the exact forgetting C-1/C-2 exist to kill. So the floor comparison is inclusive (`>=`) and only a gap clearing the whole window starts fresh.

**3. `findLatestCompletedByStream`, not `findLatestByStream`.** The existing helper returns the latest session regardless of status; at policy-resolution time that is the just-inserted RUNNING session of the current turn, which would shadow the prior episode. The new helper filters to `COMPLETED`. Covered by the "ignores the in-flight running session" test.

**4. Rolling summary untouched.** Conversation content still folds into `ConversationSummaryService` cumulatively. A fresh episode drops only the prior *tool-work digest* carry, never conversation content — true conversation compaction/reset is a separate mechanism, deferred.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/features/agents/context-window-policy.ts` | `ContextWindowPolicy` type + `resolveContextWindowPolicy` (the `Hydrate` seam) and `DEFAULT_CONTEXT_WINDOW_MESSAGES` |
| `apps/backend/tests/integration/context-window-policy.test.ts` | Integration coverage for the policy (6 tests) |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/agents/session-repository.ts` | Add `findLatestCompletedByStream` (skips the in-flight RUNNING session) |
| `apps/backend/src/features/messaging/repository.ts` | Add `findWindowFloorSequence` (oldest sequence inside a budgeted window; mirrors `list()`'s filter) |
| `apps/backend/src/features/agents/context-builder.ts` | `BuildStreamContextOptions.maxMessages` threaded through all four builders, replacing `MAX_CONTEXT_MESSAGES` |
| `apps/backend/src/features/agents/companion/context.ts` | Accept the policy; pass `maxMessages`; gate the turn-digest block on `policy.carryDigests` |
| `apps/backend/src/features/agents/persona-agent.ts` | Resolve the policy at the `Hydrate` seam and hand it to `buildAgentContext` |
| `apps/backend/src/features/agents/index.ts` | Export the policy API |
| `docs/plans/agent-runtimes-unification-redesign.md` | Reconcile §2.8 Q8 to the sequence-anchored design (sequential window = base, conversations = non-blocking highlight); record the channel→thread cross-surface continuity follow-up |

## Out of scope (deliberate)

- **Conversation highlight** — surfacing the active topic to the model is the only role the conversations system plays in this design; it's an eval-sensitive prompt-format change, kept out of this PR. In c2a the conversation could only ever be a highlight (it can't move the boundary without contradicting bias-to-continue), so deferring it costs no boundary behavior.
- **C-2b** token-budget window fill (the `truncateMessages` newest-first budget replacing the message count) — the seam is in place; the budget stays a message count here.
- **C-2c** in-enclave sealed rolling summary — blocked on §2.8 Q6 (unbuilt). The enclave path (`enclave-system-prompt.ts`) is untouched and still passes `null` for the summary by construction.
- **Channel→thread** cross-surface episode continuity — a conversation that begins in a channel and continues in a spawned thread has two episode keys; recorded as a doc follow-up, not stitched here.

## Test plan

- [x] `bun test tests/integration/context-window-policy.test.ts` — 6 pass (continue/fresh inside vs outside window, no-prior-session, sub-window history, in-flight-session shadowing guard)
- [x] `bun test tests/integration/context-builder.test.ts tests/integration/agent-session.test.ts` — 31 pass (no regression from the `maxMessages` threading)
- [x] `bun test` agent prompt + access-scope suites (enclave-system-prompt, system-prompt, message-format, turn-digests, agent-access-scope, stream-persona-participants) — 30 pass
- [x] `tsc --noEmit` backend clean; full pre-commit suite green (14-workspace typecheck, `astro check`, `generate-api-docs --check`)
- [ ] No end-to-end DM turn run against a live model — the digest-gating is verified at the policy + wiring level (the gate is `policy.carryDigests ? load : null`); a full `PersonaAgent.run` harness was not exercised

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01QNwKMU7a4ts7gZAY3iWAjM)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/934"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784029834&installation_id=129946197&pr_number=934&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F934&signature=27c7c75af4a8200ba3b0654be8e1cb6f1224cd0e4c8c9f6f5fb76ad6be499040"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/companion/context.ts
+++ b/apps/backend/src/features/agents/companion/context.ts
@@ -12,6 +12,7 @@ import { AttachmentRepository } from "../../attachments"
 import { StreamRepository, type Stream } from "../../streams"
 import { awaitAttachmentProcessing } from "../../attachments"
 import { buildStreamContext, type StreamContext } from "../context-builder"
+import type { ContextWindowPolicy } from "../context-window-policy"
 import type { ConversationSummaryService } from "../conversation-summary-service"
 import { buildSystemPrompt } from "./prompt/system-prompt"
 import { loadTurnDigestPromptBlock } from "./turn-digests"
@@ -35,6 +36,12 @@ export interface ContextParams {
   messageId: string
   persona: Persona
   trigger?: typeof AgentTriggers.MENTION
+  /**
+   * Per-turn hydration policy, resolved at the dispatch (`Hydrate`) seam by
+   * `resolveContextWindowPolicy`. Fixes the window budget and whether prior
+   * turn digests carry into this turn (DM episode recency, §2.8 Q7/Q8).
+   */
+  policy: ContextWindowPolicy
   /** Invocation time override for deterministic evals/tests. Production uses Date.now(). */
   currentTime?: Date
 }
@@ -93,7 +100,7 @@ async function resolveScratchpadCustomPrompt(
  */
 export async function buildAgentContext(deps: ContextDeps, params: ContextParams): Promise<AgentContext> {
   const { db, userPreferencesService, conversationSummaryService } = deps
-  const { workspaceId, streamId, stream, messageId, persona, trigger, currentTime } = params
+  const { workspaceId, streamId, stream, messageId, persona, trigger, policy, currentTime } = params
 
   const triggerMessage = await MessageRepository.findById(db, messageId)
   const invokingUserId = triggerMessage?.authorType === AuthorTypes.USER ? triggerMessage.authorId : undefined
@@ -137,6 +144,7 @@ export async function buildAgentContext(deps: ContextDeps, params: ContextParams
   const streamContext = await buildStreamContext(db, stream, {
     preferences,
     currentTime,
+    maxMessages: policy.maxMessages,
     triggerMessageId: messageId,
     includeAttachments: true,
   })
@@ -295,11 +303,18 @@ export async function buildAgentContext(deps: ContextDeps, params: ContextParams
   // rule. Appended after the base prompt so both first-party drivers fold the
   // identically-formatted block in at the same point (the enclave appends the
   // same formatter's output in run-turn).
-  const turnDigestBlock = await loadTurnDigestPromptBlock(db, {
-    streamId: stream.id,
-    personaId: persona.id,
-    accessibleStreamIds,
-  })
+  //
+  // Gated on the episode boundary (§2.5 / §2.8 Q8): a DM turn that opens a
+  // fresh episode — the prior session's cursor fell outside this window — must
+  // not carry that episode's digest chain. Bounded surfaces always carry
+  // (`policy.carryDigests` is true for them).
+  const turnDigestBlock = policy.carryDigests
+    ? await loadTurnDigestPromptBlock(db, {
+        streamId: stream.id,
+        personaId: persona.id,
+        accessibleStreamIds,
+      })
+    : null
 
   const composeSystemPrompt = (tools: AgentTool[]): string => {
     let systemPrompt = buildSystemPrompt(
--- a/apps/backend/src/features/agents/context-builder.ts
+++ b/apps/backend/src/features/agents/context-builder.ts
@@ -24,6 +24,7 @@ import { UserRepository } from "../workspaces"
 import { AttachmentRepository } from "../attachments"
 import { AttachmentExtractionRepository, type PdfMetadata, type PdfSection } from "../attachments"
 import { getUtcOffset, type TemporalContext, type ParticipantTemporal } from "../../lib/temporal"
+import { DEFAULT_CONTEXT_WINDOW_MESSAGES } from "./context-window-policy"
 
 /**
  * A participant in a stream (user or persona).
@@ -132,14 +133,19 @@ export interface StreamContext {
   participantTimezones?: ParticipantTemporal[]
 }
 
-const MAX_CONTEXT_MESSAGES = 20
-
 /**
  * Options for building stream context with temporal information.
  */
 export interface BuildStreamContextOptions {
   /** User preferences (used for temporal context) */
   preferences?: UserPreferences
+  /**
+   * Newest-first message budget for the conversation window. Resolved once at
+   * the dispatch (`Hydrate`) seam by `resolveContextWindowPolicy` and handed in;
+   * defaults to `DEFAULT_CONTEXT_WINDOW_MESSAGES` for callers that build context
+   * outside a turn (evals, the enclave prompt assembly).
+   */
+  maxMessages?: number
   /** Current time at invocation (for deterministic testing) */
   currentTime?: Date
   /** Trigger message ID (for determining attachment detail levels) */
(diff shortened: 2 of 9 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (minor findings, all resolved in `d645e4e`)

Backend-only change; UX and Mobile are N/A. Data-Flow and Correctness both traced the floor math and the session lifecycle end-to-end and confirmed the core mechanism is correct — `lastSeenSequence` round-trips, the in-flight RUNNING session is correctly excluded, `findWindowFloorSequence` has exact filter parity with `MessageRepository.list`, and the continue/fresh simulations behave as intended.

Three minor hardening findings surfaced and were **fixed in `d645e4e`** before this report:

1. `apps/backend/src/features/agents/context-window-policy.ts` — **bug (80):** a `maxMessages` of 0/negative made `findWindowFloorSequence` return `null` → read as "whole stream fits in window" → continue, but `list(limit:0)` is an empty window. Unreachable from the default-20 production path, but the API exposes `maxMessages`. Fixed by clamping to `Math.max(1, …)` + a regression test.
   https://github.com/threahq/threa/blob/d645e4e55575dd37e41413917b73536b61bc6068/apps/backend/src/features/agents/context-window-policy.ts#L62-L66
2. `apps/backend/src/features/agents/persona-agent.ts` — **claude-md (85):** the two policy reads ran against `pool` separately. Wrapped in `withClient` (a discrete read-pair with no AI work — unlike `buildAgentContext`, which must hold no connection across its rolling-summary AI call, INV-41).
   https://github.com/threahq/threa/blob/d645e4e55575dd37e41413917b73536b61bc6068/apps/backend/src/features/agents/persona-agent.ts#L307-L315
3. `apps/backend/src/features/agents/session-repository.ts` — **data-flow (82):** `findLatestCompletedByStream` ordered by `created_at DESC` only. Added `, id DESC` as a deterministic tiebreaker.
   https://github.com/threahq/threa/blob/d645e4e55575dd37e41413917b73536b61bc6068/apps/backend/src/features/agents/session-repository.ts#L555-L557

Dismissed (with reasoning):

- **Exclude FAILED sessions (design, 88)** — turn digests are only ever written by *completed* sessions (`persistTurnDigest` runs after a successful `runTurn`; `findRecentDigestStepsByStream` filters `status = COMPLETED`), so a failed session has no digest to carry, and §2.5 specifies "the latest completed agent session's `lastSeenSequence`." As-designed.
- **`stream` vs `sessionStreamId` at `persona-agent.ts:307` (data-flow, 80)** — the policy intentionally resolves against the same `stream` the context build uses (channel mentions build channel context + load channel digests). Switching to `sessionStreamId` would *create* the divergence the finding warns about.
- **`findWindowFloorSequence` mirror-drift (design, 82)** — Data-Flow verified exact filter parity with `list()` today; the doc comment already states the mirror. No drift to fix now.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches the agreed C-2a scope: sequence-anchored DM episode boundary + digest gating + the policy seam; conversation highlight, C-2b token budget, and C-2c sealed summary deliberately deferred.</details>
<details><summary>🔍 Bugs [1 — fixed]</summary>maxMessages-clamp edge case, fixed in d645e4e with a regression test.</details>
<details><summary>🔁 Data Flow [2 — 1 fixed, 1 dismissed]</summary>Tiebreaker fixed; stream-vs-sessionStreamId dismissed (consistent with context build). Full lifecycle trace confirmed lastSeenSequence round-trip + window parity correct.</details>
<details><summary>📋 CLAUDE.md Compliance [1 — fixed]</summary>withClient wrap for the two related reads (INV-30 / Backend Architecture pattern).</details>
<details><summary>🏗️ Design [2 — dismissed]</summary>Exclude-FAILED and floor mirror-drift, both dismissed with reasoning above.</details>
<details><summary>🔒 Security [CLEAN]</summary>SQL parameterized via tagged `sql` templates; OFFSET interpolation safe; stream/session reads are id-keyed consistent with existing repo patterns.</details>
<details><summary>🎛️ UX [N/A]</summary>Backend-only change, no user-facing surface.</details>
<details><summary>📱 Mobile [N/A]</summary>Backend-only change.</details>

---

PR #935, opened 2026-06-14
feat(agents): C-2b — token-budgeted context window fill for companion

kristofferremback:
**Design doc:** `docs/plans/agent-runtimes-unification-redesign.md` — §2.5 C-2, §2.8 Q7/Q8. Follow-on to C-2a (#934).

## Problem

After C-2a, a companion turn's conversation window is a hard message count: `policy.maxMessages` (default 20) becomes a SQL `LIMIT` in each `context-builder.ts` builder. That's the "shallow, cliff-edged window" the design calls C-2 (§1.7 / §2.5): 20 messages, then silent forgetting, with no relation to how much the model can actually hold. A scratchpad of many short one-line messages loses its beginning after a few exchanges.

The rolling-summary half of C-2 already exists and is wired on the companion path (`ConversationSummaryService.updateForContext`, called from `companion/context.ts`) — it summarizes everything older than the oldest kept message. What was missing is the budgeted window that decides where "oldest kept" falls.

## Solution

Replace the count cliff with a char-budgeted, newest-first window on the companion/plaintext path, and let the existing rolling-summary fold compose on top of it. The `ContextWindowPolicy` seam C-2a introduced is the insertion point: it gains a `maxChars` budget alongside a deepened candidate fetch ceiling. The window is filled newest-first under the char budget; whatever falls out is already picked up by the rolling summary because that service keys off the oldest kept message.

Scope matches C-2a: companion only. The enclave window stays count-based until the in-enclave sealed rolling summary (C-2c, §2.8 Q6) lands.

### How it works

1. **`ContextWindowPolicy.maxChars`** (`context-window-policy.ts`) — new required field, default `DEFAULT_CONTEXT_WINDOW_CHARS = 80_000`. `resolveContextWindowPolicy` carries it through all three return shapes; both budget params are clamped `Math.max(1, …)`.
2. **Deepened ceiling** — `CONTEXT_WINDOW_CANDIDATE_CEILING = 200` is the new default for the policy's `maxMessages`. With the count no longer the limiter, this caps how many messages are ever fetched before the char trim (fetch-volume guard) and doubles as the DM episode boundary's recency horizon.
3. **`trimToCharBudget(messages, maxChars)`** (`context-builder.ts`) — generic over `{ contentMarkdown: string }`, keeps the newest suffix that fits, always keeping at least the newest message (the trigger). Applied in `buildStreamContext` after the per-type builder returns but **before** attachment/quote/shared-message enrichment, so enrichment only runs on kept rows. Gated on an opt-in `maxChars`: callers that omit it (evals, `enclave-system-prompt.ts`) keep pure count behavior.
4. **Companion wiring** (`companion/context.ts`) — passes `maxChars: policy.maxChars` to `buildStreamContext`. `oldestKeptSequence` is the first of the trimmed set, so `ConversationSummaryService` folds everything older with no change to that service.

### Key design decisions

**1. Char budget, not a real tokenizer** — express the budget in chars using the project's ~4-chars-per-token heuristic (the same unit `MAX_MESSAGE_CHARS`, the 400k `AgentRuntime` clamp, already uses). A real tokenizer is model-specific and would add a dependency that the enclave-over-OpenRouter path can't carry uniformly; INV-36 says don't introduce it without a demonstrated need. Kris's call among the options.

**2. Trim at the fetch layer on raw `Message[]`, pre-enrichment** — the alternative (trim formatted `ModelMessage[]` after enrichment, reusing `truncateMessages`) accounts for enrichment growth exactly but loses `sequence`/`streamId`, so the rolling-summary boundary and cross-stream filtering would have to be re-derived. Trimming on raw markdown keeps `oldestKeptSequence` a clean single sequence and lets the summary fold compose untouched; the 400k clamp in `AgentRuntime` is the backstop for enrichment that inflates a kept message. Kris's call.

**3. Enclave stays count-based** — deepening the enclave's window without the sealed in-enclave rolling summary (C-2c) would just move the cliff, and competes with the 48MB sealed-assignment budget (E2EE-23). Decoupling here mirrors C-2a's companion-only discipline. The shared `DEFAULT_CONTEXT_WINDOW_MESSAGES = 20` (the builder's no-arg fallback, used by `enclave-system-prompt.ts` and evals) is deliberately **left unchanged**, so widening the companion window can't silently deepen those out-of-turn fetches.

**4. DM boundary horizon tracks the ceiling, biasing toward continue** — the boundary uses `findWindowFloorSequence(maxMessages)` where `maxMessages` is now the ceiling, i.e. the window's widest extent: "only a gap that clears the whole window starts fresh" (§2.8 Q8). A false-continue carries a stale digest chain that self-heals (the reader block re-verifies, the live turn outweighs system-context); a false-fresh drops context the user expected. The costlier mistake is the false-fresh, so the deeper horizon is the safe direction.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/agents/context-window-policy.ts` | Add `maxChars` to `ContextWindowPolicy` + params; new `CONTEXT_WINDOW_CANDIDATE_CEILING` (200) and `DEFAULT_CONTEXT_WINDOW_CHARS` (80k); `resolveContextWindowPolicy` carries+clamps the budget; DM-boundary docstring updated for the ceiling horizon. `DEFAULT_CONTEXT_WINDOW_MESSAGES` (20) kept as the builder fallback. |
| `apps/backend/src/features/agents/context-builder.ts` | New exported `trimToCharBudget`; opt-in `maxChars` on `BuildStreamContextOptions`; applied in `buildStreamContext` before enrichment. |
| `apps/backend/src/features/agents/companion/context.ts` | Pass `policy.maxChars` to `buildStreamContext`. |
| `apps/backend/src/features/agents/index.ts` | Export `trimToCharBudget`, `CONTEXT_WINDOW_CANDIDATE_CEILING`, `DEFAULT_CONTEXT_WINDOW_CHARS`. |
| `docs/plans/agent-runtimes-unification-redesign.md` | Mark C-2b shipped in §2.8 Q8; fix the stale §2.8 Q7 "`ContextWindowPolicy` does not exist yet" note. |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/features/agents/context-builder.unit.test.ts` | Pure unit contract for `trimToCharBudget` (empty/single/all-fit/newest-suffix/oversized-kept). |

## Out of scope (deliberate)

- **Enclave token-budget window** — waits for C-2c (sealed rolling summary, §2.8 Q6).
- **Conversation highlight** — `findPrimaryByMessageId` topic surfacing + window-floor expand; eval-sensitive (system-prompt/message-format), its own slice.
- **Prompt caching (C-3)** — independent optimization.

## Test plan

- [x] `bun test src/features/agents/context-builder.unit.test.ts` — 4 pass
- [x] `bun test tests/integration/context-window-policy.test.ts tests/integration/context-builder.test.ts` — 15 pass (incl. 2 new budgeted-window cases; C-2a policy tests updated for the new field), against local PG16 + pgvector
- [x] `bun test src/features/agents/conversation-summary-service.test.ts src/features/agents/enclave-system-prompt.test.ts` — 6 pass (adjacent paths)
- [x] Pre-commit suite green: prettier, 14-workspace `tsc --noEmit`, eslint, astro check, `generate-api-docs --check`
- [x] Self-review (author): verified no other code constructs a `ContextWindowPolicy` literal or calls `buildAgentContext` outside the resolver path, so `maxChars` flows automatically; confirmed `enclave-system-prompt.ts` omits `maxChars` and is unaffected; confirmed `MessageRepository.list` returns ascending so `trimToCharBudget`'s newest-suffix semantics and `oldestKeptSequence` hold.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01JtsfSnwyiwimWRu6vrTKd3)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/935"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784037828&installation_id=129946197&pr_number=935&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F935&signature=c084b3329aa5941ea701541af9c635503c87928b60ec6a476e48704d73c635a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/companion/context.ts
+++ b/apps/backend/src/features/agents/companion/context.ts
@@ -145,6 +145,7 @@ export async function buildAgentContext(deps: ContextDeps, params: ContextParams
     preferences,
     currentTime,
     maxMessages: policy.maxMessages,
+    maxChars: policy.maxChars,
     triggerMessageId: messageId,
     includeAttachments: true,
   })
--- a/apps/backend/src/features/agents/context-builder.ts
+++ b/apps/backend/src/features/agents/context-builder.ts
@@ -122,6 +122,15 @@ export interface StreamContext {
   participants?: Participant[]
   /** Conversation history - messages in chronological order, may include attachment context */
   conversationHistory: MessageWithAttachments[]
+  /**
+   * Number of leading `conversationHistory` entries that are pinned context and
+   * must survive windowing (the C-2b char-budget trim). The thread builder
+   * prepends the root/parent message as the thread anchor and sets this to 1;
+   * other surfaces leave it unset (treated as 0). Declared by the builder so
+   * the shared trim in `buildStreamContext` preserves the anchor without
+   * re-deriving it from message shape.
+   */
+  pinnedHistoryPrefix?: number
   /** For threads: path from current thread up to root channel */
   threadContext?: {
     depth: number
@@ -261,16 +270,14 @@ export async function buildStreamContext(
   // runs on kept messages. Older messages that fall out fold into the rolling
   // conversation summary downstream (it keys off the oldest kept message).
   //
-  // The thread builder prepends the root/parent message as the thread anchor —
-  // the only history entry from a different stream (`streamId !== stream.id`).
-  // Pin it so a long thread can never trim away its own anchor; the trim runs
-  // over the remaining in-stream messages. A no-op for non-thread surfaces,
-  // whose history is all from `stream.id`.
+  // `pinnedHistoryPrefix` leading entries are anchors the builder declared must
+  // survive (the thread root); trim only the rest so a long thread can never
+  // trim away its own anchor.
   if (options?.maxChars !== undefined) {
-    const history = context.conversationHistory
-    const anchor = history.length > 0 && history[0].streamId !== stream.id ? history[0] : undefined
-    const trimmed = trimToCharBudget(anchor ? history.slice(1) : history, options.maxChars)
-    context.conversationHistory = anchor ? [anchor, ...trimmed] : trimmed
+    const pinned = context.pinnedHistoryPrefix ?? 0
+    const head = context.conversationHistory.slice(0, pinned)
+    const trimmed = trimToCharBudget(context.conversationHistory.slice(pinned), options.maxChars)
+    context.conversationHistory = pinned > 0 ? [...head, ...trimmed] : trimmed
   }
 
   // Enrich with attachment context if requested
@@ -440,6 +447,8 @@ async function buildThreadContext(
       slug: stream.slug,
     },
     conversationHistory,
+    // The prepended parent is a pinned anchor: keep it through the C-2b trim.
+    pinnedHistoryPrefix: parentMessage ? 1 : 0,
     threadContext: {
       depth: threadPath.length,
       path: threadPath,
--- a/apps/backend/src/features/agents/context-builder.ts
+++ b/apps/backend/src/features/agents/context-builder.ts
@@ -146,6 +146,17 @@ export interface BuildStreamContextOptions {
    * outside a turn (evals, the enclave prompt assembly).
    */
   maxMessages?: number
+  /**
+   * Char budget for the verbatim conversation window (C-2b). When set, the
+   * fetched history (capped at `maxMessages`) is trimmed newest-first to this
+   * many chars of raw message markdown, replacing the message-count cliff with
+   * a budgeted window. Omitted by out-of-turn callers (evals, enclave-prompt
+   * assembly), which keep the pure message-count behavior. Applied before
+   * attachment/quote/shared-message enrichment, so enrichment only runs on kept
+   * messages; the 400k `MAX_MESSAGE_CHARS` clamp in `AgentRuntime` is the
+   * backstop for enrichment growth.
+   */
+  maxChars?: number
   /** Current time at invocation (for deterministic testing) */
   currentTime?: Date
   /** Trigger message ID (for determining attachment detail levels) */
@@ -165,6 +176,30 @@ export interface BuildStreamContextOptions {
    * and included as base64 data URLs so the model can see them.
    */
   loadImages?: boolean
+}
+
+/**
+ * Trim a chronological (oldest-first) message list to the newest suffix that
+ * fits within `maxChars` of raw `contentMarkdown`, always keeping at least the
+ * newest message (the trigger). This is the C-2b budgeted-window fill: a deeper,
+ * size-bounded window in place of a fixed message count. Generic over any row
+ * carrying `contentMarkdown` so it can run before the history is enriched into
+ * `MessageWithAttachments`.
+ */
+export function trimToCharBudget<T extends { contentMarkdown: string }>(messages: T[], maxChars: number): T[] {
+  if (messages.length <= 1) return messages
+
+  let total = 0
+  let kept = 0
+  for (let i = messages.length - 1; i >= 0; i--) {
+    const len = messages[i].contentMarkdown.length
+    // Always keep the newest message; stop once another would overflow.
+    if (kept > 0 && total + len > maxChars) break
+    total += len
+    kept++
+  }
+
+  return kept >= messages.length ? messages : messages.slice(messages.length - kept)
 }
 
 /**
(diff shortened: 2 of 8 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (no bugs/correctness/security/data-flow issues; 2 minor design notes, 1 actioned)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x4 (spec, correctness, design, data-flow). UX + Mobile N/A (backend-only).

Reviewed at `1b6a3f8a`. Found 2 design notes:

1. `apps/backend/src/features/agents/index.ts:155` — `trimToCharBudget` was re-exported from the agents barrel but has no cross-feature consumer (the unit test imports it relatively from `./context-builder`). Speculative public surface (INV-36). **Fixed in `8e57095c`** — removed from the barrel; the function stays exported from its module for the test.
   https://github.com/threahq/threa/blob/1b6a3f8a2df0b32f58b60bb08991457f38846d8b/apps/backend/src/features/agents/index.ts#L154-L156

2. `apps/backend/src/features/agents/context-builder.ts:271` — the anchor pin detects the prepended thread root via `history[0].streamId !== stream.id`, an implicit structural contract with `buildThreadContext`. The reviewer suggested returning the anchor as an explicit `StreamContext` field instead. **Kept as-is, with reasoning:** the detection is precise today (`buildThreadContext` at `:432` is the only builder that prepends a cross-stream row; the other three return single-stream history), and it is documented in-comment. Centralizing the trim in `buildStreamContext` keeps it on one shared path (INV-29/43); pulling the anchor into a dedicated field would split the anchor out of the formatted history it currently flows through and add surface for a hypothetical future prepend (INV-36). Revisit if a second surface ever prepends cross-stream rows.
   https://github.com/threahq/threa/blob/1b6a3f8a2df0b32f58b60bb08991457f38846d8b/apps/backend/src/features/agents/context-builder.ts#L263-L274

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches the agreed C-2b scope: char budget (not a tokenizer), applied at the fetch layer on raw `Message[]` before enrichment, companion/plaintext only, builder-fallback `DEFAULT_CONTEXT_WINDOW_MESSAGES` left unchanged so enclave-prompt/eval callers aren't deepened. Policy field threaded through resolver → `buildAgentContext` → `buildStreamContext`; rolling-summary boundary (`oldestKeptSequence`) preserved.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>`trimToCharBudget` verified across empty/single/exactly-fits/overflow-on-newest/off-by-one. Anchor pin correct and chronological order preserved. `Math.max(1, …)` clamps correct.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>`MessageRepository.list` returns ascending; `trimToCharBudget` keeps the newest suffix, so `keptMessages[0]` stays the oldest kept and the rolling-summary range `[lastSummarizedSequence+1, oldestKeptSequence-1]` has no gap/overlap. Thread anchor is filtered out of the summary scope (different streamId), consistent with pre-C-2b. DM boundary using the ceiling while the verbatim window is char-trimmed is the intentional, documented bias-to-continue (§2.8 Q8), not a correctness bug.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No violations introduced (finding 1 above was the only INV-36 surface note, now fixed).</details>
<details><summary>🏗️ Design [2 notes]</summary>Both listed above (1 fixed, 1 kept with reasoning). No config sprawl, no N+1, no parallel-implementation issues; `trimToCharBudget` (raw-markdown window fill) and `truncateMessages` (post-enrichment 400k safety clamp) are distinct stages, not duplicates.</details>
<details><summary>🔒 Security [CLEAN]</summary>The trim only drops messages; no cross-stream/cross-workspace exposure introduced. `stream` arrives pre-scoped from the handler layer.</details>
<details><summary>🎛️ UX [N/A]</summary>Backend-only change.</details>
<details><summary>📱 Mobile [N/A]</summary>Backend-only change.</details>

---

PR #936, opened 2026-06-14
fix(sync): public-root catch-up leg + notification-level outbox broadcast

kristofferremback:
Two sync-engine "never-miss / always-up-to-date" gaps from the sync-engine audit, each a case where a connected client kept stale state until reconnect. FIX #1 (control-plane role fan-out) shipped separately as #933; this PR is the remaining two, kept together on one branch per the assigned-branch constraint.

### Problem

`SyncLogRepository.listEntriesForUser` builds its `visible_streams` CTE from `stream_members` rows only. It replicates INV-62's thread→root leg but **not** the "a public root channel grants read access without a `stream_members` row" leg. So workspace sync-log catch-up dropped public-channel events (and their threads') for any reader who never joined the channel — the live socket path lets a non-member into a public room, but a reconnect/catch-up over the log silently omitted that history. This diverges from the canonical `streamAccessPredicateSql` in `streams/access.ts`, which every other set-based access check (search, attachments, sharing, labels) routes through and which *does* have the public-root leg.

### Solution

Add two `UNION ALL` legs to `visible_streams`, mirroring `streamAccessPredicateSql`'s two public branches:

- public root streams: `s.root_stream_id IS NULL AND s.visibility = 'public'`
- threads whose root is public: `JOIN streams root_s ON root_s.id = s.root_stream_id WHERE root_s.visibility = 'public'`

Both carry `bound = 0`. The `bound` is the join-position floor that stops a member from replaying a stream's pre-join history; a public channel's history is readable regardless of when — or whether — the viewer joined, so there is nothing to gate behind a join position and no pre-join leak to guard. The existing `OR EXISTS (… AND l.sync_id > vs.bound)` admit-if-any-grant-passes dedup absorbs the overlap when a viewer is *also* a member of a public channel (the bound-0 leg simply always passes).

`Visibilities.PUBLIC` is interpolated from `@threa/types` rather than a literal `'public'` (INV-33).

### Key design decision

**Inline replication over wholesale `streamAccessPredicateSql` reuse.** The CTE needs a per-stream `bound` column for the join-position gate, which the boolean `EXISTS` predicate doesn't expose. CLAUDE.md's INV-62 rule allows exactly this: *reuse the helper OR replicate the thread→root rule and pair it with a test*. The two legs are a direct structural mirror of the predicate's public branches, and the test covers the case a membership-only filter gets wrong.

The stale doc comment above the method (which stated the rule as "member of it OR member of its root" and omitted public-without-membership) is corrected (INV-25).

### Problem

`StreamService.setNotificationLevel` mutated the `stream_members` row but emitted no outbox event. A mute/notify change made in one tab stayed stale in the user's other tabs and devices until the next reconnect — `workspace-sync.ts`'s `setMutedState` only runs on the reconnect bootstrap merge, so nothing updated the badge live.

### Solution

A new author-scoped event type, `stream:notification_level_updated`, modeled exactly on `stream:read` (the existing author-scoped read-state event):

1. **Type system (end-to-end).** Added to `OutboxEventType`, `OutboxEventPayloadMap`, the new `StreamNotificationLevelUpdatedOutboxPayload` (`{ workspaceId, authorId, streamId, notificationLevel }`), and the `AuthorScopedEventType` union + `AUTHOR_SCOPED_EVENTS` array in `lib/outbox/repository.ts` (INV-31/INV-33).
2. **Routing.** `resolveDeliveryGroups` already maps any author-scoped event to `[userGroup(authorId)]`, so the event reaches only the acting user's sessions and `BroadcastHandler.sequenceEvents` lands it in the sync log under `user:${authorId}` — which means catch-up replays it too, no extra wiring.
3. **Emit.** `setNotificationLevel` now emits in the **same transaction** as the membership update (INV-4/INV-7), and only when `StreamMemberRepository.update` actually returned a membership row. It gained a `workspaceId` parameter — the unmute (`level === null`) path never fetched the stream, so the workspace id wasn't otherwise in scope; the handler passes `req.workspaceId`.
4. **Frontend.** `handleStreamNotificationLevelUpdated` writes the new level into the workspace-bootstrap membership row (sidebar mute badge), the stream-bootstrap membership mirror (open stream view), and the `db.streamMemberships` IDB row (survives remount). Registered on the engine's event gate via `socket.on`, so the **one** handler covers both live emit and catch-up `gate.dispatch` — confirmed against `socket-event-gate.ts`, where `.on` registers into the same `handlers` map both paths invoke.

### Key design decision

**Author-scoped, not user-scoped.** The two families differ only in payload field — author-scoped reads `authorId`, user-scoped reads `targetUserId` — and both resolve to `userGroup`. The notification-level choice is the acting user's own per-stream preference reflected back to their other sessions, which is exactly `stream:read`'s shape; reusing that family kept the payload and routing identical to a well-trodden path rather than introducing a `targetUserId` indirection that doesn't fit.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/sync/repository.ts` | FIX #2: two public-root `UNION ALL` legs in the `visible_streams` CTE; corrected the stale visibility doc comment; import `Visibilities` |
| `apps/backend/src/lib/outbox/repository.ts` | FIX #3: new `stream:notification_level_updated` event type, payload interface, payload-map entry, and `AuthorScopedEventType`/`AUTHOR_SCOPED_EVENTS` membership |
| `apps/backend/src/features/streams/service.ts` | FIX #3: `setNotificationLevel` takes `workspaceId` and emits the event in-transaction when a membership row was updated |
| `apps/backend/src/features/streams/handlers.ts` | FIX #3: pass `workspaceId` to `setNotificationLevel` |
| `apps/frontend/src/sync/workspace-sync.ts` | FIX #3: `StreamNotificationLevelUpdatedPayload`, `handleStreamNotificationLevelUpdated`, and the `socket.on`/`socket.off` registration |
| `apps/backend/tests/integration/sync-catch-up.test.ts` | FIX #2: `addRootStream` helper + a public/private-channel visibility test |
| `apps/backend/src/features/streams/service.test.ts` | FIX #3: three `setNotificationLevel` emit tests (scoped `spyOn`) |
| `apps/frontend/src/sync/workspace-sync.test.ts` | FIX #3: gate-dispatched catch-up replay updates the `streamMemberships` mirror |

## Out of scope (deliberate)

- The audit's other findings the user ruled out: workspace settings / feature-flags freshness (query-cache only, by design), `link_preview:dismissed` / `bot_runtime:presence` (ephemeral), `useStreamSocket` fire-and-forget join (maintainability only), and `applyMembershipRemoval` emitting nothing (full removal emits `workspace_user:removed` elsewhere). Left untouched.
- No schema change: both fixes are read-path (FIX #2) and event-emission (FIX #3) only.

## Test plan

- [x] Backend `bun test src/features/streams/service.test.ts tests/integration/sync-catch-up.test.ts` — 66 pass (3 new notification-level emit tests; 1 new public-root catch-up test)
- [x] Frontend `bunx vitest run src/sync/workspace-sync.test.ts` — 35 pass (1 new gate-dispatched notification-level replay)
- [x] Frontend `bunx vitest run src/sync/sync-engine.test.ts src/sync/socket-event-gate.test.ts` — 62 pass (no regression in the gate that carries the new event)
- [x] Full monorepo lint + per-package `tsc --noEmit` (incl. `tsconfig.tests.json`) — clean, via the pre-commit hook on both commits
- [x] Self-review: claim-verification pass over the diff (routing, in-transaction emit, gate coverage, public-root dedup) — one finding fixed mid-review (the `setNotificationLevel` test literal widened `notificationLevel` to `string` and used a `string` `joinedAt`; typed it as the repository `StreamMember` with a `Date`, caught by the stricter test tsconfig in the hook)
- [ ] No separate parallel reviewer-agent pass — the change is contained (two read/emit paths, fully unit + integration covered) so the manual passes above stood in; flag for the verify-holder session

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01SAsK8GdCubeRx6o1fsefX7)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/936"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784038071&installation_id=129946197&pr_number=936&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F936&signature=686cf63f04a42ed1563c6a4f0d450cbe4c9fd0553db0c3d8877905590c2f08ad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/streams/access.ts
+++ b/apps/backend/src/features/streams/access.ts
@@ -93,13 +93,44 @@ export async function checkStreamAccess(
 }
 
 /**
+ * The canonical "is this *already-resolved effective root* readable by the
+ * user?" leaf: public visibility OR a `stream_members` row on the root. This is
+ * the single rule that both the per-id predicate ({@link streamAccessPredicateSql})
+ * and the workspace catch-up CTE (`features/sync/repository.ts`) reduce a
+ * stream's effective root down to — extracted so the public-without-membership
+ * branch cannot live in one and be silently dropped from the other (the exact
+ * drift that previously shipped: catch-up replicated the thread→root leg but
+ * omitted the public-root leg). The caller resolves the effective root
+ * (top-level stream → itself, thread → its root) and passes its alias; this
+ * fragment only decides readability of that resolved root.
+ *
+ * `rootAlias` is injected raw (squid `raw`) because it is a SQL alias reference,
+ * not a value — it MUST be a trusted constant supplied by call-site code (e.g.
+ * `"eff_root"`), NEVER derived from user input. Built with squid `sql` so it
+ * carries `$1..$k` placeholders that {@link composeSql} renumbers when splicing
+ * it into a larger query.
+ */
+export function rootReadableConditionSql(userId: string, rootAlias: string): QueryConfig {
+  return sql`(
+    ${sql.raw(rootAlias)}.visibility = ${Visibilities.PUBLIC}
+    OR EXISTS (
+      SELECT 1 FROM stream_members
+      WHERE stream_id = ${sql.raw(rootAlias)}.id AND member_id = ${userId}
+    )
+  )`
+}
+
+/**
  * Canonical SQL predicate for the "thread → root" stream-access rule
  * (INV-62), as a reusable `EXISTS` fragment correlated to a stream-id
  * column. This is the single source of truth for the *set-based* form of
  * the same three rules `checkStreamAccess` enforces per-id (workspace
  * boundary, public root grants read without a `stream_members` row, threads
- * inherit from their root). When the thread→root resolution rule changes,
- * this fragment and the per-id helpers update together.
+ * inherit from their root). The effective root is resolved with
+ * `COALESCE(root_stream_id, id)` (a top-level stream is its own root, a thread
+ * defers to its root; a thread whose root row is missing finds no join and is
+ * denied), then {@link rootReadableConditionSql} decides readability — the same
+ * leaf the catch-up CTE applies, so the two cannot drift.
  *
  * The returned `QueryConfig` is spliced into a larger query via
  * {@link composeSql} (squid's own `sql` tag cannot nest fragments — it would
@@ -117,34 +148,14 @@ export async function checkStreamAccess(
  * workspace boundary inside the EXISTS, so it is correct even when the outer
  * query does not otherwise constrain the stream's workspace.
  */
-export function streamAccessPredicateSql(
-  workspaceId: string,
-  userId: string,
-  streamIdColumn: string
-): QueryConfig {
-  return sql`EXISTS (
+export function streamAccessPredicateSql(workspaceId: string, userId: string, streamIdColumn: string): QueryConfig {
+  return composeSql`EXISTS (
     SELECT 1
     FROM streams eff_s
-    LEFT JOIN streams eff_root ON eff_root.id = eff_s.root_stream_id
-    WHERE eff_s.id = ${sql.raw(streamIdColumn)}
+    JOIN streams eff_root ON eff_root.id = COALESCE(eff_s.root_stream_id, eff_s.id)
+    WHERE ${sql`eff_s.id = ${sql.raw(streamIdColumn)}`}
       AND eff_s.workspace_id = ${workspaceId}
-      AND (
-        (eff_s.root_stream_id IS NULL AND (
-          eff_s.visibility = ${Visibilities.PUBLIC}
-          OR EXISTS (
-            SELECT 1 FROM stream_members
-            WHERE stream_id = eff_s.id AND member_id = ${userId}
-          )
-        ))
-        OR
-        (eff_s.root_stream_id IS NOT NULL AND eff_root.id IS NOT NULL AND (
-          eff_root.visibility = ${Visibilities.PUBLIC}
-          OR EXISTS (
-            SELECT 1 FROM stream_members
-            WHERE stream_id = eff_s.root_stream_id AND member_id = ${userId}
-          )
-        ))
-      )
+      AND ${rootReadableConditionSql(userId, "eff_root")}
   )`
 }
 
--- a/apps/backend/src/features/streams/handlers.ts
+++ b/apps/backend/src/features/streams/handlers.ts
@@ -759,7 +759,7 @@ export function createStreamHandlers({
 
       await streamService.validateStreamAccess(streamId, workspaceId, userId)
 
-      const membership = await streamService.setNotificationLevel(streamId, userId, data.notificationLevel)
+      const membership = await streamService.setNotificationLevel(workspaceId, streamId, userId, data.notificationLevel)
       if (!membership) {
         return res.status(404).json({ error: "Not a member of this stream" })
       }
--- a/apps/backend/src/features/streams/index.ts
+++ b/apps/backend/src/features/streams/index.ts
@@ -10,6 +10,7 @@ export {
   checkStreamAccess,
   listAccessibleStreamIds,
   resolveEffectiveAccessStream,
+  rootReadableConditionSql,
   streamAccessPredicateSql,
 } from "./access"
 
--- a/apps/backend/src/features/streams/service.test.ts
+++ b/apps/backend/src/features/streams/service.test.ts
@@ -2,7 +2,7 @@ import { afterAll, describe, test, expect, mock, spyOn, beforeEach } from "bun:t
 import type { PoolClient } from "pg"
 import { StreamService } from "./service"
 import { StreamRepository } from "./repository"
-import { StreamMemberRepository } from "./member-repository"
+import { StreamMemberRepository, type StreamMember } from "./member-repository"
 import { StreamEventRepository } from "./event-repository"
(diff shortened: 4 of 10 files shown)


kristofferremback:
### Code review

Reviewed the final diff (4 commits, `7b7ca93`). One real bug was caught and fixed mid-review (see Design evolution); the remaining two findings are non-blocking optimistic-latency notes that match the app's established architecture.

**No blocking issues.**

#### Design evolution (caught in self-review, fixed in `d4bc8de`)
The first version of the frontend handler (`handleStreamNotificationLevelUpdated`) updated only the membership caches. But every mute badge — sidebar, quick-switcher, share modal, tab indicator — renders from `unreadState.mutedStreamIds` (a derived set live-queried from `db.unreadState`), **not** from `streamMemberships.notificationLevel`. So the badge wouldn't have moved live or after reload. Fixed by also moving the muted set in both the `WorkspaceBootstrap` query cache and `db.unreadState`, reusing the existing `setMutedState` helper; the test now asserts `mutedStreamIds` flips on mute and unmute.

#### Non-blocking notes (dismissed — architecture-consistent)
1. `apps/frontend/src/hooks/use-streams.ts:330-347` — `useSetNotificationLevel.onSuccess` updates the membership caches but not `mutedStreamIds`, so the **originating tab's** sidebar badge updates via the socket echo round-trip rather than instantly. This is the app's standard optimistic model: no mutation hook writes `db.unreadState` — IDB live-state is owned by the sync layer (echo handlers). This PR improves the originating tab from update-on-reconnect to update-on-echo; fully-optimistic would require a mutation hook to write IDB, which no other mutation does. Worth a follow-up only if the round-trip flicker is deemed user-visible.
2. `apps/frontend/src/sync/workspace-sync.ts` — the handler's `if (streamType)` / `if (stream && unread)` guards no-op when the stream isn't cached yet. This mirrors the sibling `handleStreamRead` pattern, and a stream absent from `db.streams` isn't rendered in the sidebar anyway (no badge to be stale); it resolves on the next bootstrap.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Event type wired through every registry (`OutboxEventType`, `OutboxEventPayloadMap`, `AuthorScopedEventType` + `AUTHOR_SCOPED_EVENTS`); `rootReadableConditionSql` exported from the streams barrel (INV-52); `setNotificationLevel` `workspaceId` threaded from the handler; socket `on`/`off` both registered.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Predicate rewrite (`COALESCE` effective-root + shared leaf) verified behaviour-equivalent to the old LEFT-JOIN/two-branch form across top-level streams, threads, and dangling roots. `composeSql` parameter renumbering correct (`[workspaceId, userId]` → `$1, $2`); the inner `sql\`eff_s.id = ${sql.raw(...)}\`` wrap is required because `composeSql` only inlines `QueryConfig`-shaped values, not squid raw markers.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Traced all six visibility sequences: non-member never receives private-channel events; root members receive thread events bounded by the root join position; pre-join private history stays hidden while pre-join public history is delivered by design (bound 0); message-move direct thread members keep delivery via the direct-membership leg. The originating-tab echo is idempotent (HTTP `onSuccess` and the echo carry the same committed value); rapid mute→unmute serializes correctly through Dexie.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-52 (barrel import), INV-35/37 (the refactor removes the parallel access-rule implementation), INV-4/7 (outbox emit in the membership transaction), INV-62 (catch-up now resolves access through the canonical shared leaf), INV-48 (scoped `spyOn`), INV-23 (content assertions). No violations.</details>
<details><summary>🏗️ Design [CLEAN]</summary>`rootReadableConditionSql` is a clean shared leaf; the CTE collapses 4 legs → 2 with no scan-cardinality regression (the old version also scanned all workspace streams). No circular import (`access.ts` doesn't import sync; the helper is call-time only). Frontend handler is structurally consistent with `handleStreamRead`.</details>
<details><summary>🔒 Security [CLEAN]</summary>No access-control bypass: the CTE inherited leg excludes private roots for non-members; `rootAlias`/`streamIdColumn` are call-site constants, never user input.</details>
<details><summary>🎛️ UX [2 non-blocking notes]</summary>Both above — originating-tab optimistic latency and the cold-cache guard; both consistent with the established optimistic/sync-layer split.</details>
<details><summary>📱 Mobile [CLEAN]</summary>No layout, breakpoint, touch, or viewport changes.</details>

---

PR #937, opened 2026-06-14
feat(drafts): stage 4b — thread re-pointing for drafts

kristofferremback:
**Plan:** `docs/plans/centralized-draft-system.md` §"Stage 4b". Stages 1–3 + 4a are merged; this is 4b. (No Linear ticket; tracked by the design doc.)

## Problem

A draft scoped to a not-yet-threaded message (`thread:{parentMessageId}`) or to a not-yet-created scratchpad (`stream:{localDraftId}`) was stranded the moment that scope stopped existing:

- **Thread conversion.** When messages are moved onto a target message (`EventService.moveMessagesToThread`), that message becomes a thread. Any reply draft composed against it — possibly by several authors — still pointed at `thread:{targetMessageId}`, a scope no surface renders once the thread stream exists. The reply you were typing vanished from view.
- **Scratchpad promotion.** `promoteDraft` (in `useMessageQueue`) called `purgeScopeDrafts(stream:{draftId})`, which **deleted** every draft in the scratchpad scope — including stash siblings the user hadn't sent. That contradicts the system's cardinal rule: duplicated drafts are acceptable, lost drafts are not.

## Solution

Drafts follow their message. The **server is the re-scope authority on a move** (one set-based UPDATE inside the move transaction, fanned out to each owner's `user:{userId}` room); the **client mirrors the same scope move on promotion** (it owns scratchpad promotion, which never reaches the server until the stream exists). Both reuse one local primitive that moves the draft row and the device-local "loaded" pointer together.

### How it works

1. **`DraftsRepository.rescopeByScope(db, { workspaceId, fromScope, toScope, rootStreamId })`** — a single `UPDATE drafts SET scope, root_stream_id, version = version + 1 WHERE workspace_id = $ AND scope = $ AND deleted_at IS NULL RETURNING …`. No `user_id` filter: a shared `thread:{messageId}` scope holds reply drafts from several authors and every owner's draft follows the message (each still delivered only to its own owner). Set-based (INV-56), no select-then-write (INV-20). Bumping `version` is what makes the client adopt the new scope (a clean local row sees `version > baseVersion`; a row with unpushed edits ignores the echo and its queued push splits CAS-safely).

2. **`moveMessagesToThread`** — after `MessageRepository.moveToStream`, re-scopes `thread:{targetMessageId}` → `stream:{destinationThread.id}` (+ `root_stream_id`) and emits one `draft:upserted` per re-scoped row, `targetUserId = draft.userId`, in the **same transaction** as the move (INV-4/7). `draft:upserted` was already a registered user-scoped outbox event, so no delivery-routing change.

3. **`migrateLocalDraftScope` + `migrateDraftScopeInCache`** (frontend) — re-scope a draft in place (same id). If it was the one checked out into the composer under the old scope, its `composerLoaded` pointer moves to the new scope so the reply isn't stranded under a scope nothing renders. Row write + pointer move share one IDB transaction and emit one cache signal, so a live reader never sees the draft loaded under two scopes or under none.

4. **`applyDraftUpserted`** — the clean-accept branch now branches on scope: `local.scope !== draft.scope` takes the scope-move path, same-scope edits stay a plain `putLocalDraft`. The dirty-row and stale-echo branches are unchanged (the server still arbitrates drift).

5. **`rescopeScopeDrafts` + `promoteDraft`** — promotion re-points the scratchpad scope's surviving drafts onto the real stream and pushes each (instead of purging), then kicks the op queue. The just-sent loaded draft was already resolved by the send path, so this carries the stash siblings.

6. **`draftStreamScope` / `draftThreadScope`** (`packages/types`) — shared builders for the scope string (INV-33), so the backend re-scope and the composer's `getDraftMessageKey` construct identical scopes. A re-scope on one side lines up with the other's stored rows precisely because the format lives in one place.

### Key design decisions

**1. Server owns re-scope on move; client owns it on promotion.** A move is a server operation touching other users' drafts, so it must re-scope server-side and broadcast. Promotion is a purely local transition (the stream doesn't exist server-side until promotion creates it), so the client re-scopes locally and pushes — there is nothing server-side to re-point.

**2. Re-scope is multi-user.** Drafts are private per author (INV-8), but a thread target can hold reply drafts from many authors. The UPDATE matches by workspace+scope only and emits one event per owner so each draft reaches exactly its author.

**3. Promotion re-scopes instead of purging.** The old `purgeScopeDrafts` deleted stash siblings — a loss. Re-scope + push preserves them. Strictly safer: if the loaded draft were somehow not yet resolved, re-scoping keeps it rather than deleting it. Stream promotion creates plaintext streams only (`PendingStreamCreation` carries no E2E fields), so this never pushes plaintext to an encrypted stream.

**4. Client still sends `rootStreamId: null` on push.** The server owns `root_stream_id` on re-scope; the client never needs to send it (drafts are private and listed by owner regardless of scope). Comment in `executeDraftUpsert` updated to say so.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/drafts/repository.ts` | `rescopeByScope` — set-based multi-user re-scope, version bump, `RETURNING` rows |
| `apps/backend/src/features/messaging/event-service.ts` | `moveMessagesToThread` re-scopes reply drafts + emits one `draft:upserted` per owner in the move txn |
| `apps/frontend/src/sync/draft-sync.ts` | `migrateLocalDraftScope`; `applyDraftUpserted` scope-move branch; `rootStreamId` comment |
| `apps/frontend/src/stores/draft-store.ts` | `migrateDraftScopeInCache` — re-scope + pointer move in one cache signal |
| `apps/frontend/src/hooks/use-draft-message.ts` | `rescopeScopeDrafts`; `getDraftMessageKey` uses shared builders |
| `apps/frontend/src/hooks/use-message-queue.ts` | `promoteDraft` re-scopes (was purge) + kicks op queue; widened `syncEngine` type |
| `packages/types/src/api.ts` / `index.ts` | `draftStreamScope` / `draftThreadScope` builders (INV-33) |
| `apps/backend/.../repository.test.ts`, `tests/integration/message-move.test.ts`, frontend `draft-sync.test.ts`, `use-draft-message-sync.test.ts`, `use-message-queue.test.ts` | Tests (below) |

## Out of scope (deliberate)

- **E2E drafts (Stage 4c).** `rescopeByScope` re-scopes E2E rows too, but the frontend `applyDraftUpserted` still ignores `contentJson === null` rows, and E2E drafts aren't synced before 4c — so an encrypted reply draft doesn't re-point on the client yet. No regression: E2E drafts are local-only today.
- **`thread:{messageId}` re-pointing on first *reply send*.** Only the move flow vacates `thread:{messageId}` without sending the draft. Sending a reply resolves its own draft (4a); other stash drafts on that message are not re-pointed by a send. The plan scopes 4b to move + promotion.

## Test plan

- [x] Backend `bun test src/features/drafts/repository.test.ts` — 9 pass (incl. `rescopeByScope`: multi-user, version bump, no `user_id` filter)
- [x] Backend `bun test tests/integration/message-move.test.ts` — 3 pass, deterministic across reruns (added `DELETE FROM drafts` to the suite `beforeEach`); new test asserts both authors' reply drafts follow, unrelated draft untouched, one `draft:upserted` per owner
- [x] Backend `bun test src/features/drafts src/features/messaging` — 110 pass
- [x] Frontend `draft-sync.test.ts` / `use-draft-message-sync.test.ts` — pointer-move on re-scope, stash re-scope (no pointer), `rescopeScopeDrafts` moves+pushes both rows
- [x] Frontend full suite `bunx vitest run` — 2685 pass / 241 files
- [x] `tsc --noEmit` across all 14 workspaces clean (pre-commit gate); lint + OpenAPI check clean
- [ ] Local Playwright e2e not run here (no browser harness in this session) — the "reply draft follows a message into its new thread" e2e runs in CI browser-tests

<details>

**Goal (plan §Stage 4b):** A draft that targets a not-yet-threaded message re-points to the thread stream when that message becomes a thread; a draft in a not-yet-created scratchpad re-scopes onto the real stream on promotion.

**What was built:**
- Backend `DraftsRepository.rescopeByScope` + `moveMessagesToThread` wiring (re-scope in the move txn, one user-scoped `draft:upserted` per owner).
- Frontend `migrateLocalDraftScope` / `migrateDraftScopeInCache` (re-scope + loaded-pointer move, atomic); `applyDraftUpserted` scope-move branch; `rescopeScopeDrafts` folded into `promoteDraft` (re-scope + push, replacing purge).
- Shared `draftStreamScope` / `draftThreadScope` (INV-33).

**Design evolution (self-review):** the integration test first used fixed draft ids against a persistent test DB whose `beforeEach` didn't clear `drafts`, so reruns collided on `ON CONFLICT (id) DO NOTHING` and the re-scope matched nothing. Fixed by adding `DELETE FROM drafts` to the suite teardown; verified deterministic across reruns.

**Schema changes:** none (uses the Stage 1 `drafts` table; re-scope is a column UPDATE).

**What's NOT included:** E2E draft re-pointing on the client and the reply-send re-point case (see Out of scope); E2E sealing is Stage 4c.

**Status:** complete; CI pending.

</details>

---

---
_Generated by [Claude Code](https://claude.ai/code/session_019tNYzX8aDnRbgaTyHavan7)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/937"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784038404&installation_id=129946197&pr_number=937&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F937&signature=e55076fb0e2730dc0a03190ab8fc231053b84f6d180732bbc9e221f080589da0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/drafts/repository.test.ts
+++ b/apps/backend/src/features/drafts/repository.test.ts
@@ -200,6 +200,39 @@ describe("DraftsRepository.softDelete", () => {
   })
 })
 
+describe("DraftsRepository.rescopeByScope", () => {
+  afterEach(() => mock.restore())
+
+  it("re-scopes every owner's draft for a scope (no user filter) and bumps the version", async () => {
+    const captured: Captured = { text: null, values: null }
+    const db = createQuerier(captured, [
+      { ...DRAFT_ROW, id: "draft_01", user_id: "usr_1", scope: "stream:thread_1", root_stream_id: "stream_root" },
+      { ...DRAFT_ROW, id: "draft_02", user_id: "usr_2", scope: "stream:thread_1", root_stream_id: "stream_root" },
+    ])
+
+    const rows = await DraftsRepository.rescopeByScope(db, {
+      workspaceId: "ws_1",
+      fromScope: "thread:msg_1",
+      toScope: "stream:thread_1",
+      rootStreamId: "stream_root",
+    })
+
+    expect(captured.text).toContain("UPDATE drafts SET")
+    expect(captured.text).toContain("scope =")
+    expect(captured.text).toContain("root_stream_id =")
+    expect(captured.text).toContain("version = version + 1")
+    expect(captured.text).toContain("deleted_at IS NULL")
+    // Multi-user: matches by workspace + scope only, never user_id — a shared
+    // thread scope's drafts all follow the message regardless of author.
+    expect(captured.text).not.toContain("user_id =")
+    expect(captured.values).toContain("thread:msg_1")
+    expect(captured.values).toContain("stream:thread_1")
+    expect(captured.values).toContain("stream_root")
+    // Returns each re-scoped row so the caller can emit one event per owner.
+    expect(rows.map((row) => row.userId)).toEqual(["usr_1", "usr_2"])
+  })
+})
+
 describe("DraftsRepository.listByUser", () => {
   afterEach(() => mock.restore())
 
--- a/apps/backend/src/features/drafts/repository.ts
+++ b/apps/backend/src/features/drafts/repository.ts
@@ -248,6 +248,41 @@ export const DraftsRepository = {
   },
 
   /**
+   * Re-scope every live draft from `fromScope` to `toScope` (thread re-pointing).
+   * When a not-yet-threaded message is converted to a thread, its reply drafts
+   * must follow the message into the new thread stream so they keep roaming.
+   *
+   * Multi-user by design (no `user_id` filter): a shared `thread:{messageId}`
+   * scope can hold reply drafts from several authors, and every owner's draft
+   * follows the message — drafts stay private (each is delivered only to its own
+   * `user:{userId}` room by the caller). Set-based single UPDATE (INV-56), no
+   * select-then-write (INV-20).
+   *
+   * Bumps `version` so the client's drift-aware apply accepts the new scope: a
+   * clean local row sees `version > baseVersion` and adopts it, while a row with
+   * unpushed edits ignores the echo and its queued push splits CAS-safely
+   * (expectedVersion now trails the server). Returns the re-scoped rows so the
+   * caller can emit one `draft:upserted` per owner.
+   */
+  async rescopeByScope(
+    db: Querier,
+    params: { workspaceId: string; fromScope: string; toScope: string; rootStreamId: string | null }
+  ): Promise<Draft[]> {
+    const result = await db.query<DraftRow>(sql`
+      UPDATE drafts SET
+        scope = ${params.toScope},
+        root_stream_id = ${params.rootStreamId},
+        updated_at = NOW(),
+        version = version + 1
+      WHERE workspace_id = ${params.workspaceId}
+        AND scope = ${params.fromScope}
+        AND deleted_at IS NULL
+      RETURNING ${sql.raw(COLUMNS)}
+    `)
+    return result.rows.map(mapRow)
+  },
+
+  /**
    * Unconditional soft-delete for explicit discard (no CAS) — the user threw
    * the draft away, so drift doesn't matter. Idempotent on an already-deleted
    * row. Returns the row when it tombstoned, `null` when it was missing.
--- a/apps/backend/src/features/messaging/event-service.ts
+++ b/apps/backend/src/features/messaging/event-service.ts
@@ -15,6 +15,7 @@ import {
 } from "../attachments"
 import { OutboxRepository } from "../../lib/outbox"
 import { AgentSessionRepository, StreamPersonaParticipantRepository } from "../agents"
+import { DraftsRepository, toDraftView } from "../drafts"
 import { E2eStreamsRepository } from "../e2e-streams"
 import { attachmentReferenceId, eventId, messageId, messageVersionId, streamId as generateStreamId } from "../../lib/id"
 import { MessageVersionRepository, type MessageVersion } from "./version-repository"
@@ -27,6 +28,8 @@ import {
   CompanionModes,
   StreamTypes,
   Visibilities,
+  draftStreamScope,
+  draftThreadScope,
   type AttachmentSummary,
   type AuthorType,
   type EventType,
@@ -1159,6 +1162,28 @@ export class EventService {
         updates: agentSessionEventUpdates,
       })
       await MessageRepository.moveToStream(client, destinationThread.id, updates)
+
+      // Thread re-pointing: the target message just became a thread, so any reply
+      // draft composed against it (scope `thread:{targetMessageId}`) follows the
+      // message into the new thread stream (`stream:{destinationThread.id}`).
+      // Drafts are private per author but a shared target can hold several
+      // authors' reply drafts, so this re-scopes every owner's draft and emits
+      // one user-scoped `draft:upserted` each — in this same transaction so the
+      // re-point and the move commit together (INV-4/7).
+      const rescopedDrafts = await DraftsRepository.rescopeByScope(client, {
+        workspaceId: params.workspaceId,
+        fromScope: draftThreadScope(params.targetMessageId),
+        toScope: draftStreamScope(destinationThread.id),
+        rootStreamId,
+      })
+      for (const draft of rescopedDrafts) {
+        await OutboxRepository.insert(client, "draft:upserted", {
+          workspaceId: params.workspaceId,
+          targetUserId: draft.userId,
+          draft: toDraftView(draft),
+        })
+      }
+
       await MessageRepository.updateStreamScopedReferences(client, {
         workspaceId: params.workspaceId,
         sourceStreamId: params.sourceStreamId,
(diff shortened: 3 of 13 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (one INV-56 finding fixed in 3ed825b; remaining items by-design or deferred to 4c)
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: sonnet x6 (spec, correctness, data-flow, design, UX, mobile)

Reviewed `0d15e1a`. 4 reviewers ran over the diff (UX + mobile skipped — no component/JSX/layout/touch surface changed). 4 candidate findings; 1 fixed, 3 assessed as by-design or deferred.

**Fixed**

1. `apps/backend/src/features/messaging/event-service.ts` — INV-56: the per-owner `draft:upserted` fan-out was a per-row `OutboxRepository.insert` loop; switched to `OutboxRepository.insertMany` (the established multi-owner pattern in `memos/service.ts:458`, `streams/service.ts:538`). Fixed in `3ed825b`.
   https://github.com/threahq/threa/blob/3ed825ba3ac02b85e8f2dc91ecf501cbf0641a97/apps/backend/src/features/messaging/event-service.ts#L1165-L1185

**Assessed — no change (with rationale)**

2. *Dirty-draft + server re-scope → split at old scope* (correctness, scored 90). When a user is mid-edit on a `thread:{msg}` draft (pending push, `baseVersion=1`) and a move bumps the server row to `version=2`, the queued push fails CAS and the server splits, landing the in-progress content back at `thread:{msg}`. **By-design, not a regression:** pre-PR that draft was *already* orphaned the moment its message became a thread (the composer scope moves to `stream:{thread}` and the loaded pointer is device-local). The dirty-ignore + server-as-sole-split-authority protocol is deliberate (prior-stage design); migrating mid-flight would violate it. No data loss — the split content is preserved and surfaces in the all-drafts/stash list (`use-all-drafts.ts` enumerates every draft row). The data-flow reviewer traced this exact sequence and did not flag it.

3. *`root_stream_id` clobbered on next edit* (correctness, scored 82). `casUpdate` writes `root_stream_id = $param` (full-replacement, by design) and the client always pushes `rootStreamId: null`, so the value `rescopeByScope` sets is cleared on the next edit. **Functionally inert in v1** — the client never reads `rootStreamId` (`cachedDraftFromWire` drops it) and drafts are listed by owner, not access-gated. The proper fix is to round-trip `rootStreamId` to the client, which **Stage 4c needs anyway** for E2E SSK resolution on decrypt-on-load. Deferred to 4c with that work.

4. *`rescopeScopeDrafts` partial-migration window* (design, scored 82). It re-scopes drafts in N separate IDB transactions; a crash mid-loop could leave some drafts on the old scope. **Low impact:** N is 1–2 on scratchpad promotion, no data loss (rows persist, bootstrap reconciles), and it is strictly better than the `purgeScopeDrafts` it replaced, which *deleted* stash drafts. Noted for a future single-transaction bulk re-scope; not worth a late atomicity refactor here.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches plan §Stage 4b (move re-scope + promotion re-scope + shared scope builders). E2E client re-pointing and reply-send re-pointing are deliberately out of scope (4c / not 4b).</details>
<details><summary>🔍 Bugs [2 assessed, 0 fixed]</summary>Findings #2 and #3 above — both by-design / deferred, neither a data-loss or regression.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>All four traced sequences (clean re-scope echo, dirty-draft echo+split, promotion push, cache/IDB parity) hold; `baseVersion`/`attachments`/`attachmentIds`/`contextRefs`/`clientUpdatedAt` round-trip; old-scope pointer is removed (no dangling pointer rendering the draft under two composers).</details>
<details><summary>📋 CLAUDE.md Compliance [1 violation, fixed]</summary>INV-56 (finding #1) — fixed in 3ed825b.</details>
<details><summary>🏗️ Design [1 concern, noted]</summary>Finding #4 — partial-migration window; noted, not fixed (low impact, better than the purge it replaced).</details>
<details><summary>🔒 Security [CLEAN]</summary>Parameterized SQL throughout; re-scope is workspace-scoped; multi-user fan-out delivers each draft only to its own owner's room.</details>
<details><summary>🎛️ UX [N/A]</summary>No component/JSX/interaction surface in the diff.</details>
<details><summary>📱 Mobile [N/A]</summary>No layout/touch/viewport surface in the diff.</details>

PR #938, opened 2026-06-14
fix(drafts): propagate Drafts-view deletes and unify the delete path

kristofferremback:
## Problem

Deleting a draft message from the Drafts explorer (`/drafts`) didn't propagate to the author's other devices — the row vanished locally but every other device kept showing it, in both directions (laptop→phone and phone→laptop). Reported from a live two-device session.

## Root cause

`useAllDrafts.deleteDraft` enqueued the server delete (via `syncDraftRemoval` → `enqueueDraftDelete`) but **never kicked the operation queue**. With both devices simply online (no reconnect), the `delete_draft` op sat undrained, so the deletion never reached the server and never emitted `draft:deleted`. The in-composer stash list kicked the queue after deleting; the Drafts page didn't — that asymmetry is the bug.

## Changes

- **Kick the queue on Drafts-view delete.** `useAllDrafts.deleteDraft` now calls `syncEngine?.kickOperationQueue()` after the removal (matching the stash list), so the delete drains and propagates immediately.
- **One delete path ("a draft is a draft").** Collapsed the Drafts page's loaded/stashed branch *and* the parallel `deleteStashedDraftById` into a single canonical `deleteDraftById(workspaceId, id)` in `draft-sync.ts`. Both the in-composer stash list and the Drafts explorer now delete through the same function: it removes the row, clears the loaded pointer when it pointed there, and queues an unconditional server delete. The two surfaces can no longer drift.
- **Resurrection guard.** `applyDraftUpserted` / `applyDraftsBootstrap` no longer re-create a draft that has a pending local `delete_draft` op (new `hasPendingDraftDelete` + a `pendingDeleteIds` set). This closes a reconnect-race window where a not-yet-drained delete could be undone by a `GET /drafts` re-seed or a socket echo. The queued delete is authoritative until it drains.
- **Doc note.** Recorded in `docs/plans/centralized-draft-system.md` that the Drafts view is genuinely live for channel/DM/thread message drafts but local-only for unpromoted scratchpad drafts (a deliberate Non-goal) — verified end-to-end and flagged for a possible future decision.

## Design decisions

- `deleteDraftById` lives in `draft-sync.ts` alongside the other local+sync primitives (`deleteLocalDraft`, `syncDraftRemoval`). `deleteLocalDraft` now returns the removed row so the by-id delete can mirror off its `baseVersion`; the inbound-apply callers ignore the return.
- The kick stays in the hook layer (it needs the engine from React context), consistent with the existing convention — `draft-sync.ts` is a pure module with no engine access.
- Deletion uses the unconditional `syncDraftRemoval` (the user threw the draft away). The CAS path (`syncDraftResolution`) remains reserved for resolve-on-send.

## Out of scope

The composer's own `clearDraft` (empty-the-editor) and `resolveDraft` (send) keep their scope-based helper — those are different interactions, not "clicked delete in a list". Easy to fold in later if wanted.

## Testing

- New `use-all-drafts.test.ts`: deleting from the Drafts view removes the row, queues a `delete_draft`, clears the loaded pointer, and kicks the queue.
- New `deleteDraftById` + resurrection-guard cases in `draft-sync.test.ts` (confirmed-draft server delete, never-synced cancel, pointer clear, no-op on missing row, bootstrap/socket no-resurrection).
- Removed the now-redundant `deleteStashedDraftById` test.
- 122 green across the draft suite; typecheck + lint + full pre-commit gate clean.

https://claude.ai/code/session_01V92trjwHNAxMMpNpkjpFgU

---
_Generated by [Claude Code](https://claude.ai/code/session_01V92trjwHNAxMMpNpkjpFgU)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/938"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784049898&installation_id=129946197&pr_number=938&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F938&signature=a3f7e89435094413eab2bd5861b9149333e300f274649ee27f16e16e02c8496f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---

### Follow-up commits (after the initial code review)

- `726a9df` — **perf**: restore the two indexed `pendingOperations` reads in `applyDraftsBootstrap` (addresses the self-review design finding; the prior full-table `toArray()` + JS filter was unnecessary).
- `0920e4b` — **UX parity**: the in-composer stash picker now opens the same confirm dialog as the Drafts explorer before deleting a saved draft, closing the "a draft is a draft" interaction-layer gap from the review. No success toast (matches the Drafts page — confirm already makes the delete deliberate).


kristofferremback:
> ⤴️ **Superseded** by a [newer review](https://github.com/threahq/threa/pull/938#issuecomment-4702690838).

### Code review
**Confidence: 5/7** — Good
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: Claude Sonnet ×5 (spec, correctness, data-flow, design, UX). Mobile reviewer skipped — the diff has no UI/layout/touch/viewport surface.

The core fix is sound: the missing `kickOperationQueue()` is the right root cause, the unification into one `deleteDraftById` is clean, and the resurrection guard is correct. Two non-blocking items worth acting on, plus one pre-existing UX note.

Found 2 issues (+1 pre-existing UX note):

1. `apps/frontend/src/sync/draft-sync.ts:459-465` — `applyDraftsBootstrap` now reads the whole `pendingOperations` table (`toArray()`) and filters in JS for the upsert/delete id-sets, where it previously used the `type` index. The table holds every queued op type (edits, reactions, deletes…), so this is one full-table read instead of two targeted indexed reads, and the "O(drafts) → O(1) scans" comment now undersells it. Low impact (the offline queue is small) but trivially restored with two parallel `where("type").equals(...)` reads. (design/performance)
   https://github.com/threahq/threa/blob/59e7e03e7022a29175ee6073f34283a92b634b8f/apps/frontend/src/sync/draft-sync.ts#L459-L465

2. `apps/frontend/src/sync/draft-sync.ts:387-394` — Transient stale-echo resurrection window. The live socket path (`workspace-sync.ts:1674` → `applyDraftUpserted(payload, workspaceId)` with no `pendingDeleteIds`) falls back to `await hasPendingDraftDelete`. Once a `delete_draft` op has *drained* (removed from the queue), a still-in-flight `draft:upserted` echo of a prior edit finds no pending delete and no local row, skips the `version <= baseVersion` check (only reached when a local row exists), and re-inserts the deleted draft via `putLocalDraft`. The server's own ordered `draft:deleted` (from the same soft-delete) then removes it again, so it self-heals. Timing budget for the window: the edit's socket echo must lag behind the user's *later* manual delete **and** its drain round-trip — narrow, and the resurrection path predates this PR (the new guard *reduces* it for the still-pending case). Real but low-severity; a full fix needs a client-side delete tombstone, which the centralized-draft design deliberately avoids. Flagging as a known residual, not a blocker. (data-flow)
   https://github.com/threahq/threa/blob/59e7e03e7022a29175ee6073f34283a92b634b8f/apps/frontend/src/sync/draft-sync.ts#L387-L394

**Pre-existing UX note** (not introduced by this diff): both delete surfaces now share `deleteDraftById`, but the Drafts explorer confirms ("Delete this draft? This action cannot be undone.") while the in-composer stash picker deletes immediately with no confirm and no toast. Same data outcome, different friction — worth a parity decision given the "a draft is a draft" intent, but it's a separate UX change (this PR didn't touch either wrapper).

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Unification + propagation both achieved. `deleteStashedDraftById` fully removed with no dangling callers; both the Drafts page and the stash picker route through `deleteDraftById`; the kick is added on every branch. No missing corresponding change.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>`deleteLocalDraft`'s new `CachedDraft | undefined` return is safe (its other callers ignore it); the `pendingDeleteIds` guard runs before the `!local` resurrection; routing the stash delete through `deleteDraftById` is a no-op for the loaded pointer (stash entries never match it). No security surface (local-first IDB/sync).</details>
<details><summary>🔁 Data Flow [1 residual]</summary>See #2 — transient, self-healing stale-echo resurrection. Separately checked: the bootstrap second loop doesn't consult `pendingDeleteIds`, but that's correct-by-construction — a pending-delete row's local copy is already gone, so it never appears in that loop's `localDrafts`.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-43 (one shared path for shared behavior) is satisfied by the unification; INV-38/INV-49 (dead code / no deprecated aliases) — `deleteStashedDraftById` and its test removed; INV-20 race-safety preserved (`baseVersion` read inside the delete transaction).</details>
<details><summary>🏗️ Design [1 issue]</summary>See #1 (index vs full-table read). `popStashedDraft` keeping its own get+delete+`syncDraftRemoval` rather than reusing `deleteDraftById` is correct — restore is fetch-then-delete, not discard.</details>
<details><summary>🔒 Security [CLEAN]</summary>No exploitable surface; local-first IndexedDB + offline-queue code.</details>
<details><summary>🎛️ UX [1 pre-existing]</summary>Confirm-dialog / feedback divergence between the two delete surfaces (see note above). Not introduced by this diff.</details>
<details><summary>📱 Mobile [N/A]</summary>No UI/layout/touch/viewport code in the diff — reviewer skipped.</details>

</details>

kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: Claude Sonnet ×6 (spec, correctness, data-flow, design, UX, mobile)

Re-review after the perf fix (`726a9df`) and the stash-picker confirm dialog (`0920e4b`). The earlier perf finding is fixed and the data-flow residual is accounted for; CI is green. One minor design item, corroborated by two reviewers.

Found 1 issue:

1. `apps/frontend/src/components/composer/stashed-drafts-picker.tsx:212-225` (and `apps/frontend/src/pages/drafts.tsx:165-178`) — The "Delete this draft?" confirm dialog (title + "This action cannot be undone…" + Cancel/Delete) is byte-for-byte duplicated across the stash picker and the Drafts explorer. The delete *logic* was unified into one `deleteDraftById`, but the confirm *shell* wasn't — so a future copy/style change must land in two files with nothing to catch drift, the exact failure mode the logic-layer unification avoided. A shared `<DeleteDraftConfirmDialog open onOpenChange onConfirm />` would put both surfaces on one path. (abstraction — spec INV-35/43 + design)
   https://github.com/threahq/threa/blob/0920e4b39428bf8585478dc7bf495c4c63252464/apps/frontend/src/components/composer/stashed-drafts-picker.tsx#L212-L225

→ Extracting the shared component as a follow-up commit.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>The confirm dialog correctly gates the delete (no delete fires without confirming); no success toast (deliberate — matches the Drafts page). No missing change.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Confirm state machine sound — `onDelete` fires only from `confirmDelete`, no double-fire / stale-id; the dialog is a sibling of `<Popover>` (not inside `PopoverContent`), so closing the popover doesn't unmount it; the perf reads produce identical id-sets.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>The two indexed `pendingOperations` reads preserve the exact upsert/delete id-sets; the second bootstrap loop correctly never sees a pending-delete row (already removed locally). The confirm step is pure UI — no draft-entity timing change. (The known transient stale-echo resurrection from the prior review remains a documented, accepted residual.)</details>
<details><summary>📋 CLAUDE.md Compliance [1 — see #1]</summary>INV-21 (no layout shift — popover-close + dialog-open batch into one render); INV-18 (no component defined inside a component); INV-14 (Shadcn primitive used). The dup-dialog (#1) is the INV-35/43 item.</details>
<details><summary>🏗️ Design [1 — see #1]</summary>The duplicated confirm dialog. The perf change (indexed reads) is a clean improvement.</details>
<details><summary>🔒 Security [CLEAN]</summary>Local UI + IndexedDB; no exploitable surface.</details>
<details><summary>🎛️ UX [CLEAN]</summary>Confirm wording/buttons identical to the Drafts page; focus returns to the still-mounted popover trigger; the full flow is keyboard-drivable. The cancel-then-reopen friction is a minor, understood tradeoff of closing the popover before the modal (scored below threshold).</details>
<details><summary>📱 Mobile [CLEAN]</summary>`ResponsiveAlertDialog` → Drawer on mobile; portals isolate the popover/drawer (no dismiss-cascade); `DrawerFooter` handles the safe-area; the `DrawerClose` + `confirmDelete` timing was verified correct. (The 28px trash target is pre-existing, not introduced here.)</details>

PR #940, opened 2026-06-14
fix(composer): keep the mobile keyboard open when opening the drafts/scheduled pickers

kristofferremback:
## Problem

On mobile, opening the **Drafts** or **Scheduled** picker from the composer toolbar blurred the message editor. Both pickers are Radix `Popover`s whose content is portaled to `<body>`, and Radix moves focus into that content on open (`onOpenAutoFocus`). That blur tears down the on-screen keyboard, which grows the visual viewport and re-flows the whole composer — so the popover and its primary action (Save current / Schedule send) jump to a different position mid-tap. The user has to re-orient every time.

The neighbouring toolbar affordances don't have this problem: the **Aa** style/headings menu (`StylePicker` with `keepEditorFocus`) and the link editor keep the editor focused throughout, so the keyboard stays up and nothing moves. The drafts/scheduled pickers just never adopted that pattern.

The composer already keeps focus for the *trigger* tap — the action-bar wrapper in `message-composer.tsx` `preventDefault`s `mousedown` for its own DOM subtree, and the expanded-mode FAB triggers `preventDefault` their own `pointerdown`. The gap was entirely on the portaled popover content: Radix's auto-focus on open, and focus-follows-pointer on the buttons inside it.

## Solution

Mirror what `StylePicker`/`LinkEditor` already do, factored into one helper so both pickers share it (and it's unit-testable):

`keepEditorFocusProps(isMobile)` returns props to spread onto a `PopoverContent`. On mobile it:

- prevents Radix's `onOpenAutoFocus` and `onCloseAutoFocus`, so focus never leaves the editor when the popover opens or closes;
- adds a `mousedown` handler that `preventDefault`s focus theft for buttons physically inside the popover, while **exempting** native inputs (`input, textarea, select, [contenteditable]`) so the scheduled-message date/time fields still open their platform pickers, and **ignoring** elements outside this DOM node (nested portaled menus — e.g. the per-preset alt-timezone `DropdownMenu` — whose synthetic events bubble back through the React tree but must keep their own focus).

On desktop it returns `{}` — there's no soft keyboard, and Radix's focus management is what keyboard navigation relies on.

### How it works

1. `keepEditorFocusProps(false)` → `{}`; the pickers render exactly as before on desktop.
2. `keepEditorFocusProps(true)` → `{ onOpenAutoFocus, onCloseAutoFocus, onMouseDown }`. Spread onto `PopoverContent` in both `stashed-drafts-picker.tsx` and `scheduled-messages-picker.tsx`.
3. The `onMouseDown` guard mirrors the composer action-bar wrapper: `target.closest("input, …")` short-circuits for native inputs; `!currentTarget.contains(target)` short-circuits for portaled descendants; otherwise `preventDefault()` keeps the editor focused. `preventDefault` on `mousedown` does not block the button's `click`, so every action still fires.

`scheduled-messages-picker.tsx` gained a `useIsMobile()` call (the drafts picker already had one).

### Key design decisions

**1. Mobile-only guard.** Desktop returns `{}` so Radix keeps trapping/restoring focus for keyboard users; the keyboard-teardown reflow it's solving for only exists on touch. Same split the `StylePicker` makes (it renders a `DropdownMenu` on desktop and the focus-kept `Popover` on mobile).

**2. Input + portal exemptions instead of a blanket `preventDefault`.** A blanket suppress would stop the scheduled custom-time `<input type="date">`/`<input type="time">` from opening their native pickers, and would swallow focus for the nested alt-timezone dropdown. The two guards (`closest` for inputs, `contains` for portals) are the same shape as the existing composer action-bar wrapper comment, which calls out this exact footgun.

**3. Shared helper over per-button handlers.** `StylePicker` wires `onPointerDown`/`onClick` on each option; replicating that across ~10 buttons in two files (Save current, restore/delete rows, Schedule send, presets, Pick a time, Schedule, Back, View all) would be verbose and easy to miss one. A single container-level guard on `PopoverContent` covers them uniformly (INV-35/37: reuse over parallel implementations).

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/composer/keep-editor-focus.ts` | `keepEditorFocusProps(isMobile)` — the shared focus-keeping `PopoverContent` props |
| `apps/frontend/src/components/composer/keep-editor-focus.test.ts` | Unit tests: desktop no-op, open/close auto-focus prevented, button guarded, input exempt, portaled element ignored |
| `apps/frontend/src/components/composer/stashed-drafts-picker.test.tsx` | Integration: guard reaches real Radix content on mobile, off on desktop, clicks still fire |
| `apps/frontend/src/components/composer/scheduled-messages-picker.test.tsx` | Integration: date input keeps native focus while action buttons are guarded (mobile), off on desktop |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/composer/stashed-drafts-picker.tsx` | Import + spread `keepEditorFocusProps(isMobile)` on `PopoverContent` |
| `apps/frontend/src/components/composer/scheduled-messages-picker.tsx` | Add `useIsMobile()`, import + spread `keepEditorFocusProps(isMobile)` on `PopoverContent` |

## Out of scope (deliberate)

- **Triggers untouched.** Trigger focus retention is already handled (action-bar wrapper `mousedown` guard for compact; FAB `pointerdown` `preventDefault` for expanded). No change needed.
- **Attachments / mic / format / emoji buttons.** Per the report these already behave correctly; only drafts + scheduled were the annoyance.
- **No desktop behavior change** by design.

## Test plan

- [x] `bunx vitest run src/components/composer/keep-editor-focus.test.ts src/components/composer/stashed-drafts-picker.test.tsx` — 8 pass
- [x] `bunx vitest run src/components/composer/scheduled-messages-picker.test.tsx` — 2 pass
- [x] `bunx vitest run src/components/composer src/components/editor/editor-toolbar.test.tsx` — 72 pass (no regressions)
- [x] `bun run typecheck` (monorepo) — clean; pre-commit hooks (typecheck, astro check, API-docs check, lint-staged) all passed
- [ ] On-device manual verification — not done; no physical mobile device in this environment. Behavior is asserted via the integration tests (guard fires on real Radix `PopoverContent`, date input exempted) rather than a real keyboard.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01YbKR6GARWQz4aAbGoD2Ess)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/940"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784050387&installation_id=129946197&pr_number=940&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F940&signature=93fa267465fe4615d7daf517ce033485ed381458905a9f7fef83b8cc2a657b60"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/composer/scheduled-messages-picker.test.tsx
+++ b/apps/frontend/src/components/composer/scheduled-messages-picker.test.tsx
@@ -1,0 +1,71 @@
+import { describe, it, expect, vi, beforeEach } from "vitest"
+import { render, screen, fireEvent, createEvent } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { DEFAULT_WORK_SCHEDULE } from "@threa/types"
+import { spyOnExport } from "@/test/spy"
+import { TooltipProvider } from "@/components/ui/tooltip"
+import { ScheduledMessagesPicker } from "./scheduled-messages-picker"
+import * as useMobileModule from "@/hooks/use-mobile"
+import * as hooks from "@/hooks"
+import * as workScheduleModule from "@/hooks/use-work-schedule"
+import * as contexts from "@/contexts"
+import * as editDialogModule from "@/components/scheduled/scheduled-edit-dialog"
+
+let isMobileMockValue = true
+
+function renderPicker() {
+  return render(
+    <TooltipProvider>
+      <ScheduledMessagesPicker workspaceId="ws_1" streamId="stream_1" canSchedule onSchedule={vi.fn()} />
+    </TooltipProvider>
+  )
+}
+
+describe("ScheduledMessagesPicker", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks()
+    isMobileMockValue = true
+    vi.spyOn(useMobileModule, "useIsMobile").mockImplementation(() => isMobileMockValue)
+    spyOnExport(hooks, "useScheduledList").mockReturnValue((() => ({ items: [] })) as never)
+    spyOnExport(hooks, "useCancelScheduled").mockReturnValue((() => ({ mutate: vi.fn() })) as never)
+    spyOnExport(hooks, "useSendScheduledNow").mockReturnValue((() => ({ mutate: vi.fn() })) as never)
+    vi.spyOn(workScheduleModule, "useEffectiveWorkSchedule").mockReturnValue(DEFAULT_WORK_SCHEDULE)
+    spyOnExport(contexts, "usePreferencesOptional").mockReturnValue((() => null) as never)
+    // Always-mounted edit dialog pulls in a deep react-query hook tree; stub it.
+    spyOnExport(editDialogModule, "ScheduledEditDialog").mockReturnValue((() => null) as never)
+  })
+
+  it("lets the date/time inputs take focus while guarding the action buttons (mobile)", async () => {
+    renderPicker()
+
+    await userEvent.click(screen.getByRole("button", { name: /scheduled/i }))
+    await userEvent.click(screen.getByRole("button", { name: /schedule send/i }))
+    await userEvent.click(screen.getByRole("button", { name: /pick a time/i }))
+
+    // Native date input must keep native focus so its platform picker opens.
+    // (PopoverContent is portaled to document.body, not the render container.)
+    const dateInput = document.querySelector('input[type="date"]') as HTMLInputElement
+    expect(dateInput).toBeTruthy()
+    const dateMousedown = createEvent.mouseDown(dateInput)
+    fireEvent(dateInput, dateMousedown)
+    expect(dateMousedown.defaultPrevented).toBe(false)
+
+    // The Schedule button is a plain button — focus stays on the editor.
+    const scheduleBtn = screen.getByRole("button", { name: /^schedule$/i })
+    const btnMousedown = createEvent.mouseDown(scheduleBtn)
+    fireEvent(scheduleBtn, btnMousedown)
+    expect(btnMousedown.defaultPrevented).toBe(true)
+  })
+
+  it("does not guard focus on desktop", async () => {
+    isMobileMockValue = false
+    renderPicker()
+
+    await userEvent.click(screen.getByRole("button", { name: /scheduled/i }))
+    const scheduleSend = screen.getByRole("button", { name: /schedule send/i })
+    const mousedown = createEvent.mouseDown(scheduleSend)
+    fireEvent(scheduleSend, mousedown)
+
+    expect(mousedown.defaultPrevented).toBe(false)
+  })
+})
--- a/apps/frontend/src/components/composer/scheduled-messages-picker.tsx
+++ b/apps/frontend/src/components/composer/scheduled-messages-picker.tsx
@@ -21,6 +21,7 @@ import { parseLocalDateTime, toDateInputValue, toTimeInputValue } from "@/lib/da
 import { ScheduledEditDialog } from "@/components/scheduled/scheduled-edit-dialog"
 import { ScheduledActionDrawer } from "@/components/scheduled/scheduled-action-drawer"
 import { ScheduledActions } from "@/components/scheduled/scheduled-actions"
+import { keepEditorFocusProps } from "@/lib/keep-editor-focus"
 
 interface ScheduledMessagesPickerProps {
   workspaceId: string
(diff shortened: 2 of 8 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Code | Reviewers: sonnet x6 (spec, correctness, data-flow, design, UX, mobile)

No blocking issues — the focus-keeping fix holds up across all six lenses. One adjacent, **pre-existing** consistency note (not a defect in this PR) for an optional follow-up.

---

Both pickers fixed; parity with `StylePicker`/link editor achieved (`onOpenAutoFocus`/`onCloseAutoFocus` prevented + `mousedown` guard); desktop returns `{}` unchanged; native date/time inputs exempted; nested portaled menus ignored. No missing corresponding change.
</details>

Verified: `preventDefault` on `mousedown` preserves editor focus without blocking the buttons' `click` handlers (separate events); the `closest()` guard exempts the `DateTimeField` date/time inputs; `!currentTarget.contains(target)` leaves portaled `DropdownMenu` items (alt-timezone) untouched while still guarding the in-DOM chevron trigger; no existing `onOpenAutoFocus`/`onCloseAutoFocus`/`onMouseDown` prop is clobbered.
</details>

Traced the focus lifecycle end-to-end: tap row → `setOpen(false)` → deferred `ScheduledEditDialog` (its own `onOpenAutoFocus` already prevented; TipTap autofocus reclaims focus); long-press → `ScheduledActionDrawer` (Radix modal `onFocusOutside` prevents dismiss; focus-trap ref self-heals on first tap; no Tab key on mobile); `isMobile` flip while open (handlers aren't memoized, but Radix needs no handler identity). No stranded or dropped focus.
</details>

INV-21 (the fix *removes* a layout shift rather than adding one), INV-35/37 (the shared helper is the correct DRY move — no prior helper existed), INV-36 (scoped to the need), INV-39 (picker tests mount real components; the helper test is a pure unit test), INV-48 (scoped `spyOn`/`spyOnExport`, no `vi.mock`/`mock.module`). No violations.
</details>
<details><summary>🏗️ Design [1 note — pre-existing / adjacent]</summary>

`apps/frontend/src/components/editor/editor-toolbar.tsx` — `TableControls` (L633-639) already hardcodes the same `onOpenAutoFocus`/`onCloseAutoFocus` prevention on its mobile `PopoverContent`, but does **not** use the new `keepEditorFocusProps` helper and lacks the `mousedown` item-guard — so its mobile table-row menu items can still steal editor focus. This is **pre-existing and untouched by this PR** (not a regression introduced here). Now that a canonical helper exists, `TableControls` (and arguably `StylePicker` / the `message-composer` action-bar wrapper) are candidates to converge on it in a follow-up. Optional.
https://github.com/threahq/threa/blob/5ace3aeaedbeb68e1acc846675ab8d365cc48971/apps/frontend/src/components/editor/editor-toolbar.tsx#L633-L639
</details>

Focus-management only — no user-controlled HTML, no `dangerouslySetInnerHTML`. Nothing to flag.
</details>

Both pickers now match the `StylePicker`/link-editor feel: keyboard stays up, popover stays anchored, "Save current" keeps the popover open with focus retained, scheduling closes it without yanking focus. Desktop keyboard navigation preserved via the mobile-only gate. In the custom-time flow, tapping a date/time input intentionally focuses it (expected).
</details>

`useIsMobile()` gate is consistent with the existing pickers/composer; the `mousedown` guard doesn't touch touch events, so popover-list scroll and the scheduled-row long-press gesture are unaffected; both `type="date"` and `type="time"` are exempted (no reintroduction of the Android un-editable-time issue); the popover stays anchored above the keyboard.
</details>

---

PR #948, opened 2026-06-14
feat(agents): sealed columns on agent_conversation_summaries (C-2c foundation)

kristofferremback:
**Design doc:** agent-runtimes unification redesign, §2.8 Q8 — C-2c (in-enclave sealed rolling summary). This is the persistence foundation slice.

## Problem

The companion (plaintext) agent path has a rolling conversation summary: when older messages fall out of the budgeted context window, `ConversationSummaryService` folds them into a ≤1200-char summary stored in `agent_conversation_summaries`, prepended to the next turn's system prompt as the `## Conversation Memory` block. The enclave (E2E) path has no such thing — `enclave-system-prompt.ts` passes `null` for the rolling summary because the regional backend can't read E2E message plaintext to compute one. So an E2E scratchpad still forgets its beginning at the message-count cliff (`MAX_HISTORY_MESSAGES`), with no rolling memory.

C-2c closes that by computing the rolling summary **in-enclave** on decrypted content and shipping back ciphertext the backend never opens — the auto-title/turn-digest sealed pattern generalized. That feature spans the enclave loop, the assignment wire, dispatch, and a new callback. This PR is the **first slice: the persistence foundation** — the table and repository can store and round-trip a sealed summary. Nothing in production writes the sealed columns yet; that lands with the enclave compute path in follow-ups.

## Solution

Generalize `agent_conversation_summaries` to carry an E2E-sealed summary alongside the existing plaintext one, sharing a single `(stream, persona)` row and the `last_summarized_sequence` cursor — because the sequence is cleartext metadata the enclave can report back even though the summary content is opaque.

### How it works

1. **Migration `20260614203500_agent_conversation_summaries_sealed.sql`** adds three nullable columns — `summary_ciphertext` (BYTEA), `summary_envelope` (JSONB), `key_generation` (INTEGER) — and drops the NOT NULL on `summary`, since an enclave row stores ciphertext and no plaintext. Additive and append-only (INV-17); the companion path keeps writing `summary` unchanged.
2. **`ConversationSummaryRepository.upsert`** takes a discriminated-union param — `{ summary }` (plaintext) **xor** `{ sealed: { ciphertext, envelope, keyGeneration } }`. It writes the sealed bytes as BYTEA (base64-decoded on the way in, re-encoded base64 on read) and the envelope as JSONB.
3. **The monotonic-cursor guard now moves every representation column together.** The existing `ON CONFLICT` already gated `summary` on `EXCLUDED.last_summarized_sequence > stored`; the three new columns get the same CASE, so a stale (lower-sequence) write is a no-op and advancing the cursor swaps the whole representation atomically — a row never carries both a plaintext and a sealed summary, and a flip clears the side it's leaving.
4. **`findByStreamAndPersona` / `AgentConversationSummary`** surface `summary: string | null` plus `sealed: SealedSummary | null`, exactly one populated.

### Key design decisions

**1. Extend the existing table, don't add a parallel one.** The rolling summary is one concept with two representations; the `(stream, persona)` uniqueness and the `last_summarized_sequence` cursor are identical for both hosts, so a dedicated `agent_sealed_summaries` table would duplicate that scaffolding. A sealed step row (reusing the turn-digest path literally) was also rejected: a digest is append-only and read-N-recent, but a rolling summary is a rewriteable singleton — conflating them would accumulate rows and tangle trace-step projection with summary state. (This was the storage-shape fork; extending the table was the chosen option.)

**2. The "exactly one representation" rule lives in the type, not a DB CHECK.** Consistent with the no-DB-enum / validate-in-code convention (INV-3), the upsert param is a discriminated union, so a caller can't pass both or neither. The migration carries no CHECK constraint.

**3. Sealed bytes never coexist with plaintext on a row.** Because every representation column is gated by the same cursor comparison, advancing the cursor with a plaintext summary clears `summary_ciphertext`/`summary_envelope`/`key_generation` (and vice versa). In practice a stream is either E2E or not, so a flip shouldn't happen — but the columns moving together makes a row's representation unambiguous regardless.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/agents/conversation-summary-repository.ts` | `upsert` param is now a plaintext-xor-sealed discriminated union; row/domain types carry the sealed trio; `mapRowToSummary` surfaces `sealed` (base64 ciphertext + JSONB envelope + generation); upsert gates all four representation columns on the cursor. |
| `apps/backend/src/features/agents/conversation-summary-service.test.ts` | Add `sealed: null` to the four `AgentConversationSummary` mock literals the now-required field touches (the companion service is otherwise unchanged). |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260614203500_agent_conversation_summaries_sealed.sql` | Adds the nullable sealed columns; drops `summary` NOT NULL. |
| `apps/backend/tests/integration/agent-conversation-summary-sealed.test.ts` | Real-Postgres round-trip: sealed (base64 + envelope + generation), plaintext, and the stale-write / representation-flip cursor guard. |

## Out of scope (deliberate — C-2c follow-ups)

Nothing here is wired into a turn yet. The follow-up slices: hoist the summarizer prompt into `packages/agent-runtime` (shared by both hosts); assignment-wire the prior sealed summary + the `maxChars` budget; resolve `ContextWindowPolicy` at dispatch and deepen shipped sealed-history (bounded by the 48 MB assignment cap, E2EE-23); a `/sealed-summary` callback handler; and the enclave `run-turn.ts` compute → seal → POST. Generation re-wrap across key rolls needs **no** new work — `reviveStaleActorWraps` already re-wraps every generation the enclave held, so a summary sealed under generation N rides the same wraps that cover turn digests at N (§2.8 Q6).

## Test plan

- [x] `bun test ./tests/integration/agent-conversation-summary-sealed.test.ts` (apps/backend) — 3 pass, 9 expect calls, against a real Postgres; the run applied the new migration cleanly.
- [x] `bun test ./src/features/agents/conversation-summary-service.test.ts` — 4 pass (companion path unchanged).
- [x] Pre-commit: full-monorepo `tsc --noEmit` (types, backend, frontend, enclave, +9 more), `astro check`, and the OpenAPI `--check` all clean.
- Self-review: making `AgentConversationSummary.summary` nullable + adding the required `sealed` field rippled into four mock literals in the service test; pre-commit `tsc` caught three I'd missed on the first commit attempt — fixed before the commit landed. No other consumer reads the table outside the service, its tests, and the barrel (verified by grep).

<details>

**Goal:** C-2c — give the E2E/enclave agent path a rolling conversation summary, computed in-enclave on decrypted content and sealed so the regional backend never reads it (the no-memory guarantee, INV-E1). Full parity with the companion C-2b rolling summary.

**What was built (this PR — the persistence foundation):**
- Migration adding nullable `summary_ciphertext`/`summary_envelope`/`key_generation` and dropping `summary` NOT NULL.
- Repository upsert generalized to a plaintext-xor-sealed discriminated union, with all representation columns gated by the monotonic cursor.
- Integration coverage of the sealed/plaintext round-trip and the cursor guard.

**Design decisions:** extend the existing table (vs. parallel table or sealed step rows); type-level "exactly one representation" (no DB CHECK, INV-3); columns move together so representation is unambiguous.

**What's NOT included (follow-ups):** shared summarizer prose in `packages/agent-runtime`; assignment + callback wire types; dispatch policy threading + deeper shipped sealed-history under the 48 MB cap; the enclave compute/seal/POST loop. Generation re-wrap is already covered by `reviveStaleActorWraps` (§2.8 Q6) — no new work.

**Status:** foundation merged-ready; enclave compute path is the next slice.

</details>

---

---
_Generated by [Claude Code](https://claude.ai/code/session_012tpWgoVLMnfTcrmCKz8Kct)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/948"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784063050&installation_id=129946197&pr_number=948&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F948&signature=e76ab133540d901eaa41a1297c9cceaef8fd425db49348f7a81b8af633ae48ce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (all real findings fixed in `3cd73a7b`)
**Review models:** Orchestrator: Claude Code | Reviewers: Sonnet ×4 (spec, correctness, data-flow, design). UX + Mobile lenses skipped — backend-only change, no UI/touch surface.

Reviewed the C-2c foundation diff with four lenses. Six findings; **four fixed** in `3cd73a7b`, two considered and left as-is. No security or plan-adherence issues.

**Fixed:**

1. `apps/backend/src/features/agents/index.ts` — `SealedSummary` was not re-exported from the agents barrel, though it's part of the already-exported public surface (`AgentConversationSummary.sealed` and the sealed branch of `UpsertConversationSummaryParams` both reference it), so the upcoming cross-feature `/sealed-summary` caller couldn't name it (claude-md [INV-52, _Architecture Boundaries_]: "other features import from barrels only").
   https://github.com/threahq/threa/blob/3cd73a7b686d84dd5f3f4ad5205a8d20af0415cb/apps/backend/src/features/agents/index.ts#L137-L141

2. `apps/backend/src/features/agents/conversation-summary-service.ts:47-58` — the companion path did `existing?.summary ?? ""`, which (now that this PR makes `summary` nullable) would coerce a **sealed** row's null to `""`, summarize from empty context, and `upsert` plaintext over it — flipping an E2E row to plaintext and destroying the sealed state. Unreachable today (a stream is E2E xor not, and the companion never runs on E2E streams), but the foundation introduces the null, so the boundary the enclave writer will rely on is now asserted loudly instead of silently coercing (data-flow / state-lifecycle; claude-md [INV-11, _Dependency Flow_]: "Fail loudly; no silent fallback defaults").
   https://github.com/threahq/threa/blob/3cd73a7b686d84dd5f3f4ad5205a8d20af0415cb/apps/backend/src/features/agents/conversation-summary-service.ts#L47-L58

3. `apps/backend/src/features/agents/conversation-summary-repository.ts:66` — `mapRowToSummary`'s sealed detection checked only `summary_ciphertext && key_generation`, omitting `summary_envelope`; it now requires all three, matching the documented "sealed trio all non-null" invariant rather than inferring it from two of three columns (design / round-trip; two lenses).
   https://github.com/threahq/threa/blob/3cd73a7b686d84dd5f3f4ad5205a8d20af0415cb/apps/backend/src/features/agents/conversation-summary-repository.ts#L62-L72

4. `apps/backend/src/features/agents/conversation-summary-repository.ts:101-110` — `upsert` now throws when a caller carries neither representation. The "exactly one of plaintext/sealed" rule was enforced only by the discriminated union (compile time); the table has no DB CHECK by design (claude-md [INV-3]: validate in code), so an untyped caller (e.g. an unvalidated body) could have silently inserted an all-null row. The runtime guard makes "validate in code" real (bug / INV-11).
   https://github.com/threahq/threa/blob/3cd73a7b686d84dd5f3f4ad5205a8d20af0415cb/apps/backend/src/features/agents/conversation-summary-repository.ts#L100-L110

---

Matches the stated foundation scope (extend the existing table; discriminated-union upsert; monotonic cursor moves all representation columns together). INV-36 considered: landing a migration + repo capability no production path yet writes is a legitimately-sequenced slice of an approved design (§2.8 Q8 C-2c), not speculative — not flagged.
</details>

<details><summary>🔍 Bugs [1 — fixed]</summary>

Finding 4 above (untyped-caller all-null insert) — fixed with a runtime guard.

**Considered, dismissed:** the equal-sequence write is a no-op (the `ON CONFLICT` gate uses strict `>`). For the reachable same-path case this is correct idempotence — re-summarizing the same cursor range shouldn't churn the row. The only "data-loss" framing requires a sealed↔plaintext flip at an _equal_ cursor, which needs both the enclave and companion writing the same `(stream, persona)` — unreachable by the same construction as finding 2. The `>` semantics are also pre-existing (unchanged by this PR), so nothing latent is amplified into a live bug.
</details>

<details><summary>🔁 Data Flow [1 — fixed]</summary>

Finding 2 above (companion overwriting a sealed row). Traced all four upsert paths: representation flip clears the leaving side (covered by the integration test), equal-sequence keeps the stored representation coherently, and the read guard (finding 3) now matches the write invariant.
</details>

<details><summary>📋 CLAUDE.md Compliance [CLEAN after fixes]</summary>

INV-52 (finding 1) and INV-11 (findings 2, 4) addressed. INV-1/3/17 (no FK, no enum, append-only migration) satisfied.
</details>

<details><summary>🏗️ Design [1 — noted]</summary>

The `ON CONFLICT` repeats the same cursor-gated `CASE` for four representation columns; a future 5th sealed column that forgets a branch would silently retain a stale value. PostgreSQL has no clean tuple-`CASE` alternative, so the repetition stays; the explanatory comment now names the column group (`summary, summary_ciphertext, summary_envelope, key_generation`) so the "all move together" rule is explicit at the edit site.
</details>

All values are parameterized via the `sql` tag; `sql.raw` is used only for the hardcoded `SELECT_FIELDS` constant. No plaintext/ciphertext cross-contamination — the four representation columns move atomically under one cursor gate.
</details>

---

PR #949, opened 2026-06-14
feat: render decrypted E2E sealed stream names across all label surfaces

kristofferremback:
## Problem

An E2E scratchpad's display name is sealed under the stream key (SSK), AAD-bound
to the stream: the enclave writes the auto-title into `e2e_streams.name_ciphertext`
only — never the plaintext `streams.display_name` — and a manual rename re-seals
the same way (`use-stream-or-draft.ts` rename path). Until now only the open
stream header decrypted it (`pages/stream.tsx` via `useDecryptedStreamName`).
Every other surface that resolves a label goes through the canonical resolver in
`lib/streams.ts` (`streamLabel` / `resolveStreamName` / `useStreamName` /
`getStreamName`), which had **no** sealed-name awareness. So the sidebar, search,
quick-switcher, breadcrumbs, share-picker, saved view, and memory page all showed
the plaintext fallback — which for an auto-titled E2E scratchpad is the empty
placeholder ("New scratchpad"). The real title only appeared when the scratchpad
was open.

The hard part is shape mismatch: the resolver is **synchronous and pure** (used
in list `.map()`s and sort/search keys), but the sealed-name decrypt
(`tryOpenStreamName`) is **async** and session-dependent.

## Solution

Decrypt at the store-read boundary into a memory-only cache, and overlay the
result onto `displayName` there — so the existing resolver returns the decrypted
name unchanged, with no per-call-site plumbing across the ~30 label surfaces. The
plaintext is held only in memory and never persisted; IDB keeps ciphertext, the
same contract message bodies already follow (`db.events.payload` ciphertext +
in-memory `decrypt-cache`).

### How it works

1. **`lib/crypto/stream-name-cache.ts`** — an in-memory decrypt cache for names,
   mirroring `decrypt-cache.ts` for message bodies. Keyed by
   `${workspaceId}:${streamId}:${ciphertext}`, so a rename (fresh ciphertext)
   supersedes the prior name and a never-decrypted key reads `null` (the lookup
   is self-guarding — a stale/pre-rename name can't flash before the current
   one's async decrypt lands, replacing the old hook's `useState` source-key
   guard). A decrypt that returns null (locked / not a recipient / forged AAD) is
   not cached, so a later unlock retries. `clearStreamNameCache()` drops
   everything and bumps a `generation` epoch so an in-flight decrypt that
   resolves *after* a lock refuses to write plaintext back.

2. **Overlay at the read boundary** — `useWorkspaceStreams` (workspace store) runs
   `applyDecryptedNameOverlay(workspaceId, streams)`: for each unlocked, sealed
   E2E stream whose name is cached, it replaces `displayName` with the decrypted
   copy. Identity-preserving (returns the input array when nothing changed; only
   allocates fresh objects for rewritten rows), and subscribes once to the cache
   version via `useSyncExternalStore` so it re-overlays when a decrypt lands.
   `streamLabel` / `resolveStreamName` / `useStreamName` read `displayName`, so
   they all reflect it for free.

3. **`hooks/use-decrypt-stream-names.ts`** — the trigger, mounted once per
   workspace as `StreamNameDecryptor` in `workspace-layout.tsx` (alongside
   `UnreadTabIndicator` / `FreshnessWatchers`). Once unlocked it walks the
   workspace's sealed E2E streams and fires `requestStreamName` (idempotent per
   `(stream, ciphertext)`). Trigger is separate from read so decrypts fire from
   one place, not from each render of each list row.

4. **`useDecryptedStreamName`** (open-stream header) now routes through the same
   cache instead of its own `tryOpenStreamName` + `useState`, so there is a
   single decrypt authority and the header can't drift from the lists.

5. **Persistence** — `CachedStream` gains `sealedNameCiphertext` /
   `sealedNameEnvelope`. The existing write paths already spread the wire
   `Stream`/`StreamWithPreview` (bootstrap `bulkPut`, `seedWorkspaceCache`,
   `stream:updated`, the rename `toCachedStream`), so the ciphertext flows into
   the cache + IDB once the type allows it — no new write plumbing. Not indexed,
   so no Dexie version bump.

### Key design decisions

**1. Decrypt-at-the-read-boundary, memory-only, never persisted** — Kris's call:
"reduce the likelihood of leakage … put this at the boundary where we read it and
put it into memory, i.e. memoize the shit out of it and then make sure it never is
persisted." Overlaying at the store read covers every surface (including pure
mappers and sort keys) without threading a lookup through ~30 call sites — the
drift `apps/frontend/CLAUDE.md` ("one place that knows how to turn a stream into
a label") exists to prevent. Persisting *ciphertext* at rest is safe and matches
how message bodies are handled; only the decrypted plaintext is memory-only.

**2. Decrypted name outranks plaintext only when unlocked** — the sealed name is
AAD-bound (`buildNameAad`), so a compromised server can't silently relabel what
an unlocked owner sees. A locked/missing/failed decrypt falls back to the
plaintext label, which is the placeholder for an auto-titled scratchpad — locked
⇒ placeholder is preserved, render never blocks on decrypt.

**3. Module-level cache singleton** — consistent with the codebase's accepted
memory crypto caches (`decrypt-cache`, `stream-key-cache`, `attachment-crypto`),
read via `useSyncExternalStore` and cleared at the same lock boundary
(`e2e-session-store` `lock` + `resetE2eSessionStoreCache`).

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/lib/crypto/stream-name-cache.ts` | Memory-only decrypted-name cache + `applyDecryptedNameOverlay`; lock-cleared with a generation guard |
| `apps/frontend/src/hooks/use-decrypt-stream-names.ts` | Per-workspace background decryptor that populates the cache |
| `apps/frontend/src/lib/crypto/__tests__/stream-name-cache.test.ts` | Cache, overlay, no-leak-past-lock, and resolver-integration tests |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/stores/workspace-store.ts` | `useWorkspaceStreams` overlays decrypted names onto `displayName`, subscribed to the cache version |
| `apps/frontend/src/hooks/use-decrypted-stream-name.ts` | Routes the open-stream header through the shared cache (single decrypt authority) |
| `apps/frontend/src/db/database.ts` | `CachedStream` gains `sealedNameCiphertext` / `sealedNameEnvelope` (ciphertext at rest) |
| `apps/frontend/src/stores/e2e-session-store.ts` | `clearStreamNameCache()` at the lock + reset boundary |
| `apps/frontend/src/pages/workspace-layout.tsx` | Mounts `StreamNameDecryptor` once per workspace |
| `apps/frontend/AGENTS.md` | Documents that sealed names resolve through the same label path |

## Out of scope (deliberate)

- No backend / wire / crypto changes — the seal→store→owner-decrypt loop already
  shipped on `main` (migration `20260603100000_e2e_stream_sealed_name.sql`). This
  PR is purely the frontend render path.
- Only sealed names on root scratchpads exist today (the enclave gates the
  auto-title to untitled root scratchpads), so the decryptor only needs the
  workspace streams list; threads carry no sealed name of their own.

## Test plan

- [x] `bun run test src/lib/crypto/__tests__/stream-name-cache.test.ts` — 12 pass (cache semantics, no-leak-past-lock, overlay identity-stability, and `streamLabel`/`resolveStreamName` returning the decrypted name when unlocked / placeholder when locked)
- [x] `bun run test` on the affected surfaces — `streams`, `stream-sort`, `memory`, `saved-item`, `quick-switcher`, `search`, `share`, `stream-settings` (115 pass), plus `decrypt-cache` / `use-decrypted-message-content` (15), `encryption/*` (28), `e2e-session-store` (29)
- [x] Pre-commit hook: ESLint + `tsc --noEmit` across all workspaces, astro check, API-docs check — all clean
- [ ] No automated end-to-end render test mounting a live list surface with an unlocked session — covered at the unit level via `applyDecryptedNameOverlay` + the real resolver; the hook wiring (`useDecryptStreamNames`, overlay subscription) is exercised manually-equivalent through the cache/overlay tests rather than a full DOM mount

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fq9uAVcuuALawBBaT1gUHm)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/949"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784063718&installation_id=129946197&pr_number=949&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F949&signature=cb7bede1147c0be298c55123dc0539963f4dc7928454f817a3c0297a0457a948"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/AGENTS.md
+++ b/apps/frontend/AGENTS.md
@@ -52,6 +52,20 @@ Pick the entry point by what you're holding:
 selects context-appropriate placeholder wording for unnamed streams; pass the
 one that matches the surface. `streamLabel` defaults to `"generic"`.
 
+### E2E sealed names resolve through the same path
+
+An E2E scratchpad's name can be sealed (the enclave auto-title, or a manual
+rename re-sealed under the SSK). The server-mutable plaintext `displayName` is
+only the locked-state fallback; an unlocked owner should see the tamper-evident
+decrypted copy. You do **not** thread this through call sites: `useWorkspaceStreams`
+overlays the decrypted name onto `displayName` at the store-read boundary, so
+`streamLabel`/`resolveStreamName`/`useStreamName` already reflect it everywhere.
+The plaintext is held only in the memory-only `lib/crypto/stream-name-cache`
+(decrypt authority), never persisted — IDB keeps `sealedNameCiphertext`. A locked
+session shows the placeholder; the open-stream header reads the same cache via
+`useDecryptedStreamName`. Don't re-decrypt names per surface — extend the cache /
+overlay if a new surface bypasses the workspace store.
+
 ### Why this exists
 
 DM display names are computed per-viewer on the backend at bootstrap and are
--- a/apps/frontend/src/components/thread/stream-panel.tsx
+++ b/apps/frontend/src/components/thread/stream-panel.tsx
@@ -60,6 +60,7 @@ import { ResponsiveBreadcrumbs } from "./responsive-breadcrumbs"
 import { LabelableResourceTypes, StreamTypes } from "@threa/types"
 import type { MentionStreamContext } from "@/hooks/use-mentionables"
 import { streamLabel } from "@/lib/streams"
+import { useDecryptedStreamName } from "@/hooks/use-decrypted-stream-name"
 import { copyStreamLink } from "@/lib/stream-links"
 import { LabelPicker } from "@/components/labels/label-picker"
 import { StreamLabelStack } from "@/components/labels/stream-label-stack"
@@ -100,6 +101,11 @@ export function StreamPanel({ workspaceId, onClose }: StreamPanelProps) {
   const stream = idbPanelStream ?? bootstrap?.stream
   const isThread = stream?.type === StreamTypes.THREAD
   const currentWorkspaceUserId = useWorkspaceUserId(workspaceId)
+  // The bootstrap fallback above isn't overlaid by the workspace store, so an
+  // E2E scratchpad opened in a panel before its row lands in the cache would
+  // show the placeholder. Decrypt the resolved stream's sealed name directly
+  // (same cache as the overlay) to keep the panel header consistent.
+  const decryptedPanelName = useDecryptedStreamName(workspaceId, stream)
 
   // Show loading indicator only for real streams (not drafts) and only when actively loading after initial data
   const showLoadingIndicator = !isDraft && !!panelId && getStreamState(panelId) === "loading"
@@ -432,7 +438,11 @@ export function StreamPanel({ workspaceId, onClose }: StreamPanelProps) {
   } else if (isThread && stream) {
     headerContent = <ThreadHeader workspaceId={workspaceId} stream={stream} inPanel />
   } else {
-    headerContent = <SidePanelTitle className="flex-1">{stream ? streamLabel(stream) : "Stream"}</SidePanelTitle>
+    headerContent = (
+      <SidePanelTitle className="flex-1">
+        {stream ? (decryptedPanelName ?? streamLabel(stream)) : "Stream"}
+      </SidePanelTitle>
+    )
   }
 
   return (
@@ -479,7 +489,9 @@ export function StreamPanel({ workspaceId, onClose }: StreamPanelProps) {
                 description="Choose an action for this stream."
                 header={
                   <div className="px-4 pt-2 pb-3">
-                    <p className="truncate text-base font-semibold text-foreground">{streamLabel(stream)}</p>
+                    <p className="truncate text-base font-semibold text-foreground">
+                      {decryptedPanelName ?? streamLabel(stream)}
+                    </p>
                     <p className="mt-0.5 text-xs text-muted-foreground">
                       {stream.type === StreamTypes.THREAD ? "Thread" : "Stream"} actions
                     </p>
--- a/apps/frontend/src/db/database.ts
+++ b/apps/frontend/src/db/database.ts
@@ -107,6 +107,17 @@ export interface CachedStream {
    * older cached rows still parse; treated as an empty set when absent.
    */
   e2eActors?: E2eActor[]
+  /**
+   * The stream's display name sealed under the SSK (base64 ciphertext + its
+   * `StreamEnvelope` framing). Persisting *ciphertext* at rest is safe — it
+   * mirrors how message bodies keep ciphertext in `db.events.payload` and
+   * decrypt only in memory. An unlocked client decrypts this and prefers it
+   * over the plaintext `displayName`; the decrypted plaintext is held only in
+   * the memory-only `stream-name-cache` and is never written back to IDB.
+   * Both null until the stream is renamed (or auto-titled) while E2E.
+   */
+  sealedNameCiphertext?: string | null
+  sealedNameEnvelope?: unknown | null
   _cachedAt: number
 }
 
--- a/apps/frontend/src/hooks/use-decrypt-stream-names.test.tsx
+++ b/apps/frontend/src/hooks/use-decrypt-stream-names.test.tsx
@@ -1,0 +1,135 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
+import { render, screen, act, waitFor } from "@testing-library/react"
+import { StreamTypes, type StreamType } from "@threa/types"
+import { db, type CachedStream } from "@/db"
+import { resetWorkspaceStoreCache, seedWorkspaceCache } from "@/stores/workspace-store"
+import { clearStreamNameCache } from "@/lib/crypto/stream-name-cache"
+import { useDecryptStreamNames } from "@/hooks/use-decrypt-stream-names"
+import { useStreamName } from "@/hooks/use-stream-name"
+import * as e2eSession from "@/stores/e2e-session-store"
+import * as workspaces from "@/hooks/use-workspaces"
+import * as messageEnvelope from "@/lib/crypto/message-envelope"
+
+// End-to-end of the reactive render path the unit tests can't reach: mount the
+// real per-workspace decryptor next to the canonical resolver hook and prove a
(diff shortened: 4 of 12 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (3 issues found, all fixed in `8f9044aa`; nothing blocking)

Reviewed the first commit (`5000fdfe`); the three findings below were fixed in the follow-up commit (`8f9044aa`).

Found 3 issues (all fixed):

1. `apps/frontend/src/hooks/use-stream-or-draft.ts` — **the open-stream header's decrypt was a no-op.** `useDecryptedStreamName` is called with the `VirtualStream` from `useStreamOrDraft`, but that object (and its type) omitted `sealedNameCiphertext`/`sealedNameEnvelope`, so the hook computed a null ciphertext → null key → always returned null. The header only showed the decrypted name via the `useWorkspaceStreams` overlay (when the row was already cached), leaving the first-open (bootstrap, not-yet-in-IDB) path uncovered. **Fix:** carry the sealed fields onto `VirtualStream` so the hook decrypts its own stream.
   https://github.com/threahq/threa/blob/8f9044aaaaefb793e68949dd578338c5044977b4/apps/frontend/src/hooks/use-stream-or-draft.ts#L92-L120

2. `apps/frontend/src/hooks/use-decrypt-stream-names.ts:22` — `reactivity` — the decryptor read the **overlaid** `useWorkspaceStreams`, so each name it produced bumped the cache version → the overlay recomputed → the decryptor's `streams` changed identity → its effect re-ran to find every key already cached/in-flight. Converges (no loop, `requestStreamName` is idempotent) but is avoidable O(n) churn per name during the unlock burst. **Fix:** read the new `useWorkspaceStreamsRaw` (the sealed ciphertext it needs is unchanged by the overlay), so it isn't woken by its own output.
   https://github.com/threahq/threa/blob/8f9044aaaaefb793e68949dd578338c5044977b4/apps/frontend/src/hooks/use-decrypt-stream-names.ts#L1-L40

3. `apps/frontend/src/components/thread/stream-panel.tsx:101` — `ux-consistency` — the thread side-panel falls back to `bootstrap?.stream` (not overlaid) when the stream isn't in the workspace cache, so an E2E scratchpad opened in a panel before its row lands showed the placeholder while other surfaces showed the decrypted name. **Fix:** route the panel header through `useDecryptedStreamName` (same cache as the overlay); now that the resolved stream carries sealed fields it resolves consistently.
   https://github.com/threahq/threa/blob/8f9044aaaaefb793e68949dd578338c5044977b4/apps/frontend/src/components/thread/stream-panel.tsx#L104-L108

---

Matches the agreed design: decrypt at the store-read boundary, memory-only, never persisted; extended to all label surfaces via the overlay; single decrypt authority (`stream-name-cache`); lock-cleared in both `lock` and `resetE2eSessionStoreCache`. All label-resolver entry points (`streamLabel`/`resolveStreamName`/`useStreamName`/`getStreamName`) left untouched and benefit through the overlaid `displayName`.
</details>

Correctness reviewer confirmed: the unlock→decrypt→version-bump→re-overlay cycle converges (no infinite loop); the `generation` guard prevents an in-flight decrypt from writing plaintext back after a lock. One suggestion — moving `inflight.delete` into a `finally` — was **dismissed**: the current ordering matches the reviewed `decrypt-cache` sibling, and the data-flow trace showed always-deleting would let a stale pre-lock promise evict a *new* post-lock request's in-flight slot. The functional no-op (finding 1) was the only real defect and is fixed.
</details>

Traced rename→stale-name (cache key includes the ciphertext, so a re-seal produces a new key and the old plaintext is never shown), lock mid-flight (generation guard), account switch (`resetE2eSessionStoreCache`), and every `db.streams` write path (bootstrap `bulkPut`, `seedWorkspaceCache`, `stream:updated`, rename `toCachedStream`) — all spread the wire stream, so the new sealed fields round-trip without dropping. No issues.
</details>

INV-9 (module-level cache mirrors the accepted `decrypt-cache`/`stream-key-cache` pattern), INV-15 (overlay/decrypt live in the store + cache, not in a UI component; `StreamNameDecryptor` is a renderless effect like `UnreadTabIndicator`), INV-35/37 (one decrypt authority; the header hook and overlay both route through `stream-name-cache`), and the frontend "one place resolves a label" rule (resolver untouched). No violations.
</details>

<details><summary>🏗️ Design [1 fixed]</summary>

`stream-name-cache` is correctly a sibling of `decrypt-cache`, not a duplicate (different key shape, global-version notify vs per-id, no LRU needed since stream count is bounded). The overlay subscription's blast radius is bounded and identity-preserving (returns the same array when nothing changed). The decryptor-reads-its-own-output churn (finding 2) is fixed.
</details>

Decrypted plaintext lives only in the in-memory `names` Map; IDB stores ciphertext only. No path persists plaintext or lets it survive a lock. Nothing exploitable introduced.
</details>

<details><summary>🎛️ UX [1 fixed]</summary>

Cross-surface consistency was the main risk; the thread-panel bootstrap gap (finding 3) is fixed. Placeholder-when-locked and the brief placeholder→name swap on first decrypt are intentional and coherent.
</details>

Pure data/label-resolution change — no layout, touch, viewport, or safe-area surface touched. Longer decrypted names hit the pre-existing truncation on each surface (no regression introduced).
</details>

---

PR #950, opened 2026-06-15
feat(agents): in-enclave sealed rolling summary + deepened window (C-2c compute)

kristofferremback:
**Design doc:** agent-runtimes-unification-redesign §2.5 (C-2), §2.8 Q6/Q7. Builds on the C-2c persistence foundation (#948).

## Problem

The enclave (Ariadne on E2E scratchpads) ran every turn against a flat 30-message history cliff with **no rolling memory**. The companion (plaintext) already had both — a token-budgeted window and a rolling conversation summary (C-2a/C-2b, #934/#935) — but the enclave couldn't reuse any of it: the backend never sees E2E plaintext, so it can't compute a summary or read the conversation to deepen the window. #948 generalized `agent_conversation_summaries` to hold a sealed summary beside the plaintext one, but nothing wrote or read those columns — it was schema + repository capability only.

The result was a hard amnesia boundary: past ~30 messages an encrypted scratchpad forgot everything, while the equivalent plaintext scratchpad carried a continuous conversation.

## Solution

Move the C-2 mechanism *into* the enclave, where the plaintext lives, and ship the sealed result back over the existing callback transport — the same read-back round-trip the turn digests (C-1) already use, not the write-only auto-title path. The summarizer itself is hoisted into `@threa/agent-runtime` so the companion and the enclave fold over one shared implementation (no drift); the backend only ever relays ciphertext, so the §1.6 no-memory guarantee holds (INV-E7).

### How it works

1. **Shared summarizer** (`packages/agent-runtime/src/runtime/rolling-summary.ts`) — the fold prompt, the 1200-char/600-token bounds, the 40/batch × 5-batch limits, the per-message format, and the `## Conversation Memory` injection formatter, mirroring `turn-digest.ts`. `foldRollingSummary` runs over `AgentRuntimeAI.generateTextWithTools` — the narrow surface both hosts implement. The companion's `ConversationSummaryService` now delegates to it (dropping `ai.generateObject`), and the enclave calls the same function over its `rawChat`-backed AI, so a summary built on either surface reads identically.

[...]

PR #892, opened 2026-06-12
refactor(backend): extract DebouncedOutboxHandler base for feature outbox handlers

kristofferremback:
## Summary

Audit finding #6 (`docs/codebase-audit-2026-06-12.md`, P1): every feature outbox handler copy-pasted the same ~60-line scaffold — a 7-field `DEFAULT_CONFIG`, CursorLock + DebounceWithMaxWait wiring, `ensureListener()`/`handle()`, and the `fetchAfterId` batch loop with seen-tracking and the partial-progress error path.

This extracts one abstract `DebouncedOutboxHandler` base class in `lib/outbox/` and migrates 16 handlers onto it. Handlers keep only their event dispatch and domain logic; per-handler config deviations ride through `super()` overrides (e.g. push keeps `batchSize: 10` with its lock-during-network-I/O rationale comment).

**Net: −1,000 lines** (854 insertions, 1,854 deletions).

## Design decisions

- `processEvents()` is `protected` (not private) so a rare handler needing a different batch shape can override it while reusing construction/wiring. None of the 16 migrated handlers needed to.
- `processEvent(event)` is the single abstract method — implementations filter on `event.eventType` and skip-and-return, exactly like the bodies of the old per-handler loops. Throwing aborts the batch at that event with partial progress preserved (same semantics as every old copy).
- Constructor signatures of all migrated handlers are unchanged, so `server.ts` needed no edits.

## Deliberately not migrated

- `activity/outbox-handler.ts` — owned by in-transit #886
- `enclave-runtimes/dispatch/enclave-dispatch-handler.ts` — active enclave work (#889 area)
- `bot-runtimes/invocation-outbox-handler.ts` — agent-session adjacent, deviant shape
- `lib/outbox/broadcast-handler.ts` — deviant batch shape: it sequences and routes the whole batch (`sequenceEvents()`) before per-event handling, so the per-event `processEvent` contract doesn't fit

These migrate trivially once their PRs land.

## Verification

- `bun run typecheck` clean
- New base-class unit test (per-event processing, partial-progress error path, empty batch)
- Backend unit suite: 1,746 pass / 0 fail
- Integration suite: parity with main (3 pre-existing env-dependent security-regression failures, identical on untouched main)
- E2E suite: 308 pass / 0 fail

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #952, opened 2026-06-15
feat(enclave): nudge owners to re-wrap turns no live EIK can serve

kristofferremback:
**Workstream:** agent-runtimes C-2c follow-up #2 · INV-E7 / E2EE-7 · `docs/plans/agent-runtimes-unification-redesign.md` §2.8 Q6

## Problem

An E2E scratchpad's enclave turn only runs if a *live* agent instance holds the stream's SSK wrap addressed to its EIK. But the enclave mints a fresh EIK on every start (at least once per deploy, often more under scaling/health churn), and a fresh EIK holds no wraps until the owner re-wraps to it. Only the owner's unlocked device can do that — the enclave can't seal the SSK to itself (INV-E7). So after a full-fleet turnover, a turn enqueued in that scratchpad sits `pending` until `ENCLAVE_PENDING_PARK_AFTER_MS` (15 min) parks it with no reply.

Today the heal is lazy and narrow: `useActorWrapRevive` (`stream-encryption-affordance.tsx`) only re-wraps the stream the owner currently has **open**. A turn stuck in a scratchpad the owner isn't looking at strands until they happen to open it. There was no signal telling the owner "go re-key this one."

## Solution

A proactive, two-tier nudge driven off the claim-poll sweep, gated precisely on "a live fleet exists but no instance can serve this turn" — so a healthy fleet that already holds the wraps never triggers it, and a deploy where old instances keep serving never triggers it.

The trigger is **turn-unservable**, not **EIK-registration** (a deliberate pivot — see decision 1): registration fires on every deploy regardless of whether anything is actually stuck and races instance churn, whereas unservability is the exact "owner must re-wrap" signal and is immune to how many EIKs churn.

### How it works

1. **Detect** — `EnclaveInvocationsRepository.findUnservablePending` runs in `EnclaveClaimService.claimTurn` alongside `parkExhausted`, same cadence (and the existing claim-nudge `NOTIFY` wakes a poller promptly on new work). It's `claimNext`'s predicate inverted: a pending turn surfaces when at least one live EIK is registered (a zero-instance gap is ops, not an owner re-wrap) **and** no single live EIK holds wraps for *both* the reply's current generation and the prompt's trigger generation. Joined to `e2e_streams.owner_user_id` — the one party who can heal it.
2. **Socket tier** (`enclave:rewrap_needed`, user-scoped) — emitted immediately on detection. `EnclaveRewrapNudgeListener`, mounted once per workspace, heals an online unlocked owner in place via `reviveStaleActorWraps` regardless of which stream they have open. The lazy per-stream affordance becomes the locked-session fallback.
3. **Web-push tier** (`enclave:rewrap_nudge`, push-only) — emitted only after a grace window (`REWRAP_WEBPUSH_GRACE_MS` = 2 min), so an online owner's socket heal or a still-capable instance reappearing resolves it first without waking a device. `PushService.deliverRewrapNudge` reuses the existing focus-suppression (`getTargetSubscriptions`) and respects notification prefs/DND like a saved reminder. The SW (`sw.ts`) shows "Your assistant is waiting" and deep-links to the stream via the existing `notificationclick` path.
4. **Dedup** — both tiers compare-and-set on independent clocks in `enclave_rewrap_notifications` (`last_socket_emit_at` / `last_webpush_emit_at`), so a fleet minting a fresh EIK on every start can't spam the owner. Socket re-arms at 5 min, web-push at 30 min. The dedup claim and the outbox insert share a transaction (INV-7), so a clock never advances without the event that justifies it.

The sweep is best-effort: a nudge failure logs and the poll continues — a turn still parks visibly on its own timer if no one heals it.

### Key design decisions

**1. Trigger on turn-unservable, not EIK registration.** The follow-up was scoped as "push on EIK registration," but EIKs mint on every enclave start, so registration would fan out on every deploy mostly when nothing is stuck, and "re-wrap to EIK X" races a fleet where X is already gone. Unservability fires exactly when work is blocked, is per-stream/per-owner, and is churn-independent. (Kris's call after I surfaced the frequency.)

**2. Socket can heal; web-push can only summon.** A re-wrap needs the owner's in-memory UIK private key, which lives only in an unlocked tab — a service-worker push runs keyless and physically can't re-wrap. So the socket tier does the heal and web-push is strictly a "come back to the app" nudge. (Kris's framing.)

**3. Grace + dedup on web-push, immediate on socket.** Time and presence have opposite failure modes: an online owner heals over the socket in seconds, so web-pushing immediately would wake a device for a state about to resolve itself. Web-push waits out the grace; socket fires now. Both dedup so churn doesn't spam.

**4. `enclave:rewrap_nudge` is push-only (delivery-groups returns `null`).** A user-scoped nudge would also broadcast to the owner's socket (harmless but a phantom) and hit the sync log; returning `null` keeps it off the wire entirely — the push handler owns it, its socket sibling `enclave:rewrap_needed` carries the live signal.

**5. Tracking table, not state on `enclave_invocations` (INV-57).** Dedup is per-(workspace, root stream) workflow state, not domain state on the work item, and threads coalesce onto their root (the wraps and heal are the root's).

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260615120000_enclave_rewrap_notifications.sql` | Per-channel dedup ledger keyed by (workspace, root stream); two `*_emit_at` clocks (INV-57, INV-8) |
| `apps/backend/src/features/enclave-runtimes/rewrap-notifications-repository.ts` | `claimSocketNudge` / `claimWebpushNudge` — compare-and-set emit gates (INV-20) |
| `apps/frontend/src/components/encryption/enclave-rewrap-nudge-listener.tsx` | Workspace-level socket listener; heals via `reviveStaleActorWraps` for any owned stream |
| `apps/backend/tests/integration/enclave-rewrap-nudge.test.ts` | Real-Postgres servability matrix for `findUnservablePending` |
| `apps/backend/src/features/enclave-runtimes/rewrap-notifications-repository.test.ts` | SQL-shape + dedup-signal unit tests |
| `apps/frontend/src/components/encryption/enclave-rewrap-nudge-listener.test.tsx` | Listener behavior: heals when unlocked owner, ignores locked / wrong user / wrong workspace |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/enclave-runtimes/invocations-repository.ts` | `findUnservablePending` query + `UnservablePendingInvocation` type |
| `apps/backend/src/features/enclave-runtimes/claim-service.ts` | `nudgeUnservableOwners` + `emitRewrapNudge` sweep in `claimTurn`; grace/re-emit constants |
| `apps/backend/src/features/enclave-runtimes/index.ts` | Export the new repository (INV-52) |
| `apps/backend/src/lib/outbox/repository.ts` | Register `enclave:rewrap_needed` (user-scoped) + `enclave:rewrap_nudge` event types + payloads |
| `apps/backend/src/lib/outbox/delivery-groups.ts` | Route the nudge push-only (`null`); the socket event auto-routes user-scoped |
| `apps/backend/src/lib/outbox/index.ts` | Re-export the two new payload types |
| `apps/backend/src/features/push/service.ts` | `deliverRewrapNudge` — focus-suppressed, pref/DND-respecting web-push |
| `apps/backend/src/features/push/outbox-handler.ts` | Dispatch `enclave:rewrap_nudge` to the push service |
| `apps/frontend/src/sw.ts` | `rewrap_needed` push-display branch ("Your assistant is waiting") |
| `apps/frontend/src/pages/workspace-layout.tsx` | Mount `EnclaveRewrapNudgeListener` inside the socket + E2E-session providers |
| `packages/types/src/api.ts` + `index.ts` | `EnclaveRewrapNeededPayload` wire type |

## Out of scope (deliberate)

- **No live enclave-vs-OpenRouter run** — verified by unit + real-Postgres integration + typecheck; no end-to-end enclave harness exists (same honest gap noted on #950).
- **Window-floor expansion (Effect B)** and **proactive in-enclave re-wrap on key roll** — the other C-2c follow-ups, untouched here.
- The lazy per-stream `useActorWrapRevive` stays as the locked-session fallback; this adds the workspace-level path, it doesn't replace it.

## Test plan

- [x] `bun test src/features/enclave-runtimes/ src/lib/outbox/ src/features/push/` — 192 pass
- [x] `bun test ./tests/integration/enclave-rewrap-nudge.test.ts` (real Postgres) — 6 pass (servable / unservable / zero-instance / both-gen correlation / split-coverage)
- [x] Frontend `bunx vitest run enclave-rewrap-nudge-listener.test.tsx` — 4 pass
- [x] `tsc --noEmit` clean across types / backend / frontend (also via pre-commit: all workspaces + astro + api-docs `--check`)
- [x] Pre-PR review (1 reviewer agent over the diff) — no blockers/should-fix; both-gen correlation confirmed against `claimNext`. One nit: `findUnservablePending`'s "a live EIK exists" gate is global (`enclave_runtimes` is a global infra table, INV-8 exception), so the integration test is order-sensitive — mitigated by the per-stream `findFor` filter and the one `DELETE FROM enclave_runtimes` that isolates the zero-instance case. Not a production concern.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P4RZNwCy5DmVereDEEGUEg)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/952"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784149456&installation_id=129946197&pr_number=952&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F952&signature=3b30d70aaaf39be1067c2c06c58247168a4a1eb978b7f7b6bedbc877f5d786df"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
> [!NOTE]
> Superseded by a newer review: https://github.com/threahq/threa/pull/952#issuecomment-4715461771

### Code review
**Confidence: 5/7** — Good (no bugs/security; a few non-blocking lifecycle + cleanup findings, one worth a follow-up)
**Review models:** Orchestrator: Claude Code | Reviewers: sonnet x4 (spec, correctness, data-flow, design). UX + Mobile skipped as N/A — the only frontend additions are a headless side-effect listener, a static push string, and a mount; no interactive UI/layout.

Found 6 issues (none blocking):

1. `apps/backend/src/features/enclave-runtimes/rewrap-notifications-repository.ts` — **Ledger is never reset when a stream becomes servable, so a fast heal-then-restick is suppressed.** After the owner heals, the row keeps recent `*_emit_at` timestamps; a *new* stuck episode within the re-emit window gets no fresh nudge. The web-push window (`REWRAP_WEBPUSH_REEMIT_MS` = 30 min) exceeds the park window (`ENCLAVE_PENDING_PARK_AFTER_MS` = 15 min), so an **offline** owner whose stream re-sticks within 30 min can have the turn park before a fresh web-push is allowed — plausible given the fresh-EIK-per-start churn this feature targets. (data-flow / state-lifecycle) Options: clear the ledger row when a turn for that stream is next successfully claimed (stream is servable again), or set the web-push re-emit ≤ the park window.
   ``https://github.com/threahq/threa/blob/efac7ec4b5db1a6fa2bb2995b35bf6482d9a8e2a/apps/backend/src/features/enclave-runtimes/rewrap-notifications-repository.ts#L11-L37``

2. `apps/frontend/src/components/encryption/enclave-rewrap-nudge-listener.tsx:14-22` — **Comment overclaims the locked-session fallback.** It says a locked session "heals on its next unlock via the per-stream affordance," but `useActorWrapRevive` only mounts when the owner *opens that specific stream*. An owner locked at nudge time who unlocks without visiting the stuck stream isn't healed until the next socket re-emit (≤5 min). Bounded, but the comment should say so. (state-lifecycle)
   ``https://github.com/threahq/threa/blob/efac7ec4b5db1a6fa2bb2995b35bf6482d9a8e2a/apps/frontend/src/components/encryption/enclave-rewrap-nudge-listener.tsx#L11-L23``

3. `apps/frontend/src/components/encryption/enclave-rewrap-nudge-listener.tsx:48-66` — **No reconnect re-trigger (borderline INV-53).** If the socket drops while a turn is stuck and the nudge is emitted during the gap, the tab misses it until the next re-emit (≤5 min). This is an action subscription, not a data read-model, and the re-emit clock bounds the gap — noted rather than flagged as a hard INV-53 violation. (plan)
   ``https://github.com/threahq/threa/blob/efac7ec4b5db1a6fa2bb2995b35bf6482d9a8e2a/apps/frontend/src/components/encryption/enclave-rewrap-nudge-listener.tsx#L48-L66``

4. `apps/backend/src/features/push/outbox-handler.ts:91-100` — **`enclave:rewrap_nudge` branch omits the trailing `return`** the other four branches all have. Harmless today (it's the last branch) but a silent fall-through hazard if a branch is added below it. (consistency — flagged by two lenses)
   https://github.com/threahq/threa/blob/efac7ec4b5db1a6fa2bb2995b35bf6482d9a8e2a/apps/backend/src/features/push/outbox-handler.ts#L91-L100

5. `apps/backend/src/features/enclave-runtimes/claim-service.ts:217-253` — **Two transactions per stream per poll.** `emitRewrapNudge` opens one tx for the socket tier, then a second for the web-push tier. The reviewer's "double-emit because the owner healed between the txns" doesn't hold (the two txns are back-to-back server-side — no time for a client round-trip). The real residue: a turn already older than the grace at first detection fires *both* tiers in one sweep, mitigated by push focus-suppression. Minor; a single tx updating both clocks conditionally would halve round-trips. (design)
   https://github.com/threahq/threa/blob/efac7ec4b5db1a6fa2bb2995b35bf6482d9a8e2a/apps/backend/src/features/enclave-runtimes/claim-service.ts#L217-L253

6. `apps/backend/src/features/push/service.ts:418-454` — **`deliverRewrapNudge` repeats the pref/DND → `getTargetSubscriptions` → evict → `sendAndEvictStale` skeleton** of `deliverPushForActivity`/`deliverPushForSavedReminder`. It matches the existing convention (so not introduced drift), but a shared `buildAndSendPush(workspaceId, targetUserId, payload)` helper would stop the skeleton multiplying. (design — pre-existing pattern, optional)
   https://github.com/threahq/threa/blob/efac7ec4b5db1a6fa2bb2995b35bf6482d9a8e2a/apps/backend/src/features/push/service.ts#L418-L454

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Trigger-on-unservable (not registration), two-tier socket/web-push, per-(workspace, root stream) dedup in a tracking table, best-effort sweep that never breaks the poll — all present and wired (event registered → routed → delivered; payload type → usage). Issue 3 is the one plan-adherence soft spot (reconnect gap).</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No correctness bugs. Verified: the `findUnservablePending` `NOT EXISTS` correctly requires ONE live EIK (`r.key_id`) to cover BOTH the reply and trigger generation (no split-coverage false negative); the dedup CAS `rowCount` semantics are right on first-insert / within-window / past-window; the web-push grace uses the oldest pending `createdAt`; a nudge failure is caught and never breaks the claim poll; the listener's ref + `[socket, workspaceId]` deps avoid re-subscribe churn and clean up on unmount.</details>
<details><summary>🔁 Data Flow [2 issues]</summary>Issues 1 (ledger not reset → re-stick suppression) and 2 (comment overclaim). The ledger round-trip across socket→web-push claims is otherwise correct (web-push `IS NULL` guard claims its slot even after the socket insert created the row).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-4/7 (outbox + dedup-clock in one tx), INV-8 (tracking table workspace-scoped; `enclave_runtimes` is the documented global-infra exception), INV-20 (CAS dedup), INV-52 (barrel export), INV-57 (tracking table, not state on `enclave_invocations`), INV-46 (display text in the SW, not the backend) all hold. Two reviewer "violations" dismissed: the `enclave_runtimes` EXISTS being fleet-global is by design (it *enables* the nudge, doesn't suppress cross-workspace), and the `messages` PK join needs no workspace filter (`m.id` is a globally-unique ULID → PK seek).</details>
<details><summary>🏗️ Design [3 notes]</summary>Issues 4–6, plus: `findUnservablePending` runs on every claim poll × fleet size, but at the same cadence as the existing `parkExhausted` and gated on `status='pending'` rows — accepted pattern, not a new storm.</details>
<details><summary>🔒 Security [CLEAN]</summary>Events carry only `workspaceId`/`targetUserId`/`rootStreamId` — no plaintext. No injection (numeric interval params are bound, not interpolated). No auth bypass.</details>
<details><summary>🎛️ UX [N/A]</summary>No interactive UI surface — headless listener + static push string.</details>
<details><summary>📱 Mobile [N/A]</summary>No layout or touch surface.</details>

---

</details>

kristofferremback:
### Code review
**Confidence: 5/7** — Good (a few non-blocking issues, all in the service-worker notification surface)
**Review models:** Orchestrator: claude-opus-4-8[1m] | Reviewers: sonnet x6 (spec, correctness, data-flow, design, UX, mobile)

The backend half of this change is solid: data-flow traced the dedup ledger through concurrent pollers, the heal-then-restick EIK-churn sequence, and clock independence — all correct, and the 10-min web-push re-arm provably lands a re-stuck turn's nudge (~T+12) before it parks (~T+20). Spec, correctness, and security came back clean. The three surviving issues are all polish on the new `rewrap_needed` web-push path in `sw.ts`.

Found 3 issues:

1. `apps/frontend/src/sw.ts:646-656` — The `rewrap_needed` notification omits `renotify: true`, which every other repeatable notification in this handler sets (test path L617, activity path L705). The inline comment at L709-717 documents why it matters on mobile: replacing a same-tag notification without `renotify` silently swaps the tray entry with no sound/vibration/banner. The nudge re-emits every 10 min (`REWRAP_WEBPUSH_REEMIT_MS`) while a turn stays stuck, so an owner who swipes the first one away gets subsequent nudges silently. (mobile-regression)
   https://github.com/threahq/threa/blob/55f8f73c3b0ad2c3be62237b9d0f5399920daf88/apps/frontend/src/sw.ts#L646-L656

2. `apps/frontend/src/sw.ts:652` — The nudge uses `tag: rewrap:${streamId}`, but the `SW_MSG_CLEAR_NOTIFICATIONS` handler (L795-801) only closes tags `${streamId}` and `${streamId}:mention`. So once the owner heals the stream (socket tier on any device, or opening it), the "Your assistant is waiting" notification is **not** auto-dismissed — it lingers in the OS tray until the next 10-min re-emit overwrites it or the user manually swipes. Every other stream notification clears on view; this one doesn't. Fix: add `rewrap:${streamId}` to the clear set. (ux-consistency)
   https://github.com/threahq/threa/blob/55f8f73c3b0ad2c3be62237b9d0f5399920daf88/apps/frontend/src/sw.ts#L794-L801

3. `apps/frontend/src/sw.ts:648-649` — Copy/expectation gap (soft, score 80): the web-push fires precisely when the socket tier did *not* heal the owner in place, i.e. the owner is offline **or** online-but-locked (the listener drops locked sessions). The body says "Open Threa to let your assistant reply…" but a locked owner who taps it lands on the stream behind a lock icon and must *unlock* before `useActorWrapRevive` re-wraps — the copy never says so. Wording like "Unlock Threa to let your assistant reply…" would close the gap. The unlock CTA is present in the stream header, so this is friction, not a dead-end. (ux-feedback)
   https://github.com/threahq/threa/blob/55f8f73c3b0ad2c3be62237b9d0f5399920daf88/apps/frontend/src/sw.ts#L646-L656

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Every element of the described design is present and wired: `findUnservablePending` (tested), socket tier `enclave:rewrap_needed` (user-scoped), web-push tier `enclave:rewrap_nudge` (push-only, delivery-groups returns null), 2-min grace, 5-min socket / 10-min web-push re-arm, CAS dedup sharing the outbox transaction (INV-7), listener mounted once in `workspace-layout`, SW handler + `EnclaveRewrapNeededPayload` type. The PR body still says "web-push at 30 min" (now 10 min in code) — body-only drift, not in the diff.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>SQL predicate (live-EIK-exists AND no-single-EIK-covers-both-generations), the CAS dedup, transaction boundaries, socket routing, and owner targeting all check out. The frontend listener's ref-for-mutable-state pattern is correct and the effect re-subscribes on `workspaceId` change with no leak. The per-stream best-effort error handling degrades to next-sweep retry as intended.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Traced three sequences. (1) Concurrent pollers: the PK `(workspace_id, root_stream_id)` plus the conditional `ON CONFLICT DO UPDATE … WHERE` serialize to exactly one emit and one row. (2) Heal-then-restick across EIK churn: re-stuck turn B's web-push clears at ~T+12 and B parks at ~T+20 — ~8 min headroom, confirming the 30→10 min change. (3) The two `*_emit_at` clocks are fully independent (socket INSERT leaves the web-push column NULL, so the web-push CAS proceeds).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-1 (no FKs), INV-7 (CAS + outbox same tx), INV-8 (workspace_id in PK, every query filtered), INV-17 (new migration), INV-20 (race-safe upsert), INV-51/52 (colocated + barrel), INV-57 (workflow state off the domain entity) all satisfied. Outbox event-type strings are compile-checked union members, not magic strings (INV-33 N/A); SW notification text is frontend, explicitly carved out of INV-46.</details>
<details><summary>🏗️ Design [CLEAN]</summary>Constants centralized in `claim-service.ts`; dedup lookups are PK seeks; the new push branch fits the `DebouncedOutboxHandler` base (#892) and the claim path (#906/#918/#950). `findUnservablePending` has no LIMIT but returns only the naturally-small *unservable* pending set on a best-effort sweep already running at park cadence — acceptable. (Design agent stalled mid-run; these checks completed by the orchestrator.)</details>
<details><summary>🔒 Security [CLEAN]</summary>Queries are fully parameterized (incl. the `$1 * INTERVAL` and `reemitMs` numerics — no user-input path). Both tiers target only `e2e_streams.owner_user_id`; the socket event is user-scoped and the push payload carries only the owner's own `rootStreamId` as metadata. No injection, auth-bypass, or data-exposure path found.</details>
<details><summary>🎛️ UX [2 issues]</summary>Issues 2 (lingering notification — clear-tag mismatch) and 3 (locked-arrival copy gap) above.</details>
<details><summary>📱 Mobile [1 issue]</summary>Issue 1 (missing `renotify` → silent re-nudges) above. `icon`/`badge`/`tag`/`notificationclick` focus-vs-openWindow are all consistent with the existing paths.</details>

PR #955, opened 2026-06-15
refactor(e2e): collapse sealed-name loading-state into a single authority

kristofferremback:
## Problem

For E2E sealed stream names, the decrypted **value** is already centralized: one mounted driver (`useDecryptStreamNames`) kicks decrypts, and one store-boundary overlay (`applyDecryptedNameOverlay` in `useWorkspaceStreams`) reflects the result wherever a label resolves. The **status** — "is this sealed name still resolving, so show a loader instead of flashing the placeholder" — was not. It needs the session, and three surfaces each re-derived the same wiring by hand:

- `components/layout/sidebar/sidebar.tsx` — `useE2eSession(...).status` + a `useSyncExternalStore(subscribeStreamNameCache, ...)` version subscription, fed per row into `resolveSealedNamePending(workspaceId, stream, status)`.
- `contexts/coordinated-loading-context.tsx` — the same `useWorkspaceUserId` + `useE2eSession` + version-subscription triple, `.some()`-ed across streams to gate the initial reveal.
- `hooks/use-decrypted-stream-name.ts` `useStreamNameDecrypting` — the same triple for the open-stream header.

This is the "fix one surface, miss another" drift the frontend notes warn about: the decrypt machine itself is centralized, but every consumer that wants the loading state re-pairs `useE2eSession` with the cache subscription, so a fourth surface would copy it a fourth time.

## Solution

Collapse the triple into one session-aware authority and route every surface through it. `resolveSealedNamePending` stays the pure decision (unchanged logic); the wiring lives in exactly one hook.

### How it works

1. **`useSealedNamePendingResolver(workspaceId)`** (`hooks/use-decrypted-stream-name.ts`) wires `useE2eSession` (via `useWorkspaceUserId`) and the `subscribeStreamNameCache` version subscription once, and returns a resolver `(stream) => boolean` memoized on `[workspaceId, session.status, version]`. Its identity changes when the session flips or a decrypt lands (or the cache clears on lock), so consumers re-run.
2. **List surfaces apply the resolver across rows.** The sidebar builder calls it once and uses `isSealedNamePending(streamWithPreview)` per row for `nameDecrypting`; the coordinated-loading reveal gate does `idbStreams.some(isSealedNamePending)`. Both drop their own `useE2eSession` + version-subscription wiring.
3. **The header reads one stream** through `useStreamNameDecrypting`, now a thin wrapper over the same resolver.

The pure `resolveSealedNamePending` keeps the deadlock-safety the reveal gate depends on: it returns `true` only while the session is settling (`unknown`/`unlocking`) or unlocked-and-decrypting, and `false` for a locked/no-key session and once a decrypt settles (success or definitive failure) — so a locked user reveals immediately with placeholders.

### Key design decisions

**1. One authority hook, not status-on-the-row.** The fuller shape — carry `{ value, status }` on the `useWorkspaceStreams` row so a consumer reads decrypted data with no awareness it's encrypted — would require coupling the value overlay (and thus `useWorkspaceStreams`) to the session. That coupling is explicitly off-limits (it breaks the store-only tests that mount `useWorkspaceStreams` without a session) and would force a phase-in-cache rework that touches the reveal gate's deadlock-safety. The resolver hook gets the divergence fix without that risk; the transparent-row projection is a separable follow-up (see Out of scope). This is posture **C** from the design discussion (ciphertext stays at rest; reads centralized, not plaintext-at-rest-in-IDB).

**2. Keep `resolveSealedNamePending` pure and exported.** The hook is the wiring; the decision stays a pure function so it's unit-testable and the hook is a thin shell. No behavior change to the decision itself.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-decrypted-stream-name.ts` | Add `useSealedNamePendingResolver` (the single authority); `useStreamNameDecrypting` becomes a wrapper over it; `resolveSealedNamePending` unchanged logic, doc updated. |
| `apps/frontend/src/components/layout/sidebar/sidebar.tsx` | Drop `useE2eSession` + `useSyncExternalStore`/cache-version wiring; resolve `nameDecrypting` via the authority. Keeps `getE2eSessionState` (still used for the encrypted-scratchpad affordance). |
| `apps/frontend/src/contexts/coordinated-loading-context.tsx` | Drop `useWorkspaceUserId` + `useE2eSession` + cache-version wiring; reveal gate `.some()`s the authority. Removed the now-dead `e2eSessionStatus` debug field. |
| `apps/frontend/src/components/layout/sidebar/types.ts` | Refresh the `nameDecrypting` doc comment to name the new authority. |
| `apps/frontend/src/hooks/use-decrypt-stream-names.test.tsx` | Add a `ListHarness` + two cases covering the list-application contract (pending across rows until a decrypt lands, ignoring plaintext rows; never pending while locked). |
| `apps/frontend/AGENTS.md` | Document the rule: loading state goes through `useSealedNamePendingResolver`; don't re-pair `useE2eSession` with the cache version. |

## Out of scope (deliberate)

- **Transparent `{ value, status }` on the row** (the fuller projection) — deferred; requires coupling the overlay to the session, which is the constraint this change respects. Clean follow-up if reads should be fully session-unaware.
- **The open-stream header still calls `useE2eSession` directly** at `pages/stream.tsx:154` — but for the rename affordance gate (`e2eUnlocked`), not for name loading state. The name status there now goes through the authority. Left as-is.
- **Message-body / attachment decrypt caches** — already have their own single read path (`useDecryptedMessageContent`); not touched.

## Test plan

- [x] `bunx vitest run` on the touched + neighbor suites (use-decrypt-stream-names, stream-name-cache, coordinated-loading-context, scratchpad-item, general-tab.rename) — 56 pass. Existing locked / unlocked / decrypting reveal-gate and header cases stay green (behavior unchanged); two new cases lock the list-application contract.
- [x] `bunx tsc --noEmit` (frontend) — clean.
- [x] `bunx eslint` + `bunx prettier --check` on touched files — clean.
- [ ] No full `bun run test:e2e` — this is a render-path refactor with no wire/route change; covered by the integration-style component tests above rather than Playwright.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01BJimp1unpy4gmbJmsxBF2Z)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/955"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784152264&installation_id=129946197&pr_number=955&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F955&signature=06df081fa0bb656a00c3480c8c7ebf49a2a4067f5983cf1a4c53c1775d9783c4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/AGENTS.md
+++ b/apps/frontend/AGENTS.md
@@ -66,6 +66,18 @@ session shows the placeholder; the open-stream header reads the same cache via
 `useDecryptedStreamName`. Don't re-decrypt names per surface — extend the cache /
 overlay if a new surface bypasses the workspace store.
 
+**Loading state goes through the single authority.** The decrypted _value_ is
+transparent through the overlay above, but "is this sealed name still resolving"
+(show a loader vs. the placeholder) needs the session, so it has exactly one
+owner: `useSealedNamePendingResolver(workspaceId)` in `use-decrypted-stream-name`.
+It wires `useE2eSession` + the name-cache subscription once and returns a resolver;
+list surfaces apply it across rows (the sidebar builder's `nameDecrypting`, the
+coordinated-loading reveal gate's `.some`) and the open-stream header reads one
+stream via `useStreamNameDecrypting`. Do **not** re-pair `useE2eSession` with the
+cache version yourself — that triple drifted across three surfaces before it was
+collapsed here. A new surface that needs the loading state calls the resolver; it
+never reaches for the session directly.
+
 ### Why this exists
 
 DM display names are computed per-viewer on the backend at bootstrap and are
--- a/apps/frontend/src/components/layout/sidebar/sidebar.tsx
+++ b/apps/frontend/src/components/layout/sidebar/sidebar.tsx
@@ -1,13 +1,12 @@
-import { useEffect, useMemo, useRef, useState, useSyncExternalStore } from "react"
+import { useEffect, useMemo, useRef, useState } from "react"
 import { toast } from "sonner"
 import { FileText, Lock, RefreshCw, StickyNote } from "lucide-react"
 import { useLocation, useNavigate, useParams } from "react-router-dom"
 import { useAuth } from "@/auth"
 import { useCreateEncryptedScratchpad } from "@/hooks/use-create-encrypted-scratchpad"
 import { useE2eUnlockOptional } from "@/components/encryption/e2e-unlock-provider"
-import { getE2eSessionState, useE2eSession } from "@/stores/e2e-session-store"
-import { resolveSealedNamePending } from "@/hooks/use-decrypted-stream-name"
-import { getStreamNameCacheVersion, subscribeStreamNameCache } from "@/lib/crypto/stream-name-cache"
+import { getE2eSessionState } from "@/stores/e2e-session-store"
+import { useSealedNamePendingResolver } from "@/hooks/use-decrypted-stream-name"
 import {
   useActivityCounts,
   useAllDrafts,
@@ -99,16 +98,11 @@ export function Sidebar({ workspaceId }: SidebarProps) {
   const currentUser = workspaceUsers.find((u) => u.workosUserId === user?.id) ?? null
   const createEncryptedScratchpad = useCreateEncryptedScratchpad(workspaceId, currentUser?.id ?? null)
   const e2eUnlock = useE2eUnlockOptional()
-  // Resolve the sealed-name loading state for the whole list in one place (the
-  // session + shared name cache), so each row reads it as plain data instead of
-  // re-wiring the session. `nameCacheVersion` re-evaluates when a decrypt settles
-  // — including a failure, which leaves the overlaid row unchanged.
-  const e2eSessionStatus = useE2eSession(workspaceId, currentUser?.id ?? "").status
-  const nameCacheVersion = useSyncExternalStore(
-    subscribeStreamNameCache,
-    getStreamNameCacheVersion,
-    getStreamNameCacheVersion
-  )
+  // Resolve the sealed-name loading state for the whole list off the single
+  // authority (session + shared name cache, wired once), so each row reads it as
+  // plain data instead of re-wiring the session. The resolver re-binds when a
+  // decrypt settles — including a failure, which leaves the overlaid row unchanged.
+  const isSealedNamePending = useSealedNamePendingResolver(workspaceId)
 
   const draftCount = allDrafts.length
   const savedCount = useLiveSavedCount(workspaceId)
@@ -167,7 +161,7 @@ export function Sidebar({ workspaceId }: SidebarProps) {
           urgency,
           section,
           dmPeerUserId,
-          nameDecrypting: resolveSealedNamePending(workspaceId, streamWithPreview, e2eSessionStatus),
+          nameDecrypting: isSealedNamePending(streamWithPreview),
         }
       })
   }, [
@@ -182,8 +176,7 @@ export function Sidebar({ workspaceId }: SidebarProps) {
     workspaceUsers,
     unreadState,
     workspaceId,
-    e2eSessionStatus,
-    nameCacheVersion,
+    isSealedNamePending,
   ])
 
   // System streams are auto-created infrastructure — don't count toward "has content"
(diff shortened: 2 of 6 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (one minor, non-blocking design note)

Found 1 minor issue (non-blocking):

1. `apps/frontend/src/hooks/use-decrypted-stream-name.ts:132-135` — `useStreamNameDecrypting` (the single-stream header path) now routes through `useSealedNamePendingResolver`, which `useMemo`-allocates a closure only to invoke and discard it immediately; the pre-PR version was a single `useMemo` returning the boolean directly (design: one extra memoized closure, recomputed only when `session.status`/cache `version` change — negligible cost). **Assessed as a deliberate tradeoff, left as-is:** the alternative (calling the pure `resolveSealedNamePending` directly here) re-introduces the `useE2eSession` + cache-version wiring this PR exists to collapse. Routing the header through the one authority is the point.
   https://github.com/threahq/threa/blob/03e14783d214f453ebdc9e13c42d39a510eb9871/apps/frontend/src/hooks/use-decrypted-stream-name.ts#L132-L135

---

All three surfaces (sidebar builder, coordinated-loading reveal gate, open-stream header) route through `useSealedNamePendingResolver`/`useStreamNameDecrypting`. The hard constraint holds — `useWorkspaceStreams`/the value overlay stays session-free (`workspace-store.ts` subscribes only to the name-cache version, never the session). `resolveSealedNamePending` remains exported but has no external callers. `AGENTS.md` documents the new authority. No dead code left behind.
</details>

Memo identity changes exactly when it must (`[workspaceId, session.status, version]`), so consumers' `useMemo`s recompute correctly and not spuriously. No stale-closure, no hook-rules violation (the returned function is a plain call, not a hook), no subscription leak. Cold-load (unknown→unlocked→decrypt-lands) and locked-session paths both verified sound.
</details>
<details><summary>🔁 Data Flow [CLEAN for this diff — 1 pre-existing item noted]</summary>

Traced cold-load, locked-start, and decrypt-failure sequences end-to-end — every terminal state settles to "not pending" (no reveal-gate deadlock); a decrypt landing bumps `version` → new resolver identity → all consumers recompute.

One pre-existing quirk surfaced and was **ruled out of scope**: on session lock, `clearStreamNameCache()` bumps `version`; if that render lands before `useE2eSession` reports `locked`, the resolver briefly sees stale `unlocked` + cleared `attempted` and returns `true` for one render cycle (a transient `nameDecrypting` flash on sidebar rows / header). The **pre-PR code had identical wiring** (`useSyncExternalStore(subscribeStreamNameCache)` + `e2eSessionStatus` both in the sidebar memo deps), so this diff neither introduces nor amplifies it. The reveal gate is unaffected regardless (its `sealedNamesPending` term is guarded by `!isReady`). Left for a separate change if it ever proves user-visible.
</details>

No newly-introduced invariant violations. Tests use scoped `vi.spyOn` against namespace imports (INV-48), not `vi.mock`. The session-decoupling constraint on `useWorkspaceStreams` is respected.
</details>
<details><summary>🏗️ Design [1 note — see above]</summary>

The single-authority hook is clean: returning a `useMemo`-stable function gives referential stability between decrypt-settle events, so list consumers (`processedStreams`, the reveal gate `.some`) re-run only when the resolver actually changes. The one note (header closure allocation) is a deliberate tradeoff. No parallel implementation left un-collapsed; consistent with the value-overlay/decrypt-authority pattern established in #949/#951.
</details>

Pure React hook/memo refactor — no new XSS surface, no `dangerouslySetInnerHTML`, no secret exposure.
</details>

Strictly more consistent than before (one wiring vs three). The pure predicate is unchanged, so reveal-gate deadlock-safety, skeleton↔name timing, and INV-21 layout-shift behavior are all preserved. New tests cover the locked (no-deadlock) and settle cases.
</details>

Logic-only consolidation — no layout, CSS, breakpoints, touch handlers, or `isMobile` branches touched. Nothing mobile-specific to regress.
</details>

---

PR #956, opened 2026-06-15
fix(link-previews, agents): defeat Reddit anti-bot wall for previews and Ariadne's read_url

kristofferremback:
## Problem

Reddit link previews render as a card titled **"Reddit – Please wait for verification"** that never resolves into real content. The cause isn't a stuck fetch — the fetch *succeeds* against Reddit's anti-bot wall.

When `fetchGenericMetadata` (`apps/backend/src/features/link-previews/worker.ts`) requests a Reddit URL with our polite crawler UA (`Threa/1.0 (Link Preview; +https://threa.io/bot)`), Reddit returns **HTTP 200** with an interstitial challenge page whose only metadata is `<title>Reddit - Please wait for verification</title>`. The `!response.ok` guard doesn't fire (it's a 200), so `parseHtmlMeta` scrapes that title and the worker caches it as a `completed` preview with a 24h expiry. The `+`-prefixed contact-URL convention that gets us past Amazon-class gates is not on Reddit's allowlist.

The same wall hits **Ariadne's `read_url` tool** (`packages/agent-runtime/src/tools/read-url-tool.ts`), which fetches pages with `threaFetchUserAgent("Agent Reader")` — so the agent would read "Please wait for verification" as the page content for any Reddit link.

## Solution

Promote the Reddit decision into the shared outbound-fetch identity module (`packages/types/src/outbound-fetch.ts`) — which already declares itself the single source of truth for fetch identity (INV-33) — so every outbound fetcher handles Reddit the same way. Two behavioral changes, both scoped to Reddit hosts (`reddit.com`, its subdomains, `redd.it`):

1. **Recognized social-unfurl crawler UA for Reddit.** Reddit allowlists the well-known link-preview crawlers (this is how Slack/Discord/iMessage show Reddit cards), so Reddit requests now go out as `facebookexternalhit/1.1` to receive the real server-rendered OpenGraph HTML. Mirrors the `Twitterbot/1.0` precedent already in the link-preview worker for X image fetches.

2. **Interstitial guard so it's never cached (link-preview worker only).** A real Reddit post/comment page always carries `og:image`; the challenge page carries none. After parsing, a Reddit page with no `og:image` is returned `status: "failed"` (retryable) instead of cached — so even if the UA stops working, the challenge page can't ship as a preview.

### How it works

- `@threa/types` now exports `isRedditUrl(url)`, `REDDIT_CRAWLER_USER_AGENT` (`"facebookexternalhit/1.1"`), and `resolveFetchUserAgent(component, url)` — the recognized token for Reddit hosts, else the polite `threaFetchUserAgent(component)`. This is the one entry point outbound fetchers call, so Reddit handling can't drift between callers.
- **Link-preview worker:** the fetch `User-Agent` is `resolveFetchUserAgent("Link Preview", url)`. After `parseHtmlMeta`, `if (isRedditUrl(url) && !metadata.imageUrl)` returns `{ status: "failed", expiresAt: minutesFromNow(30) }`. Failed previews aren't on a retry timer — they refetch only when the same URL is next posted/edited — so the 30-minute window just avoids re-hammering Reddit for a quickly-reposted URL.
- **Agent `read_url`:** the per-request `User-Agent` is `resolveFetchUserAgent("Agent Reader", url)`. Resolved per fetch (not once up front) so a redirect *into* Reddit — `redd.it → reddit.com` — is covered by the redirect-following loop in `fetchWithRedirectValidation`.

### Key design decisions

**1. Shared module, not a per-feature copy.** `packages/agent-runtime` can't import from `apps/backend`, and duplicating the host list / UA in two places invites drift (INV-33, INV-35). The outbound-fetch module is the existing home for this identity, so `isRedditUrl` + the UA + the resolver live there and both callers consume them. The link-preview feature dropped its local `REDDIT_FETCH_USER_AGENT` const and its duplicate `isRedditUrl`.

**2. Structural `og:image` check, not a title-string match.** Detecting the wall by matching the literal "Please wait for verification" title would be English-only and brittle to Reddit copy changes. Keying on the absence of `og:image` — a structural property every real Reddit page has and the interstitial lacks — keeps the decision language-independent (INV-54) and avoids caching the challenge page as a silent fallback (INV-11). Trade-off: a genuine Reddit page with no image would be marked failed-and-retryable; post/comment permalinks always carry one, so this is an accepted edge.

**3. Guard is link-preview-only; the agent tool relies on the UA.** The interstitial guard exists because the worker *caches* a bad result for 24h. `read_url` doesn't cache — if the UA ever fails, the model sees the wall text once and can react — so replicating the `og:image` parse into the markdown-oriented tool would be scope creep for no durable harm. The recognized UA (the part that actually prevents the wall) is shared; the cache-specific guard is not.

## New / promoted exports

| File | Change |
| ---- | ------ |
| `packages/types/src/outbound-fetch.ts` | Add `isRedditUrl`, `REDDIT_CRAWLER_USER_AGENT`, `resolveFetchUserAgent` |
| `packages/types/src/index.ts` | Re-export the three new symbols |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/link-previews/worker.ts` | Use `resolveFetchUserAgent("Link Preview", url)`; `og:image`-absent interstitial guard; import Reddit helpers from `@threa/types` |
| `apps/backend/src/features/link-previews/config.ts` | Drop local `REDDIT_FETCH_USER_AGENT` (moved to `@threa/types`) |
| `apps/backend/src/features/link-previews/url-utils.ts` | Drop local `isRedditUrl` (moved to `@threa/types`) |
| `packages/agent-runtime/src/tools/read-url-tool.ts` | Resolve UA per-request via `resolveFetchUserAgent("Agent Reader", url)` |
| `packages/types/src/outbound-fetch.test.ts` | `isRedditUrl` + `resolveFetchUserAgent` coverage |
| `apps/backend/src/features/link-previews/worker.test.ts` | Reddit recognized-UA + completed path; interstitial → failed |
| `apps/backend/src/features/link-previews/url-utils.test.ts` | Drop the `isRedditUrl` block (moved to types test) |
| `packages/agent-runtime/src/tools/read-url-tool.test.ts` | Reddit URL → recognized crawler UA |

## Out of scope (deliberate)

- **No production verification of the UA.** Outbound network is blocked in the dev/CI sandbox (egress policy returns 403), so I could not live-confirm that `facebookexternalhit/1.1` currently defeats Reddit's wall. The guard makes the link-preview path safe either way; both paths are worth a quick check on staging by posting / reading a Reddit link once deployed.
- **No backfill of already-cached interstitial previews.** Existing rows showing "Please wait for verification" expire on their own 24h TTL and refetch on next encounter. A one-shot prune wasn't deemed worth a migration; say the word if you want one.
- **`web_search` untouched.** It delegates crawling to the Tavily API rather than fetching Reddit directly, so the wall doesn't apply.

## Test plan

- [x] `bun test ./packages/types/src/outbound-fetch.test.ts ./packages/agent-runtime/src/tools/read-url-tool.test.ts` — 43 pass, 0 fail
- [x] `bun test apps/backend/src/features/link-previews/` — 133 pass, 0 fail (Reddit UA + interstitial cases included)
- [x] `bun run --filter @threa/{types,backend,agent-runtime} typecheck` — clean
- [x] Pre-push hook: full monorepo lint + typecheck (all workspaces) + OpenAPI `--check` — 0 errors, spec up to date
- [ ] Live Reddit fetch — not run; sandbox egress is blocked (see Out of scope)

---

### DIFF
--- a/apps/backend/src/features/link-previews/config.ts
+++ b/apps/backend/src/features/link-previews/config.ts
@@ -15,17 +15,6 @@ export const FETCH_TIMEOUT_MS = 10_000
  */
 export const FETCH_USER_AGENT = threaFetchUserAgent("Link Preview")
 
-/**
- * User-Agent for Reddit fetches.
- *
- * Reddit serves a "Please wait for verification" anti-bot interstitial to generic crawler UAs —
- * even polite ones carrying the `+`-prefixed contact URL that gets us past Amazon-class gates. It
- * allowlists the well-known social link-unfurl crawlers (this is how Slack/Discord/iMessage show
- * Reddit previews), so Reddit requests go out under a recognized crawler token to receive the real
- * server-rendered OpenGraph HTML. Mirrors the existing `Twitterbot/1.0` precedent for X image fetches.
- */
-export const REDDIT_FETCH_USER_AGENT = "facebookexternalhit/1.1"
-
 /** Maximum HTML bytes to read before stopping (some sites like YouTube put meta tags 600KB+ in) */
 export const MAX_HTML_BYTES = 512 * 1024
 
--- a/apps/backend/src/features/link-previews/config.ts
+++ b/apps/backend/src/features/link-previews/config.ts
@@ -15,6 +15,17 @@ export const FETCH_TIMEOUT_MS = 10_000
  */
 export const FETCH_USER_AGENT = threaFetchUserAgent("Link Preview")
 
+/**
+ * User-Agent for Reddit fetches.
+ *
+ * Reddit serves a "Please wait for verification" anti-bot interstitial to generic crawler UAs —
+ * even polite ones carrying the `+`-prefixed contact URL that gets us past Amazon-class gates. It
+ * allowlists the well-known social link-unfurl crawlers (this is how Slack/Discord/iMessage show
+ * Reddit previews), so Reddit requests go out under a recognized crawler token to receive the real
+ * server-rendered OpenGraph HTML. Mirrors the existing `Twitterbot/1.0` precedent for X image fetches.
+ */
+export const REDDIT_FETCH_USER_AGENT = "facebookexternalhit/1.1"
+
 /** Maximum HTML bytes to read before stopping (some sites like YouTube put meta tags 600KB+ in) */
 export const MAX_HTML_BYTES = 512 * 1024
 
--- a/apps/backend/src/features/link-previews/url-utils.test.ts
+++ b/apps/backend/src/features/link-previews/url-utils.test.ts
@@ -4,10 +4,27 @@ import {
   extractUrls,
   detectContentType,
   isBlockedUrl,
+  isRedditUrl,
   parseMessagePermalink,
   parseGitHubUrl,
   parseLinearUrl,
 } from "./url-utils"
+
+describe("isRedditUrl", () => {
+  test("matches reddit.com, its subdomains, and the redd.it shortener", () => {
+    expect(isRedditUrl("https://www.reddit.com/r/ClaudeAI/comments/abc/title/")).toBe(true)
+    expect(isRedditUrl("https://reddit.com/r/ClaudeAI")).toBe(true)
+    expect(isRedditUrl("https://old.reddit.com/r/ClaudeAI")).toBe(true)
+    expect(isRedditUrl("https://redd.it/abc123")).toBe(true)
+  })
+
+  test("does not match lookalike or unrelated hosts", () => {
+    expect(isRedditUrl("https://notreddit.com/r/ClaudeAI")).toBe(false)
+    expect(isRedditUrl("https://reddit.com.evil.example/r/x")).toBe(false)
+    expect(isRedditUrl("https://example.com")).toBe(false)
+    expect(isRedditUrl("not a url")).toBe(false)
+  })
+})
 
 describe("normalizeUrl", () => {
   test("lowercases hostname", () => {
--- a/apps/backend/src/features/link-previews/url-utils.test.ts
+++ b/apps/backend/src/features/link-previews/url-utils.test.ts
@@ -4,27 +4,10 @@ import {
   extractUrls,
   detectContentType,
   isBlockedUrl,
-  isRedditUrl,
   parseMessagePermalink,
   parseGitHubUrl,
   parseLinearUrl,
 } from "./url-utils"
-
-describe("isRedditUrl", () => {
-  test("matches reddit.com, its subdomains, and the redd.it shortener", () => {
-    expect(isRedditUrl("https://www.reddit.com/r/ClaudeAI/comments/abc/title/")).toBe(true)
-    expect(isRedditUrl("https://reddit.com/r/ClaudeAI")).toBe(true)
-    expect(isRedditUrl("https://old.reddit.com/r/ClaudeAI")).toBe(true)
-    expect(isRedditUrl("https://redd.it/abc123")).toBe(true)
-  })
-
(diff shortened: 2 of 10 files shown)


kristofferremback:
### Code review

Backend-only diff — UX and Mobile lenses skipped (no frontend files). 2 findings: 1 fixed during review, 1 optional follow-up.

1. `apps/backend/src/features/link-previews/worker.ts:322` — **(fixed during review, commit 2199513)** The interstitial guard keyed on `!metadata.imageUrl` alone. With the new recognized UA, real Reddit **text** posts, subreddit landing pages, and profiles return genuine `og:title`+`og:description` but often no `og:image` — so the guard would discard those good previews and mark them `failed` with a permanent 30-min retry loop (data-flow regression introduced by this PR). Fixed by requiring **both** `!imageUrl && !description`: the interstitial carries neither, while every real Reddit page carries a description. (`title` is unusable as a discriminant — it falls back to the URL slug, so it's never absent.) Regression test added for the text-only case.
   https://github.com/threahq/threa/blob/21995133dd6dddb2785207012c94750c85c9fc46/apps/backend/src/features/link-previews/worker.ts#L318-L326

2. `apps/backend/src/features/link-previews/worker.ts:20` — **(optional, pre-existing)** `TWITTER_IMAGE_FETCH_USER_AGENT = "Twitterbot/1.0"` stays a local feature constant while this PR promotes the parallel Reddit crawler UA into `@threa/types` (INV-33 source-of-truth / INV-37 no parallel implementations). Both impersonate a recognized crawler to bypass a host's bot gate, so the shared module is arguably the home for both. Left out of scope here: the constant pre-dates this PR and is used inline for the X image-fallback fetch (not routed through `resolveFetchUserAgent`), so moving it is a separable cleanup rather than part of the Reddit fix.
   https://github.com/threahq/threa/blob/21995133dd6dddb2785207012c94750c85c9fc46/apps/backend/src/features/link-previews/worker.ts#L20-L20

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Move to `@threa/types` is complete — no dangling references to the removed `REDDIT_FETCH_USER_AGENT` or the local `isRedditUrl`; all call sites updated; `read_url` change matches the "same handling" intent.</details>
<details><summary>🔍 Bugs [1 fixed]</summary>Finding #1 (text-only Reddit false-positive) — fixed during review. `isRedditUrl` host matching verified correct (accepts subdomains + redd.it, rejects `reddit.com.evil.example`/`notreddit.com`); the FQDN-trailing-dot edge scored below threshold.</details>
<details><summary>🔁 Data Flow [CLEAN after fix]</summary>pending → fetch → complete/failed → publish traced; `failed` + 30-min expiry interacts correctly with `isPreviewCacheFresh` and `completePreviewsAndPublish`; no fields dropped. `read_url` per-request UA resolution is correct across redirect legs (redd.it → reddit.com); non-Reddit behavior identical to before.</details>
<details><summary>📋 CLAUDE.md Compliance [1 optional note]</summary>Finding #2 (INV-33/INV-37, Twitter UA). Guard uses a structural metadata check rather than an English title match — consistent with INV-54 (no language-specific heuristics).</details>
<details><summary>🏗️ Design [CLEAN]</summary>`resolveFetchUserAgent` shape is sound; `facebookexternalhit/1.1` impersonation matches the existing `Twitterbot/1.0` precedent.</details>
<details><summary>🔒 Security [CLEAN]</summary>SSRF protection (`isBlockedUrl` pre/post-redirect in the worker; `validateUrlWithDns` in read_url) is untouched by the UA swap and guard.</details>

PR #958, opened 2026-06-15
feat(agents): let personas read images on the web and in Threa

kristofferremback:
## Problem

Agent personas couldn't actually *see* images — on the web or in Threa.

Two failure modes hid behind one symptom. When a user uploads an image, the `image-caption` pipeline (`apps/backend/src/features/attachments/image-caption/`) runs a vision model at upload time and stores a 1–2 sentence `summary`. `formatAttachmentDescription` (`apps/backend/src/features/agents/companion/prompt/message-format.ts:150`) injects that summary as **text** into the persona's context — so Ariadne was reading a caption, never the pixels. The tool that hands over the real bytes, `load_attachment`, exists and works, but was **not** in Ariadne's `enabledTools` (`built-in-agents.ts`). So "what's the exact value in this cell?" or "what's wrong with this screenshot?" — anything the caption didn't happen to capture — was unanswerable.

Web images were dropped entirely: `read_url` (`packages/agent-runtime/src/tools/read-url-tool.ts`) hard-rejected any `image/*` content type ("Unsupported content type") and `node-html-markdown` strips `` tags, so images embedded in pages never surfaced.

## Solution

Feed real image bytes into the runtime's existing multimodal channel. `AgentToolResult.multimodal` already carries `{ type: "image"; url }` and the runtime injects it as a user-turn image block (`agent-runtime.ts` ~L679); `load_attachment` was already the working reference. This change extends that path to three surfaces, gated on the model's vision capability.

### How it works

1. **Images shared in Threa** — add `AgentToolNames.LOAD_ATTACHMENT` to Ariadne's `enabledTools`. `tool-set.ts:141` already gates it on `workspace && supportsVision`, and Ariadne runs `openrouter:anthropic/claude-sonnet-4.6` (`inputModalities: [text, image]`), so the tool now builds. She gets the attachment id from the `[Image: file (att_x)]` tag already in her context, so `load_attachment` alone is enough to view a conversation image.

2. **Image URLs on the web / pasted links** — `createReadUrlTool` takes `{ supportsVision }`. When set and the fetched `content-type` is `image/*`, `readImageContent` returns the bytes as a multimodal image block instead of rejecting them: restricted to PNG/JPEG/GIF/WebP (`SUPPORTED_IMAGE_MIME_TYPES`), capped at `MAX_IMAGE_BYTES` (5 MB — checked against the declared `content-length` first to avoid downloading an oversized body, then against actual bytes). The existing SSRF/DNS guards and redirect validation run unchanged before any of this.

3. **Images embedded in HTML pages** — for vision callers, `extractImageUrls` pulls `` values from the HTML, resolves them to absolute http(s) URLs, dedupes, and caps at `MAX_IMAGE_URLS` (20), returned under `ReadUrlResult.images`. The model then chooses to view one via a follow-up `read_url` (which only fetches over http(s); `data:` and non-http srcs are skipped).

`supportsVision` is wired through the companion tool-set (`createReadUrlTool({ supportsVision })`, derived from `modelRegistry.supportsVision(persona.model)`) and the enclave web tools.

### Key design decisions

**1. Gate everything on `supportsVision`.** A text-only persona gets the prior behavior — images reported unreadable, no `images` list — rather than wasted tokens or an error mid-turn. The companion already computed `supportsVision` for `load_attachment`; `read_url` reuses the same signal. Tool description and prompt prose are conditional on it, so a non-vision model isn't told it can view images.

**2. Return images as base64 data URLs, not raw provider URLs.** Mirrors `load_attachment`. Fetching the bytes ourselves keeps the SSRF guards in the loop and lets us enforce the size cap before the model ever sees the image, rather than trusting the provider to fetch an arbitrary URL.

**3. Surface embedded image URLs as text, not auto-injected images.** Auto-fetching every `` on a page would blow context and cost. Listing the URLs and letting the agent pick one keeps it agent-driven and composes with surface #2 on a single path.

**4. Enclave `read_url` gets `supportsVision: true` unconditionally.** The enclave only serves e2e-capable personas (today only Ariadne, vision-capable) and already feeds images to its model via the ungated `load_attachment` (`apps/enclave/src/agent/attachment-tool.ts`). Threading a registry into the enclave just to re-derive a constant it already assumes elsewhere would be inconsistent; the comment records the assumption so both tools move together if a non-vision e2e persona is ever added.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/agent-runtime/src/tools/read-url-tool.ts` | `createReadUrlTool({ supportsVision })`; `readImageContent` (image/* → multimodal block, mime allowlist, 5 MB cap); `extractImageUrls` + `ReadUrlResult.images`; conditional description/prompt prose |
| `apps/backend/src/features/agents/built-in-agents.ts` | Add `LOAD_ATTACHMENT` to Ariadne's `enabledTools` |
| `apps/backend/src/features/agents/companion/tool-set.ts` | Pass `supportsVision` to `createReadUrlTool` |
| `apps/enclave/src/agent/tools.ts` | Build enclave `read_url` with `supportsVision: true` (rationale in comment) |
| `packages/agent-runtime/src/tools/read-url-tool.test.ts` | 6 new cases: image→multimodal, no-vision rejection, unsupported type, oversize-by-declared-length, embedded-URL listing, list omitted without vision |
| `apps/backend/src/features/agents/built-in-agents.test.ts` | Assert Ariadne's config keeps `LOAD_ATTACHMENT` |

## Out of scope (deliberate)

- **The `general_research` sub-loop stays `supportsVision: false`** (`persona-agent.ts:560`), matching the existing `load_attachment` treatment there — deep-research passes won't pull images, only top-level companion turns will. Easy to flip later.
- **No new `get_attachment`/`search_attachments` for Ariadne.** Reading a conversation image needs only the id already in context, so `load_attachment` alone covers the request without broadening her surface.
- **SVG and other image types remain unsupported** — limited to the four formats the providers accept natively.

## Test plan

- [x] `packages/agent-runtime` `bun test src/tools/read-url-tool.test.ts` — 42 pass (6 new)
- [x] `packages/agent-runtime` full suite — 205 pass
- [x] `apps/backend` `bun test src/features/agents/` — 295 pass
- [x] `apps/enclave` `bun test src/agent/tools.test.ts` — 6 pass
- [x] `tsc --noEmit` clean for agent-runtime, backend, enclave (also the full monorepo typecheck via the pre-push hook)
- [x] Self-review: caught `SourceItem` has no `domain` field (the multimodal source literal) and dropped it; the pre-existing read_url sources only compiled via an intermediate `const`
- [ ] Not manually exercised against a live model — covered by unit tests at the tool boundary (the runtime's multimodal injection is unchanged and already in production for `load_attachment`)

> Note: `apps/enclave` has 29 pre-existing failures in `createBackendCallbacks.pollMessages`, identical on a clean `origin/main` checkout and unrelated to this change.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01U8B2tbu4w5VDnYezHMETWa)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/958"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784153388&installation_id=129946197&pr_number=958&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F958&signature=291fcd95f6a87228a4900f8b58f9a71874c85ee75c4fa11eefc2270556368e25"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/built-in-agents.test.ts
+++ b/apps/backend/src/features/agents/built-in-agents.test.ts
@@ -1,5 +1,14 @@
 import { describe, expect, it } from "bun:test"
-import { ARIADNE_AGENT_ID, EMPTY_AGENT_ID, isE2eCapablePersona } from "./built-in-agents"
+import { AgentToolNames } from "@threa/types"
+import { BUILT_IN_AGENTS, ARIADNE_AGENT_ID, EMPTY_AGENT_ID, isE2eCapablePersona } from "./built-in-agents"
+
+describe("Ariadne built-in config", () => {
+  it("can load attachments for visual analysis (vision-capable model)", () => {
+    // load_attachment is what lets Ariadne actually SEE an uploaded image, not
+    // just read its caption — keep it enabled so that capability can't regress.
+    expect(BUILT_IN_AGENTS[ARIADNE_AGENT_ID].enabledTools).toContain(AgentToolNames.LOAD_ATTACHMENT)
+  })
+})
 
 describe("isE2eCapablePersona", () => {
   it("is true for Ariadne (the enclave persona)", () => {
--- a/apps/backend/src/features/agents/built-in-agents.ts
+++ b/apps/backend/src/features/agents/built-in-agents.ts
@@ -70,6 +70,7 @@ Keep responses short and direct. Default to a few sentences unless the user asks
       AgentToolNames.GENERAL_RESEARCH,
       AgentToolNames.DESCRIBE_MEMO,
       AgentToolNames.REACT_TO_MESSAGE,
+      AgentToolNames.LOAD_ATTACHMENT,
       AgentToolNames.GITHUB_LIST_REPOS,
       AgentToolNames.GITHUB_LIST_BRANCHES,
       AgentToolNames.GITHUB_LIST_COMMITS,
--- a/apps/backend/src/features/agents/companion/tool-set.ts
+++ b/apps/backend/src/features/agents/companion/tool-set.ts
@@ -117,7 +117,7 @@ export function buildToolSet(config: ToolSetConfig): AgentTool[] {
     tavilyApiKey && isToolEnabled(enabledTools, AgentToolNames.WEB_SEARCH)
       ? createWebSearchTool({ tavilyApiKey, currentTime, timezone })
       : null,
-    isToolEnabled(enabledTools, AgentToolNames.READ_URL) ? createReadUrlTool() : null,
+    isToolEnabled(enabledTools, AgentToolNames.READ_URL) ? createReadUrlTool({ supportsVision }) : null,
 
     // Workspace search tools
     workspace && isToolEnabled(enabledTools, AgentToolNames.SEARCH_MESSAGES)
--- a/apps/enclave/src/agent/tools.ts
+++ b/apps/enclave/src/agent/tools.ts
@@ -68,7 +68,10 @@ export function buildEnclaveTools(deps: EnclaveToolDeps): AgentTool[] {
         createWebSearchTool({ tavilyApiKey: deps.tavilyApiKey, currentTime: deps.currentTime, timezone: deps.timezone })
       )
     }
-    tools.push(createReadUrlTool())
+    // The enclave only serves e2e-capable personas, which run vision-capable
+    // models — and it already feeds images to the model via the ungated
+    // load_attachment tool above — so read_url may return images here too.
+    tools.push(createReadUrlTool({ supportsVision: true }))
     return tools
   }
 
--- a/packages/agent-runtime/src/tools/read-url-tool.test.ts
+++ b/packages/agent-runtime/src/tools/read-url-tool.test.ts
@@ -95,6 +95,110 @@ describe("read-url-tool", () => {
     expect(parsed.error).toContain("application/pdf")
   })
 
+  describe("image handling", () => {
+    // A 1x1 PNG — content is irrelevant, we only assert it round-trips as base64.
+    const pngBytes = Buffer.from(
+      "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+M8AAAMCAQDeY7HoAAAAAElFTkSuQmCC",
+      "base64"
+    )
+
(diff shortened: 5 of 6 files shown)


kristofferremback:
@coderabbitai review

---

kristofferremback:
Thanks for the review — all five addressed in `52ed3f4`.

1. **Mismatched-quote regex** — fixed. Switched `extractImageUrls` to an alternation that enforces matching delimiters and lets a value contain the opposite quote: `/]*?\bsrc\s*=\s*(?:"([^"]+)"|'([^']+)')/gi`, taking `match[1] ?? match[2]`. (Your first `(["'])([^"']+)\1` suggestion still drops values like `O'Reilly.png` because `[^"']+` excludes both quotes; the alternation handles it.) Test added.

2. **Private/internal image URLs in context** — fixed. Factored the synchronous host checks out of `validateUrlWithDns` into a shared `staticInternalHostReason(hostname)` (localhost, internal suffixes, bare private/reserved IPs via the existing `isPrivateOrReservedIP`), and `extractImageUrls` now drops any URL it flags. DNS-based cases still fall to `read_url` as you noted. Test added with a `169.254.169.254` ``.

3. **Missing actual-bytes-over-cap test** — added, with no `content-length` header so the post-download guard is what fires.

4. **SVG rejection test** — added `expect(multimodal).toBeUndefined()`.

5. **Image source title** — now uses the filename (`new URL(url).pathname.split("/").pop()`) instead of the raw URL.

`read-url-tool.test.ts` is at 45 pass; agent-runtime suite 208 pass; monorepo typecheck clean.

---

PR #963, opened 2026-06-16
feat(composer): convert large pastes into snippet attachments

kristofferremback:
## Problem

Pasting a large blob into the composer — a stack trace, a config file, a few hundred lines of code — dumps the whole thing inline. ProseMirror has to parse and render every line as editable rich text, the message becomes unreadable, and the editor bogs down on genuinely large clipboards. There was no "this is too big to be a message, make it a file" path the way Slack snippets work.

## Solution

Intercept oversized text pastes in the composer and route them through a plain snippet editor; on save the content becomes a `.txt` attachment chip at the exact spot it was pasted. The editor is a native `<textarea>`, not a rich editor, so the browser renders and scrolls arbitrarily large content natively — no virtualization, no parse cost.

The whole feature rides the attachment path that pasted images and dropped files already use (`handleFileInsert` → `onFileUpload`), so the inline chip, upload-scope-drift guarding, and E2E client-side encryption all come for free — a snippet pasted into an E2E stream is encrypted exactly like any other attachment.

### How it works

1. **Detection** — `shouldConvertPasteToSnippet(text)` (`snippet-paste.ts`) returns true when a paste is `>= 1500` chars OR `>= 12` lines (`SNIPPET_PASTE_CHAR_THRESHOLD` / `SNIPPET_PASTE_LINE_THRESHOLD`). Line count is `(text.match(/\n/g)?.length ?? 0) + 1`, which counts both LF and CRLF.
2. **Trigger** — in `RichEditor`'s `handlePaste`, after the file-paste and memo-link branches, a large text paste calls `event.preventDefault()`, records the caret (`editor.state.selection.from`) in `snippetInsertPosRef`, bumps a per-session counter, and sets `snippetDraft`. Gated on `onFileUploadRef.current` being present — composers without attachment support (edit forms) fall through to the existing inline-paste behavior.
3. **Editor** — `SnippetEditorDialog` is a `ResponsiveDialog` (centered Dialog on desktop, keyboard-safe content-height Drawer on mobile via `disableSnapPoints`). Monospace `<textarea wrap="off">` for code-column layout, an editable filename input (defaults `snippet-N.txt`), a live char/line counter, and Cmd/Ctrl+Enter to save (plain Enter stays free for newlines). It re-seeds from the incoming paste on each open so a cancelled snippet can't leak into the next.
4. **Save** — `handleSnippetSave` synthesizes `new File([text], filename, { type: "text/plain" })`, restores the caret to the remembered paste position (clamped to `doc.content.size`, since the dialog stole focus), and hands the file to `handleFileInsert` — the same call pasted files make. The chip uploads, swaps its temp id for the real attachment id, and sends as a normal attachment; the backend runs text extraction on it like any uploaded text file.

### Key design decisions

**1. Native `<textarea>`, not ProseMirror** — the trigger for this feature is "the clipboard is too big to render as rich text", so the editor cannot itself be a rich-text surface. A `<textarea>` is the one widget browsers virtualize natively, which is what "works with a silly amount of text" requires. No syntax highlighting (out of scope, and highlighting megabytes would reintroduce the cost we're avoiding).

**2. Trigger on lines OR characters** — the two have opposite blind spots. A 2,000-char single-line minified blob has one line; a 30-line shell script can be under 1,500 chars. Either alone misses half the cases, so both thresholds are checked.

**3. Convert even inside code blocks** — a user ruling. The intuition that a code block "wants" inline code loses to the fact that the paste is still too large to read inline regardless of cursor context; the snippet attachment is the better home either way.

**4. Reuse `handleFileInsert` rather than a parallel path** — the alternative was a bespoke snippet upload, which would have to re-implement temp-id tracking, scope-drift guarding (`uploadScopeVersionRef`), and E2E encryption, and would drift from the file path over time. Routing a synthesized `File` through the existing path means a snippet *is* an attachment, with no new backend surface.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/editor/snippet-paste.ts` | Pure trigger heuristic (`shouldConvertPasteToSnippet`), thresholds, and `defaultSnippetFilename` — shared by the paste handler and tests (INV-33). |
| `apps/frontend/src/components/editor/snippet-editor-dialog.tsx` | The monospace snippet editor (responsive Dialog/Drawer, filename field, char/line counter). |
| `apps/frontend/src/components/editor/snippet-paste.test.ts` | Unit tests for the threshold logic (char, line, CRLF, filename numbering). |
| `apps/frontend/src/components/editor/snippet-editor-dialog.test.tsx` | Component tests: seeding, save payload (with filename trim), disabled-save guard, cancel-without-save. |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/editor/rich-editor.tsx` | Detect large pastes in `handlePaste`; hold `snippetDraft` state + caret/counter refs; `handleSnippetSave` synthesizes the `File` and inserts at the paste caret via `handleFileInsert`; render `SnippetEditorDialog` (only when `onFileUpload` is wired). |

## Out of scope (deliberate)

- **Syntax highlighting** — explicitly excluded; see decision 1.
- **Tab-to-indent in the editor** — would trap keyboard focus (a11y); left out of the "very simple" v1.
- **Cancel does not re-insert the text inline** — the paste is discarded (the clipboard still holds it, so re-paste works). No "insert as plain text anyway" escape hatch in v1.
- **No backend changes** — a snippet is an ordinary `text/plain` attachment; the upload, storage, extraction, and timeline-rendering paths are untouched.

## Test plan

- [x] `bun run test src/components/editor/snippet-paste.test.ts src/components/editor/snippet-editor-dialog.test.ts` — 9 pass
- [x] `bun run test src/components/editor` (full editor suite incl. paste path) — 326 pass
- [x] `bun run test src/components/composer/message-composer.test.tsx` — passes (composer host unaffected)
- [x] `tsc --noEmit` across the monorepo — clean (pre-commit hook)
- [x] `eslint` on all touched files + monorepo lint (pre-commit hook) — clean
- [x] Self-review of the diff against the claims above — verified (reuse of `handleFileInsert`, `onFileUpload` gating, no code-block guard, caret clamp)
- [ ] Manual in-app paste of a large blob — not run in this session; the flow is covered by unit/component tests but not yet exercised end-to-end in the browser

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01DHrmfPn1MYURXRM26FyCSi)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/963"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784185504&installation_id=129946197&pr_number=963&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F963&signature=1835eec922877ca926000a6cea299cffc4c6f28c2635dabbcd6f59ebb4977eb5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/editor/rich-editor.tsx
+++ b/apps/frontend/src/components/editor/rich-editor.tsx
@@ -32,6 +32,8 @@ import { useMentionables } from "@/hooks/use-mentionables"
 import { useWorkspaceEmoji } from "@/hooks/use-workspace-emoji"
 import { useGiphyEnabled } from "@/hooks/use-giphy-enabled"
 import { GiphyPickerDialog } from "./giphy-picker-dialog"
+import { SnippetEditorDialog } from "./snippet-editor-dialog"
+import { shouldConvertPasteToSnippet, defaultSnippetFilename } from "./snippet-paste"
 import type { GiphyGif } from "@threa/types"
 import { cn } from "@/lib/utils"
 import { usePreferences } from "@/contexts"
@@ -244,6 +246,14 @@ export const RichEditor = forwardRef<RichEditorHandle, RichEditorProps>(function
   const giphyEnabled = useGiphyEnabled(workspaceId) && enableCommands
   const [giphyOpen, setGiphyOpen] = useState(false)
 
+  // A large paste opens the snippet editor instead of inserting inline; on save
+  // it becomes a `.txt` attachment chip at the original paste position. The
+  // draft holds the pasted text + suggested filename; the caret position and a
+  // per-session counter live in refs so they don't re-render the editor.
+  const [snippetDraft, setSnippetDraft] = useState<{ text: string; filename: string } | null>(null)
+  const snippetInsertPosRef = useRef<number | null>(null)
+  const snippetCountRef = useRef(0)
+
   // Unfiltered for type-lookup: ensures all broadcast slugs always resolve correctly
   const { mentionables } = useMentionables()
   // Filtered for autocomplete dropdown only
@@ -437,6 +447,35 @@ export const RichEditor = forwardRef<RichEditorHandle, RichEditorProps>(function
     }
   }, [])
 
+  // Save the snippet editor's contents as a text attachment, inserting the chip
+  // back at the caret position the paste happened at. Reuses the same
+  // upload-and-insert path as pasted images/files (handleFileInsert), so E2E
+  // encryption, scope-drift guarding, and the inline chip all come for free.
+  const handleSnippetSave = useCallback(
+    ({ text, filename }: { text: string; filename: string }) => {
+      const editorInstance = editorRef.current
+      const insertPos = snippetInsertPosRef.current
+      setSnippetDraft(null)
+      snippetInsertPosRef.current = null
+      const uploadFn = onFileUploadRef.current
+      if (!editorInstance || editorInstance.isDestroyed || !uploadFn) return
+
+      const safeName = filename.trim() || "snippet.txt"
+      const file = new File([text], safeName, { type: "text/plain" })
+
+      // Restore the caret to where the paste landed (the dialog stole focus).
+      const chain = editorInstance.chain().focus()
+      if (insertPos != null) {
+        const clamped = Math.min(insertPos, editorInstance.state.doc.content.size)
+        chain.setTextSelection(clamped)
+      }
+      chain.run()
+
+      void handleFileInsert(file, editorInstance)
+    },
+    [handleFileInsert]
+  )
+
   // Insert the chosen GIF as an inline embed rendered straight from Giphy's CDN
   // (no download/upload — that's how Giphy intends embeds to work). Synchronous,
   // so there's no scope-drift window to guard against.
@@ -530,6 +569,17 @@ export const RichEditor = forwardRef<RichEditorHandle, RichEditorProps>(function
             event.preventDefault()
             return true
           }
+        }
+
+        // Text too large to read inline becomes a snippet attachment: open the
+        // editor seeded with the paste, remembering the caret so the chip lands
+        // where the paste did. Needs an upload handler to attach through.
+        if (onFileUploadRef.current && shouldConvertPasteToSnippet(text)) {
+          event.preventDefault()
+          snippetInsertPosRef.current = editorRef.current.state.selection.from
+          snippetCountRef.current += 1
+          setSnippetDraft({ text, filename: defaultSnippetFilename(snippetCountRef.current) })
+          return true
         }
 
         const handled = insertPastedText(
@@ -994,6 +1044,20 @@ export const RichEditor = forwardRef<RichEditorHandle, RichEditorProps>(function
           onOpenChange={setGiphyOpen}
           workspaceId={workspaceId}
           onSelect={handleGifSelect}
+        />
+      ) : null}
+      {onFileUpload ? (
+        <SnippetEditorDialog
+          open={snippetDraft !== null}
+          onOpenChange={(open) => {
+            if (!open) {
+              setSnippetDraft(null)
+              snippetInsertPosRef.current = null
+            }
+          }}
+          initialText={snippetDraft?.text ?? ""}
+          defaultFilename={snippetDraft?.filename ?? "snippet.txt"}
+          onSave={handleSnippetSave}
         />
       ) : null}
     </div>
--- a/apps/frontend/src/components/editor/rich-editor.tsx
+++ b/apps/frontend/src/components/editor/rich-editor.tsx
@@ -33,7 +33,7 @@ import { useWorkspaceEmoji } from "@/hooks/use-workspace-emoji"
 import { useGiphyEnabled } from "@/hooks/use-giphy-enabled"
 import { GiphyPickerDialog } from "./giphy-picker-dialog"
 import { SnippetEditorDialog } from "./snippet-editor-dialog"
-import { shouldConvertPasteToSnippet, defaultSnippetFilename } from "./snippet-paste"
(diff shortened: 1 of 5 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (several non-blocking polish items; no bugs, data-integrity, or security findings)
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: sonnet x4 (spec+design, correctness+data-flow, UX, mobile)

Found 5 issues — all minor/non-blocking. The correctness pass found no logic bugs: the caret clamp to `doc.content.size` is safe (ProseMirror's `TextSelection.near` resolves it to the nearest valid text position), the `handleSnippetSave` `useCallback` deps are correct, and `onOpenChange` clears both `snippetDraft` and `snippetInsertPosRef` on cancel.

1. `snippet-editor-dialog.tsx:46,113` — No autofocus on the textarea. `textareaRef` is captured but never focused, so Radix focuses the first tabbable element — the **filename `<Input>`** — not the snippet body. A user who just pasted a large blob lands in the filename field instead of their content. (accessibility)
   https://github.com/threahq/threa/blob/1a1d100eb3f08a86a9604c95123a5cdf866ff1d6/apps/frontend/src/components/editor/snippet-editor-dialog.tsx#L112-L120

2. `snippet-editor-dialog.tsx:102` — Body container is missing `overflow-y-auto`. `responsive-dialog.tsx:54-58` explicitly requires `flex-1 min-h-0 overflow-y-auto` on the body when `disableSnapPoints` is set. On a phone with the keyboard open, the `min-h-[280px]` textarea plus filename input can exceed the drawer's `max-h-[92dvh]` clip with no scroll container, pushing the filename field or the Cancel/Attach buttons off-screen. (mobile-layout)
   https://github.com/threahq/threa/blob/1a1d100eb3f08a86a9604c95123a5cdf866ff1d6/apps/frontend/src/components/editor/snippet-editor-dialog.tsx#L102-L102

3. `rich-editor.tsx:468,1069` + `snippet-editor-dialog.tsx:107` — INV-33 (API Contracts §"Avoid magic strings by centralizing constants at the source of truth"): the fallback filename `"snippet.txt"` is hardcoded in three places, yet `snippet-paste.ts` was created expressly to centralize snippet constants (its own header cites INV-33). The blank-filename fallback should be a named export (e.g. `SNIPPET_FALLBACK_FILENAME`) so there's one source of truth. (claude-md)
   https://github.com/threahq/threa/blob/1a1d100eb3f08a86a9604c95123a5cdf866ff1d6/apps/frontend/src/components/editor/rich-editor.tsx#L467-L469

4. `snippet-editor-dialog.tsx:82` — `aria-describedby={undefined}` disconnects the `ResponsiveDialogDescription` ("This is too long to paste inline…") from the dialog's accessible description. That sentence is the one automatic explanation for why the dialog interrupted the paste; screen-reader users won't hear it on open. (Copied from `SavedEditDialog`, which has the same line — but that one has no meaningful description to lose.) (accessibility)
   https://github.com/threahq/threa/blob/1a1d100eb3f08a86a9604c95123a5cdf866ff1d6/apps/frontend/src/components/editor/snippet-editor-dialog.tsx#L78-L83

5. `snippet-editor-dialog.tsx:50-55` — The re-seed effect depends on `[open, initialText, defaultFilename]`, so a prop change while the dialog is open would clobber in-progress edits. Benign today (the parent freezes `snippetDraft` while open and mirrors the `SavedEditDialog` pattern), but a latent foot-gun: seeding should key on the `false→true` open transition only. Flagged independently by the spec and correctness lenses. (state-lifecycle)
   https://github.com/threahq/threa/blob/1a1d100eb3f08a86a9604c95123a5cdf866ff1d6/apps/frontend/src/components/editor/snippet-editor-dialog.tsx#L50-L56

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches the stated intent: large-paste detection in `handlePaste`, native-textarea editor, reuse of `handleFileInsert` → `onFileUpload`, editable `snippet-N.txt` filename, converts even inside code blocks, no toast on success (INV-63). Dialog correctly gated on `onFileUpload` being wired.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No logic bugs. Caret clamp safe, callback deps correct, ref cleanup on cancel correct, double-trim of filename harmless.</details>
<details><summary>🔁 Data Flow [1 issue]</summary>Issue 5 (re-seed effect) — latent only.</details>
<details><summary>📋 CLAUDE.md Compliance [1 violation]</summary>Issue 3 (INV-33 magic string).</details>
<details><summary>🏗️ Design [CLEAN]</summary>Reuses the existing attachment path rather than a parallel implementation; constants centralized except for issue 3.</details>
<details><summary>🔒 Security [CLEAN]</summary>Client-side textarea→File; no injection/XSS surface (React-safe, no `dangerouslySetInnerHTML`).</details>
<details><summary>🎛️ UX [2 issues]</summary>Issues 1 and 4 (focus target, description wiring).</details>
<details><summary>📱 Mobile [1 issue]</summary>Issue 2 (missing body scroll container under keyboard).</details>

---

PR #976, opened 2026-06-16
fix(drafts): stop a resolved-on-send draft resurrecting into the composer

kristofferremback:
**Context:** follow-up to the drafts resolve-on-send work (#932, stage 4a). No Linear ticket — found in production: a DM message sent twice (`msg_01KV92E42H…` and its now-deleted sibling, byte-identical `content_json`, two distinct `client_message_id`s, 1.7s apart).

## Problem

After sending a message, the just-sent draft could reload **into the composer**, and a confused second send produced a duplicate the backend couldn't dedupe (a fresh `client_message_id` per `sendMessage`, so `ON CONFLICT (stream_id, client_message_id)` doesn't fire).

The restore is a local race, not a re-invalidation. Send clears the editor and calls `composer.resolveDraft()` **fire-and-forget**, which tears down the draft row + the device-local `composerLoaded` pointer asynchronously. A debounced `saveDraft` armed by the last keystroke can fire in that window; `upsertLoadedDraft` **independently re-reads** the loaded pointer (`getLoadedDraftId`), and if that read lands *after* the pointer was cleared it takes the create branch — writing a brand-new draft with the just-sent content **and a new pointer**. `useDraftComposer`'s late-hydrate effect then refills the editor. Whether it fires depends on Dexie read/delete interleaving, which is why it was intermittent.

Two narrower gaps compounded it:
- The inbound-apply resurrection guard only honored a pending `delete_draft`, never `resolve_draft` — so an in-flight `draft:upserted` echo, or a reconnect `syncDrafts` bootstrap that re-seeds the still-present server row before the resolve op drains, re-created the draft (as a stash entry).
- Inbound sync legitimately never writes the `composerLoaded` pointer, so a roamed/echoed draft can only reach the stash — but nothing asserted that, so it could regress.

## Solution

A short-lived, device-local **resolution guard** keyed two ways — by scope (the local stale-save race) and by `(draftId, version)` (inbound echoes) — so a just-sent draft can't be revived by either path. No backend dedup backstop (deliberately rejected) and no change to the debounce.

### How it works

1. `resolveLoadedDraft` calls `markScopeResolved(scope)` as its **first** statement, before the async teardown. `upsertLoadedDraft`'s create branch then returns early on `isScopeRecentlyResolved(scope)`. The ordering is the correctness argument: the mark happens-before the pointer delete, which happens-before any read that returns a null pointer and routes a racing save into the create branch — so the flag is always set by the time it matters, regardless of Dexie interleaving.
2. The guard is lifted (`clearScopeResolved`) the instant the user re-engages — a keystroke (`saveDraftDebounced`) or an attachment (`addAttachment`) — so the next message starts a fresh draft. `removeAttachment` returns early without a loaded draft, so it never hits the create branch and needs no clear.
3. `resolveLoadedDraft` also records `markDraftResolved(loadedId, baseVersion)`. `applyDraftUpserted` (and thus `applyDraftsBootstrap`, which loops through it) drops an inbound row via `isResolvedDraftEcho(id, version)` at or below the resolved version. A **strictly newer** version is a real edit from another device and still applies — no data loss.
4. TTL is 60s — long enough to cover the debounce + send round-trip and an in-flight echo / reconnect bootstrap, short enough to be irrelevant once the user moves on. In-memory (module-local, mirroring the draft-store cache), so a full page reload is out of its window — an accepted gap (see Out of scope).

### Key design decisions

**1. Guard inside `upsertLoadedDraft`'s create branch, not at `saveDraft`'s entry** — the entry check can pass before `resolve` marks the scope (the save started during `await sendMessage`). Only the create branch has the happens-before guarantee (`null` pointer ⟹ teardown committed ⟹ mark set). The update-existing branch is left alone; it's the legitimate "update the loaded draft" path, and resolve deletes that row afterward anyway.

**2. Version-aware echo suppression (`<= resolved`), not blanket-by-id** — blanket suppression would also drop a genuine newer edit roamed from another device, violating the no-loss rule the draft system is built on ("duplicates acceptable, lost drafts not"). Matches the user's framing: ignore my own `(id, version)`, respect anyone else's newer one.

**3. No backend content/time-window dedup** — explicitly rejected by the user as its own source of bugs. The fix lives entirely in the draft layer.

**4. Debounce left as-is** — it's already a trailing 500ms-after-last-edit debounce (`saveDraftDebounced`), which is the desired shape; the bug was the teardown race, not the debounce cadence.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/sync/draft-resolution-guard.ts` | In-memory, TTL'd guard: `markScopeResolved`/`isScopeRecentlyResolved`/`clearScopeResolved` (local stale-save race) and `markDraftResolved`/`isResolvedDraftEcho` (inbound echo by id+version). |
| `apps/frontend/src/sync/draft-resolution-guard.test.ts` | Unit coverage: scope mark/lift/expiry, version-aware echo suppression, TTL. |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-draft-message.ts` | `resolveLoadedDraft` marks scope + records `(id, version)`; `upsertLoadedDraft` create branch refuses a just-resolved scope; `saveDraftDebounced`/`addAttachment` lift the guard on re-engagement. |
| `apps/frontend/src/sync/draft-sync.ts` | `applyDraftUpserted` drops `isResolvedDraftEcho` rows before the existing delete/dirty guards. |
| `apps/frontend/src/hooks/use-draft-message.test.ts` | Stale-save-after-resolve does not resurrect; re-engagement allows a fresh draft. |
| `apps/frontend/src/sync/draft-sync.test.ts` | Echo/bootstrap of a resolved draft dropped; strictly-newer version survives; inbound sync never writes the `composerLoaded` pointer. |

## Out of scope (deliberate)

- **Backend dedup backstop** — rejected by the user.
- **The CAS-fail tail** — if a draft's version drifted (e.g. an ack hadn't landed at send time), the server keeps it and a reconnect *after* the 60s window re-seeds it as a stash entry. Correct per no-loss, but a possible surprise zombie; separate, narrower issue.
- **Page-reload window** — the guard is in-memory; a reload inside the window loses it. The reconnect bootstrap then re-seeds at most a stash entry (never the composer — the pointer is local and was cleared), so it can't reproduce the double-send.
- **Formalizing "inbound sync never activates the composer" as a numbered invariant in `CLAUDE.md`** — enforced here by test + comments; deferred pending a call on the INV id.

## Test plan

- [x] `apps/frontend` — `draft-resolution-guard.test.ts` + `draft-sync.test.ts` + `use-draft-message.test.ts`: 87 pass
- [x] `apps/frontend` — adjacent composer/draft suites (`use-draft-composer`, `use-stash-composer`, `use-stashed-drafts`, `use-all-drafts`, `message-input`): 73 pass
- [x] `apps/frontend` `tsc --noEmit` clean; full-monorepo typecheck clean (pre-commit hook, exit 0)
- [ ] No headless multi-tab/reconnect harness — the resurrection paths are covered at the unit level (guard + apply functions), not as a live end-to-end repro

---

https://claude.ai/code/session_01RDkgonSQip5n93w24DdV6D

---
_Generated by [Claude Code](https://claude.ai/code/session_01RDkgonSQip5n93w24DdV6D)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/976"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784239236&installation_id=129946197&pr_number=976&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F976&signature=311d0609a81d11906469c1bac87c514282b1a6acc7533f1e3726ce5f98d3e576"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/hooks/use-draft-message.test.ts
+++ b/apps/frontend/src/hooks/use-draft-message.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
 import { renderHook, act, waitFor } from "@testing-library/react"
-import { useDraftMessage, getDraftMessageKey, upsertLoadedDraft } from "./use-draft-message"
+import { useDraftMessage, getDraftMessageKey, upsertLoadedDraft, resolveLoadedDraft } from "./use-draft-message"
+import { clearScopeResolved, resetDraftResolutionGuard } from "@/sync/draft-resolution-guard"
 import { ContextRefKinds, type JSONContent } from "@threa/types"
 import type { DraftContextRef } from "@/lib/context-bag/types"
 import { db } from "@/db"
@@ -49,6 +50,7 @@ describe("useDraftMessage", () => {
   beforeEach(async () => {
     vi.restoreAllMocks()
     resetDraftStoreCache()
+    resetDraftResolutionGuard()
     await db.drafts.clear()
     await db.composerLoaded.clear()
     await db.pendingOperations.clear()
@@ -410,6 +412,53 @@ describe("useDraftMessage", () => {
 
       expect(await loadedDraft(draftKey)).toBeUndefined()
       expect(await db.pendingOperations.where("type").equals("resolve_draft").count()).toBe(0)
+    })
+
+    it("a stale debounced save firing after resolve-on-send does NOT resurrect the draft", async () => {
+      // The exact production race: the user's last keystroke armed a debounced
+      // save; the send tears the draft down (resolveLoadedDraft), then the stale
+      // save fires with the just-sent content. Without the guard it re-creates the
+      // draft + composer pointer and the editor refills — so the user sends twice.
+      await db.drafts.put({
+        id: "draft_confirmed",
+        workspaceId,
+        scope: draftKey,
+        contentJson: makeDoc("sent message"),
+        attachments: [],
+        baseVersion: 2,
+        clientUpdatedAt: Date.now(),
+      })
+      await db.composerLoaded.put({ scope: draftKey, workspaceId, draftId: "draft_confirmed" })
+
+      await resolveLoadedDraft(workspaceId, draftKey)
+      // Stale save fires (its keystroke predates the send): no loaded pointer now,
+      // so this takes the create path — which the guard must refuse.
+      await upsertLoadedDraft(workspaceId, draftKey, { contentJson: makeDoc("sent message"), attachments: [] })
+
+      expect(await db.composerLoaded.get(draftKey)).toBeUndefined()
+      const scoped = (await db.drafts.toArray()).filter((d) => d.scope === draftKey)
+      expect(scoped).toHaveLength(0)
+    })
+
+    it("allows a brand-new draft once the user engages the composer again after send", async () => {
+      await db.drafts.put({
+        id: "draft_confirmed",
+        workspaceId,
+        scope: draftKey,
+        contentJson: makeDoc("sent message"),
+        attachments: [],
+        baseVersion: 2,
+        clientUpdatedAt: Date.now(),
+      })
+      await db.composerLoaded.put({ scope: draftKey, workspaceId, draftId: "draft_confirmed" })
+
+      await resolveLoadedDraft(workspaceId, draftKey)
+      // The user types again — engagement lifts the post-send guard (this is what
+      // saveDraftDebounced / addAttachment do on real input).
+      clearScopeResolved(draftKey)
+      await upsertLoadedDraft(workspaceId, draftKey, { contentJson: makeDoc("a different message"), attachments: [] })
+
+      expect(await db.composerLoaded.get(draftKey)).toBeDefined()
     })
   })
 
--- a/apps/frontend/src/hooks/use-draft-message.test.ts
+++ b/apps/frontend/src/hooks/use-draft-message.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
 import { renderHook, act, waitFor } from "@testing-library/react"
 import { useDraftMessage, getDraftMessageKey, upsertLoadedDraft, resolveLoadedDraft } from "./use-draft-message"
-import { clearScopeResolved, resetDraftResolutionGuard } from "@/sync/draft-resolution-guard"
+import { getScopeResolveSeq, resetDraftResolutionGuard } from "@/sync/draft-resolution-guard"
 import { ContextRefKinds, type JSONContent } from "@threa/types"
 import type { DraftContextRef } from "@/lib/context-bag/types"
 import { db } from "@/db"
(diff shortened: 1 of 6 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (all findings fixed in-branch)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3 (spec+design, correctness+data-flow, UX)

Reviewed the original diff (`4af5f42`); 3 issues found, **all addressed in `f066eff`**. Mobile lens skipped — sync-logic-only diff, no UI/touch/viewport surface.

1. `apps/frontend/src/hooks/use-draft-message.ts` — **(UX, 82 — data loss)** The first version's scope guard denied *all* create-new-draft operations for a recently-resolved scope for 60s, lifted only by `saveDraftDebounced`/`addAttachment`. Other legitimate create paths — `saveShareContent` (OS share-sheet → stream), the DM-scope migration — would be **silently dropped** within the window. The opposite of "never lose data."
   → Fixed by replacing the time-windowed, lift-per-caller guard with a **monotonic per-scope resolve sequence** that only the debounced save opts into (`observedResolveSeq`). Every other create path passes no seq and is unaffected by construction.

2. `apps/frontend/src/hooks/use-draft-message.ts` — **(design, 82)** `return row` on the guarded path handed back an unpersisted `CachedDraft`, misleading to callers that use the return.
   → After the refactor the guarded early-return is reachable *only* from the debounced save, which ignores the return; no other caller opts in. Contained.

3. `apps/frontend/src/sync/draft-resolution-guard.ts` — **(design, 80)** `TTL_MS` was coupled to `DEBOUNCE_MS` with no enforced relationship.
   → The scope guard no longer has a TTL (it's a plain counter). The remaining TTL is only the inbound-echo guard, whose window reflects reconnect/echo timing, not the debounce.

**Correctness & data-flow: clean.** The reviewer verified the happens-before (synchronous seq bump precedes the pointer delete precedes the racing read), the guard placement, the `?? 0` version default, the echo check's inertness when nothing was resolved, and the no-loss "strictly-newer survives" rule. No issues.

---
<details><summary>📐 Plan Adherence — CLEAN</summary>Delivers the fix in the draft layer only; honors every stated constraint (no backend dedup; debounce unchanged; newer remote version doesn't clobber local edits; inbound never activates the composer).</details>
<details><summary>🔍 Bugs — CLEAN</summary>No correctness issues; the riskiest interleavings were simulated.</details>
<details><summary>🔁 Data Flow — CLEAN</summary>Draft round-trip + echo/self-event interleavings traced; no field-drop or resurrection survives.</details>
<details><summary>📋 CLAUDE.md Compliance — CLEAN</summary>Module-level Maps mirror the existing draft-store cache pattern (not a new hidden singleton); comments are load-bearing (INV-25).</details>
<details><summary>🏗️ Design — 2 concerns, both resolved</summary>Findings 2 & 3 above, fixed in `f066eff`.</details>
<details><summary>🔒 Security — CLEAN</summary>Local sync logic; no surface.</details>
<details><summary>🎛️ UX — 1 issue, resolved</summary>Finding 1 above (silent draft loss), fixed in `f066eff`.</details>
<details><summary>📱 Mobile — N/A</summary>No UI/touch/viewport changes.</details>

---

PR #974, opened 2026-06-16
fix: isolate Pi remote sessions by instance id

kristofferremback:
## Problem

The Pi remote extension used one global `config.instanceId` for every linked Pi session in `~/.pi/agent/threa-remote.json`. Multiple local Pi sessions could still claim and complete invocations because claims include `runtimeSessionId`, but each session heartbeat overwrote the same `bot_runtime_instances` row with its own `capabilities.runtimeSessionId`. Frontend presence only renders connected when the session link's `instanceId` and `runtimeSessionId` match the current presence row, so a second session made the first scratchpad show "Not connected" while it could still respond.

## Solution

Make the runtime instance identity session-scoped for Pi remote links, and add a public API rebind path so existing scratchpads can migrate without creating replacement scratchpads.

### How it works

1. `extensions/pi-remote/src/threa-remote.ts` now stores `instanceId` on each `linkedSessions[runtimeSessionId]` entry. New `/remote-control` links call `createInstanceId()` before `POST /bot-runtime/sessions` and persist that id with the link.
2. Heartbeat, WebSocket `bot:hello`, claim, renew, trace-step, complete, and fail calls use the current session or claim's instance id instead of the legacy global id.
3. Claimed invocations carry `claimedInstanceId` locally so completion/renew/fail continues to authorize with the instance id that won the claim.
4. `POST /api/v1/workspaces/:workspaceId/bot-runtime/sessions/rebind` moves an active session link from the legacy instance id to a new per-session id. The plugin calls this lazily on `session_start` / `/remote-control on` when it sees an old link with no stored `instanceId`.
5. `BotRuntimeService.rebindPiRemoteSessionInstance` updates the link and upserts presence in one transaction, preserving the existing BIK behavior with `retainBik`.

### Key design decisions

**1. Rebind existing links instead of creating new scratchpads** — Old local config already has many linked scratchpads. Recreating them would strand conversation history and active stream links. The rebind endpoint updates `bot_runtime_session_links.instance_id` for the existing `linkId` only when workspace, bot, old instance id, runtime session id, runtime kind, and active status all match.

**2. Keep the legacy global id only as a compatibility fallback** — Existing config may not have per-link ids until it talks to a backend that supports `/sessions/rebind`. The plugin still falls back to `config.instanceId` when a link has no `instanceId`, but warns that connected status can be wrong with multiple sessions.

[...]

PR #978, opened 2026-06-16
feat(channels): drive Claude Code from a Threa scratchpad (claude-code-channel runtime)

kristofferremback:
## Problem

You can drive **Pi** from a Threa scratchpad (`extensions/pi-remote/`): type in the scratchpad, Pi works against your real files, the reply posts back. There was no equivalent for **Claude Code**, even though Claude Code's [channels](https://code.claude.com/docs/en/channels) feature (research preview, ≥ 2.1.80) is built for exactly this — an MCP server that pushes events into a running session and lets Claude reply through a tool.

The backend half-anticipated it: `claude-code-channel` already existed in `BOT_RUNTIME_KINDS`, but as a stub — `runtime-kind-config` had it as `sessionLinking: "none"` and `createRuntimeSessionSchema` was pinned to `z.literal("pi-local")`. So the only runtime kind that could create a scratchpad, become its active actor, and receive targeted active-scratchpad turns was `pi-local`. A Claude Code channel had no first-class way onto those rails.

## Solution

Two halves:

1. **`extensions/claude-code-remote/`** — a Claude Code *channel* (an MCP stdio server Claude Code spawns) that bridges a running session to a Threa scratchpad. It's the Claude Code counterpart to `pi-remote`, riding the same bot-runtime API.
2. **Make `claude-code-channel` a first-class linked runtime kind** on the backend, so the channel uses its real kind rather than masquerading as `pi-local`.

The two systems push opposite ways — a Claude Code channel *pushes* events into the session and exposes a reply tool; Threa's bot-runtime is a *pull/claim* queue. The extension is the adapter between them.

### How it works

**The bridge (`extensions/claude-code-remote/src/channel-server.ts`).** On start it reads config, derives a per-cwd-stable `instanceId`/`runtimeSessionId` (so relaunching in the same directory reuses the same scratchpad), creates a runtime session (scratchpad + active-actor link), registers presence, connects the `/bot` socket (falling back to a poll if there's no `wsUrl`), and drains claims. For each claimed invocation it pushes `notifications/claude/channel` with the prompt + the claim's hydrated context, so Claude reads it as `<channel source="threa" invocation_id="…">…</channel>`; Claude calls the `reply` tool, whose handler calls `…/bot-invocations/:id/complete` and the answer posts into the scratchpad. Claims are renewed at ⅓ of the 120s lease while in flight; an unanswered turn is force-closed after `replyTimeoutMs` (default 30 min).

**Permission relay (opt-in, default on).** The channel declares `claude/channel/permission`. When Claude hits a tool-approval prompt, the request is posted into the scratchpad ("Reply `yes <id>` / `no <id>`"); the user's reply comes back as an ordinary scratchpad message — hence a claimed invocation — which the channel recognizes as a verdict (`parsePermissionVerdict`) and routes to Claude Code as `notifications/claude/channel/permission` instead of pushing it as a fresh prompt.

**Backend — `claude-code-channel` as a linked kind.**
- `createRuntimeSessionSchema` now accepts `z.enum(["pi-local", "claude-code-channel"])` — the two kinds whose turns are pinned to a session link.
- `createOrLinkPiRemoteSession[InTransaction]` takes a required `runtimeKind` and stamps it on both the presence and link rows, gating the Pi-only session-control capabilities (`/compact`, `/model`, …) the channel can't drive.
- `runtime-kind-config` gives `claude-code-channel` `sessionLinking: "required"` with its own missing-session-link notice.
- No migration: `runtime_kind` is `TEXT` (INV-3) and `claude-code-channel` was already a valid `BOT_RUNTIME_KINDS` value.

### Key design decisions

**1. Report the real `claude-code-channel` kind, not `pi-local`.** The first cut presented as `pi-local` to ride the active-scratchpad rails with zero backend change — but that's a masquerade that would mis-attribute the runtime and offer it Pi-only session-control commands. Making the kind first-class is a small, additive backend change and is the honest shape. The kind is invisible in the UI (the presence pill shows the display name).

**2. The session-link finder identifies a client by `(instanceId, runtimeSessionId)`, not by kind.** `findActivePiRemoteSession` hardcoded `runtime_kind = 'pi-local'`. A runtime client is uniquely identified by its instance+session; the kind is an attribute of that client, and the `(workspace_id, bot_id, runtime_kind, instance_id, runtime_session_id)` unique key already keeps rows distinct. Filtering the lookup on kind is what caused the reuse bug below.

**3. One turn at a time.** The channel can't observe Claude's turn lifecycle, only inbound + the reply. It claims serially (a follow-up sent mid-turn waits for the current reply), with one exception: while a permission request is open it keeps draining so the verdict can reach the blocked turn.

## Design evolution (found in live testing)

Running the real channel against a live local stack surfaced a bug the unit tests and typecheck missed: the **second** `createSession` for the same channel (i.e. every restart in the same directory) returned **500 — `duplicate key value violates unique constraint "bot_runtime_session_links_workspace_id_bot_id_runtime_kind__key"`**. Cause: `findActivePiRemoteSession` hardcoded `runtime_kind = 'pi-local'`, so it never found the existing `claude-code-channel` link, re-ran the create path, and the link insert (whose `ON CONFLICT` targets `(…, root_stream_id, active_stream_id)`) violated the `(…, runtime_kind, instance_id, runtime_session_id)` unique key. Fixed by dropping the kind filter from `findActiveByRuntimeSession` (decision 2); re-verified live (the second launch now reuses the same scratchpad) and locked with an idempotency e2e test. A side effect of the 500 loop — `createScratchpad` runs before the transaction, so each failed attempt left an empty scratchpad — is noted in Out of scope.

## New files

| File | Purpose |
| ---- | ------- |
| `extensions/claude-code-remote/src/channel-server.ts` | The MCP channel ⇄ bot-runtime bridge (claim → push → reply tool → complete, renewal, permission relay) |
| `extensions/claude-code-remote/src/threa-client.ts` | Bearer HTTP client for the bot-runtime API + `/bot` ws-hint helpers |
| `extensions/claude-code-remote/src/config.ts` | Env/file config + per-cwd-stable id derivation |
| `extensions/claude-code-remote/src/index.ts` | Entry: load config, connect stdio, start, signal handling |
| `extensions/claude-code-remote/src/*.test.ts` | Unit tests (config, ws-hint/socket-url, verdict regex, prompt format, instructions) |
| `extensions/claude-code-remote/{package.json,tsconfig.json,README.md,.mcp.json.example}` | Standalone package (outside the bun workspaces, like `pi-remote`) + setup docs |
| `docs/claude-code-channel.md` | End-to-end explainer of the mechanism |
| `apps/backend/src/features/bot-runtimes/runtime-kind-config.test.ts` | Unit test: `claude-code-channel` is `required` + has its notice; link-free kinds stay `none` |
| `apps/backend/tests/e2e/claude-code-channel-session.test.ts` | e2e: session create, idempotent reuse, reject link-free kind, full round-trip |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/public-api/schemas.ts` | `createRuntimeSessionSchema.runtimeKind` → `z.enum(["pi-local","claude-code-channel"])` |
| `apps/backend/src/features/public-api/handlers.ts` | Pass `runtimeKind: data.runtimeKind` into the session-link create |
| `apps/backend/src/features/bot-runtimes/runtime-kind-config.ts` | `claude-code-channel` → `sessionLinking: "required"` + missing-link notice |
| `apps/backend/src/features/bot-runtimes/service.ts` | `createOrLinkPiRemoteSession*` takes `runtimeKind` (presence + link rows, gates session-control caps); `findActivePiRemoteSession` no longer injects `pi-local` |
| `apps/backend/src/features/bot-runtimes/repository.ts` | `findActiveByRuntimeSession` drops the `runtime_kind` filter (the reuse-bug fix) |
| `apps/backend/src/features/public-api/schemas.test.ts` | Cases: accepts both linked kinds, rejects link-free kinds |
| `apps/backend/src/features/bot-runtimes/invocation-outbox-handler.test.ts` | Case: unlinked `claude-code-channel` posts its notice |
| `docs/public-api/openapi.json` | Regenerated (the `runtimeKind` enum) |

## Out of scope (deliberate)

- **Per-tool trace.** A channel sees only inbound + reply, not Claude's tool calls, so the Threa trace card shows one "working" step, not the per-tool trace a Pi session produces.
- **Session-control commands** (`/compact`, `/model`, …) — a channel can't drive Claude's host session, so the kind doesn't advertise them.
- **`renameBotRuntimeSession` for `claude-code-channel`.** The finder fix makes rename kind-agnostic, but the channel doesn't call rename; only Pi does today.
- **`createScratchpad` runs before the link transaction**, so a failed link orphans an empty scratchpad. Pre-existing structure shared with `pi-local`; the finder fix removes the only path that hit it repeatedly. Not refactored here.

## Follow-ups (planned)

The next milestone is real parity with `pi-remote`:

- **Attachments, both directions** — download attachments posted in the scratchpad into the working directory so Claude can read them, and send local files back as attachments on the reply (Pi does this with a `THREA_ATTACH: <path>` directive that uploads the file and rewrites it to an attachment link). This is the main remaining gap to "on-par with Pi for real".
- **Per-tool trace** — a channel can't observe Claude's tool calls, so matching Pi's rich trace card needs a different mechanism than Pi's lifecycle hooks.

The README now carries a `Roadmap` and a `Troubleshooting` section — the latter documenting the rough edge that bit during live testing: declining the *"Use this MCP server?"* prompt silently blacklists the server in `disabledMcpjsonServers`, after which it never appears in `/mcp`.

## Test plan

- [x] **Run live against a local stack** — the channel links its scratchpad, registers presence as `claude-code-channel`, and claims active-scratchpad invocations; the full claim→push→reply→complete path is covered by the e2e round-trip test
- [x] Extension `bun test` — 29 pass; `bunx tsc --noEmit` clean
- [x] Backend e2e `tests/e2e/claude-code-channel-session.test.ts` — 3 pass (session create, idempotent reuse, reject link-free, full round-trip dispatch→claim→reply)
- [x] Backend regression: pi-local `tests/e2e/bot-runtime-websocket.test.ts` + `commands.test.ts` + the new file — 11 pass across 3 files (pi-local flows unaffected)
- [x] Backend unit: `schemas.test.ts`, `runtime-kind-config.test.ts`, `service.test.ts`, `invocation-outbox-handler.test.ts` — 32 pass
- [x] Full monorepo `tsc --noEmit` + `tsc -p tsconfig.tests.json` clean (enforced by the pre-commit hook); `eslint` clean on changed files; OpenAPI `generate:api-docs:check` clean
- [x] **Live**: the real channel run against a local stack creates a `claude-code-channel` session and reuses the same scratchpad on relaunch (the 500 is gone); presence + link rows persist `runtime_kind = 'claude-code-channel'`
- [x] Self-review: 3 parallel reviewer workflows (Threa-contract conformance, channel-protocol conformance, correctness/concurrency, design). Findings fixed: double-complete race (clear deadline before `complete`), renew cadence vs lease, shutdown bounding, socket double-attach guard, permission-map leak, dead `defaultDisplayName` (wired in). The reuse-500 was caught later in live testing and fixed (above).
- [ ] Live full round-trip with a real Claude Code reply in the browser — **not done**: local dev uses real WorkOS, so a user message couldn't be posted autonomously. The full dispatch→claim→reply path is instead proven by the e2e round-trip test against the in-process test server. See the morning runbook in the PR thread.

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/978"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784240232&installation_id=129946197&pr_number=978&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F978&signature=57057675cd175e6263610ed49982952351b889ef9a5498a58da191ee6afb5ad5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/bot-runtimes/invocation-outbox-handler.test.ts
+++ b/apps/backend/src/features/bot-runtimes/invocation-outbox-handler.test.ts
@@ -263,4 +263,20 @@ describe("BotInvocationOutboxHandler active-scratchpad session-link policy", () 
       expect.objectContaining({ actorId: "bot_1", trigger: "active-scratchpad" })
     )
   })
+
+  it("posts the Claude Code channel notice and skips dispatch when an unlinked claude-code-channel bot is active", async () => {
+    const { createInvocation, createMessage } = setupActiveScratchpad({
+      instance: { runtimeKind: "claude-code-channel" },
+    })
+
+    await runProcessMessageCreated(plainUserMessage)
+
+    expect(createInvocation).not.toHaveBeenCalled()
+    expect(createMessage).toHaveBeenCalledWith(
+      expect.objectContaining({
+        contentMarkdown: expect.stringContaining("Claude Code with the Threa channel"),
+        metadata: { "bot_runtime.notice": "missing_session_link" },
+      })
+    )
+  })
 })
--- a/apps/backend/src/features/bot-runtimes/repository.ts
+++ b/apps/backend/src/features/bot-runtimes/repository.ts
@@ -486,18 +486,22 @@ export const BotRuntimeSessionLinkRepository = {
     return result.rows[0] ? mapSessionLink(result.rows[0]) : null
   },
 
+  // A runtime client is identified by (instanceId, runtimeSessionId) — that pair is
+  // unique per client regardless of runtimeKind. The lookup must NOT filter on kind:
+  // doing so made session reuse miss a non-pi-local link and re-run the create path,
+  // which then violated the (…, runtime_kind, instance_id, runtime_session_id) unique
+  // constraint on the runtime's second launch.
   async findActiveByRuntimeSession(
     db: Querier,
     params: {
       workspaceId: string
       botId: string
-      runtimeKind: BotRuntimeKind
       instanceId: string
       runtimeSessionId: string
     }
   ): Promise<BotRuntimeSessionLink | null> {
     const result = await db.query<BotRuntimeSessionLinkRow>(
-      sql`SELECT * FROM bot_runtime_session_links WHERE workspace_id = ${params.workspaceId} AND bot_id = ${params.botId} AND runtime_kind = ${params.runtimeKind} AND instance_id = ${params.instanceId} AND runtime_session_id = ${params.runtimeSessionId} AND status = 'active'`
+      sql`SELECT * FROM bot_runtime_session_links WHERE workspace_id = ${params.workspaceId} AND bot_id = ${params.botId} AND instance_id = ${params.instanceId} AND runtime_session_id = ${params.runtimeSessionId} AND status = 'active'`
     )
     return result.rows[0] ? mapSessionLink(result.rows[0]) : null
   },
--- a/apps/backend/src/features/bot-runtimes/repository.ts
+++ b/apps/backend/src/features/bot-runtimes/repository.ts
@@ -486,11 +486,13 @@ export const BotRuntimeSessionLinkRepository = {
     return result.rows[0] ? mapSessionLink(result.rows[0]) : null
   },
 
-  // A runtime client is identified by (instanceId, runtimeSessionId) — that pair is
-  // unique per client regardless of runtimeKind. The lookup must NOT filter on kind:
-  // doing so made session reuse miss a non-pi-local link and re-run the create path,
-  // which then violated the (…, runtime_kind, instance_id, runtime_session_id) unique
-  // constraint on the runtime's second launch.
+  // A runtime client is identified by (instanceId, runtimeSessionId). Callers that
+  // know the kind they're acting for (session create) pass it so reuse is kind-exact
+  // — the original bug was the inverse, a hardcoded `runtime_kind = 'pi-local'` that
+  // made a claude-code-channel relaunch miss its own link and re-run the create path,
+  // violating the (…, runtime_kind, instance_id, runtime_session_id) unique key.
+  // `runtimeKind` omitted (rename) matches any kind; ORDER BY keeps that deterministic
+  // even though the (instance, session) pair is unique per kind in practice.
   async findActiveByRuntimeSession(
     db: Querier,
     params: {
@@ -498,10 +500,20 @@ export const BotRuntimeSessionLinkRepository = {
       botId: string
       instanceId: string
       runtimeSessionId: string
+      runtimeKind?: BotRuntimeKind
     }
   ): Promise<BotRuntimeSessionLink | null> {
+    const kind = params.runtimeKind ?? null
     const result = await db.query<BotRuntimeSessionLinkRow>(
-      sql`SELECT * FROM bot_runtime_session_links WHERE workspace_id = ${params.workspaceId} AND bot_id = ${params.botId} AND instance_id = ${params.instanceId} AND runtime_session_id = ${params.runtimeSessionId} AND status = 'active'`
+      sql`SELECT * FROM bot_runtime_session_links
+        WHERE workspace_id = ${params.workspaceId}
+          AND bot_id = ${params.botId}
+          AND instance_id = ${params.instanceId}
+          AND runtime_session_id = ${params.runtimeSessionId}
+          AND status = 'active'
+          AND (${kind}::text IS NULL OR runtime_kind = ${kind})
+        ORDER BY updated_at DESC
+        LIMIT 1`
     )
     return result.rows[0] ? mapSessionLink(result.rows[0]) : null
   },
--- a/apps/backend/src/features/bot-runtimes/repository.ts
+++ b/apps/backend/src/features/bot-runtimes/repository.ts
@@ -486,13 +486,11 @@ export const BotRuntimeSessionLinkRepository = {
     return result.rows[0] ? mapSessionLink(result.rows[0]) : null
   },
 
-  // A runtime client is identified by (instanceId, runtimeSessionId). Callers that
-  // know the kind they're acting for (session create) pass it so reuse is kind-exact
-  // — the original bug was the inverse, a hardcoded `runtime_kind = 'pi-local'` that
-  // made a claude-code-channel relaunch miss its own link and re-run the create path,
-  // violating the (…, runtime_kind, instance_id, runtime_session_id) unique key.
-  // `runtimeKind` omitted (rename) matches any kind; ORDER BY keeps that deterministic
-  // even though the (instance, session) pair is unique per kind in practice.
+  // A runtime client is identified by (instanceId, runtimeSessionId). Session-create
+  // callers pass `runtimeKind` so reuse matches the link of the kind being created —
+  // the unique key includes runtime_kind, so two kinds can hold the same instance/
+  // session pair. Rename callers omit it to match that runtime identity across kinds;
+  // `ORDER BY updated_at DESC LIMIT 1` keeps the kindless case deterministic.
   async findActiveByRuntimeSession(
     db: Querier,
     params: {
--- a/apps/backend/src/features/bot-runtimes/runtime-kind-config.test.ts
+++ b/apps/backend/src/features/bot-runtimes/runtime-kind-config.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from "bun:test"
+import { resolveRuntimeKindConfig } from "./runtime-kind-config"
+
+describe("resolveRuntimeKindConfig", () => {
+  it("requires a session link for pi-local and the Claude Code channel", () => {
+    const pi = resolveRuntimeKindConfig("pi-local")
+    const cc = resolveRuntimeKindConfig("claude-code-channel")
+    expect(pi.sessionLinking).toBe("required")
+    expect(cc.sessionLinking).toBe("required")
+    if (cc.sessionLinking === "required") {
+      const notice = cc.missingSessionLinkNotice("Scout")
+      expect(notice).toContain("Scout")
+      expect(notice).toContain("Claude Code")
+    }
+  })
+
+  it("leaves untargeted kinds link-free", () => {
+    expect(resolveRuntimeKindConfig("hermes").sessionLinking).toBe("none")
+    expect(resolveRuntimeKindConfig("openclaw").sessionLinking).toBe("none")
+    expect(resolveRuntimeKindConfig("custom").sessionLinking).toBe("none")
+  })
+
+  it("defaults a bot with no runtime instance to the pi-local policy", () => {
+    expect(resolveRuntimeKindConfig(null).sessionLinking).toBe("required")
+  })
+})
--- a/apps/backend/src/features/bot-runtimes/runtime-kind-config.ts
+++ b/apps/backend/src/features/bot-runtimes/runtime-kind-config.ts
@@ -1,11 +1,12 @@
 /**
  * Per-runtime-kind dispatch policy for the active-scratchpad path.
  *
- * Pi drives long-lived local sessions, so its active-scratchpad turns must be
- * pinned to an explicit session link (created via Pi's `/remote-control`
- * flow) — without one we post a notice telling the user how to link. Other
- * kinds never create session links; their invocations dispatch untargeted and
- * any live instance of the bot may claim them.
+ * Pi and the Claude Code channel drive long-lived local sessions, so their
+ * active-scratchpad turns must be pinned to an explicit session link (created
+ * via Pi's `/remote-control` flow or the Claude Code channel's startup) —
+ * without one we post a notice telling the user how to link. Other kinds never
+ * create session links; their invocations dispatch untargeted and any live
+ * instance of the bot may claim them.
  */
 
 import type { BotRuntimeKind } from "@threa/types"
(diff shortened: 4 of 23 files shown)


kristofferremback:
### Morning runbook — the live click-test

The backend change + the full mechanism are proven by the e2e round-trip test; this is the optional "watch it work in the browser" pass. Local dev uses real WorkOS, so this one needs your session — it's ~2 minutes.

1. **Stack** — if `bun run dev` isn't still up, start it (`bun run db:start` first if Postgres/MinIO are down). Backend on `:3002`, app on `:3000`.
2. **Creds** — you already have a personal bot + `threa_bk_…` key in your workspace (`ws_01KFDY969162D2MRKN827NCCSX`). Put them in `~/.claude/threa-channel/config.json`:
   ```json
   { "baseUrl": "http://localhost:3002", "workspaceId": "ws_01KFDY969162D2MRKN827NCCSX", "apiKey": "threa_bk_…" }
   ```
   (`:3002` direct — local dev doesn't hand out a `wsUrl`, so the channel polls; that's fine.)
3. **Register** — in any project dir, drop a `.mcp.json` (see `extensions/claude-code-remote/.mcp.json.example`, point `args` at the absolute path of `src/index.ts`).
4. **Run** — `claude --dangerously-load-development-channels server:threa`. The channel logs the scratchpad URL; it shows in the Threa sidebar as `Claude Code - <dir>`.
5. **Drive** — open that scratchpad, type "what files are in this directory?". Claude works against the real files and replies in the scratchpad. Add `--dangerously-skip-permissions` for unattended, or leave permission relay on and approve with `yes <id>` in the scratchpad.

Notes from the overnight session: a real `claude-code-channel` session already exists in your dev workspace from verification (scratchpad `stream_01KV96A30…`); 172 empty orphan scratchpads created by the reuse-500 loop (before the fix) were **archived**, not deleted.

kristofferremback:
### Code review
**Confidence: 5/7** — Good (non-blocking)
**Review models:** Orchestrator: opus-4-8 (1M) | Reviewers: sonnet x2 (spec+design, correctness+data-flow)

Found 2 issues, both non-blocking:

1. `apps/backend/src/features/bot-runtimes/repository.ts:494` — **kind-agnostic session-link finder is nondeterministic by convention.** `findActiveByRuntimeSession` dropped its `runtime_kind` filter (the session-reuse-500 fix), but the unique key is `(workspace_id, bot_id, runtime_kind, instance_id, runtime_session_id)` — two active links *can* exist for the same `(instance_id, runtime_session_id)` differing only by kind. With no `ORDER BY` and `rows[0]`, `createBotRuntimeSession` and `renameBotRuntimeSession` would resolve nondeterministically if both ever existed (a claude-code-channel restart could match a stale pi-local link → wrong scratchpad). **Not currently reachable** — the channel derives `cc-`/`ccs-` ids and Pi uses `pi-*`/uuid, so the tuples never collide — but the finder relies on that convention rather than enforcing it. (data-flow/82; also a naming-clarity issue — the method is still `…PiRemoteSession…` though it now serves both kinds.)
   _Fix:_ thread the request's `runtimeKind` through the create-path lookup so reuse is kind-exact (rename can stay kind-agnostic — it's a benign write), and add `ORDER BY updated_at DESC LIMIT 1`.
   https://github.com/threahq/threa/blob/665387d562ee85dfcded04b67388f520b7b9e9b7/apps/backend/src/features/bot-runtimes/repository.ts#L494-L509

2. `apps/backend/src/features/bot-runtimes/service.ts:176` — **stale `Pi`-prefixed names after broadening to all linked kinds.** `findActivePiRemoteSession`, `createOrLinkPiRemoteSession[InTransaction]`, and the private `upsertPiRemoteSessionPresenceInTransaction` keep the `Pi` prefix though they're now parametrized over any `BotRuntimeKind`. (design/80) The `Pi` prefix is the bot-runtime-session subsystem's established name (reinforced by #974's `rebindPiRemoteSessionInstance`), so a full family rename is **out of scope** here — flagging the naming lag for a future cleanup rather than diverging from the just-merged convention.
   https://github.com/threahq/threa/blob/665387d562ee85dfcded04b67388f520b7b9e9b7/apps/backend/src/features/bot-runtimes/service.ts#L176-L183

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches intent: the extension + first-class `claude-code-channel` kind + the reuse-500 fix. `runtimeKind` is threaded through the #974-extracted presence helper at every call site; the rebind path correctly stays `pi-local`; OpenAPI regenerated for the schema change.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No logic/race/lifecycle bugs. Checked the channel's in-flight map, renew-vs-120s-lease cadence, per-invocation deadline + reply re-arm, the serial-claim gate with its permission-verdict exception, socket double-attach guard, and bounded shutdown.</details>
<details><summary>🔁 Data Flow [1 issue]</summary>See #1 — kind-agnostic finder nondeterminism, gated by the id-prefix convention.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No invariant violations introduced. (The extension is a standalone package outside the bun workspaces, like `pi-remote`, so backend INVs apply in spirit.)</details>
<details><summary>🏗️ Design [2 issues]</summary>See #1 and #2 — the kind-agnostic finder + the stale `Pi`-prefixed naming after the contract widened to all linked kinds.</details>
<details><summary>🔒 Security [CLEAN]</summary>Permission relay lets any scratchpad poster approve tool calls — documented, and the 5-letter request id (24⁵ ≈ 8M) makes an accidental match negligible. Bot-key (`threa_bk_`) auth over HTTP, same as the rest of the bot-runtime API. No high-confidence vuln.</details>

_UX + Mobile lenses not run — backend/extension-only diff, no frontend changes._

kristofferremback:
Addressed review finding #1 in `3b3802d7` — `findActiveByRuntimeSession` now takes the request's `runtimeKind` so the session-create reuse lookup is kind-exact (rename stays kind-agnostic; added `ORDER BY updated_at DESC LIMIT 1` for determinism). Still fixes the original reuse-500. Finding #2 (Pi-prefixed naming) left as the deliberate, documented defer — it's the bot-runtime-session subsystem's established name.

kristofferremback:
**CI note — the `Tests` red is a pre-existing flaky frontend test, not from this change.**

Failure: `apps/frontend/src/sync/sync-engine.test.ts > SyncEngine sync cursor (active mode) > resumes the gate (drains the buffer) even when the batch flush throws` (`AssertionError: expected undefined to be defined`, :1294) — 257 passed, 1 failed.

Evidence it's not this PR:
- This diff is **backend + extension only — zero frontend files** (`git diff --name-only … | grep apps/frontend` → empty). A frontend sync-gate test can't be broken by it.
- Run **in isolation on a dev machine** (no backend), this exact test goes **pass / fail / pass** across 3 runs — flaky (~⅓ fail).
- `main`'s last 3 CI runs are green, and this PR's prior commit `665387d5` (identical frontend) passed Tests.

So: a timing-sensitive flake in the sync-engine drain-on-throw test, surfaced under CI contention. Out of scope to fix in this channel PR; flagging for the ongoing CI-flakiness effort. Re-running the Tests job.

kristofferremback:
Thanks @coderabbitai — addressed in `16578a52`:

**Fixed:**
- **Permission prompt → wrong stream** (`channel-server.ts`): now targets the executing turn via a tracked `activeTurnStreamId` (set when a non-verdict invocation is pushed, cleared at turn end), instead of `at(-1)` of the in-flight map — a follow-up message claimed during an open-permission window can no longer mis-target the prompt.
- **Presence stale on renew 404** (`channel-server.ts`): the 404 branch now `syncPresence()` after dropping the claim, so a last-in-flight 404 can't strand the runtime as busy.
- **Swallowed `server.start()` error** (`index.ts`): removed the catch — it propagates to `main().catch`, which logs and exits non-zero, so a connected-but-dead channel fails loud (INV-11).
- **INV-25 change-narration** in the finder comment (`repository.ts`) and the e2e header — rewritten to the current contract/intent.
- **Doc contradiction** (`docs/claude-code-channel.md`): dropped the stale "no backend changes" line.

**Intentionally not changed (with reasons):**
- **Cross-platform path split** (`config.ts`): the channel is launched via `bun` on a developer machine and the codebase assumes POSIX paths; a Windows split adds surface for a case this never runs in. Skipped.
- **More `ThreaClient` request/error unit tests**: the client's full request lifecycle is exercised end-to-end by the backend e2e suite (`claude-code-channel-session.test.ts` drives create → claim → complete against a real server). Mocking `fetch` to re-test the same paths adds maintenance for little signal; the URL/hint helpers that have non-trivial logic are unit-tested. Deferred.

PR #969, opened 2026-06-16
feat(e2ee): gate bot claims on BIK wrap coverage for sealed streams

kristofferremback:
## Summary

Adds the **claim-side gate** for sealed external-bot turns: `BotInvocationRepository.claimOne` now refuses to hand an E2E (sealed) invocation to a bot instance whose registered identity key (BIK) can't actually open it. This is the next slice of the "wire half" for letting an owner-granted external/self-hosted bot harness participate in an E2E stream (follows #967, which shipped the `SealedTurnContext` type + pure builder).

Everything stays **dark in production** — see "Inert today" below.

## What changed

`apps/backend/src/features/bot-runtimes/repository.ts` — `claimOne` gains a conditional sealed-stream predicate that mirrors the enclave's `EnclaveInvocationsRepository.claimNext` two-`EXISTS`:

- **Plaintext invocation** (no `e2e_streams` row for the root) → claimable by any instance of the bot, exactly as before.
- **E2E invocation** → claimable only when the *claiming instance's* registered BIK (`bot_runtime_instances.public_key_id`) has `stream_e2e_key_wraps` (`recipient_kind = 'bot'`) covering **both**:
  - the reply generation — `e.current_key_generation` (what the bot must seal its reply under), and
  - the prompt generation — `(m.envelope ->> 'keyGeneration')::int` for the trigger `source_message_id` (what the bot must open).

The candidate is aliased to `i` and the lock is now explicit (`FOR UPDATE OF i SKIP LOCKED`), since the new arm joins `e2e_streams`, `stream_e2e_key_wraps`, `messages`, and `bot_runtime_instances` — all inside `EXISTS` subqueries, so only `bot_invocations` is locked.

## Design decisions

- **Conditional, not unconditional.** Unlike the enclave's `claimNext` (which only ever serves E2E streams), this same `claimOne` serves plaintext streams too — hence the `NOT EXISTS (e2e_streams) OR (...both wraps...)` shape. Plaintext behavior is byte-unchanged.
- **Keyed to the instance's BIK, not a passed key id.** The bot path resolves the recipient key from the claiming instance row (`ri.public_key_id`), so a keyless instance (`public_key_id` NULL) never matches and can't claim a sealed turn it has no key to open.
- **Both generations, like the enclave.** Re-evaluated at claim time against the live wraps table (never snapshotted) so a key roll / revoke between enqueue and claim is reflected.

## Inert today

[...]

PR #983, opened 2026-06-17
feat(e2ee): sealed bot trace steps (/sealed-steps + /started) behind the policy flag

kristofferremback:
**Workstream:** E2EE for external bot harnesses — `docs/plans/agent-runtimes-unification-redesign.md` §2.6. Builds on #981 (sealed bot **claim**) and #969/#967. Phase 2 of 6.

## Problem

A sealed-capable external bot harness that wins a sealed claim (#981) has nowhere to send its trace steps. The enclave streams sealed steps back over `POST /internal/enclave-runtimes/sessions/:id/steps` and `/steps/started` (`session-handlers.ts`), persisting ciphertext the backend can't read; the public bot API has only the **plaintext** `/bot-invocations/:invocationId/steps` path, which runs content through the `TraceProjector` and stores cleartext. There is no sealed wire variant for bots, so an owner-granted E2E bot's reasoning/tool trace can't land at all.

§2.6 also names the binding the plaintext path lacks: per-runner callback identity. The plaintext `/steps` authenticates with `claimToken` in the body matched against `bot_invocations.claim_token`; a sealed turn carries the owner's plaintext, so it needs the stronger model-A binding the enclave already uses — a per-claim token verified against `agent_sessions.callback_token_hash` (E2EE-21/22).

## Solution

Add the external siblings of the enclave's sealed step callbacks, reusing the machinery #981 already built (the `callback_token_hash` / `reply_key_generation` columns set at claim, the `content_ciphertext` / `content_envelope` step columns from migration `20260530051907`). The whole path stays dark: no delivery verdict resolves to `sealed` until `externalSealedDelivery` flips (Phase 6), so these routes are unreachable in production.

### How it works

1. **Neutral callback header.** `THREA_CALLBACK_TOKEN_HEADER = "X-Threa-Callback-Token"` (`packages/types/src/constants.ts`) — the per-claim binding for *any* sealed external driver, deliberately **not** enclave-named (§2.6 rule 1). The enclave keeps its own `X-Enclave-Callback-Token`; this is a separate header for the bot transport.

2. **Shared sealed-session guards.** The three guards a sealed callback must pass — `assertSessionRunning`, `verifyCallbackToken`, `assertReplyKeyGeneration` — move into `apps/backend/src/features/agents/sealed-session-guards.ts` and are exported from the agents barrel. Each is a property of the `agent_sessions` row, not of one transport, so only the *header* the token rides in differs between drivers. The enclave's `assertCallbackBound` now reads its own header and delegates to `verifyCallbackToken`; its `assertRunning` / `assertReplyGeneration` become aliased imports of the shared functions (no call-site churn).

3. **`POST /bot-invocations/:invocationId/sealed-steps/started`** (`startBotInvocationSealedStep`) — opens an in-flight sealed step: insert the row (no `completed_at`) + `updateCurrentStepType` in one transaction (INV-6; `step_number` computed atomically in the INSERT, INV-20), then broadcast `agent_session:step:started` to the session room and `agent_session:progress` to the stream room. Content ciphertext/envelope are optional here (a tool step has no result yet), mirroring the enclave.

4. **`POST /bot-invocations/:invocationId/sealed-steps`** (`recordBotInvocationSealedStep`) — finalizes that step in place via `updateStep` (sealed content + `completed_at`, preserving `started_at` so duration is real). If the start POST was dropped, fall back to a race-safe completed `appendStep` + `updateCurrentStepType` so the trace still lands. Broadcasts `agent_session:step:completed`.

5. **Auth.** Both go through `authorizeSealedStepCallback`: bot invocations reuse the invocation id as the agent session id, so the path's `:invocationId` is the session id. The callback-token verify is the binding (a NULL row hash — every plaintext session — is rejected); on top of it, `stream.workspaceId !== req.workspaceId` and `session.personaId !== bot.id` enforce workspace/persona isolation (INV-8). The server only ever persists or relays ciphertext (INV-E7).

6. **Schemas + docs.** `recordSealedInvocationStepSchema` / `startSealedInvocationStepSchema` (`schemas.ts`) validate the body as the `SealedStep` / `SealedStepStart` wire shapes, with base64 decodability checked at the boundary (a malformed byte that slips through is a permanently unreadable step). Route descriptors declare the required `X-Threa-Callback-Token` header param; `docs/public-api/openapi.json` regenerated via `bun run generate:api-docs`.

### Key design decisions

**1. Header-bound auth, not body `claimToken`** — the sealed steps authenticate with the per-claim callback token in `X-Threa-Callback-Token` verified against `agent_sessions.callback_token_hash`, not the plaintext path's `claimToken`-in-body matched against `bot_invocations.claim_token`. §2.6: "per-runner identity so the grant binds to the bot it names … the callback auth … E2EE-21/22's fix covers both." Possession of the per-claim secret (delivered only inside the winning sealed claim, #981 Decision #1) is a stronger binding than a self-reported instance id any workspace-key holder could copy.

**2. Two dedicated routes, not a union-branch on `/steps`** — the plaintext `/steps` runs the `TraceProjector` and doubles as a busy-presence heartbeat; the sealed path writes ciphertext rows and emits start/completed frames directly. They share almost nothing, and `/steps/started` has no plaintext equivalent. Mirroring the enclave's start/finalize split keeps each wire format clean rather than overloading one schema with a discriminated union.

**3. Guards shared, not duplicated (INV-35/37)** — rather than reimplement the running/token/generation checks for bots, the enclave's locals were lifted into one module both drivers import. This is also what §2.6 rule 1 asks for structurally: a shared sealed vocabulary, not an enclave-owned one re-prefixed per consumer.

**4. No presence touch on sealed steps** — unlike plaintext `/steps`, the sealed handlers don't refresh presence; that would require `instanceId` in the body, which the header-auth model deliberately omits. The bot's `/renew` loop owns claim liveness, exactly as the enclave's heartbeat does. Mirrors the enclave (its step callbacks touch no presence).

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/features/agents/sealed-session-guards.ts` | The three sealed session-callback guards, shared by enclave + bots |
| `apps/backend/src/features/public-api/sealed-steps.test.ts` | Unit coverage for both new handlers |

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/constants.ts` | Add `THREA_CALLBACK_TOKEN_HEADER` |
| `packages/types/src/index.ts` | Export the new constant |
| `apps/backend/src/features/agents/index.ts` | Barrel-export the three guards |
| `apps/backend/src/features/enclave-runtimes/session-handlers.ts` | Delegate `assertCallbackBound` to `verifyCallbackToken`; alias-import `assertRunning`/`assertReplyGeneration` from the shared module; drop the local definitions + the now-unused `timingSafeEqual`/`hashCallbackToken` imports |
| `apps/backend/src/features/public-api/handlers.ts` | Add `startBotInvocationSealedStep`, `recordBotInvocationSealedStep`, and the `authorizeSealedStepCallback` / `emitBotSealedProgress` closures |
| `apps/backend/src/features/public-api/schemas.ts` | Add `recordSealedInvocationStepSchema` + `startSealedInvocationStepSchema` |
| `apps/backend/src/features/public-api/routes.ts` | Route descriptors for both endpoints + the callback-token header param; widen the `parameters` type to allow `in: "header"` |
| `apps/backend/src/routes.ts` | Wire the two express routes |
| `docs/public-api/openapi.json` | Regenerated spec |

## Out of scope (deliberate)

- **Sealed `/complete`** (Phase 3): accept `SealedReply` → `eventService.createMessage` with ciphertext/envelope/`e2eVersion:2`. The plaintext `completeBotInvocation` keeps its `assertNotE2eStream` guard (E2EE-2) — that gate is correct for the plaintext sink and stays until the sealed completion lands.
- **Consent UX** (Phase 4), **Pi client sealed harness** (Phase 5), **flipping `externalSealedDelivery`** (Phase 6).
- **No DB-backed e2e for the sealed path** — the policy const is hardcoded off and the HTTP e2e harness has no crypto-seeding helpers, so the sealed verdict can't be produced end-to-end yet (same constraint as #981). Coverage is unit-level; DB-backed e2e waits until the flag is injectable (Phase 6).

## Test plan

- [x] `bun test src/features/public-api/sealed-steps.test.ts` — 7 pass (sealed persistence, dropped-start fallback, all four rejection codes: missing/mismatched token, non-running session, wrong key generation)
- [x] `bun test src/features/public-api src/features/bot-runtimes` — 127 pass (no regression)
- [x] `bun test src/features/agents src/features/enclave-runtimes` — 426 pass (guards refactor preserves enclave behavior)
- [x] `bun run typecheck` — clean across all packages incl. `tsconfig.tests.json`
- [x] `eslint` clean on all changed backend files
- [x] `bun run generate:api-docs` — OpenAPI regenerated, no drift
- [x] Pre-PR review (2 reviewer agents — correctness/data-flow + spec/§2.6/invariant): **0 findings**. Confirmed: aliased `asserts` import preserves narrowing at all enclave call sites; `SessionStatuses` still imported where used; workspace/persona isolation chain sound; INV-E7 (ciphertext-only) and INV-6/20 (atomic step insert) hold.
- [ ] No live-DB harness in this environment, so the step-insert SQL is exercised via stubbed repos like its siblings — not integration-tested here (runs in CI).

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01ADPwawiQ84QhHS8o7DoRB4)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/983"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784273371&installation_id=129946197&pr_number=983&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F983&signature=3184cdba470df76683187de5c952cefdf21a8b92d505fbc445d8dfa346f47655"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/index.ts
+++ b/apps/backend/src/features/agents/index.ts
@@ -113,6 +113,7 @@ export { AgentConfigOverrideRepository } from "./agent-config-override-repositor
 export type { AgentConfigOverride } from "./agent-config-override-repository"
 
 export { hashCallbackToken } from "./callback-token"
+export { assertSessionRunning, verifyCallbackToken, assertReplyKeyGeneration } from "./sealed-session-guards"
 
 export { AgentSessionRepository, SessionStatuses } from "./session-repository"
 export type {
--- a/apps/backend/src/features/agents/sealed-session-guards.ts
+++ b/apps/backend/src/features/agents/sealed-session-guards.ts
@@ -1,0 +1,64 @@
+import { timingSafeEqual } from "node:crypto"
+import { HttpError } from "@threa/backend-common"
+import { hashCallbackToken } from "./callback-token"
+import { SessionStatuses, type AgentSession } from "./session-repository"
+
+/**
+ * The guards every sealed session callback must pass, shared by every
+ * sealed-capable driver — the enclave's session callbacks today and an
+ * owner-granted external bot harness's sealed `/steps`/`/complete` next (§2.6).
+ * They live in the neutral agents feature because each one is a property of the
+ * `agent_sessions` row, not of any one transport: only the header the token
+ * rides in differs between drivers, so each call site reads its own header and
+ * hands the value here.
+ */
+
+export function assertSessionRunning(session: AgentSession | null): asserts session is AgentSession {
+  if (!session) throw new HttpError("Session not found", { status: 404, code: "SESSION_NOT_FOUND" })
+  if (session.status !== SessionStatuses.RUNNING) {
+    throw new HttpError("Session is not running", { status: 409, code: "SESSION_NOT_RUNNING" })
+  }
+}
+
+/**
+ * Bind a sealed callback to the runner the session was assigned to (Phase 2.4b,
+ * E2EE-21). The cleartext token was minted at claim and delivered only inside
+ * the sealed assignment/claim to the winning instance — possession proves the
+ * caller is that runner, a stronger binding than a self-reported instance id
+ * that any workspace-key holder could copy. The row holds only the sha256
+ * digest, so the presented value is hashed before the timing-safe compare (both
+ * sides are fixed-length hex). The token is mandatory: an absent token — like a
+ * NULL row hash, which no token can match — is rejected outright.
+ */
+export function verifyCallbackToken(
+  session: Pick<AgentSession, "callbackTokenHash">,
+  presented: string | undefined
+): void {
+  if (!presented) {
+    throw new HttpError("Missing callback token", { status: 403, code: "CALLBACK_TOKEN_MISSING" })
+  }
+  const expected = session.callbackTokenHash
+  const presentedHash = Buffer.from(hashCallbackToken(presented))
+  if (!expected || !timingSafeEqual(presentedHash, Buffer.from(expected))) {
+    throw new HttpError("Callback token mismatch", { status: 403, code: "CALLBACK_TOKEN_MISMATCH" })
+  }
+}
+
+/**
+ * A seal under any generation other than the one the assignment/claim prescribed
+ * would persist as a permanently undecryptable row — reject it loudly so the turn
+ * fails visibly instead (E2EE-21). Sessions dispatched before the column shipped
+ * (NULL) are exempt.
+ */
+export function assertReplyKeyGeneration(
+  session: Pick<AgentSession, "replyKeyGeneration">,
+  envelope: { keyGeneration: number } | undefined
+): void {
+  if (!envelope || session.replyKeyGeneration == null) return
+  if (envelope.keyGeneration !== session.replyKeyGeneration) {
+    throw new HttpError(
+      `Sealed payload uses key generation ${envelope.keyGeneration}; this session seals under ${session.replyKeyGeneration}`,
+      { status: 400, code: "E2E_WRONG_KEY_GENERATION" }
+    )
+  }
+}
--- a/apps/backend/src/features/agents/session-repository.ts
+++ b/apps/backend/src/features/agents/session-repository.ts
@@ -739,6 +739,13 @@ export const AgentSessionRepository = {
       messageId?: string
       completedAt?: Date
       /**
+       * Scope the update to one session so a caller-controlled `stepId` can only
+       * touch a step of the session it owns. The public bot sealed-step callback
+       * passes the authorized session id (the step id is fully caller-supplied
+       * there); omitted on the trusted internal enclave path, where it no-ops.
+       */
+      sessionId?: string
+      /**
        * Guard against overwriting a finalized step. A mid-run substep snapshot can
        * race a finalize (`/steps`): network reordering or a retry can land the
        * snapshot after completion, clobbering the final content with a partial one.
@@ -759,6 +766,7 @@ export const AgentSessionRepository = {
           message_id = COALESCE(${params.messageId ?? null}, message_id),
           completed_at = COALESCE(${params.completedAt ?? null}, completed_at)
         WHERE id = ${stepId}
+          AND (${params.sessionId ?? null}::text IS NULL OR session_id = ${params.sessionId ?? null})
           ${sql.raw(params.requireRunning ? "AND completed_at IS NULL" : "")}
         RETURNING ${sql.raw(STEP_SELECT_FIELDS)}
       `
--- a/apps/backend/src/features/enclave-runtimes/session-handlers.ts
+++ b/apps/backend/src/features/enclave-runtimes/session-handlers.ts
@@ -2,7 +2,6 @@ import { z } from "zod"
 import type { Request, Response } from "express"
 import type { Pool } from "pg"
 import type { Server } from "socket.io"
-import { timingSafeEqual } from "node:crypto"
 import {
   AGENT_STEP_TYPES,
   AuthorTypes,
@@ -26,7 +25,9 @@ import {
   SessionStatuses,
   ARIADNE_AGENT_ID,
   getBuiltInAgentConfig,
-  hashCallbackToken,
+  assertSessionRunning as assertRunning,
+  verifyCallbackToken,
+  assertReplyKeyGeneration as assertReplyGeneration,
   failSessionWithLifecycle,
   type AgentSession,
 } from "../agents"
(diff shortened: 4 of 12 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Code | Reviewers: sonnet x2 (spec+design, correctness+data-flow)

Backend-only diff, so UX and Mobile lenses were not run.

Found 1 issue — **fixed in this review pass** (commit `4735632`):

1. `apps/backend/src/features/public-api/handlers.ts:1352` — `recordBotInvocationSealedStep` passed a caller-supplied `data.stepId` to `AgentSessionRepository.updateStep`, which filtered on `id` alone (no `session_id` scope). On this **public** bot endpoint the step id is fully caller-controlled, so an authenticated bot holding a valid callback token for its own session could finalize/overwrite a step row belonging to a different session. The enclave's `/steps` handler has the same unscoped pattern, but it runs under the trusted internal credential — the public bot surface amplifies a latent gap into a real cross-session data-integrity write (`bug` / `data-flow`, 82).
   https://github.com/threahq/threa/blob/4735632308c6e663df0559d71150940d1a96a0eb/apps/backend/src/features/public-api/handlers.ts#L1351-L1361
   ✅ Fixed: added an optional, null-guarded `sessionId` scope to `updateStep` (trusted enclave callers omit it, so their behavior is unchanged) and the bot finalize now passes the authorized `session.id`. Test updated to assert the scoping.
   https://github.com/threahq/threa/blob/4735632308c6e663df0559d71150940d1a96a0eb/apps/backend/src/features/agents/session-repository.ts#L769

---

Matches §2.6 intent: neutral `X-Threa-Callback-Token` header (not enclave-named, rule 1), shared sealed-session guards reused by both drivers, sealed `/steps` + `/steps/started` mirroring the enclave, dark behind `externalSealedDelivery`. Schema → route → express wiring → regenerated OpenAPI all present and consistent.
</details>
<details><summary>🔍 Bugs [1 issue — fixed]</summary>

The unscoped `updateStep` finalize (above), fixed in `4735632`. The alias-imported `asserts` guards preserve narrowing at every enclave call site; `SessionStatuses` remains imported where the enclave still uses it; the dropped-start fallback's `started_at`/duration math and progress emit match the enclave; `assertReplyKeyGeneration` is invoked correctly in both handlers (the start envelope is legitimately optional for tool steps and short-circuits).
</details>
<details><summary>🔁 Data Flow [CLEAN post-fix]</summary>

Step insert is race-safe (atomic `MAX(step_number)+1` + `ON CONFLICT DO NOTHING`, INV-20); insert + `updateCurrentStepType` share one transaction (INV-6). The session callback binding (`callback_token_hash` / `reply_key_generation`) set at claim is verified on every step. With the scope fix, a finalize can only touch a step of the authorized session.
</details>

No new invariant violations. HttpErrors carry status+code (INV-32/11); bodies are Zod-validated (INV-55); guards reused rather than duplicated (INV-35/37); server persists/relays ciphertext only, never decrypts (INV-E7).
</details>

The three guards were centralized to avoid a parallel implementation; the enclave delegates via thin aliased imports. No config sprawl, no N+1, no leaky abstraction.
</details>
<details><summary>🔒 Security [CLEAN post-fix]</summary>

The callback-token compare is timing-safe and rejects a NULL row hash (every plaintext session). Workspace/persona isolation (`stream.workspaceId !== req.workspaceId`, `session.personaId !== bot.id`) holds. The one cross-session write vector (above) is closed by the `sessionId` scope.
</details>
<details><summary>🎛️ UX [N/A]</summary>

No frontend changes.
</details>
<details><summary>📱 Mobile [N/A]</summary>

No frontend changes.
</details>

PR #987, opened 2026-06-17
refactor(frontend/crypto): unify client decrypt decode entry (slice 2)

kristofferremback:
## Summary

Slice 2 of the client decrypt-layer unification ([design doc](https://github.com/threahq/threa/blob/main/docs/plans/client-decrypt-layer-unification.md)). Slice 1 (PR #980) extracted the shared cache primitive; this slice extracts the **unified decode entry** that the read surfaces over encrypted content each hand-rolled.

Three read paths (`useDecryptedMessageContent`, `useDecryptedStepContent`, `useStreamSearch`) independently called `useE2eSession`, resolved `rootStreamId` from the stream row, and copied the same cache-poisoning guard: *a thread shares its root scratchpad's SSK and carries no wraps of its own, so a decrypt against the bare (thread) id finds no wrap, fails, and — because the id is immutable — caches that failure permanently.* Three copies of one footgun.

This collapses them to one `resolveDecryptContext`.

## What changed

- **New `apps/frontend/src/lib/crypto/decrypt-context.ts`:**
  - `resolveDecryptContext(workspaceId, streamId, session, streamRow)` → `{ ready: false, reason: "locked" | "unhydrated" }` or `{ ready: true, opts }`. Owns the session-unlocked check, root-SSK resolution (`streamRow.rootStreamId ?? undefined`), and the hold-until-the-row-hydrates guard.
  - `isSessionUnlocked(session)` — a type guard narrowing `privateKey`/`keyId` non-null; the single unlocked predicate.
- **`useDecryptedMessageContent` / `useDecryptedStepContent`:** drop their inline `sessionUnlocked` + `streamHydrated` + root resolution and route through `resolveDecryptContext`. Effect deps stay primitive (opts fields, not the per-render opts object), so a settled decrypt doesn't re-fire on an unrelated stream-row update.
- **`useStreamSearch`:** its decrypt callback now resolves context through the same gate (session + stream row snapshotted into refs). This *adds* the hold-until-hydrated guard to search, which previously lacked it.
- **`lib/drafts/decryption.ts`:** `isE2eUnlocked` now delegates to `isSessionUnlocked` — the 4th copy of the session check folds into the shared predicate.

## Design decision: why drafts don't route through the row→root step

Drafts share the session predicate but **not** `resolveDecryptContext`'s row→root resolution. Their root (`e2eStreamId`) is already resolved at the composer (`stream?.rootStreamId ?? streamId`) before reaching the draft layer, which then operates on a pre-resolved root in batch with no per-draft stream row. So drafts carry no row→root footgun to share — forcing them through the row step would mean re-deriving a root they already hold. Recorded in the design doc.

## Behavior preservation & tests

Behavior-preserving for the message/step paths; the existing suites are the guard. All affected tests pass (`use-decrypted-message-content`, `use-decrypted-step-content`, `use-stream-search`, `use-draft-message`, draft content/previews, crypto).

- New `__tests__/decrypt-context.test.ts` covers the predicate and the locked/unhydrated/ready resolution.
- One test-artifact change: the `useStreamSearch` integration test now stubs a hydrated stream row, since the hold-until-hydrated guard now applies to search too (it previously decrypted against an unhydrated row).

Pre-commit ran the full monorepo lint + typecheck clean.

## Reviewer notes

- Search gains the hydration guard it lacked — in production search runs on the open (hydrated) stream, so no user-visible change, but it now skips a sealed row whose stream row hasn't hydrated rather than risk a doomed thread-id decrypt.
- Next: slice 3 (attachments onto the layer + the uniform `{ value, status }` read shape).

---
_Generated by [Claude Code](https://claude.ai/code/session_01FowoQvmZxVq4HUZcqFJhs1)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/987"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784286551&installation_id=129946197&pr_number=987&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F987&signature=382f45d0a82556fb55731d0b8d9a86725e7f714460922925d4ee80183ac8a601"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/hooks/use-decrypted-message-content.ts
+++ b/apps/frontend/src/hooks/use-decrypted-message-content.ts
@@ -6,6 +6,7 @@ import {
   subscribeToDecryption,
   type DecryptCacheEntry,
 } from "@/lib/crypto/decrypt-cache"
+import { resolveDecryptContext } from "@/lib/crypto/decrypt-context"
 import { useE2eSession } from "@/stores/e2e-session-store"
 import { useStreamFromStore } from "@/stores/stream-store"
 import type { AttachmentRef } from "@/lib/crypto/attachment-crypto"
@@ -70,17 +71,12 @@ export function useDecryptedMessageContent(
 ): DecryptedMessageContent {
   const session = useE2eSession(workspaceId, userId ?? "")
   const eventId = event.id
-  // A thread shares its root scratchpad's SSK and carries no wraps of its own,
-  // so the key must be resolved against the root (the wrap AAD is bound to the
-  // root id). `undefined` here is ambiguous: a top-level stream (its own root,
-  // resolve against self) reads the same as a thread whose row hasn't hydrated
-  // yet (root unknown). Resolving the latter against the thread id finds no
-  // wrap and fails — and the failure is cached permanently. So gate the decrypt
-  // on the stream row being present: until then we don't know the root and hold
-  // at `pending` rather than poison the cache with a doomed attempt.
+  // The session + root-SSK resolution and the "hold until the stream row hydrates"
+  // guard (a doomed thread-id decrypt poisons the cache forever) live in
+  // `resolveDecryptContext` — see it for why an unhydrated row must hold.
   const stream = useStreamFromStore(event.streamId)
-  const streamHydrated = stream !== undefined
-  const rootStreamId = stream?.rootStreamId ?? undefined
+  const ctx = resolveDecryptContext(workspaceId, event.streamId, session, stream)
+  const opts = ctx.ready ? ctx.opts : null
 
   const cached = useSyncExternalStore<DecryptCacheEntry | undefined>(
     (listener) => subscribeToDecryption(eventId, listener),
@@ -91,12 +87,10 @@ export function useDecryptedMessageContent(
   // Memoize so useEffect deps don't churn on every render — a fresh wrapper
   // object every render would re-fire the effect even though nothing changed.
   const envelopePayload = useMemo(() => readEnvelopePayload(event), [event])
-  const sessionUnlocked = session.status === "unlocked" && !!session.privateKey && !!session.keyId
-  const canDecrypt = !!envelopePayload && sessionUnlocked && streamHydrated
-  const needsDecrypt = canDecrypt && (cached === undefined || cached.status === "pending")
+  const needsDecrypt = !!envelopePayload && opts !== null && (cached === undefined || cached.status === "pending")
 
   useEffect(() => {
-    if (!needsDecrypt || !envelopePayload || !session.privateKey || !session.keyId) return
+    if (!needsDecrypt || !envelopePayload || !opts) return
     void requestDecryption(
       eventId,
       {
@@ -104,23 +98,19 @@ export function useDecryptedMessageContent(
         envelope: envelopePayload.envelope,
         ciphertext: envelopePayload.ciphertext,
       },
-      {
-        privateKey: session.privateKey,
-        recipientKeyId: session.keyId,
-        workspaceId,
-        streamId: event.streamId,
-        rootStreamId,
-      }
+      opts
     )
+    // Depend on the opts fields (primitives) rather than the per-render opts object
+    // so an unrelated stream-row update doesn't re-fire a settled decrypt.
   }, [
     needsDecrypt,
     envelopePayload,
-    session.privateKey,
-    session.keyId,
+    opts?.privateKey,
+    opts?.recipientKeyId,
+    opts?.workspaceId,
+    opts?.streamId,
+    opts?.rootStreamId,
     eventId,
-    workspaceId,
-    event.streamId,
-    rootStreamId,
   ])
 
   if (!envelopePayload) {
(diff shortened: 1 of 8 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-8[1m] | Reviewers: sonnet x3 (spec+design, correctness+data-flow, UX)

_Mobile lens skipped: the diff is pure decode-layer hooks/crypto logic — no components, JSX, CSS, layout, touch, or viewport code._

Found 1 issue:

1. `apps/frontend/src/lib/crypto/decrypt-context.ts:25-32` — `UnlockedSession` is `export`ed but has zero importers outside its own file; it serves only as the `session is UnlockedSession` annotation on `isSessionUnlocked`'s type guard, and TypeScript propagates that narrowing to call sites without the name being visible externally. (claude-md [INV-36, *Maintainability and Scope Control*]: "Do not add speculative features, configuration, or comments for imagined requirements.") The `export` costs nothing to add when a caller actually needs the narrowed type — drop it to a local `interface` until then.
   https://github.com/threahq/threa/blob/38945e97191a303d9f5038de597a384483e3d7b9/apps/frontend/src/lib/crypto/decrypt-context.ts#L25-L32

---

Matches the slice-2 spec in `docs/plans/client-decrypt-layer-unification.md`: `resolveDecryptContext` owns session + root-SSK resolution + the hold-until-hydrated guard; `useDecryptedMessageContent`, `useDecryptedStepContent`, and `useStreamSearch` route through it; `isE2eUnlocked` delegates to the shared `isSessionUnlocked` (the 4th copy of the session check). The draft decision (keeps its composer-pre-resolved root, no row→root step) is consistent with the doc and verified against `lib/drafts/decryption.ts`. The behavior-preserving claim holds for the message/step paths (the locked/unhydrated/pending/decrypted/failed mapping is equivalent to the old inline logic).
</details>

Effect-dep change (hooks now depend on `opts?.privateKey, opts?.recipientKeyId, …` instead of the per-render `opts` object) preserves the old behavior: each listed field is the same value/reference the old deps carried (`opts?.privateKey` is `session.privateKey`, a stable `CryptoKey` ref; the rest are strings), so a settled decrypt doesn't re-fire and lock→unlock still re-fires. The redundant `!envelopePayload || !opts` checks in the effect body are harmless TS narrowing. Search `streamRowRef`/`event.streamId` are consistent (search is single-stream-scoped, so `event.streamId === streamId`).
</details>
<details><summary>🔁 Data Flow [CLEAN — 1 considered & dismissed]</summary>

**Considered:** the new hold-until-hydrated guard in `useStreamSearch` (absent before) means a sealed event whose stream row is `undefined` now resolves to `null` (skipped) rather than being decrypted. In theory, during the IDB-hydration window a user could search before the row lands and get incomplete E2E results.

**Dismissed (below threshold):** (a) it is the *correct, intentional* trade-off the PR documents — and the proposed "decrypt anyway for top-level streams" fix is **unsound**: you cannot distinguish a top-level stream from a thread *without* the row, so decrypting against a bare id is exactly the footgun, and since `requestDecryption` warms the shared cache the timeline reads, a doomed thread-id decrypt would poison timeline rendering, not just search; (b) practically unreachable — the open stream's row is already hydrated by the time a user opens search, types a query, and triggers it, because the timeline mounted in the same view depends on that very row. Skipping a not-yet-readable row is strictly safer than the old doomed-decrypt path.
</details>
<details><summary>📋 CLAUDE.md Compliance [1 — see headline]</summary>

The one INV-36 nit above. No INV-35/37 issues (this IS the reuse fix), no INV-9/11 (the helper is pure, no module state/singletons), comment discipline is good (the hooks redirect to the canonical `resolveDecryptContext` rather than narrating the change, per INV-25).
</details>

`resolveDecryptContext` serves the three row-based callers without distortion; drafts deliberately don't route through the row→root step because their root is composer-pre-resolved (no row→root footgun to share) — a sound boundary, not a partial abstraction. No re-render storms introduced.
</details>

E2EE decode path: `resolveDecryptContext` returns `ready: true` only for an unlocked session with both keys present, and the lock-epoch write-back guard lives in the cache primitive (unchanged here), so no path was opened for plaintext to be cached past lock or for a wrong-key decrypt.
</details>

The user-perceivable decrypt states (locked / pending / decrypted / failed) surface in the same situations as before; unlocked-but-unhydrated correctly shows pending, not a stuck/failed state. The search behavior change is safer than before and transiently/practically unreachable for an open stream.
</details>

---

PR #992, opened 2026-06-17
refac​tor(frontend): attachments onto the decrypt layer + uniform { status, value } cache shape (decrypt-layer slice 3)

kristofferremback:
**Design doc:** [`docs/plans/client-decrypt-layer-unification.md`](../blob/claude/decrypt-layer-slice-3-m18r18/docs/plans/client-decrypt-layer-unification.md) — slice 3 (final slice). Follows slice 0 (#955), slice 1, slice 2 (#987).

## Problem

Every client-side encrypted field decrypts through the same skeleton — an in-memory cache keyed by a ciphertext id, inflight-dedup, a lock-epoch guard, a subscribe signal, and a session + key gate — but it was re-implemented per field. Slices 1–2 collapsed the cache primitive (`createDecryptedCache`) and the session/hydration gate (`resolveDecryptContext`). Two divergences remained:

1. **Attachments had no read cache at all.** `e2e-attachment-list.tsx` fetched the opaque S3 ciphertext and decrypted it **on every mount** — so an image re-decrypted each time it scrolled back into view. It was the one surface that most needed a shared cache lifecycle and the only one with none.
2. **The cache-entry shape was not uniform.** `stream-name-cache` already used `{ status, value }`; `decrypt-cache` (message bodies + trace steps + drafts + search) used `{ status, content }`. The rename to unify on `{ status, value }` was deliberately deferred from slices 1–2 to here.

## Solution

Give attachment bytes a real cache instance over the slice-1 primitive, a read hook that mirrors `useDecryptedMessageContent`, and unify the last cache-entry field on `{ status, value }`. No new cache machinery, no new session wiring — the layer already existed; this is the final field moving onto it.

### How it works

1. **`lib/crypto/attachment-cache.ts`** — a per-key `createDecryptedCache<{ status, value: Blob | null }>` instance. `subscription: "per-key"` (one resolve must not re-render every attachment), `retryFailed: true` (a fetch/decrypt miss is a transient network blip, not terminal like a message-id decrypt), `lru: 64` (blobs are whole files — cap memory; the goal is making a recently-viewed image re-render free, not retaining everything scrolled past). It auto-registers its `clear` via the slice-1 registry, so `clearAllDecrypted()` on lock drops attachment plaintext through the single existing call site — no new clear wiring. `requestAttachmentBytes(id, fetchDecrypt)` records a thrown fetch/decrypt as a `failed` entry the UI can react to.

2. **`hooks/use-decrypted-attachment.ts`** — returns a `pending | ready | failed` union. The decrypted `Blob` lives in the shared cache; the object URL is created and revoked **in the hook** (effect keyed on the blob identity), never cached — a cached URL would pin its blob alive past lock. On lock the cache emits, `cached` flips off `decrypted`, the effect cleanup revokes the URL. `fetchAndDecryptAttachment` (the network fetch + in-memory decrypt with the per-file key/iv) moved here from the component.

3. **`e2e-attachment-list.tsx`** — `E2eImageAttachment` reads through the hook instead of its own `useState(url/failed)` + per-mount `useEffect` fetch. `E2eFileAttachment`'s download routes through `requestAttachmentBytes` too, so a previewed image that's then downloaded reuses the cached blob (single fetch, deduped).

4. **Rename `DecryptCacheEntry.content` → `value`** — message bodies, stream names, and attachment bytes now all carry `{ status, value }`. The step/search/draft readers move through the rename; the step hook's own public `content` return field is untouched (only the cache-entry accessor changed). `tsc --noEmit` is the completeness guard — a missed accessor wouldn't compile.

5. **`apps/frontend/AGENTS.md`** — the "how to add an encrypted field" recipe (cache instance → `resolveDecryptContext` gate → per-domain `{ status }` hook), so the next field reuses the layer instead of copying the nearest example.

### Key design decisions

**1. Cache the `Blob`, not the object URL.** A cached object URL leaks: eviction or lock would need to find and revoke every URL. Caching bytes lets the hook own URL lifecycle (React-managed, revoked on unmount/blob-change) while the cache stays a plain GC-friendly byte store. The expensive part — network fetch + decrypt — is what's cached.

**2. `retryFailed: true` for attachments** (vs. `false` for message bodies). A message id's decrypt failure is terminal (wrong recipient / tampered AAD never starts succeeding). An attachment fetch fails on transient network blips, so a later request — the download-chip retry the image fallback offers — must re-attempt. Same reasoning the name cache uses.

**3. Keep per-domain hooks, share the helpers (resolves doc Q3).** A single generic `useDecryptedField` would flatten the typed returns — a message carries `attachmentRefs`/`sources`, an attachment a `url`, a step `TraceSource[]`. The uniformity that mattered is the `{ status, value }` **cache-entry** shape, shared at the cache layer; each hook keeps its own typed discriminated union.

**4. Steps keep sharing `messageCache` keyed by `step.id` (resolves doc Q2).** Same entry shape, same LRU pressure profile; a separate instance would duplicate for no behavior difference.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/lib/crypto/attachment-cache.ts` | Per-key `createDecryptedCache` instance for decrypted attachment bytes |
| `apps/frontend/src/hooks/use-decrypted-attachment.ts` | Read hook (pending/ready/failed) + `fetchAndDecryptAttachment`; owns object-URL lifecycle |
| `apps/frontend/src/lib/crypto/__tests__/attachment-cache.test.ts` | Cache: caching, dedup, transient retry, clear |
| `apps/frontend/src/hooks/use-decrypted-attachment.test.ts` | Hook: pending/ready/failed + object-URL mapping |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/e2e-attachment-list.tsx` | Image reads via the hook (no per-mount re-decrypt); download routes through the shared cache |
| `apps/frontend/src/lib/crypto/decrypt-cache.ts` | `DecryptCacheEntry.content` → `value` |
| `apps/frontend/src/hooks/use-decrypted-message-content.ts` | Read `cached.value` (was `cached.content`) |
| `apps/frontend/src/hooks/use-decrypted-step-content.ts` | Read `cached.value`; public `content` return field unchanged |
| `apps/frontend/src/hooks/use-stream-search.ts` | Read `entry.value` |
| `apps/frontend/src/lib/drafts/decryption.ts` | Read `cached.value` / `entry.value` (4 sites) |
| `apps/frontend/AGENTS.md` | "How to add an encrypted field" recipe |
| `docs/plans/client-decrypt-layer-unification.md` | Slice 3 marked DONE; Q2/Q3 resolved |
| `*.test.ts(x)` (5 files) | Cache-entry mocks/accessors `content` → `value` |

## Out of scope (deliberate)

- **The download path's `setTimeout(revoke, 10_000)`** (`E2eFileAttachment.handleDownload`) is unchanged from before slice 3 — it can retain a downloaded file's object URL up to ~10s past unmount and isn't revoked on lock. Flagged in self-review; left as-is because (a) it's pre-existing, not introduced here, and (b) it covers a file the user explicitly downloaded to disk, where the plaintext is already persisted by user action. A focused hygiene follow-up (clear the timer on unmount) can take it if we want the window gone.
- **E2E-draft attachments / context refs / stash** — sealed-draft *bodies* roam (#946); these stay session-local pending their own design (per the doc's "Out of scope").
- **The `DraftScratchpad` vs `CachedDraft` "draft" naming collision** — a separate non-decrypt-layer rename follow-up the doc tracks.

## Test plan

- [x] `bunx vitest run` over the 17 affected suites (attachment cache + hook, decrypt-cache, decrypt-context, message/step/search content, all draft hooks, trace-step, event-item, message-event, stream-sync, sync-engine) — **290 pass**
- [x] `bun run typecheck` (`tsc --noEmit` across the monorepo incl. astro `public-site`) — clean; this is the rename's completeness guard
- [x] `bunx eslint` + `prettier --check` on all changed files — clean (also enforced by the pre-commit hook on both commits)
- [x] Pre-PR self-review (1 focused correctness agent over the new attachment code): 1 finding fixed (effect depended on the per-render `ref` object; aligned to primitive-field deps per the frontend convention — see second commit), 1 noted as pre-existing/out-of-scope (the download `setTimeout` revoke window above). Cache-clear registration, hook URL revoke on unmount/blob-change, `retryFailed` wiring, no-loop effect guard, no double-fetch, and INV-18 all verified correct.
- [ ] No manual browser pass — behavior-preserving for existing surfaces (the existing message/step/search/draft/sync suites are the regression guard); the only behavior *change* is attachments no longer re-decrypting per mount, covered by the cache + hook tests.

<details>

**Goal.** Final slice of the client decrypt-layer unification: move E2E attachment bytes onto the shared decrypt layer and unify every cache entry on `{ status, value }`.

**What was built.**
1. `attachment-cache.ts` — per-key `createDecryptedCache` instance (retryFailed, lru 64), auto-registered for lock-clear.
2. `use-decrypted-attachment.ts` — pending/ready/failed hook mirroring `useDecryptedMessageContent`; owns object-URL lifecycle so bytes never leak past lock; carries `fetchAndDecryptAttachment`.
3. `e2e-attachment-list.tsx` — image preview reads the hook (no per-mount re-decrypt); file download warms the same cache.
4. `DecryptCacheEntry.content` → `value` rename across message/step/search/draft readers + their tests.
5. AGENTS.md recipe; doc slice-3 DONE + Q2/Q3 resolved.

**Design decisions.** Cache the Blob not the URL (leak-free eviction); `retryFailed: true` (transient fetch failures); keep per-domain hooks for typed returns (Q3); steps keep sharing `messageCache` (Q2).

**Design evolution (self-review).** First version of the hook depended on the per-render `ref` object in the request effect; a future caller constructing refs inline could churn it. Fixed to depend on the primitive fields (`attachmentId`, `workspaceId`, `status`), matching `useDecryptedMessageContent`.

**Schema changes.** None (frontend-only; no IDB shape change — ciphertext stays at rest, plaintext stays memory-only).

**What's NOT included.** Draft attachments/context refs/stash; the `DraftScratchpad`/`CachedDraft` rename; the pre-existing download `setTimeout` revoke window.

**Status.** Complete — slice 3 closes the workstream (slices 0–3 all DONE).

</details>

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01SZw7DVoEjgCgp1m7ZwHpPQ)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/992"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784305608&installation_id=129946197&pr_number=992&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F992&signature=cc6c4af748e2c416257faa2e6a1d9d030611acf04e43de43510e084921a8e515"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/AGENTS.md
+++ b/apps/frontend/AGENTS.md
@@ -107,3 +107,48 @@ always written the sealed columns only.
   searchable only client-side over the decrypted copy. Streams renamed _before_
   this shipped still hold plaintext in `display_name`; the server can't re-seal
   them, so they stay until the owner renames again while unlocked.
+
+## How to add an encrypted field (the decrypt layer)
+
+Every client-side encrypted field (message bodies, trace steps, stream names,
+drafts, attachment bytes) decrypts through one shared layer under
+`lib/crypto/`. Ciphertext stays at rest in IDB; plaintext is memory-only and
+cleared on lock (E2EE-4). Don't hand-roll a new cache, session gate, or
+hydration guard — reuse the three pieces below, and a new field needs no new
+session wiring and no copied footgun.
+
+1. **A cache instance, not a new cache.** Call
+   `createDecryptedCache<{ status, value }>(...)` from `lib/crypto/decrypted-cache.ts`.
+   Every entry is `{ status: "pending" | "decrypted" | "failed"; value: V | null }` —
+   keep that uniform shape (`value`, not a field named after the domain). Pick:
+   - `subscription`: `"per-key"` for many keys where one resolve must not re-render
+     the rest (messages, attachments); `"global"` for few keys read as a list (names).
+   - `retryFailed`: `false` when a miss is terminal (a message id's decrypt can't
+     start succeeding); `true` when a null open is transient (locked / not-yet-a-
+     recipient / a network blip — names, attachment bytes).
+   - `lru`: cap when entries are unbounded in count or large in size (messages 500,
+     attachment blobs 64); omit for a small bounded set (names).
+
+   The instance auto-registers its `clear` in the lock-clear registry, so
+   `clearAllDecrypted()` (the single lock / account-switch site in
+   `e2e-session-store`) wipes it — you do **not** add a clear call. A cache can only
+   hold plaintext if its module is loaded, and loading registers it.
+
+2. **Gate the decrypt through `resolveDecryptContext`** (`lib/crypto/decrypt-context.ts`)
+   for any field keyed under a stream's SSK. It owns the unlocked check
+   (`isSessionUnlocked`), root-SSK resolution, and the hold-until-the-stream-row-
+   hydrates guard — decrypting a thread's content against its own id finds no wrap,
+   fails, and caches that failure **forever**, so never decrypt against a bare
+   `streamId` when the row is unhydrated. Read the row with `useStreamFromStore` and
+   pass it in; `{ ready: false }` means hold at `locked`/`pending`, never attempt.
+
+3. **A per-domain read hook returning a `{ status }` discriminated union**
+   (`plaintext | locked | pending | decrypted | failed`). Subscribe to the cache
+   with `useSyncExternalStore` (per-key) and fire `request(...)` from an effect
+   whose deps are the **primitive opts fields** (`opts?.privateKey`, …,
+   `opts?.rootStreamId`), not the per-render `opts` object — a fresh object every
+   render would re-fire settled decrypts on unrelated row updates. Keep hooks
+   per-domain (typed returns) rather than one generic hook; they share the helpers
+   above, not the hook. `useDecryptedMessageContent` is the reference shape;
+   `useDecryptedAttachment` mirrors it (and owns the object-URL lifecycle so the
+   blob in the cache never leaks past unmount/lock).
--- a/apps/frontend/src/components/timeline/e2e-attachment-list.tsx
+++ b/apps/frontend/src/components/timeline/e2e-attachment-list.tsx
@@ -1,8 +1,9 @@
-import { useEffect, useState } from "react"
+import { useState } from "react"
 import { toast } from "sonner"
 import { Download, FileLock2, Loader2 } from "lucide-react"
-import { attachmentsApi } from "@/api"
-import { decryptAttachmentBytes, type AttachmentRef } from "@/lib/crypto/attachment-crypto"
+import { type AttachmentRef } from "@/lib/crypto/attachment-crypto"
+import { requestAttachmentBytes } from "@/lib/crypto/attachment-cache"
+import { fetchAndDecryptAttachment, useDecryptedAttachment } from "@/hooks/use-decrypted-attachment"
 import { triggerDownload } from "@/lib/image-utils"
 import { Button } from "@/components/ui/button"
 
@@ -12,20 +13,12 @@ import { Button } from "@/components/ui/button"
  * ciphertext with no thumbnails, variants, or real metadata, so we fetch the
  * raw ciphertext, decrypt it in-memory with the per-file key/iv from the ref,
  * and render from a blob URL. Images preview inline; everything else is a
- * decrypt-on-click download. Real filename/size come from the ref, never the
- * server's placeholder row.
+ * decrypt-on-click download. The fetched bytes live in the shared
+ * `attachment-cache`, so an image re-mounting on scroll doesn't re-fetch and
+ * re-decrypt. Real filename/size come from the ref, never the server's placeholder.
  */
 
 const PREFIX = "🔒 "
-
-async function fetchAndDecrypt(workspaceId: string, ref: AttachmentRef): Promise<Blob> {
-  const url = await attachmentsApi.getDownloadUrl(workspaceId, ref.attachmentId, { variant: "raw" })
-  const res = await fetch(url)
-  if (!res.ok) throw new Error(`Attachment fetch failed (${res.status})`)
-  const ciphertext = new Uint8Array(await res.arrayBuffer())
-  const plaintext = await decryptAttachmentBytes({ ciphertext, key: ref.key, iv: ref.iv })
-  return new Blob([plaintext], { type: ref.mimeType })
-}
 
 function formatSize(bytes: number): string {
   if (bytes < 1024) return `${bytes} B`
@@ -34,29 +27,11 @@ function formatSize(bytes: number): string {
 }
 
 function E2eImageAttachment({ workspaceId, attachmentRef }: { workspaceId: string; attachmentRef: AttachmentRef }) {
-  const [url, setUrl] = useState<string | null>(null)
-  const [failed, setFailed] = useState(false)
+  const decrypted = useDecryptedAttachment(workspaceId, attachmentRef)
 
-  useEffect(() => {
-    let cancelled = false
-    let objectUrl: string | null = null
-    fetchAndDecrypt(workspaceId, attachmentRef)
-      .then((blob) => {
-        if (cancelled) return
-        objectUrl = URL.createObjectURL(blob)
-        setUrl(objectUrl)
-      })
-      .catch(() => {
-        if (!cancelled) setFailed(true)
-      })
-    return () => {
-      cancelled = true
-      if (objectUrl) URL.revokeObjectURL(objectUrl)
-    }
-  }, [workspaceId, attachmentRef])
-
-  if (failed) return <E2eFileAttachment workspaceId={workspaceId} attachmentRef={attachmentRef} />
-  if (!url) {
+  if (decrypted.status === "failed")
+    return <E2eFileAttachment workspaceId={workspaceId} attachmentRef={attachmentRef} />
+  if (decrypted.status === "pending") {
     return (
       <div className="flex h-40 w-full max-w-sm items-center justify-center rounded-lg border bg-muted/40">
         <Loader2 className="h-5 w-5 animate-spin text-muted-foreground" aria-label="Decrypting attachment" />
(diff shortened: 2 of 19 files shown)


kristofferremback:
### Code review

Found 2 issues — both fixed in `92ed625`. One higher-scored finding was investigated and dismissed as not reachable (reasoning below).

1. `apps/frontend/src/hooks/use-decrypted-attachment.ts:58-68` — **Cache-hit pending flash** (reactivity / ux-feedback, scored 84 + 82 by two lenses). The `blob` arrived synchronously from `useSyncExternalStore`, but the object URL landed a commit later via `useState`/`useEffect`, so a cache-hit re-mount (scroll a decrypted image back into view) flashed the pending spinner for one frame before painting — undercutting this slice's "re-mount is free" goal. **Fixed:** the URL is now derived synchronously with `useMemo`; the effect keeps ownership of revocation (unmount / url change / lock-clear).
   https://github.com/threahq/threa/blob/92ed62591021bb76fbb6d79b1c72b9907f33903e/apps/frontend/src/hooks/use-decrypted-attachment.ts#L57-L69

2. `apps/frontend/src/hooks/use-decrypted-attachment.test.ts:67` — **Inaccurate comment** (claude-md [INV-25, §7 Maintainability]: "Comments explain why the current implementation works; they do not narrate … a stale comment is worse than none"). The comment called a retryable failure "terminal," but the cache is `retryFailed: true`. **Fixed:** the comment now states the real reason (the hook's effect only fires on undefined/pending; the retry is driven by the download button).
   https://github.com/threahq/threa/blob/92ed62591021bb76fbb6d79b1c72b9907f33903e/apps/frontend/src/hooks/use-decrypted-attachment.test.ts#L66-L70

---

Matches the slice-3 spec in `docs/plans/client-decrypt-layer-unification.md`: attachment-bytes cache instance, the read hook, `e2e-attachment-list` off per-mount re-decrypt, the `content`→`value` rename across all readers, and the AGENTS.md recipe. Q2/Q3 resolved in the doc. The `content`→`value` rename is complete — `tsc --noEmit` is the completeness guard (a missed cache-entry accessor wouldn't compile), and no accessor was left in an untyped cast.
</details>
<details><summary>🔍 Bugs [1 investigated, dismissed]</summary>

**`use-decrypted-attachment.ts` request effect — "wedge at pending after a mid-flight lock" (correctness agent, scored 88): investigated and dismissed as not reachable.** The concern: after `clearAllDecrypted()` on lock, a request that was in flight leaves the hook with `cached === undefined` and `status === undefined` unchanged, so the effect (deps `[attachmentId, workspaceId, status]`) doesn't re-fire — unlike `useDecryptedMessageContent`, whose deps include `opts.privateKey` and restart on unlock.

This is a real fragility of the hook **in isolation**, but it is not reachable in the tree: `message-event.tsx:1808` passes `attachmentRefs` only when `useDecryptedMessageContent` is `decrypted`, so a lock flips that to `locked` and **unmounts** `E2eAttachmentList` (and this hook). The hook is therefore only ever mounted under an unlocked session and is torn down on lock — it never observes a lock while mounted, and remounts fresh on unlock. I added a comment documenting this precondition (why the hook needs no `resolveDecryptContext` session gate) so a future caller that keeps it mounted across a lock is warned.
</details>

Traced the `{status, value}` rename through every reader (message-content, step-content, stream-search, drafts: `resolveDraftDecryption` / `rememberDecryptedRefs` / `cachedDraftBody` / `cachedDraftAttachments`) — none reads the old field. The lock-epoch generation guard in the shared primitive correctly drops a decrypt (message or attachment) that resolves after `clear()`, so no plaintext is written back post-lock.
</details>
<details><summary>📋 CLAUDE.md Compliance [1 fixed]</summary>

INV-25 (stale/inaccurate comment) — issue #2 above, fixed. No INV-18 violation (all components module-level). The module-level cache instance is the established decrypt-layer pattern (slices 1–2), consistent with INV-9's approved exceptions. INV-63: download success stays silent (browser download is its own signal), `toast.error` only on failure — compliant.
</details>

Cache-the-Blob-not-the-URL keeps eviction/lock leak-free (URL lifecycle in the hook). `retryFailed: true` + `lru: 64` justified for whole-file blobs. Per-domain hook with shared helpers matches the resolved Q3.
</details>

E2EE-4 (plaintext memory-only, cleared on lock): the decrypted Blob lives only in the cache (dropped by `clearAllDecrypted`); the object URL is revoked on unmount / blob-change / lock-clear. No path keeps a decrypted blob or live URL past lock. The download path's `setTimeout(revoke, 10s)` is pre-existing and covers a file the user explicitly downloaded to disk (noted "Out of scope" in the PR).
</details>
<details><summary>🎛️ UX [1 fixed]</summary>

The pending-flash (issue #1) was the one UX finding — fixed. The image→download-chip fallback on decrypt failure and the pending→ready layout transition are unchanged from the prior implementation (pre-existing, not amplified). No `toast.success` added (INV-63).
</details>

Logic-only refactor: every mobile-relevant className in `e2e-attachment-list.tsx` (`max-w-sm`, `max-h-80`, `h-40 w-full`, the `h-7` download button) is unchanged. No mobile regression.
</details>

PR #991, opened 2026-06-17
feat(e2ee): consent gate for adding a bot to an encrypted scratchpad

kristofferremback:
**Design doc:** `docs/plans/agent-runtimes-unification-redesign.md` §2.6 — forward-compatibility for E2EE external agents (Phase 4: consent UX). Builds on the merged sealed wire (#969 claim-coverage, #981 claim, #983 steps, #988 complete).

## Problem

`StreamBotsSection` (the Bots list in stream settings → Members) grants a bot to a stream by calling `botsApi.grantStreamAccess`, which writes a `bot_channel_access` row. That's the right grant for a plaintext stream, but it's the wrong — and silently broken — grant for an **end-to-end encrypted** scratchpad:

- On an E2E stream, a bot is dispatched a turn only when `resolveSealingContext` finds an `e2e_stream_actors` row for it (`apps/backend/src/features/e2e-streams/sealing-context.ts:32-37`); a missing row resolves the delivery verdict to `denied` and the turn is never dispatched. The stream SSK is wrapped to the bot in the same invite path. `bot_channel_access` does neither.
- The sealed callback path doesn't even consult `bot_channel_access`: `authorizeSealedInvocationCallback` (`apps/backend/src/features/public-api/handlers.ts:721-740`) authorizes steps/complete by bot API key + per-claim callback token + persona match, never `assertStreamAccessible`.

So today "Add bot" on an encrypted scratchpad records a grant that lets the bot do nothing. And unlike the enclave (Ariadne), whose key-wrap is automatic on scratchpad creation, granting a third-party bot read access to an encrypted conversation is a decision the owner should make deliberately — there was no consent surface for it.

## Solution

On an E2E scratchpad, route the "Add bot" action through the existing `inviteActorToStream` path (the SSK wrap — the owner's deliberate act) behind a confirm dialog, **in addition to** the standing channel grant. Plaintext streams keep the silent immediate grant unchanged.

### How it works

- `StreamBotsSection` now takes the `stream` (from the Members-tab bootstrap) and derives `isE2e = stream?.e2eEnabled === true`.
- **Plaintext:** selecting a bot calls `grantMutation.mutate(bot.id)` immediately — same as before.
- **E2E:** selecting a bot sets `pendingGrantBot`, opening a `ResponsiveAlertDialog` ("This bot will be able to read every message in this end-to-end encrypted scratchpad. Only add bots you trust…"). On confirm, `confirmE2eGrant`:
[...]

PR #988, opened 2026-06-17
feat(e2ee): sealed bot completion (/sealed-complete) behind the policy flag

kristofferremback:
## Summary

Phase 3 of E2EE for external bot harnesses. Adds a sealed completion endpoint so an owner-granted external bot harness can complete a **sealed** turn the way the enclave does. **Inert in production** — no delivery verdict resolves to `sealed` until `externalSealedDelivery` flips (Phase 6), so this endpoint is dark today.

Builds on #983 (sealed trace steps), #981 (sealed claim), #969/#967. Workstream: design doc `docs/plans/agent-runtimes-unification-redesign.md` §2.6.

New route: `POST /api/v1/workspaces/{workspaceId}/bot-invocations/{invocationId}/sealed-complete` — the external sibling of the enclave `/complete` and the sealed variant of the plaintext bot `/complete`.

## What it does

`completeBotInvocationSealed` (`apps/backend/src/features/public-api/handlers.ts`): in **one transaction** (INV-7) it persists the turn's final sealed reply (or `noResponse`), flips the `bot_invocations` claim, and finalizes the agent session (`completeSession` + the `agent_session:completed` stream event/outbox), then emits the live session-room frame **outside** the tx (INV-41). Race-safe on the `status = 'claimed'` predicate (INV-20); the server only ever persists/relays ciphertext (INV-E7). Auth is the bot API key + the per-claim callback token in `X-Threa-Callback-Token`, verified against the session's `callback_token_hash` binding, plus workspace/persona isolation (INV-8) — no `instanceId`/`claimToken` body fields (the header-auth model, "model A").

Unlike the enclave (which streams each reply to `/messages` and acks here), the bot path delivers its single sealed reply **inline**, exactly as the plaintext bot complete does — only the body is ciphertext.

## Design decisions

- **Neutral, shared sealed vocabulary (§2.6 rule 1):** adds `SealedComplete` (reusing `SealedReply`), not an enclave-named type. No `assertNotE2eStream` sprinkle and no `instanceof` branch on the sealed path (§2.6 rule 2) — the route inherits darkness because no sealed claim is ever issued while the flag is off.
[...]

PR #997, opened 2026-06-17
feat(pi-remote): sealed (E2EE) external-bot harness — Phase 5

kristofferremback:
**Tracking:** `docs/plans/agent-runtimes-unification-redesign.md` §2.6 (forward-compatibility: E2EE for external agents) — **Phase 5** of the sealed external-bot workstream. No Linear ticket.

## Problem

The sealed (end-to-end-encrypted) wire for an external bot exists end-to-end on the **backend** — the claim hands a `SealedTurnContext` (#967/#981), `/sealed-steps` (#983) and `/sealed-complete` (#988) accept sealed callbacks, and claims are gated on BIK wrap coverage (#969). Nothing on the **client** consumed it. The Pi remote harness (`extensions/pi-remote/`) only spoke the plaintext wire: it had no identity key to receive an SSK wrap, couldn't open a sealed claim, and posted replies/steps as cleartext. So an owner could invite the Pi bot to an E2E scratchpad (the consent gate landed in #991) but the bot could never actually serve a turn there.

This is Phase 5: build the client half — generate + register a BIK, open the sealed claim, run the turn locally, and seal replies/steps back. It stays dark until Phase 6 flips `externalSealedDelivery` (the gate is still `false`, so no E2E claim resolves to `sealed` for a third-party bot yet).

## Solution

Mirror the enclave's sealed runner (`apps/enclave/`) inside the standalone Pi harness, reusing the exact crypto and AAD scheme so the owner's client opens the bot's replies the same way it opens the enclave's. The plaintext path is untouched; everything sealed branches on a per-claim `sealing` state derived from `sealedContext`.

Because `extensions/pi-remote` ships standalone (copied to `~/.pi/agent/extensions/` + a local `npm install`), it can't import the private, unpublished `@threa/crypto` workspace package. Per Kris's call ("Let's vendor it for now. We can change this later."), the needed crypto slice is vendored into `src/crypto.ts`, with a parity test guarding it against the canonical package.

### How it works

1. **BIK (Bot Identity Key)** — `ensureBik()` lazily generates an X25519 keypair and persists it to `~/.pi/agent/threa-remote-bik.json` (mode `0600`), `publicKeyId` = `bik_<ulid>`. It rides **every** `bot:hello` and HTTP presence write via `bikPresenceFields()`. This is load-bearing: the backend's `bot_runtime_instances` upsert overwrites `public_key`/`public_key_id` by default (`features/bot-runtimes/repository.ts:428-430`; only server-internal writes pass `retainBik`), so a heartbeat that omitted the key would clear what `bot:hello` registered and break wrap coverage.
2. **Open the claim** — `hydrateSealedClaim` → `openSealedTurnContext` unwraps the SSK for each generation in `sealedContext.wraps` (`unwrapStreamKey`, AAD `buildWrapAad({ streamId: rootStreamId, keyGeneration, recipientKeyId: bik.publicKeyId })`), opens the trigger + history (`openMessageAsString` + `parseSealedPayload`), sets the decrypted trigger as `promptMarkdown`, and stashes a `SealingState` (reply SSK, generation, sender id, callback token, decrypted context). The turn then runs identically to a plaintext one.
3. **Seal back** — replies (`sealReplyWith`, `msg_<ulid>`) and trace steps (`sealStepWith`, `step_<ulid>`) seal under the reply SSK with `buildMessageAad({ streamId: rootStreamId, messageId|stepId, senderId })` — the same `streamId|id|senderId` binding the enclave's `trace-observer` uses, so a step id rides the `messageId` slot. Sealed callbacks authorize with the `X-Threa-Callback-Token` header (the claim's `callbackToken`), not body `instanceId`/`claimToken`. Steps POST to `/sealed-steps`, completion to `/sealed-complete`.
4. **Stream id** — wraps and the owner's message AAD bind to the **root** stream that owns the E2E key (the enclave assignment uses `e2e.streamId`, the root). The harness uses `rootStreamId` from the claim response, falling back to `activeStreamId`.
5. **Trace detail scales with sealing** — on a sealed turn the harness sends the **full** tool args / file contents / output (the real command, the patch, stdout); on a plaintext turn it keeps the existing redactions ("omitted for safety"). `shouldEmitFullTrace` gates it (see decision 6).

### Key design decisions

**1. Noble X25519 KEM, not `@hpke/core`'s native one.** Bun's WebCrypto lacks X25519 `deriveBits`, so the native KEM throws `EncapError`/`DecapError: The algorithm is not supported` on both seal and open (verified empirically in Bun 1.3.11). Pi's runtime is unknown and could be Bun, so `crypto.ts` composes the suite from `@hpke/dhkem-x25519` (pure-JS, noble-backed) while keeping `@hpke/core`'s `Aes256Gcm` + `HkdfSha256`. It's the identical RFC 9180 DHKEM(X25519, HKDF-SHA256) and is wire-interoperable with the owner's native KEM (keys serialize to the same 32 raw bytes; `@hpke` supports mixing implementations). This makes the vendored copy *more* runtime-portable than canonical `@threa/crypto`, which assumes Node.

**2. Vendor with a byte-parity drift guard.** `crypto.ts` is a faithful subset of `packages/crypto`'s `{encoding,hpke,stream-key,envelope,sealed-payload}`. The INV-35 duplication risk is mitigated by `crypto.parity.test.ts`, which imports **both** the vendored module and canonical `@threa/crypto` (relative path — the bare specifier won't resolve in a non-workspace package) and asserts identical AAD bytes, envelope/payload versions, cross seal/open round-trips, and that a vendored unwrap recovers a wrapped SSK. If the canonical AAD layout or envelope version ever changes, this fails. Only the recipient/seal half is vendored — the harness never wraps a key to another recipient, so HPKE `seal`/`wrapStreamKey` are omitted (INV-38).

**3. Don't leak plaintext on the E2E path.** Sealed completion strips `THREA_ATTACH:` directives instead of uploading bytes to plaintext S3 (attachment sealing is out of scope; a one-line note tells the user). Sealed `/fail` scrubs the error to its class name (`scrubSealedError`) so a decrypt/turn error can't echo decrypted content; the enclave's failure path is the same shape (`errorName` only). `buildInvocationPrompt` never calls the plaintext messages API for a sealed turn — context comes from what was already decrypted at claim time.

**4. Hydration failure fails the claim, loudly.** If the BIK can't open `sealedContext` (a BIK/wrap race, or a key rotated away), `claimNextInvocation` fails the invocation with a generic reason and returns `null` rather than throwing up through the poll and leaving the claim to TTL-recycle in a loop.

**5. Self-installed package.** Like its sibling `extensions/claude-code-remote`, pi-remote is not a Bun workspace member; it declares its own deps and commits its own `bun.lock`. Added `@hpke/core`, `@hpke/dhkem-x25519`, `ulid`. It has no `typecheck` script — `threa-remote.ts` type-imports the *global* `@earendil-works/pi-coding-agent`, which isn't an installed dep, so `tsc` can't resolve it in-repo (`bun test` runs fine — type-only imports are erased).

**6. Full trace detail when sealed, redacted when plaintext (configurable, default full-on-sealed).** The point of E2E trace steps: give the *owner* more without giving the *server* anything. On a sealed turn each step is ciphertext the server can't read, so `formatToolCallTrace`/`formatToolResultTrace` emit the real command/patch/output (`fullToolArgumentSummary` + the real result body). On a plaintext turn the steps reach the server in the clear, so they always stay redacted — the toggle **cannot** turn full detail on for plaintext (that would reintroduce the exact "whoops, a secret hit the server" liability E2EE removes). `shouldEmitFullTrace(invocation)` = `invocation.sealing !== undefined && config.sealedFullTrace !== false`; `config.sealedFullTrace` (default `true`, editable via `/remote-control configure`) only lets a user opt a sealed turn back to redacted. `full` defaults to `false` at the formatter so a missed call site fails safe (redacted).

## New files

| File | Purpose |
| ---- | ------- |
| `extensions/pi-remote/src/crypto.ts` | Vendored `@threa/crypto` subset (noble X25519 KEM): HPKE keygen/open, AES-GCM stream seal/open, AAD builders, sealed-payload parse/serialize |
| `extensions/pi-remote/src/crypto.parity.test.ts` | Drift guard: byte-parity of the vendored module against canonical `@threa/crypto` |
| `extensions/pi-remote/src/threa-remote.sealed.test.ts` | Sealed-path behavior: open a sealed claim, seal a reply/step, AAD binding, error scrub |
| `extensions/pi-remote/bun.lock` | Standalone lockfile (matches `claude-code-remote`) |

## Modified files

| File | Change |
| ---- | ------ |
| `extensions/pi-remote/src/threa-remote.ts` | BIK keystore; sealed-claim hydration; sealed steps/complete/fail routing; BIK on `bot:hello` + presence; full-vs-redacted trace detail + `config.sealedFullTrace`; `__testing` exports |
| `extensions/pi-remote/src/threa-remote.test.ts` | Full-detail (sealed) trace cases alongside the existing redaction tests |
| `extensions/pi-remote/package.json` | Add `@hpke/core`, `@hpke/dhkem-x25519`, `ulid`; `@types/bun`; `test` script |
| `extensions/pi-remote/README.md` | Note the new deps + vendored crypto |
| `.agents/skills/update-pi-remote-plugin/SKILL.md` | Document the sealed wire, BIK, vendored crypto, full-trace toggle, self-install model, sealed checklist |

## Out of scope (deliberate)

- **Flipping `externalSealedDelivery`** — Phase 6. Until then no E2E claim resolves to `sealed` for a third-party bot, so this code is dark in production.
- **Sealed attachments** — the `SealedTurnContext` ships no attachment ciphertext and the harness uploads none on the sealed path (directives stripped). Additive later.
- **Interjection pull, rolling summary, auto-title, turn digests** — enclave-only continuity features; `SealedTurnContext` deliberately omits them (INV-36).
- **`extensions/claude-code-remote`** — the other standalone channel will need the same crypto vendoring when it grows an E2E path; not touched here.

## Design evolution

- **Rebased onto the latest `main`** as it advanced under the PR: #996 (keep intermediate narration out of the posted reply) and #1000 (per-key merge-aware, locked `saveConfig` + JSON error bodies) both touch `threa-remote.ts`. Both auto-merged with the sealed code; a nice interaction — #996's new `message_end → recordTraceStep("thinking", …)` now flows through the sealed branch, so the model's running narration is sealed on an E2E turn. #1004 (backend presence grace) left the `bot_runtime_instances` upsert unchanged, so the "BIK rides every presence write" reasoning still holds.
- **Added the full-vs-redacted trace toggle (decision 6)** after the initial sealed harness, on owner request: "the point of E2EE on the steps is to give the owner more information without giving the server more information."

## Test plan

- [x] `cd extensions/pi-remote && bun test` — 100 pass (86 adapter + 7 crypto parity + 5 sealed + 2 full-trace), 0 fail
- [x] Pre-commit hook green on both commits: root `typecheck` (all 14 packages), `astro check`, OpenAPI `--check`, migration + Dockerfile-workspace checks
- [x] Vendored `crypto.ts` typechecks clean in isolation; the feature added no new type errors (a strict scoped `tsc` shows only the harness's pre-existing strictness gaps + the global Pi type, unrelated to this change)
- [x] Claim-verification pass: every `file.ts`/symbol/behavioral claim above checked against the diff
- [ ] Native↔noble X25519 interop not exercised in-suite — native encap/decap throw under Bun, so the parity test wraps with noble; cross-impl interop relies on RFC 9180 conformance + identical 32-byte key serialization (the production owner runs native WebCrypto X25519, which browsers support)
- [ ] Full E2E (sealed delivery through Pi against a live backend) — gated by `externalSealedDelivery=false` (Phase 6); no live-DB/Pi harness in this environment

<details>

**Goal:** Phase 5 of the sealed external-bot workstream (design §2.6) — the Pi client harness that drives the sealed wire built in #969/#967/#981/#983/#988.

**What was built:**
- Vendored crypto module (`crypto.ts`) — recipient/seal subset of `@threa/crypto`, noble X25519 KEM for runtime portability.
- BIK keystore — per-install X25519 keypair, persisted `0600`, registered on `bot:hello` + every presence write.
- Sealed claim hydration — unwrap SSK (root-stream AAD), open trigger/history, derive `SealingState`.
- Sealed callbacks — replies/steps sealed under the SSK (`buildMessageAad`), `X-Threa-Callback-Token` auth, `/sealed-steps` + `/sealed-complete`; `/renew` + `/fail` shared (claim token = callback token, model A); errors scrubbed; attachment directives dropped.
- Full-vs-redacted trace detail — full (real args/output) on sealed turns, redacted on plaintext; `config.sealedFullTrace` toggle (default on for sealed).
- Tests — parity drift guard + sealed behavior + full-trace cases. Skill + README updated.

**Design decisions:** noble KEM (Bun lacks native X25519); vendor + parity guard (INV-35); no plaintext leak on E2E (strip attachments, scrub errors, no plaintext context fetch); fail hydration loudly; self-installed package; trace detail scales with sealing.

**Schema changes:** none.

**What's NOT included:** the `externalSealedDelivery` flip (Phase 6); sealed attachments; enclave-only continuity (interjection/summary/title/digests); `claude-code-remote` vendoring.

**Status:** rebased onto latest `main`, pushed (commits `4712ec9` + `e1b9433`), dark behind the policy flag.

</details>

---


kristofferremback:
_Superseded by a [newer review](https://github.com/threahq/threa/pull/997#issuecomment-4750405352)._

### Code review
**Review models:** Orchestrator: Claude Code | Reviewers: sonnet x2 (spec+design, correctness+data-flow)

Reviewed at `e1b9433`; every finding below is addressed through `05d7425`.

**My reviewers (2 — backend-only diff, so UX/Mobile skipped):**

1. `extensions/pi-remote/src/threa-remote.ts` — BIK lifecycle (state-lifecycle). Create/persist was all-or-nothing, and a concurrent first-time `ensureBik` could mint two keypairs and register different ones. → Fixed: `createBik` now persists best-effort (still returns the in-memory key, logs on failure), and `ensureBik` memoizes the first create in a shared promise so concurrent callers share one key. The originally-flagged "a later heartbeat clears a registered BIK" is not reachable — `ensureBik` returns the cached key once set, so once registered the BIK rides every presence write.

Spec & design reviewer: no issues — the security invariant holds (full trace detail is unreachable on a plaintext turn: `shouldEmitFullTrace` gates on `invocation.sealing`, set only by `hydrateSealedClaim`; `config.sealedFullTrace` can only opt a *sealed* turn back to redacted), and INV-2/25/35/36/38/47 are clean.

---

**CodeRabbit triage**

Fixed in `05d7425`:
- **Attachment-only sealed reply** no longer collapses to `noResponse` — it seals the "attachments unsupported" notice so the owner isn't left with silence (`completeSealedInvocation`).
- **`THREA_ATTACH` no longer advertised on sealed turns** — sealed completion strips it, so the prompt now tells the agent attachments are unsupported there (`buildInvocationPrompt`).
- **`ensureBik` first-time race** serialized via a shared in-flight promise (above).
- **SKILL install-wording** npm/bun mix removed.

Declined, with reasons:
- **`openMessage` caller-computed AAD** and **`buildMessageAad` input validation** (`crypto.ts`): this file is a byte-faithful, parity-tested mirror of `@threa/crypto`, and the enclave reference runner (`apps/enclave`) relies on the identical `openMessage`/`buildMessageAad` behavior. Hardening these primitives belongs **upstream in `packages/crypto`** (then re-vendored here), not as a unilateral fork divergence that the parity test exists to prevent. In this harness the inputs are always prefixed ULIDs (no `|`, never empty), so the proposed validation never fires; and a caller-computed AAD on `openMessage` adds no real protection here, because the server delivers the whole `sealedContext` — the harness would compute the "expected" AAD from the same untrusted source. Happy to do the upstream `packages/crypto` hardening as a separate change if you'd like it.
- **Fail-closed presence/hello when `ensureBik` fails**: addressed via the create-race fix instead. With memoization a registered BIK can never be cleared (once `bik` is set `ensureBik` always returns it; before any success there is no key to clear). Failing closed would also drop *plaintext* presence on a catastrophic keygen failure — worse, since the plaintext path needs no BIK.

</details>

kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Code | Reviewers: sonnet x2 (spec+design, correctness+data-flow)

Re-review covering the fix commits since the last pass (HEAD `39714c9`; backend-only diff, so UX/Mobile skipped). Found 1 issue — fixed in `39714c9`:

1. `extensions/pi-remote/src/threa-remote.ts` — `createBik` logged a *persistence* failure but a *key-generation* failure (e.g. a WebCrypto build without X25519) propagated out and was swallowed by the heartbeat/hello `.catch(() => undefined)`, so sealed turns could become unservable with no diagnostic — and the comment over-claimed "Logged, never silent" (INV-11 fail-loud + INV-25 accurate-comment). Fixed: generation wrapped in a try/catch that logs loudly and rethrows; comment corrected.
   https://github.com/threahq/threa/blob/39714c952e4d01a474c32d8e581becdbb0e8e8f5/extensions/pi-remote/src/threa-remote.ts#L470-L485

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Phase 5 + the full-vs-redacted trace toggle implemented as intended. Full trace detail is unreachable on a plaintext turn — `shouldEmitFullTrace` gates on `invocation.sealing` (set only by `hydrateSealedClaim` on a claim carrying `sealedContext`), and `config.sealedFullTrace` can only opt a sealed turn back to redacted.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>The `ensureBik` first-time-create memoization is correct: concurrent callers share one in-flight promise (one keypair), and a rejection resets `bikInFlight` so the next call retries. `completeSealedInvocation` attachment-only / empty / NO_RESPONSE branches are correct.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Once a BIK is registered it can't be cleared by a later BIK-less presence write (cached `bik` always rides presence); `sealedContext` is never forwarded to the server; the `sealing`/`full` flag gates identically across both formatters and the step-routing dispatch.</details>
<details><summary>📋 CLAUDE.md Compliance [1 — fixed]</summary>INV-11 / INV-25: silent key-generation failure + over-claiming comment in `createBik` (above), fixed in `39714c9`.</details>
<details><summary>🏗️ Design [CLEAN]</summary>Vendored crypto is a parity-test-guarded mirror of `@threa/crypto`; full-trace decision lives in one gate (`shouldEmitFullTrace`).</details>
<details><summary>🔒 Security [CLEAN]</summary>Full detail is E2E-only (ciphertext); the BIK private key is never logged or sent; sealed `/fail` is scrubbed to the error class name.</details>

kristofferremback:
Addressed the INV-24 nit in `d208053` (consolidated the sealed-test assertions).

On the two re-surfaced `crypto.ts` items (`openMessage` caller-AAD, `buildMessageAad` validation) — keeping the earlier decision, and there's a decisive reason beyond "it never fires for ULID inputs": **`crypto.ts` is a synced copy of `@threa/crypto`** (the header says so, and `crypto.parity.test.ts` enforces byte-parity). A local-only hardening here would be **reverted on the next re-vendor from `packages/crypto`**, so it wouldn't survive — and it would diverge from the enclave reference runner, which calls the identical `openMessage`/`buildMessageAad`. If we want the slot-binding-on-open and the message-AAD validation, the right move is to add them **upstream in `packages/crypto`** (canonical `buildMessageAad` in `envelope.ts` is the one that lacks the validation `buildWrapAad` has) and then re-vendor. Happy to do that as a separate PR if you'd like it — it's just out of scope for the Pi-harness change.

kristofferremback:
Superseded by #1166 — same crypto design (vendored noble-KEM slice, BIK keystore, AAD scheme, full-trace-when-sealed), but rebuilt against the post-#1067 WS-transport architecture with the sealed support in @threa/bot-runtime-client so both harnesses share it, plus the pieces this PR predated: sealed interim messages, a WS sealed-steps frame, harness-created encrypted scratchpads, the consent UI, and the EXTERNAL_SEALED_DELIVERY flip. Leaving this open for Kris to close.

kristofferremback:
Acknowledged — superseded by #1166. Standing down from watching/rebasing this branch; leaving it open for you to close. Happy to help on #1166 if useful.

---

PR #1000, opened 2026-06-18
fix(pi-remote): stop concurrent Pi instances clobbering each other's links + surface API error bodies

kristofferremback:
fix(pi-remote): stop concurrent Pi instances clobbering each other's links + surface API error bodies

Two related fixes found while running two Pi remotes side-by-side (a long-chat
perf worktree Pi alongside the main Pi), both sharing
`~/.pi/agent/threa-remote.json`.

## 1. Merge-aware, atomic config writes

`linkedSessions` is keyed by `runtimeSessionId` (derived from cwd), so each Pi
instance owns exactly one entry. But `saveConfig()` wrote the whole in-memory
config over the file — and an instance only holds its own link in memory. So a
save from instance A silently wiped instance B's `linkedSessions` entry.

Symptom: the other Pi showed "Threa remote: not linked" after a reload and
stopped polling, so it never claimed messages sent to its scratchpad.

Fix: `saveConfig()` now re-reads the file on disk and merges `linkedSessions`
per-key (this instance's own keys win over stale on-disk copies; other
instances' links are preserved), then writes atomically via temp-file + rename.
Non-mergeable fields (`baseUrl`/`workspaceId`/`apiKey`/...) keep the in-memory
value. `streamCursors` (legacy) is merged the same way.

This narrows the race to near-zero and makes a half-written file impossible; a
true cross-process lock isn't added (each instance only ever mutates its own
key, so last-writer-wins per key is correct).

## 2. Surface the server error body on non-2xx responses

`request()` deliberately threw away the response body on non-2xx to avoid
dumping huge HTML proxy/CGI error pages into Pi. That also discarded the
server's exact Zod `fieldErrors` / error `code`, so a rejected `/complete`
surfaced only as `Threa API 400: Bad Request` — undiagnosable.

Fix: for `application/json` responses, read and format the error body
(`error`, `code`, `details`), capped at 500 chars. Non-JSON responses (proxy
HTML) are still not read, so a 502 markup page can't flood Pi.

Example after the fix: `Threa API 400: Bad Request — Validation failed
[CAPABILITY_NOT_DECLARED] details={...}`.

## Tests

85 pass, bundles clean. No behavior change to the happy path.

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1000"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784374906&installation_id=129946197&pr_number=1000&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1000&signature=45a18677e5fc1ac0a2e4211841b4c09dbd29c5596876eb8b8ab6d1262da02072"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/extensions/pi-remote/src/threa-remote.ts
+++ b/extensions/pi-remote/src/threa-remote.ts
@@ -1,6 +1,6 @@
 import { execFile, spawn } from "node:child_process"
 import { promisify } from "node:util"
-import { existsSync, mkdirSync, readFileSync, renameSync, statSync, writeFileSync } from "node:fs"
+import { closeSync, existsSync, mkdirSync, openSync, readFileSync, renameSync, statSync, unlinkSync, writeFileSync } from "node:fs"
 import { homedir, hostname, platform } from "node:os"
 import { basename, dirname, join, resolve } from "node:path"
 import { io as openSocket, type Socket } from "socket.io-client"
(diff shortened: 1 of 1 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: opencode-go/glm-5.2 | Reviewers: sonnet x2 (spec+design, correctness+data-flow) — run inline by the orchestrator (subagent fan-out unavailable in this session); backend-only diff, so UX + Mobile were not spawned per the skill's tiering.

Found 1 issue:

1. `extensions/pi-remote/src/threa-remote.ts:239-259` — Residual read-modify-write race means a concurrent Pi instance's link can still be dropped (the exact bug this PR targets). The new `saveConfig()` does `readStoredConfig()` → merge → `writeFileSync(tmp)` → `renameSync`. The read and the rename are not atomic across processes, so a classic TOCTOU window remains: instance B can `readStoredConfig()` after A has read but before A's `renameSync` lands. B then merges against the *pre-A* snapshot, and B's later `renameSync` overwrites A's just-written file — dropping A's `linkedSessions` entry (or reverting A's cursor update to its stale on-disk value). The PR text acknowledges a cross-process lock isn't added and asserts "last-writer-wins per key is correct," but that only holds when the winning write actually carries the key's latest value; in this interleaving the winner carries a *stale* snapshot of the loser's key. Impact is the original symptom: the dropped instance shows "Threa remote: not linked" on its next reload and stops polling. It self-heals on the owning instance's next `saveConfig()` (cursor advance on turn completion), but a reload between saves reproduces the bug. The change is a strict improvement over the guaranteed-clobber before it (hence 6/7 not lower), but it isn't fully INV-20-compliant. Fix: wrap the read-merge-write in a cross-process lock — `proper-lockfile`'s `lockSync` (already a dependency of pi's own `FileSettingsStorage` for `settings.json`) on `CONFIG_PATH`, mirroring that pattern; that closes the window entirely while keeping the merge. (CLAUDE.md [INV-20, "Write paths must be race-safe"]: "Never do select-then-update without locking/concurrency control"; bug: file-based RMW is still check-then-act without a lock.)
   https://github.com/threahq/threa/blob/b569a0167c008e923ef4e9a16ec0d5a17aa1b4eb/extensions/pi-remote/src/threa-remote.ts#L239-L259

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>The two stated fixes both land and match intent: (1) merge-aware `linkedSessions` so concurrent instances stop wiping each other's links; (2) surface JSON error bodies on non-2xx (capped, JSON-gated) so 400s are diagnosable. No missing corresponding changes — this is a runtime-loaded Pi package with no backend/frontend/type/migration counterparts.</details>
<details><summary>🔍 Bugs [1 issue]</summary>See issue 1 — the residual RMW race. No other logic errors: temp file is pid-namespaced (no cross-instance tmp collision), `renameSync` is atomic on POSIX, the JSON-gated body read can't flood Pi (sliced to 500), and `readStoredConfig` failing leaves `detail=""` (graceful).</details>
<details><summary>🔁 Data Flow [1 issue]</summary>See issue 1. Traced the `linkedSessions[runtimeSessionId]` lifecycle: create (`createRemoteSession` → save), mutate (enable/disable/rename/cursor → save), reload (`session_start` reads disk). The defect is a field (a peer instance's key) dropped across the save path under interleaving — the canonical round-trip-preservation miss, just at the file layer. Self-heal exists via the owner's next save, which softens but doesn't eliminate the window.</details>
<details><summary>📋 CLAUDE.md Compliance [1 violation]</summary>INV-20 ("write paths must be race-safe / never select-then-update without locking") — issue 1. No other invariants engaged (INV-17 migrations, INV-51 colocation, INV-4 outbox, INV-62 stream access are backend-only and don't apply to this extension file).</details>
<details><summary>🏗️ Design [CLEAN, 2 minor notes]</summary>Two minor cleanups below the ≥80 bar, not blocking: (a) the `streamCursors` merge branch is dead code — `persisted.streamCursors` is `delete`d a few lines above, so `if (onDisk?.streamCursors && persisted.streamCursors)` can never be true; the legacy global cursor is intentionally not re-persisted (migrated into per-link cursors by `migrateSessionState`), so this is harmless but misleading — drop the branch or comment it. (b) A crash between `writeFileSync(tmp)` and `renameSync` leaves a `.pid.tmp` orphan; harmless litter, worth a `try/finally` cleanup if you touch this again. No leaky abstractions or config sprawl introduced.</details>
<details><summary>🔒 Security [CLEAN]</summary>No high-confidence vuln. The error-body path reads server JSON and injects it into a Pi `Error` message / `ui.notify` — it's rendered as terminal text by Pi, not as HTML/markdown in Threa, so no XSS sink. Content is capped at 500 chars. No secrets are logged (the body is the server's error, not request payloads). The merge does not widen access to other workspaces' data — `linkedSessions` is local-only config.</details>
<details><summary>🎛️ UX [skipped]</summary>Backend-only diff (no frontend files); UX agent not spawned per tiering.</details>
<details><summary>📱 Mobile [skipped]</summary>Backend-only diff (no frontend files); Mobile agent not spawned per tiering.</details>

kristofferremback:
Pushed `936803c8` addressing both review findings:

**1. Cross-process lock (🟠 Major, INV-20):** `saveConfig()` now wraps the entire `readStoredConfig()` → merge → temp write → `renameSync` in a zero-dependency O_EXCL lockfile (`acquireConfigLockSync`), mirroring pi's own `FileSettingsStorage.lockSync` on `settings.json` — I avoided pulling `proper-lockfile` into this runtime-loaded package. Includes a stale-lock guard (>5s mtime reclaim) and a best-effort unlocked fallback so a lock failure never wedges the extension. The read-merge-write is now fully serialized across Pi instances. Verified: two concurrent `bun` saves both merge their own `linkedSessions` key without dropping pre-existing entries.

**2. Dead `streamCursors` merge (⚡ Quick win):** captured `inMemoryStreamCursors` before the `delete persisted.streamCursors`, and the merge now runs when either side has cursors (`onDisk?.streamCursors || inMemoryStreamCursors`). The branch is reachable now.

85 tests pass, bundles clean, no new dependencies. Happy to re-review.

kristofferremback:
Pushed `7b98db4d` addressing both new findings:

**1. Never write unlocked (🟠 Major, INV-20):** the best-effort unlocked fallback was wrong — on lock timeout/error the full read-merge-rename still ran and could clobber another instance's just-written sessions, reintroducing the original bug. Now both failure paths in `acquireConfigLockSync` log `skipping save` and return `undefined`, and `saveConfig` returns early (`if (!release) return`) when no lock was acquired. A skipped cursor/link save self-heals on the owning instance's next turn; an unlocked clobber does not, so skipping is strictly safer.

**2. Drop change-history narration (🧪 Nitpick):** rewrote the `inMemoryStreamCursors` comment to state the current invariant (`removed from the canonical write shape before the legacy cursor merge`) instead of narrating `the prior version of this merge was dead.`

85 tests pass, bundles clean. No unlocked write path remains.

PR #1005, opened 2026-06-18
fix(drafts): suppress sent draft resurrection races

kristofferremback:
## Problem

Draft resolve-on-send still had race windows that could leave an identical stash entry after the user sent a message:

1. `upsertLoadedDraft` only guarded the no-existing-row create path. A debounced save that had already read the loaded draft before `resolveLoadedDraft()` deleted it could still write that same row back afterward, turning the sent message into a stash entry.
2. `enqueueDraftResolve` deleted the queued `upsert_draft` op. If that op was already in flight, its socket/bootstrap echo could arrive with a version newer than the resolve CAS. With no local row and no dirty marker left, `applyDraftUpserted` treated the echo as another device's drift and accepted it.
3. Never-confirmed drafts (`baseVersion = 0`) used the unconditional delete path. If that delete reached the server before the first draft insert, the old `softDelete` matched zero rows and vanished; the late insert then created a live `version = 1` zombie draft.

The second race also has a split variant: a stale in-flight write can be accepted by the server under a fresh split id, so the client needs a way to recognize that split row as this device's own sent write rather than unrelated drift.

## Solution

Keep resolve-on-send as a no-loss CAS path, but teach both sides to distinguish unrelated drift from this device's own superseded sent-write echoes, and make delete-before-insert ordering durable for never-confirmed drafts.

### How it works

1. `upsertLoadedDraft` now checks the scope resolve sequence inside the IndexedDB write transaction before writing either an existing row or a new loaded row. If `resolveLoadedDraft` advanced the sequence while a debounced save was in flight, the stale save returns without touching IDB, cache, decrypt state, or the operation queue.
2. `enqueueDraftResolve` no longer removes a pending `upsert_draft`. It keeps the op as an in-flight echo marker and copies its `writeId` into the resolve payload as `supersededWriteIds`.
3. `applyDraftUpserted` now suppresses no-local-row echoes while a pending upsert exists, and CAS-resolves inbound rows whose `lastClientWriteId` is in the pending resolve's `supersededWriteIds` set. Bootstrap builds the same pending-id and pending-write-id sets once and passes them through the apply path.
4. The backend resolve endpoint accepts `supersededWriteIds`. `DraftsRepository.softDeleteCas` remains one workspace/user/id-scoped `UPDATE`, but its guard is now `(version = expectedVersion OR last_client_write_id = ANY(supersededWriteIds))`, so this device's lost-ack write can be tombstoned without deleting unrelated drift.
5. `DraftsRepository.softDelete` is now an `INSERT ... ON CONFLICT DO UPDATE` tombstone write. If delete wins before a never-confirmed draft's first insert, it plants a negative tombstone; when the late `expectedVersion = 0` upsert collides with that tombstone, `DraftsService.upsert` drops it without splitting.
6. If `executeDraftUpsert` receives an ack after the local draft was removed, it queues the matching cleanup: CAS resolve for sends, unconditional delete for explicit discards. Split rows are locally removed before queuing their cleanup so a socket echo that won the race cannot linger in IDB.

### Key design decisions

**1. Keep the queued upsert as the dirty marker** — deleting the op made the client unable to tell “my write is still in flight” from “there is no local work.” The queue still no-ops the upsert if the row is gone; keeping it only extends the dirty marker through the dangerous echo window.

**2. Match sent-write echoes by `lastClientWriteId`, not content** — content equality would be brittle across markdown/ProseMirror serialization and attachments. `lastClientWriteId` already exists as the server-side idempotency key, so exposing it on the owner-scoped draft wire row gives the authoring device a precise identifier for its own late write.

**3. Preserve resolve-vs-discard semantics** — confirmed send cleanup remains CAS-guarded and only removes matching versions or superseded write ids. Explicit discard still uses the unconditional delete path. Never-confirmed send cleanup also uses the unconditional path, but that path now plants a tombstone instead of disappearing when the row is absent.

## New files

None.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-draft-message.ts` | Adds the transactional stale-resolve check for debounced saves that already read the loaded draft. |
| `apps/frontend/src/sync/draft-sync.ts` | Carries superseded write ids through resolve ops, suppresses own in-flight echoes, resolves split ghosts, and preserves discard cleanup semantics. |
| `apps/frontend/src/sync/operation-queue.ts` | Replays `resolve_draft` with `supersededWriteIds`. |
| `apps/frontend/src/api/drafts.ts` | Documents the resolve wire contract. |
| `apps/backend/src/features/drafts/handlers.ts` | Validates optional `supersededWriteIds` on resolve. |
| `apps/backend/src/features/drafts/service.ts` | Passes superseded write ids into resolve and drops fresh upserts that collide with tombstones. |
| `apps/backend/src/features/drafts/repository.ts` | Extends the CAS soft-delete predicate and changes unconditional delete to plant negative tombstones. |
| `apps/backend/src/features/drafts/view.ts` | Includes `lastClientWriteId` on owner-scoped draft wire rows. |
| `packages/types/src/api.ts` | Adds `Draft.lastClientWriteId` and `ResolveDraftInput.supersededWriteIds`. |
| `apps/frontend/src/hooks/use-draft-message.test.ts` | Covers the stale save that already read the old loaded row. |
| `apps/frontend/src/sync/draft-sync.test.ts` | Covers pending-upsert echo suppression, split echo cleanup, CAS-vs-delete cleanup, and resolve payload write ids. |
| `apps/backend/src/features/drafts/repository.test.ts` | Covers the extended CAS predicate and negative tombstone write. |
| `apps/backend/src/features/drafts/service.test.ts` | Verifies resolve forwards superseded write ids, tombstoned fresh upserts are dropped, and delete publish/no-op behavior. |
| `docs/features/architecture/drafts.md` | Updates the draft architecture doc with the lost-ack, superseded-write-id, and negative-tombstone edges. |

## Deleted files

None.

## Out of scope (deliberate)

- No database migration: `last_client_write_id` and `deleted_at` already exist; the negative tombstone uses the existing `drafts` table.
- No prod data cleanup/backfill for already-created ghost drafts. Existing identical stashes still need manual discard or a separate forensic cleanup.
- No content-based dedupe. It risks deleting real user drafts that happen to match the sent message.

## Self-review

Manual pre-PR review focused on the resolve/send lifecycle and own-echo interleavings. The first pass fixed the pending-upsert marker race; the review pass then found the split-id variant, which is why the patch exposes `lastClientWriteId` and queues CAS cleanup for split echoes. Claude Code review later found the never-confirmed delete-before-insert variant with production evidence; that was accepted and fixed by the negative tombstone path in commit `a5152478`.

## Test plan

- [x] `bun scripts/test-silent.ts frontend src/sync/draft-sync.test.ts src/hooks/use-draft-message.test.ts --run` — 90 pass
- [x] `bun scripts/test-silent.ts backend-unit src/features/drafts/repository.test.ts src/features/drafts/service.test.ts` — 19 pass
- [x] `bun run typecheck` — all workspaces clean
- [x] Pre-commit hooks — `bun run lint`, `bun run typecheck`, `bun apps/backend/scripts/generate-api-docs.ts --check`, Dockerfile workspace check, migration check all clean

<details>

## Goal

Stop sent message drafts from reappearing as identical stash entries without weakening the draft system's no-loss split-on-drift rule.

## What was built

- A transactional stale-save guard in `upsertLoadedDraft` for debounced saves that began before send teardown.
- Resolve-op preservation of in-flight upsert markers plus copied `writeId`s.
- Client reconciliation that suppresses/cleans up this device's own sent-write echoes, including rows that arrive under a server split id.
- Backend CAS resolve support for `supersededWriteIds` via the existing `last_client_write_id` column.
- Backend negative tombstones for never-confirmed deletes, plus fresh-upsert dropping when `expectedVersion = 0` collides with such a tombstone.
- Unit coverage for frontend reconciliation and backend resolve/delete predicates.

## Design decisions

- Use `last_client_write_id` as the identity for own sent-write echoes.
- Keep explicit discard as unconditional delete; keep confirmed send cleanup as CAS resolve.
- Make unconditional delete durable on absent rows so mobile request ordering cannot create `version = 1` zombies.
- Keep the solution local to draft reconciliation; no schema change or content dedupe.

## Design evolution

The initial investigation found two races: stale debounced saves and pending-upsert echoes. Self-review found a third form of the second race: the server can split a stale in-flight write under a fresh id, so pending draft id alone is insufficient. Claude Code review then found the never-confirmed delete-before-insert variant that matched production `version = 1` rows. The final version carries write ids through resolve, exposes `lastClientWriteId` on owner-scoped draft wire rows, and plants negative tombstones for delete-before-insert ordering.

## Schema changes

None.

## Not included

- Cleanup of already-created production ghost drafts.
- A merge/dedupe UI for duplicate drafts.
- Garbage collection of draft tombstones.

## Status

Implemented, committed, pushed, and covered by targeted frontend/backend tests plus full typecheck and pre-commit checks.

</details>

---

🤖 _PR by [Codex](https://github.com/openai/codex)_

### DIFF
--- a/apps/backend/src/features/drafts/handlers.ts
+++ b/apps/backend/src/features/drafts/handlers.ts
@@ -39,6 +39,7 @@ const upsertSchema = z
 
 const resolveSchema = z.object({
   expectedVersion: z.number().int().min(1),
+  supersededWriteIds: z.array(z.string().min(1)).optional(),
 })
 
 interface Dependencies {
@@ -106,6 +107,7 @@ export function createDraftsHandlers({ draftsService }: Dependencies) {
         userId,
         id,
         expectedVersion: parsed.data.expectedVersion,
+        supersededWriteIds: parsed.data.supersededWriteIds ?? [],
       })
       res.json(result)
     },
--- a/apps/backend/src/features/drafts/repository.test.ts
+++ b/apps/backend/src/features/drafts/repository.test.ts
@@ -167,7 +167,7 @@ describe("DraftsRepository.casUpdate", () => {
 describe("DraftsRepository.softDeleteCas", () => {
   afterEach(() => mock.restore())
 
-  it("tombstones only when the version matches so a drifted copy survives", async () => {
+  it("tombstones on a matching version or superseded write id so unrelated drift survives", async () => {
     const captured: Captured = { text: null, values: null }
     const db = createQuerier(captured)
 
@@ -176,12 +176,15 @@ describe("DraftsRepository.softDeleteCas", () => {
       userId: "usr_1",
       id: "draft_01",
       expectedVersion: 2,
+      supersededWriteIds: ["write_sent"],
     })
 
     expect(captured.text).toContain("deleted_at = NOW()")
     expect(captured.text).toContain("deleted_at IS NULL")
     expect(captured.text).toContain("version =")
+    expect(captured.text).toContain("last_client_write_id = ANY")
     expect(captured.values).toContain(2)
+    expect(captured.values).toContainEqual(["write_sent"])
   })
 })
 
--- a/apps/backend/src/features/drafts/repository.test.ts
+++ b/apps/backend/src/features/drafts/repository.test.ts
@@ -191,15 +191,20 @@ describe("DraftsRepository.softDeleteCas", () => {
 describe("DraftsRepository.softDelete", () => {
   afterEach(() => mock.restore())
 
-  it("tombstones unconditionally (explicit discard) without a version guard", async () => {
+  it("plants a negative tombstone so a delete that wins the race against first insert still sticks", async () => {
     const captured: Captured = { text: null, values: null }
     const db = createQuerier(captured)
 
     await DraftsRepository.softDelete(db, "ws_1", "usr_1", "draft_01")
 
+    expect(captured.text).toContain("INSERT INTO drafts")
+    expect(captured.text).toContain("ON CONFLICT (id) DO UPDATE")
     expect(captured.text).toContain("deleted_at = NOW()")
     expect(captured.text).toContain("deleted_at IS NULL")
     expect(captured.text).not.toContain("AND version =")
+    expect(captured.values).toContain("draft_01")
+    expect(captured.values).toContain("ws_1")
+    expect(captured.values).toContain("usr_1")
   })
 })
 
--- a/apps/backend/src/features/drafts/repository.ts
+++ b/apps/backend/src/features/drafts/repository.ts
@@ -292,20 +292,24 @@ export const DraftsRepository = {
   },
 
   /**
-   * Unconditional soft-delete for explicit discard (no CAS) — the user threw
-   * the draft away, so drift doesn't matter. Idempotent on an already-deleted
-   * row. Returns the row when it tombstoned, `null` when it was missing.
+   * Unconditional soft-delete for explicit discard and resolve-on-send of a
+   * never-confirmed draft (no CAS). Plants a negative tombstone when the id has
+   * not been seen yet, so a delete that reaches the server before the draft's
+   * first insert still wins the race. Returns the row when this call tombstoned
+   * it (live row transitioned, or negative marker inserted) and `null` when the
+   * row was already tombstoned.
    */
   async softDelete(db: Querier, workspaceId: string, userId: string, id: string): Promise<Draft | null> {
     const result = await db.query<DraftRow>(sql`
-      UPDATE drafts SET
+      INSERT INTO drafts (id, workspace_id, user_id, scope, client_updated_at, deleted_at)
+      VALUES (${id}, ${workspaceId}, ${userId}, '', NOW(), NOW())
+      ON CONFLICT (id) DO UPDATE SET
         deleted_at = NOW(),
         updated_at = NOW(),
-        version = version + 1
-      WHERE id = ${id}
-        AND workspace_id = ${workspaceId}
-        AND user_id = ${userId}
-        AND deleted_at IS NULL
+        version = drafts.version + 1
+      WHERE drafts.workspace_id = ${workspaceId}
+        AND drafts.user_id = ${userId}
+        AND drafts.deleted_at IS NULL
       RETURNING ${sql.raw(COLUMNS)}
     `)
     return result.rows[0] ? mapRow(result.rows[0]) : null
--- a/apps/backend/src/features/drafts/repository.ts
+++ b/apps/backend/src/features/drafts/repository.ts
@@ -223,15 +223,23 @@ export const DraftsRepository = {
   },
 
   /**
-   * CAS soft-delete for resolve-on-send and explicit discard. Tombstones the
-   * row only when `version` still matches, so a copy that drifted since the
-   * send started survives as a stash entry instead of being collaterally
-   * destroyed. Returns the tombstoned row on success, `null` on version drift.
+   * CAS soft-delete for resolve-on-send. Tombstones the row when `version` still
+   * matches, or when its last write id is one this device already superseded by
+   * sending the message. Unrelated drift survives as a stash entry instead of
+   * being collaterally destroyed. Returns the tombstoned row on success, `null`
+   * on version drift.
    */
   async softDeleteCas(
     db: Querier,
-    params: { workspaceId: string; userId: string; id: string; expectedVersion: number }
+    params: {
+      workspaceId: string
+      userId: string
+      id: string
+      expectedVersion: number
+      supersededWriteIds?: string[]
+    }
   ): Promise<Draft | null> {
+    const supersededWriteIds = params.supersededWriteIds ?? []
     const result = await db.query<DraftRow>(sql`
       UPDATE drafts SET
         deleted_at = NOW(),
@@ -241,7 +249,8 @@ export const DraftsRepository = {
         AND workspace_id = ${params.workspaceId}
         AND user_id = ${params.userId}
         AND deleted_at IS NULL
-        AND version = ${params.expectedVersion}
+        AND (version = ${params.expectedVersion}
+          OR (${supersededWriteIds.length > 0} AND last_client_write_id = ANY(${supersededWriteIds})))
       RETURNING ${sql.raw(COLUMNS)}
     `)
     return result.rows[0] ? mapRow(result.rows[0]) : null
--- a/apps/backend/src/features/drafts/repository.ts
+++ b/apps/backend/src/features/drafts/repository.ts
@@ -302,6 +302,7 @@ export const DraftsRepository = {
   async softDelete(db: Querier, workspaceId: string, userId: string, id: string): Promise<Draft | null> {
     const result = await db.query<DraftRow>(sql`
       INSERT INTO drafts (id, workspace_id, user_id, scope, client_updated_at, deleted_at)
+      -- Empty scope is a tombstone-only sentinel; real draft scopes are never blank.
       VALUES (${id}, ${workspaceId}, ${userId}, '', NOW(), NOW())
       ON CONFLICT (id) DO UPDATE SET
         deleted_at = NOW(),
--- a/apps/backend/src/features/drafts/service.test.ts
+++ b/apps/backend/src/features/drafts/service.test.ts
@@ -128,6 +128,25 @@ describe("DraftsService.upsert", () => {
     expect(result.draft.id).toBe("draft_split")
   })
 
+  it("drops a fresh insert landing on a tombstone without splitting", async () => {
+    const service = setupService()
+    const insertIfAbsent = spyOn(DraftsRepository, "insertIfAbsent").mockResolvedValue(null)
+    spyOn(DraftsRepository, "findByIdForUpdate").mockResolvedValue(
+      fakeDraft({ version: 1, lastClientWriteId: null, deletedAt: NOW })
+    )
+    const casUpdate = spyOn(DraftsRepository, "casUpdate")
+    const outbox = spyOn(OutboxRepository, "insert").mockResolvedValue({} as any)
+
+    const result = await service.upsert({ ...baseUpsertParams(), writeId: "write_late", expectedVersion: 0 })
+
+    expect(result.split).toBe(false)
+    expect(result.originalId).toBeUndefined()
+    expect(result.draft.id).toBe(DRAFT_ID)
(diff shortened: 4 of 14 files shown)


kristofferremback:
@coderabbitai review

kristofferremback (commented):
## Review: solid on the confirmed-draft races, but it misses the row variant that's actually stuck in prod

I investigated the same report against the prod DB and landed on a complementary fix ([`claude/draft-reconciliation-bugs-ab05vn`](https://github.com/threahq/threa/tree/claude/draft-reconciliation-bugs-ab05vn)). Sharing the one gap I think this PR leaves open, with evidence.

### What this PR gets right

The `confirmed-draft` (`baseVersion > 0`) echo races are handled well:
- keeping the queued `upsert_draft` as an in-flight echo marker,
- threading `supersededWriteIds` so a lost-ack write can be tombstoned by **write id** even after its version drifts,
- exposing `lastClientWriteId` so the client recognizes its own **split-id** echo.

Those are real races and the approach is sound.

### The gap: the never-confirmed delete-before-insert zombie

The live ghosts in prod are **`version = 1`**, `updated_at == created_at`, with server `created_at` landing **~2s _after_** the message was sent. `version = 1` means only the insert ran (no CAS update) — i.e. the draft was **never confirmed** at resolve time. In prod for this one workspace there were **32** such draft/message pairs (draft inserted within 30s _after_ a byte-identical sent message); 2 still live.

That path is the one this PR deliberately routes around:

1. A never-confirmed send takes `enqueueDraftDelete` (per `syncDraftResolution`, `baseVersion 0`), not the resolve path — so it records **no** `supersededWriteIds` and enqueues a *delete* op, not a *resolve* op.
2. `softDelete` (the unconditional path) is **unchanged** here — only `softDeleteCas` was touched. It's still `UPDATE … WHERE id = …`, so when the delete reaches the server **before** the slow insert it matches **zero rows and vanishes** — no tombstone.
3. The insert then lands and creates a live `version 1` row. The delete op already drained as a no-op; the in-flight continuation that would re-clean it is lost when the phone freezes right after send. Nothing remains to suppress it, so it re-seeds into the stash on every bootstrap.

The write-id echo suppression added here can't catch this case, because that write id was never recorded as superseded (the delete path doesn't carry it) and there's no resolve op for it.

### Suggestion

These two fixes barely overlap and compose cleanly:

| Area | This PR | The complementary change |
| --- | --- | --- |
| `softDeleteCas` + `supersededWriteIds` / split echoes | ✅ | — |
| wire `lastClientWriteId`, frontend echo handling | ✅ | — |
| `softDelete` → negative tombstone (durable on absent row) | — | ✅ |
| `upsert` → drop a fresh (`expectedVersion 0`) insert landing on a tombstone | — | ✅ |

The server-side negative tombstone makes the delete-before-insert zombie **structurally impossible** regardless of request ordering or app lifecycle, which the in-flight-continuation cleanup can't fully guarantee on mobile. I'd suggest folding it in here (or stacking it) so the never-confirmed `v1` rows are covered too. Happy to push it onto this branch or open it stacked — whichever you prefer.

Not blocking — this PR is a net improvement on its own; just flagging that it doesn't fully close the specific rows in the report.

---

kristofferremback:
Accepted and folded into this PR in `a5152478`.

The additional race is real: a never-confirmed draft (`baseVersion = 0`) can send/delete before the first draft insert reaches the backend, and the old unconditional `softDelete` could match zero rows and leave no durable marker. The follow-up commit changes `DraftsRepository.softDelete` to plant a negative tombstone with `INSERT ... ON CONFLICT DO UPDATE`, and `DraftsService.upsert` now drops a fresh `expectedVersion = 0` upsert that collides with a tombstoned id instead of splitting it into a live zombie.

Added/updated coverage:
- backend draft repository/service tests now cover the negative tombstone and fresh-upsert drop path — 19 pass
- PR description updated with the never-confirmed delete-before-insert variant and design evolution

🤖 _Response by [Codex](https://github.com/openai/codex)_

kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (no blocking issues; a few non-blocking notes)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3 (spec+design, correctness+data-flow, mobile/lifecycle)

Reviewed the combined PR (Codex's confirmed-draft echo/split work + the folded-in negative-tombstone fix, commit `a5152478`). The reconciliation is sound — I traced the high-risk sequences end-to-end and they hold. **No blocking issues.** Three non-blocking notes below, plus what was verified.

No findings cleared the ≥80 bar after verification. Two reviewers independently flagged `hasPendingSupersededWriteId` and both proposed scoping it by `draft.id` — that fix is **incorrect** and would reintroduce a bug (see Note 1), so it's deliberately not recommended.

---

### Non-blocking notes

**1. `hasPendingSupersededWriteId` is intentionally unscoped — add a comment so it stays that way.**
The scan over *all* `resolve_draft` ops (not filtered by `draft.id`) is **correct by design**: a split echo arrives under a freshly-minted server id whose own resolve op doesn't exist yet, so the superseded `writeId` must be matched against the *original* draft's resolve op. A future "optimization" that filters by `draft.id` would silently rebreak split-ghost cleanup. The only residual risk is a cross-draft false match, which requires a `write_${Date.now()}_${random}` collision across devices — astronomically improbable. Recommend a one-line comment stating the scan is intentionally global and why, to protect it from a well-meaning refactor.
https://github.com/threahq/threa/blob/a5152478a06041293cd94fd3b1c131e2f3f51c85/apps/frontend/src/sync/draft-sync.ts#L254-L259

**2. `scope = ''` negative-tombstone sentinel is load-bearing but undocumented.**
`softDelete`'s `INSERT ... ON CONFLICT` plants a tombstone with `scope = ''` and null content. It never surfaces — `listByUser` and `rescopeByScope` both filter `deleted_at IS NULL` — so this is safe today. But the empty-string scope is a reserved sentinel with no schema enforcement; a one-line comment in `softDelete` noting it's a content-less tombstone marker (never a real scope) would prevent a future regression.
https://github.com/threahq/threa/blob/a5152478a06041293cd94fd3b1c131e2f3f51c85/apps/backend/src/features/drafts/repository.ts#L300-L315

**3. Residual (pre-existing, multi-device only): split-ghost cleanup still leans on a non-durable continuation.**
When a confirmed draft's in-flight push splits server-side, the cleanup that resolves the new split id is enqueued in `executeDraftUpsert`'s post-ack continuation. If the JS context is frozen/killed between the HTTP ack and that enqueue *and* a second device concurrently advanced the split, the split row can briefly survive until a bootstrap lands while the original resolve op (carrying the superseded `writeId`) is still queued. This is **not introduced by this PR** — the prior `enqueueDraftDelete(res.draft.id)` had the same continuation dependency and was weaker — and the new `supersededWriteIds` bootstrap match actually *narrows* the window. Flagging only as a known residual for a future durable-cleanup pass; not for this PR.

---
<details><summary>✅ Verified sound (traced sequences)</summary>

- **Never-confirmed delete-before-insert (`version 1` zombie):** negative tombstone + `expectedVersion 0`-onto-tombstone drop closes it for every interleaving of {insert lands, delete lands, app frozen}. The durable `delete_draft` op replays on cold start and the tombstone makes the late insert lose regardless of ordering.
- **Confirmed in-flight upsert + send:** `enqueueDraftResolve` captures the writeId while the upsert op is still in IDB; `softDeleteCas`'s `last_client_write_id = ANY(...)` arm tombstones this device's lost-ack write, and the inbound echo at `N+1` is suppressed by the version guard (`draft.version <= pendingResolveVersion`). No-loss preserved: another device's drift (different `last_client_write_id`) is not matched.
- **Kept upsert op:** on replay `executeDraftUpsert` reads the torn-down row, finds it gone, and no-ops — it cannot re-push a server row (queue is sequential; op deleted only after the continuation returns).
- **Split echo:** caught by `ownSentWrite` and CAS-resolved at the split row's version 1; idempotent if the continuation also enqueued one.
- **`scope=''` tombstone:** never returned by `listByUser` / `rescopeByScope`; `draft:deleted` publish on a freshly-planted marker is harmless (no local row on other devices).
- **SQL:** `ANY($supersededWriteIds)` is parameterized; `expectedVersion >= 1` schema guard keeps the never-confirmed path off the resolve endpoint.
</details>

<details><summary>📐 Plan Adherence [CLEAN]</summary>The `supersededWriteIds` path is wired end-to-end (handler schema → service → repo predicate → frontend op payload → queue replay → applyDraftUpserted/bootstrap), and `lastClientWriteId` is added to both the wire view and the `Draft` type. No missing corresponding change.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No ≥80 logic/null/race defects introduced. The reordered `dirty` check in `applyDraftUpserted` (now before the no-local-row accept) is the intended ghost-suppression and is consistent with the kept-upsert-marker design.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Field round-trips verified across create/confirm/split/resolve/delete. Negative-tombstone row is inert in every read path.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-20 (race-safe upsert) satisfied by the `ON CONFLICT` tombstone; no migration needed (existing columns), consistent with INV-17.</details>
<details><summary>🏗️ Design [2 notes]</summary>Notes 1 and 2 above — both documentation/guard-rail suggestions, no behavior change.</details>
<details><summary>🔒 Security [CLEAN]</summary>Parameterized array predicate; owner/workspace-scoped writes; no injection or exposure.</details>
<details><summary>📱 Mobile [1 residual]</summary>Note 3 above — pre-existing, multi-device, narrowed by this PR.</details>

---

kristofferremback:
Followed up on the unified-review notes in `5ced2b55`.

Disposition:
- **Accepted note 1:** added a guardrail comment explaining that `hasPendingSupersededWriteId` intentionally scans all pending resolves because split echoes arrive under a fresh draft id.
- **Accepted note 2:** added a guardrail comment explaining that `scope = ''` is a tombstone-only sentinel, never a real draft scope.
- **Acknowledged note 3:** leaving the residual multi-device durable-cleanup edge out of this PR. It is pre-existing, this PR narrows it, and fixing it properly needs a separate durable continuation/cleanup design.

Verification:
- `bun scripts/test-silent.ts backend-unit src/features/drafts/repository.test.ts src/features/drafts/service.test.ts` — 19 pass
- pre-commit lint/typecheck/API-doc/migration/Dockerfile checks — clean

🤖 _Response by [Codex](https://github.com/openai/codex)_

PR #1009, opened 2026-06-19
fix(timeline): fetch and re-anchor when date jump target is out of window

kristofferremback:
## Problem

Jumping to a date in a stream timeline (the floating date pill → preset or calendar) usually did nothing useful: it relocated the scroll to the *oldest already-loaded* message instead of fetching and landing on the chosen day.

The fast path in `handleJumpToDate` (`apps/frontend/src/components/timeline/stream-content.tsx`) decided "is the target already loaded?" with:

```js
events.find(event => localStartOfDayMs(new Date(event.createdAt)) >= targetDayMs)
```

`find` returns the *first* event satisfying `day >= target`. When you're parked at the live tail and jump weeks back, **every** loaded message satisfies `day >= target`, so the match is `events[0]` — the oldest loaded row. The handler scrolled there and returned, so the out-of-window fetch path almost never ran. From the user's seat the jump "just nudges the scroll up a little" and never relocates.

`#1008` (already in `main`) wired up the post-jump scroll driver, the shift-baseline reset, and stale-target clearing — but left this false-positive in the guard, so all that machinery sat unused for the common case.

## Solution

Only take the in-window scroll when the loaded window genuinely **straddles** the target day; otherwise fall through to the existing fetch-and-re-anchor path.

### How it works

The decision moves into a pure, exported `resolveDateJumpAnchor({ events, targetDayMs, hasOlderEvents })` that returns the message id to scroll to, or `null` to mean "fetch". It walks events in ascending (chronological) order and returns the in-window anchor only when:

- a loaded message sits **strictly before** the target day (`sawEarlierMessage`) — so the first on-or-after match is the genuine anchor with nothing earlier missing; or
- there is **no older history left to fetch** (`hasOlderEvents === false`) — the window already starts at the beginning of the stream.

Otherwise it returns `null`, and `handleJumpToDate` calls `jumpToEventByDate` (`apps/frontend/src/hooks/use-events.ts`), which fetches a window of surrounding messages around the day's first message via `getEventsAroundDate`, enters jump mode (bidirectional pagination — `hasOlder`/`hasNewer`, separate older/newer infinite queries), and re-anchors the scroll on the server-resolved `anchorMessageId` through the existing `pendingScrollTarget` + `pendingScrollRequestVersion` driver. So after the jump, scrolling up *and* down keeps loading history.

### Key design decisions

**1. Gate on "window straddles the day", not "any loaded message is newer".** The straddle test is exactly the condition under which `find`'s first match is provably the earliest on-or-after-target message: the preceding loaded message is strictly before the day, and the window is contiguous, so nothing earlier can be missing. The two opposite failure modes the gate closes: a target older than the whole window (every row matches → wrong oldest-row anchor) versus a target genuinely inside the window (correct, no fetch). The `!hasOlderEvents` branch covers the start-of-stream case where the first row legitimately has no predecessor.

**2. Extract a pure helper rather than inline the logic.** Matches the sibling timeline predicates (`classifyDeepLinkScrollTick`, `shouldPrefetchOlderHistory`, `computeScrollEdges`) that already live as exported pure functions with unit coverage in `stream-content.test.ts`. The scroll/fetch wiring itself is untestable without a live virtualizer; the decision is not, so it's the part worth isolating.

**3. Reuse the existing fetch + re-anchor machinery untouched.** `jumpToEventByDate`, jump mode, and the post-jump driver are the same paths deep-links and search already use. The fix is purely making the guard fall through to them at the right time — no new fetch or scroll code.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Add pure `resolveDateJumpAnchor` helper; rewrite `handleJumpToDate` to use it (replaces the `events.find` false-positive); add `hasOlderEvents` to the callback deps |
| `apps/frontend/src/components/timeline/stream-content.test.ts` | 7 unit tests for `resolveDateJumpAnchor`: straddle, live-tail (regression), start-of-stream, forward-jump, non-message events, companion-response anchor, missing-messageId |

## Out of scope (deliberate)

- The fetch/jump-mode/re-anchor machinery and the date pill UI (`stream-date-header.tsx`) are unchanged — the bug was the guard in front of them, not the machinery itself.
- No backend change. `getEventsAroundDate` / `findFirstMessageOnOrAfter` already resolve the correct anchor and surrounding window.

## Test plan

- [x] `bunx vitest run src/components/timeline/stream-content.test.ts` — 31 pass (7 new)
- [x] `bunx tsc --noEmit -p apps/frontend/tsconfig.json` — clean
- [ ] No live virtualizer harness, so the scroll/fetch wiring is exercised only through the pure `resolveDateJumpAnchor` helper (as with its sibling predicates) — the DOM-level jump is not integration-tested

---

---
_Generated by [Claude Code](https://claude.ai/code/session_018c7uzs8qTSrmyUaojB5NHV)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1009"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784423593&installation_id=129946197&pr_number=1009&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1009&signature=544e1727173eead5afdf425e06c7ea71cdfa308dbec3b77cdf41a8f0e2e17958"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/stream-content.test.ts
+++ b/apps/frontend/src/components/timeline/stream-content.test.ts
@@ -5,7 +5,9 @@ import {
   computeScrollEdges,
   shouldPrefetchOlderHistory,
   shouldShowOlderSkeletons,
+  resolveDateJumpAnchor,
 } from "./stream-content"
+import { localStartOfDayMs } from "@/lib/dates"
 
 const DEADLINE = 4000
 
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (one minor, non-blocking note)

The changed logic is clean: correctness, design, CLAUDE.md compliance, plan adherence, security, and mobile all came back with no findings. One sub-threshold UX observation is recorded below as a non-blocking note.

**Highlights from the review**
- The straddle guard (`sawEarlierMessage` + `hasOlderEvents`) was traced against all jump sequences — live-tail→old-date (the regression), target-inside-window, start-of-stream with no older history, forward-jump past the window, and missing-`messageId` payload — and is correct in each. When in doubt it errs toward fetching, which is always safe.
- Adding `hasOlderEvents` to the `useCallback` deps fixes what was a latent stale-closure read in the prior code, not just churn.
- Nulling `pendingScrollTarget.current` on the fast path cleanly supersedes any in-flight post-jump driver loop.

---

All stated deliverables present: `resolveDateJumpAnchor` helper, rewritten `handleJumpToDate`, corrected `useCallback` deps (incl. `hasOlderEvents`), fall-through to `jumpToEventByDate` (fetch + bidirectional jump mode + re-anchor), and 7 unit tests covering the decision branches.
</details>

No logic errors, off-by-one, null hazards, stale-closure, or missing-dep issues. The non-message-event-before-target case is mildly conservative (may fetch when a direct scroll was possible) but always safe, and is exercised by a test.
</details>
<details><summary>🔁 Data Flow — CLEAN</summary>

`pendingScrollTarget` / `pendingScrollRequestVersion` correctly drive the post-jump re-anchor; the fall-through path enters jump mode so scroll-up and scroll-down both keep paginating (the user's explicit requirement).
</details>

The `resolveDateJumpAnchor` JSDoc explains *why* the naïve `events.find(day >= target)` is wrong (a load-bearing correctness rationale, INV-25-compliant), not change-history. `toast.info` for the empty-result case is informational, not a happy-path success toast (INV-63 OK). No nested ternaries, dead code, or speculative additions.
</details>

Extracting a pure, exported helper matches the sibling timeline predicates (`classifyDeepLinkScrollTick`, `shouldPrefetchOlderHistory`). The fetch/scroll machinery is reused untouched.
</details>

No DOM-injection paths touched; logic-only diff.
</details>
<details><summary>🎛️ UX — 1 note (below blocking threshold)</summary>

`apps/frontend/src/components/timeline/stream-content.tsx` (`handleJumpToDate` fetch path) — **No loading indicator while the fetch is in flight (score 75, non-blocking).** When `resolveDateJumpAnchor` returns `null`, the popover has already closed and the timeline sits unchanged during `await jumpToEventByDate(...)` — no spinner, skeleton, or disabled state on the pill. For a far-back date on a slow connection this can feel like the tap was dropped. This is pre-existing (the date header and fetch path are untouched here), but the fix *amplifies* it: the fetch path now actually runs for out-of-window dates where it previously almost never did. A fix (thread an `isJumpFetching` flag into `StreamDateHeader` to show a spinner) is a small follow-up, out of scope for this minimal bug fix.

https://github.com/threahq/threa/blob/e58ca6faac15e02cf0fc0fd17a7a3f0502a800ae/apps/frontend/src/components/timeline/stream-content.tsx#L1300-L1322
</details>
<details><summary>📱 Mobile — CLEAN</summary>

Logic-only diff; no JSX/CSS/breakpoint/touch surface changed. The date pill's touch-scroll forwarding and the refine loop's `touchmove` abort are pre-existing and unmodified.
</details>

---

PR #1011, opened 2026-06-19
feat(frontend): External agent state in the scratchpad header pill

kristofferremback:
## Problem

A scratchpad's header pill had two states — **Companion** (Sparkles, Ariadne replies) and **Quiet** (Moon, silent capture) — driven solely by `stream.companionMode`. An external agent attached to a scratchpad (a Pi remote bot runtime, granted channel access via `getBotMemberIds`) runs with companion mode off, so it showed up as **Quiet** — indistinguishable from a scratchpad with no agent at all. Its real signal (is the runtime connected?) lived only in `ActiveBotStatusStrip`, a separate floating pill rendered `absolute inset-x-4 top-2` over the timeline — exactly where the jump-to-date pill (`StreamDateHeader`) floats, so the two overlapped.

## Solution

Fold a simplified version of the runtime-status strip into the existing header pill as a third state, and delete the floating strip.

### How it works

- **New `useActiveBotPresence(workspaceId, streamId)` hook** (`apps/frontend/src/hooks/use-active-bot-presence.ts`) resolves the bot runtime attached to a scratchpad. It's a cache-only observer (`useQuery` with `enabled: false`, `staleTime: Infinity`, `queryFn` reading `queryClient.getQueryData`) over `streamKeys.bootstrap` — the same query `StreamContent` populates and the `bot_runtime:presence` socket handler keeps fresh via `setQueryData` (`stream-sync.ts`). So the header reflects live connection changes without issuing its own fetch or room-join. The selection logic (prefer a bot id with presence, else the first member id; require the bot to exist in `useWorkspaceBots`) is moved verbatim from the deleted `StreamContent` memo.
- **The header pill** (`apps/frontend/src/pages/stream.tsx`) gains an **External** branch: a connection dot — `bg-emerald-500` when `presence.status` is `available`/`busy` (`BotRuntimeStatuses.AVAILABLE`/`BUSY`), `bg-muted-foreground/40` otherwise — plus the label "External". Priority is `Encrypted → External → Companion → Quiet`.
- **The floating `ActiveBotStatusStrip`** and its computation in `StreamContent` are removed; the component file is deleted (INV-38).
- **Encrypted collapses to a bare lock** (no "Encrypted" label, no chevron, `px-1.5` instead of `px-2.5`) to reclaim header width, per the follow-up request.

### Key design decisions

**1. Encrypted keeps top priority over External.** An E2E scratchpad's only *persistent* encryption signal is this pill: `StreamHeaderEncryptionAction` renders an unlock/setup CTA only while `session.status` is `locked`/`no-key`/`unlocking` and renders *nothing* once unlocked (`useStreamEncryptionAction` returns null). Demoting Encrypted below External would drop that signal for an unlocked encrypted pad that also has an external agent. Trade-off: an external agent on an *encrypted* scratchpad reads as the lock, not "External" (the agent stays reachable via the pill's settings popover). Non-encrypted scratchpads — the Pi-remote case that motivated this — now read "External". This is the conservative, non-regressive ordering; flagged to the user.

**2. Cache-only observer, not a second live query.** `StreamContent` (rendered under the same page) owns the real `useStreamBootstrap` with its `enabled` gating and room-join. The header must not duplicate that, so the hook observes the cache rather than re-fetching — the documented TanStack v5 cache-only pattern in the root `CLAUDE.md`.

**3. Simplified strip.** The old strip showed bot name + status copy + the runtime's project path with RTL truncation. The request was "a simplified version … green or grey status and says external", so the header pill carries only the dot + "External"; the connection detail lives in the button's `aria-label`.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/hooks/use-active-bot-presence.ts` | Cache-only observer resolving the external agent + its runtime presence for a scratchpad |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/pages/stream.tsx` | Add External pill state (connection dot + label); collapse Encrypted to an icon-only lock; call `useActiveBotPresence` |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Remove the floating strip render, its `activeBotPresence` memo, and the now-unused `useWorkspaceBots` import |
| `apps/frontend/src/hooks/index.ts` | Export `useActiveBotPresence` / `ActiveBotPresence` |

## Deleted files

| File | Reason |
| ---- | ------ |
| `apps/frontend/src/components/timeline/active-bot-status-strip.tsx` | Replaced by the External pill state; the only importer was `StreamContent` (INV-38) |

## Out of scope (deliberate)

- **Sidebar scratchpad decoration** (`scratchpad-item.tsx`) still shows only Lock/Sparkles — no "external" dot. Not part of the request; the header is where the live status was crowding the date pill.
- **Encrypted + external combined state** — not added (see decision 1).

## Test plan

- [x] `bunx tsc --noEmit` on `apps/frontend` clean; full-monorepo typecheck clean via the pre-commit hook on both commits
- [x] `bun run test src/sync/stream-sync.test.ts` — 50 pass (covers the `bot_runtime:presence` → `setQueryData` path the hook observes)
- [ ] No unit harness for the header pill — it's inline in the `StreamPage` component, and no existing test mounts it, so the External branch is not covered by a dedicated render test. The presence-resolution and cache-update paths it depends on are tested above.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01ViroRxscGrGpMJVGvqw8wb)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1011"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784436690&installation_id=129946197&pr_number=1011&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1011&signature=4aad3f60b1400cd2a105e067c25641ef95629debc256ba68fbe9ad48a28e805d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review
**Confidence: 5/7** — Good (one blocking bug found and fixed in-review; remaining items are non-blocking tradeoffs tied to explicit product decisions)

Found 6 issues — **2 fixed in commit `2245a2a`**, 4 surfaced for a product call (all direct consequences of the requested header minimization).

**Fixed in-review:**

1. `apps/frontend/src/pages/stream.tsx` — **Rules-of-Hooks violation (bug).** `useActiveBotPresence` was called *below* the `!workspaceId/!streamId` null-guard and the `streamError` early return, so a render that hit either return would call fewer hooks than a normal render → React throws "Rendered more hooks than during the previous render" and crashes on any stream 404/403. Hoisted the call above the early returns alongside the other hooks (it already tolerates undefined args).
   https://github.com/threahq/threa/blob/2245a2a4cf17d1a87466c9f8eba6983896af2df2/apps/frontend/src/pages/stream.tsx#L164-L165

2. `apps/frontend/src/pages/stream.tsx` — **INV-25** ("Change narration … belongs in the commit message, never in the code"). The pill comment said "Replaces the old floating status strip" — narration of deleted code. Trimmed to the surviving rationale.

**Surfaced for a product call** (each follows directly from "minimize the encryption hint to just the lock" and "green/grey status + says External" — flagging the tradeoffs, not asserting they're wrong):

3. `apps/frontend/src/pages/stream.tsx` — **Encrypted pill discoverability (ux-affordance, 85).** Encrypted now renders as a bare lock with no label and no chevron. Every other state keeps the chevron as the "this is a dropdown" cue; the lock looks like a static badge, so the companion/tool-settings popover it still opens is no longer visibly discoverable (only `cursor-pointer`/hover hint at it).
   https://github.com/threahq/threa/blob/2245a2a4cf17d1a87466c9f8eba6983896af2df2/apps/frontend/src/pages/stream.tsx#L372-L378

4. `apps/frontend/src/pages/stream.tsx` — **Encrypted tap target on touch (mobile-touch, 85).** Icon-only at `px-1.5 py-0.5` around a 12px lock ≈ 24×16px — below the 44×44 guideline. The other states stay wide via their label. (Pill height is shared across all states and pre-existing; only the Encrypted width regressed.) A transparent expanded hit-area would keep the compact look without re-widening the bar.
   https://github.com/threahq/threa/blob/2245a2a4cf17d1a87466c9f8eba6983896af2df2/apps/frontend/src/pages/stream.tsx#L408-L432

5. `apps/frontend/src/pages/stream.tsx` — **External connection state is color-only (accessibility, 82).** Connected vs not is conveyed solely by the green/grey dot for sighted users; the label is always "External". `aria-label` covers screen readers, but the old strip also carried the status as *text* ("Available"/"Not connected"), so for color-vision-deficient users this is a minor regression. A filled-vs-ring dot would satisfy WCAG 1.4.1 without layout change. This matches the explicit "green/grey status" ask, so flagging only as a nicety.
   https://github.com/threahq/threa/blob/2245a2a4cf17d1a87466c9f8eba6983896af2df2/apps/frontend/src/pages/stream.tsx#L380-L389

6. `apps/frontend/src/components/timeline/active-bot-status-strip.tsx` (deleted) — **bot name + runtime path no longer surfaced (ux-feedback, 80).** The strip showed the bot name and the runtime project path (RTL-truncated); the External pill shows only "External", and `LiveAgentSettings` doesn't surface either. Power users who run multiple Pi sessions lose the only place that disambiguated *which* runtime/path. Deliberate per the "simplified" ask; noted in the PR's Out-of-scope.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>All three asks delivered (External state, simplified status folded into the header, floating strip removed) plus the follow-up (Encrypted → lock-only). No missing corresponding changes.</details>
<details><summary>🔍 Bugs [1 — fixed]</summary>Rules-of-Hooks ordering (#1), fixed in `2245a2a`.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Presence flows socket → `setQueryData(streamKeys.bootstrap)` → cache-only observer → memo → pill. Verified: bot disconnect flips the dot grey but keeps "External"; bot removed from `botMemberIds` falls back to Quiet/Companion. Draft scratchpads (no bootstrap row) return null safely.</details>
<details><summary>📋 CLAUDE.md Compliance [1 — fixed]</summary>INV-25 change-narration comment (#2), fixed in `2245a2a`. Cache-only observer pattern correctly applied (CodeRabbit concurred). INV-35/37 satisfied — the hook consolidates the previously-inline `StreamContent` memo rather than duplicating it. INV-38 satisfied — strip + dead imports removed.</details>
<details><summary>🏗️ Design [CLEAN]</summary>No leaky abstractions or parallel implementations. `useMemo` deps are correct (`setQueryData` returns a new presence object only on change).</details>
<details><summary>🔒 Security [CLEAN]</summary>Presentational frontend diff; no injection/XSS/auth surface.</details>
<details><summary>🎛️ UX [3 — see #3, #5, #6]</summary>All consequences of the requested minimization; surfaced for a product call.</details>
<details><summary>📱 Mobile [1 — see #4]</summary>Encrypted tap target; consequence of the lock-only request.</details>

---

PR #1014, opened 2026-06-19
feat(timeline): persist the unread divider for the reading session

kristofferremback:
## Problem

The "New" unread divider in a stream's timeline was transient. `useUnreadDivider`
(`apps/frontend/src/hooks/use-unread-divider.ts`) showed the line for 3s, faded it
to `opacity-0` over 500ms, then unmounted it. A second effect also cleared it the
moment the stream was marked read — and `useAutoMarkAsRead` fires ~500ms after the
stream is viewed, so in practice the line often vanished before its own fade timer
ran. The result: no durable marker of where the current reading session started.
Other chat apps keep that line pinned until you leave the stream.

## Solution

Turn the divider into a persistent, in-place marker for the reading session: it
latches at the first-unread position, stays red for a few seconds, settles to muted
gray, and keeps reserving its row until the user switches streams. It transitions
color rather than opacity, so it never unmounts and never shifts layout (INV-21).

### How it works

1. **Latch in render, per stream.** A `latchRef` (`{ streamId, eventId }`) resets
   when `streamId` changes and then captures the first non-empty `firstUnreadEventId`.
   `displayedUnreadId` reads straight off that ref, so the divider stays pinned to
   its event for the whole session even after `useAutoMarkAsRead` clears the live
   unread. `displayedUnreadId` is threaded through `stream-content.tsx` (as
   `firstUnreadEventId`) → `event-list.tsx` to place the row.
2. **Dim, don't fade.** A single effect keyed on `displayedUnreadId` resets `isDimmed`
   to false on (re)latch and arms one 3s timer to flip it true. `UnreadDivider`
   switches `text-destructive` → `text-muted-foreground` under `transition-colors
   duration-500`, with the rules using `border-current` to follow the text color. No
   `opacity-0`, no unmount.
3. **Gate on a resolved read position.** `firstUnreadEventId` returns `undefined`
   until `readStateResolved` is true; `stream-content.tsx` passes
   `idbStream !== undefined || membership !== undefined`. Until a read source has
   hydrated, `lastReadEventId` is not authoritative, and treating the first message
   as unread would latch a divider that then sticks.
4. **Reset on stream change.** The render-phase ref reset clears the latch on
   `streamId` change; the dim effect (keyed on `displayedUnreadId`) clears `isDimmed`.
   Re-entering a now-read stream shows nothing — the latch only re-arms when a fresh
   `firstUnreadEventId` is computed for the new stream's events.

### Key design decisions

**1. Latch in render, not in an effect.** The first version latched/reset in
`useEffect`s. Switching to a stream that already has unread changes `streamId` and
`firstUnreadEventId` in the same commit; the latch effect ran before the reset effect
(saw a stale guard) and then never re-fired, so the new stream's divider never showed.
Doing the latch in render via a ref (the same pattern as the existing `deepLinkedRef`
in this file) is immune to effect ordering. (Found in self-review + flagged by
CodeRabbit; commit 2.)

**2. Gate the unread computation on `readStateResolved`.** With the divider now
persistent there is no clear-on-read to undo a wrong guess, so a transient where
events have loaded but the read position has not would latch the first message and
strand it. The gate withholds any unread decision until membership or the stream row
has hydrated. (Root cause of the browser-test failure; commit 3.)

**3. Transition color, not opacity.** Fading to `opacity-0` and unmounting both
removed the line's physical presence; animating `color`/`border-color` keeps the row
reserved, so the red→gray settle causes no layout shift (INV-21).

**4. Rename `isFading` → `isDimmed` (and `isDividerFading` → `isDividerDimmed`).** The
prop no longer means "fading toward gone" but "settled to its gray resting state" — a
real semantic change, renamed through `event-list.tsx`/`stream-content.tsx` rather
than left as a misleading alias (INV-25, INV-49).

## Design evolution (self-review + CI)

- **Commit 1** shipped the persistent divider with an effect-based latch.
- **Commit 2** replaced it with a render-phase latch after the effect ordering left
  the divider blank when switching into a stream that already had unread; also fixed
  the reaction E2E, which marked read via a raw server-only API call that never
  updated client state (it passed before only because the old divider auto-dismissed).
- **Commit 3** added the `readStateResolved` gate — the actual cause of the browser
  `evaluate-results` failure was the divider latching the first message during the
  read-position hydration window on stream open.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-unread-divider.ts` | Render-phase per-stream latch; `readStateResolved` gate on the unread computation; dim timer keyed on the latched id; `isFading` → `isDimmed` |
| `apps/frontend/src/components/timeline/unread-divider.tsx` | Color transition (`text-destructive` → `text-muted-foreground`, `border-current`, `transition-colors`) instead of opacity fade; `isFading` → `isDimmed` prop |
| `apps/frontend/src/components/timeline/event-list.tsx` | `isDividerFading` → `isDividerDimmed` through props/context/memo equality; pass `isDimmed` to `UnreadDivider` |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Rename to `isDividerDimmed`; pass `readStateResolved: idbStream !== undefined || membership !== undefined` to the hook |
| `apps/frontend/src/hooks/use-unread-divider.test.ts` | Persist-then-dim, read-state-resolved gate, re-latch-on-switch, already-read-stream, and the two existing deep-link/scroll cases |
| `tests/browser/message-reactions.spec.ts` | "reaction should not trigger unread divider": read through the app (await the client's own `/read` POST) instead of a raw server-only read |

## Out of scope (deliberate)

- **Scroll/snap behavior on open.** The follow-up Kris flagged — don't force-scroll
  past the first unread message on stream open (Snapchat-style) — is untouched.
  `useScrollToElement` wiring is unchanged. Tracked as a separate change.

## Test plan

- [x] `bunx vitest run src/hooks/use-unread-divider.test.ts` — 6 pass (read-state gate, persist-then-dim, re-latch on switch, already-read-stream, two deep-link/scroll cases)
- [x] `bunx vitest run src/components/timeline/event-list.test.ts` — 46 pass (covers the `isDividerDimmed` context-equality path)
- [x] Full monorepo `tsc --noEmit` + ESLint clean (pre-push hook across all apps/packages)
- [x] Browser E2E suite green in CI on `419a1da` — all 4 shards + `evaluate-results` pass, including `message-reactions.spec.ts:257`
- [ ] No manual browser pass in this session — the red→gray timing and no-layout-shift settle are asserted via the hook test + the `transition-colors` change and the E2E, not visually confirmed by hand

---

### DIFF
--- a/apps/frontend/src/components/timeline/event-list.tsx
+++ b/apps/frontend/src/components/timeline/event-list.tsx
@@ -595,9 +595,8 @@ function TimelineItemContentImpl({ item, ctx, deferSecondaryHydration }: Timelin
     }
   }
 
-  return (
+  const rowContent = (
     <>
-      {showUnreadDivider && <UnreadDivider isDimmed={ctx.isDividerDimmed} />}
       {item.type === "day_divider" && <DayDivider dayStartMs={item.dayStartMs} />}
       {item.type === "command_group" && (
         <div className="px-3 sm:px-6">
@@ -644,6 +643,17 @@ function TimelineItemContentImpl({ item, ctx, deferSecondaryHydration }: Timelin
         </div>
       )}
       {eventNode}
+    </>
+  )
+
+  return (
+    <>
+      {showUnreadDivider && <UnreadDivider isDimmed={ctx.isDividerDimmed} />}
+      {/* The first-unread row gets extra top padding so the absolutely-positioned
+          divider has room to breathe above the message instead of crowding it.
+          Only wrap when the divider shows so every other row keeps its spacing
+          and DOM shape. */}
+      {showUnreadDivider ? <div className="pt-3">{rowContent}</div> : rowContent}
     </>
   )
 }
--- a/apps/frontend/src/components/timeline/event-list.tsx
+++ b/apps/frontend/src/components/timeline/event-list.tsx
@@ -36,7 +36,7 @@ interface EventListProps {
   streamId: string
   highlightMessageId?: string | null
   firstUnreadEventId?: string
-  isDividerFading?: boolean
+  isDividerDimmed?: boolean
   agentActivity?: Map<string, MessageAgentActivity>
   /** Hide session group cards (used in channels where responses go to threads) */
   hideSessionCards?: boolean
@@ -516,7 +516,7 @@ export interface TimelineItemRenderContext {
   streamId: string
   highlightMessageId?: string | null
   firstUnreadEventId?: string
-  isDividerFading?: boolean
+  isDividerDimmed?: boolean
   agentActivity?: Map<string, MessageAgentActivity>
   hideSessionCards?: boolean
   newMessageIds?: Set<string>
@@ -597,7 +597,7 @@ function TimelineItemContentImpl({ item, ctx, deferSecondaryHydration }: Timelin
 
   return (
     <>
-      {showUnreadDivider && <UnreadDivider isFading={ctx.isDividerFading} />}
+      {showUnreadDivider && <UnreadDivider isDimmed={ctx.isDividerDimmed} />}
       {item.type === "day_divider" && <DayDivider dayStartMs={item.dayStartMs} />}
       {item.type === "command_group" && (
         <div className="px-3 sm:px-6">
@@ -731,7 +731,7 @@ export function timelineRowPropsEqual(prev: TimelineItemContentProps, next: Time
     p.workspaceId !== n.workspaceId ||
     p.streamId !== n.streamId ||
     p.hideSessionCards !== n.hideSessionCards ||
-    p.isDividerFading !== n.isDividerFading ||
+    p.isDividerDimmed !== n.isDividerDimmed ||
     p.onAbortResearch !== n.onAbortResearch
   ) {
     return false
@@ -802,7 +802,7 @@ export function EventList({
   streamId,
   highlightMessageId,
   firstUnreadEventId,
-  isDividerFading,
+  isDividerDimmed,
   agentActivity,
   hideSessionCards,
   newMessageIds,
@@ -885,7 +885,7 @@ export function EventList({
     streamId,
     highlightMessageId,
     firstUnreadEventId,
-    isDividerFading,
+    isDividerDimmed,
     agentActivity,
     hideSessionCards,
     newMessageIds,
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -1490,6 +1490,10 @@ export function StreamContent({
     streamId,
     isLoading,
     highlightMessageId,
+    // Until a read source has hydrated, lastReadEventId is not authoritative;
+    // gate the unread computation so a not-yet-known read position can't latch
+    // a divider on the first message (which would then stick for the session).
+    readStateResolved: idbStream !== undefined || membership !== undefined,
   })
 
   const queryClient = useQueryClient()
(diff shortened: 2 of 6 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (minor, non-blocking)

Found 2 issues, both minor:

1. `apps/frontend/src/hooks/use-unread-divider.ts:60` — `readStateResolved` defaults to `true`, so the hydration gate is opt-in. The one real caller (`stream-content.tsx:1496`) passes it, but a future call site (thread overlay, popout) that omits it silently bypasses the guard and can latch a divider on the first message before the read position resolves — exactly the sticky-divider bug the gate was added to prevent (design: fail-safe direction inverted). Low impact today (single caller); worth considering `= false` + an explicit `true` at the call site.
   https://github.com/threahq/threa/blob/419a1da835e7b3a983fa732864268e9de69b550c/apps/frontend/src/hooks/use-unread-divider.ts#L57-L63

2. `apps/frontend/src/components/timeline/unread-divider.tsx:16-18` — Comment leans toward restatement (INV-25 [Maintainability]: *"a comment that says in prose what the next line already says in code … the first to rot"*). "transitions color (red → muted) rather than opacity" restates the `transition-colors` / `text-*` classes; the load-bearing part is the INV-21 "keeps reserving its row / never shifts layout" rationale. Trim to just that.
   https://github.com/threahq/threa/blob/419a1da835e7b3a983fa732864268e9de69b550c/apps/frontend/src/components/timeline/unread-divider.tsx#L15-L18

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Diff fully delivers the stated intent: persistent session-scoped divider, red→gray color settle, no unmount, reset on stream switch. `isFading`→`isDimmed` / `isDividerFading`→`isDividerDimmed` rename is complete across the hook, `unread-divider.tsx`, `event-list.tsx` (props/context/memo comparator), and both `stream-content.tsx` call sites.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Correctness lens found none. Verified: render-phase `latchRef` mutation is deterministic and StrictMode-safe (idempotent second pass); re-latch on stream switch works in a single render (reset-then-latch); latch persists through auto-mark-as-read (`firstUnreadEventId`→`undefined` never clears `latchRef.current.eventId`); already-read stream switch latches nothing.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>`displayedUnreadId` / `latchRef` / `isDimmed` / `lastReadEventId` traced across renders and stream switches. Invariants hold: latch-at-most-once-per-view, persist-through-read, no-divider-on-already-read, re-latch-on-unread-stream.</details>
<details><summary>📋 CLAUDE.md Compliance [1 minor]</summary>INV-25 comment-restatement nit (issue 2). INV-21 (no layout shift) satisfied — color/border-color transition only, absolutely-positioned element never enters document flow. INV-49 satisfied — no deprecated alias left after the rename. `useRef` is component-scoped, not an INV-9 singleton.</details>
<details><summary>🏗️ Design [1 minor]</summary>`readStateResolved` opt-in default (issue 1). Render-phase latch via ref mirrors the existing `deepLinkedRef` pattern in the same file — accepted.</details>
<details><summary>🔒 Security [CLEAN]</summary>Frontend-only; no injection/auth/XSS surface (no `dangerouslySetInnerHTML`).</details>
<details><summary>🎛️ UX [CLEAN]</summary>Two observations surfaced but both below threshold: a possible late-appearing red line on slow read-state hydration (inherent to the correct gate; ~55) and the persistent gray "New" label reading as semantically stale (the user explicitly requested this; ~50).</details>
<details><summary>📱 Mobile [CLEAN]</summary>Viewport-agnostic change; no `isMobile`/breakpoint/touch handlers. `pointer-events-none` preserved; absolute positioning keeps the line out of flow and out of virtualized row measurement.</details>

PR #1015, opened 2026-06-19
fix(conversations): use quote-replies and back-and-forth continuity in boundary classification

kristofferremback:
## Problem

The conversation boundary classifier (`apps/backend/src/features/conversations/boundary-extraction/`) was over-fragmenting long DM threads. Forensics on one production DM (read-only prod query): **3,597 messages split across 452 conversations** — 52 singletons, 20 empty. Two root causes in the classifier, which is a single per-message LLM call (`gpt-5.4-nano`, temp 0.2) fed only rendered markdown:

1. **Explicit quote-replies carried no weight.** A `quoteReply` node in `content_json` records `attrs.messageId` — the exact message the user chose to reply to — but the service never resolved it, and the quoted message's conversation wasn't guaranteed to be a candidate the model could pick. Measured on that DM: **19 of 90 quote-replies (21%) were filed in a different conversation than the message they quote**, despite the reply being an explicit user action.

2. **The prompt biased toward fragmentation.** `config.ts` said *"When in doubt … start fresh"* and *"A short or ambiguous message … is a new conversation"*. In a rapid back-and-forth this split a single exchange: short acks ("samma här", ":fire:") became their own singletons, and each participant's turns drifted into different conversations (e.g. one side's messages in "Kimi k2.7-code i Pi", the other's in "exsules API/GraphQL" — one conversation, torn in two).

## Solution

Give the classifier the one structural signal it was ignoring (the quote-reply edge) and rebalance the prompt toward continuing the live exchange — while leaning on the existing reassignment mechanism to split or merge as broader context arrives, rather than trying to be right on the first message.

### How it works

1. **Resolve quote-replies to a conversation.** `BoundaryExtractionService.resolveReplyTargets` reads the new message's `quoteReply.attrs.messageId` via the new `collectQuoteReplyMessageIds` (`@threa/prosemirror`), fetches those messages **scoped to the current stream** (`MessageRepository.findByIdsInStreams(client, ids, [streamId])` — conversations are per-stream, so a quote into another stream has no valid target here), and looks up their primary conversation with `ConversationRepository.findPrimariesByMessageIds`.
2. **Force the quoted conversation into the candidate set.** `mergeConversationsById` folds those conversations into `candidateConversations`, which feeds both `activeConversations` (so the model can assign to it even when it scrolled out of the surrounding-message window) and `validUpdateTargets`. The resolved edges are passed as `ExtractionContext.replyTargets`.
3. **Surface it in the prompt.** `LLMBoundaryExtractor.buildPrompt` renders an `## Explicit reply` block listing each quoted message, its conversation, topic, and a 100-char snippet. The prompt instructs the model to treat a quote-reply as the strongest continuity signal — default the reply into the quoted conversation, overridable only when the content clearly opens a new topic.
4. **Rebalance continuity.** The "start fresh when in doubt" guidance is replaced with: prefer continuing the active exchange; a short reply/reaction/follow-up from *either* participant continues it (author is irrelevant); never split one exchange across authors. The reassignment section now explicitly covers **both** directions — merge two threads that are one topic, and split out messages a continuity guess lumped together once a later message reveals a distinct topic.

### Key design decisions

**1. Strong signal, not a hard override.** A quote-reply biases the decision but the model can still start a new conversation when the content plainly changes subject. This matches the user's framing: *"a reply may not necessarily be of the same conversation … but at the very least it should be a strong signal."* Hard-forcing would misfile the "quoting you, but new topic" case.

**2. Same-stream scoping over cross-stream resolution.** The agents quote-resolver walks cross-stream with access checks; here the target must be a conversation in *this* stream (conversations carry a `streamId`), so resolving cross-stream quotes would only produce unusable candidates. Scoping the fetch to `[streamId]` is both correct and cheaper.

**3. Lean into early continuity, let reassignment settle it.** Per the user's ruling, it's fine for classification to be jumpy early and settle later — conversations can be split or merged retroactively. So the prompt prefers continuity up front and the reassignment guidance is strengthened to split out mistakes, rather than trying to avoid every premature merge. Moderate (not maximal) continuity: still starts a new conversation on a clear topic change.

**4. Reuse the extractor, don't duplicate it (INV-35).** `collectQuoteReplyMessageIds` lands in `@threa/prosemirror` next to `collectMentionSlugs`/`collectAttachmentReferenceIds`, reading node attrs rather than markdown (INV-54). The agents `quote-resolver.ts` now imports it and its private `extractQuoteReplyMessageIds`/`walkJsonNodes` duplicate is deleted (INV-38).

## New files

None.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/prosemirror/src/extractors.ts` | Add `collectQuoteReplyMessageIds` (reads `quoteReply.attrs.messageId`, deduped, first-seen order) |
| `packages/prosemirror/src/index.ts` | Export `collectQuoteReplyMessageIds` |
| `packages/prosemirror/src/extractors.test.ts` | Tests: document order, dedupe, missing/empty id, no-quote |
| `apps/backend/src/features/agents/quote-resolver.ts` | Use the shared helper; delete the private `extractQuoteReplyMessageIds`/`walkJsonNodes` |
| `apps/backend/src/features/conversations/boundary-extraction-service.ts` | `resolveReplyTargets` + `mergeConversationsById`; thread `replyTargets` and the merged candidate set through phase 1 |
| `apps/backend/src/features/conversations/boundary-extraction/types.ts` | `ReplyTarget` type; `ExtractionContext.replyTargets` |
| `apps/backend/src/features/conversations/boundary-extraction/llm-extractor.ts` | Render the `## Explicit reply` block; fill `{{REPLY_CONTEXT}}` |
| `apps/backend/src/features/conversations/boundary-extraction/config.ts` | `{{REPLY_CONTEXT}}` section; quote-reply + continuity guidance; bidirectional reassignment |
| `apps/backend/src/features/conversations/index.ts` | Export `ReplyTarget` |
| `apps/backend/evals/suites/boundary-extraction/types.ts` | `EvalReplyTarget`; `replyTargets` on input; `reply`/`continuity` categories |
| `apps/backend/evals/suites/boundary-extraction/suite.ts` | Pass `replyTargets` into the production `ExtractionContext` |
| `apps/backend/evals/suites/boundary-extraction/cases.ts` | Two cases: quote-reply beats the more-recent conversation; short ack stays in the exchange |

## Out of scope (deliberate)

- **No backfill / reclassification of existing data.** Per the user's call, this is forward-only — the 452 existing conversations are left as-is. A reclassification job would be a separate, costed write operation (and this branch only had read-only DB access).
- **No structural use of *sequential* (non-quote) replies.** Only explicit quote-replies are resolved structurally; plain back-and-forth continuity is handled by the prompt, not a metadata edge.
- **No model or temperature change.** Same `gpt-5.4-nano` @ 0.2.

## Test plan

- [x] `bunx tsc -p apps/backend/tsconfig.json --noEmit` — clean (covers `evals/**` too)
- [x] `bunx tsc -p packages/prosemirror/tsconfig.json --noEmit` — clean
- [x] `bun test ./packages/prosemirror/src/extractors.test.ts` — 14 pass
- [x] `bun test apps/backend/src/features/agents/ apps/backend/src/features/conversations/` — 327 pass (one logged ERROR is a deliberate error-path test)
- [ ] `bun run eval -- -s boundary-extraction` — **not run**: hits OpenRouter (live API/cost), so the new eval cases are committed but unscored here. Worth running before merge to confirm the prompt changes score as intended.
- [x] Self-review (inline): correctness + claim-verification pass over the diff. Confirmed no stale `relevantConversations` read after the merge, and `validConvIds` (in `validateResult`) derives from `activeConversations`, so quoted conversations are assignable.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qw9PhTCx9VGzvUJapN15Xy)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1015"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784452170&installation_id=129946197&pr_number=1015&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1015&signature=555404dd43b91ee3166620efb75fa833cbbcfd6cfd7980bfa47cac7a6bb69797"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/evals/suites/boundary-extraction/cases.ts
+++ b/apps/backend/evals/suites/boundary-extraction/cases.ts
@@ -464,4 +464,98 @@ Any ideas what's causing this?`,
       minConfidence: 0.7,
     },
   },
+
+  // Quote-reply: explicit reply continues the quoted message's conversation,
+  // even when a different conversation is the most recent one in the stream.
+  {
+    id: "reply-quote-continues-quoted-conv-001",
+    name: "Reply: quote-reply joins quoted message's conversation, not the most recent",
+    input: {
+      newMessage: {
+        authorId: "user_def456",
+        authorType: "user",
+        contentMarkdown: "samma här, kör på det",
+      },
+      activeConversations: [
+        {
+          id: "conv_buss",
+          topicSummary: "Buss nio imorgon",
+          messageCount: 8,
+          lastMessagePreview: "samma här",
+          participantIds: ["user_abc123", "user_def456"],
+          completenessScore: 3,
+        },
+        {
+          id: "conv_gpt",
+          topicSummary: "gpt down?",
+          messageCount: 2,
+          lastMessagePreview: "nvm, fungerar nu",
+          participantIds: ["user_abc123", "user_def456"],
+          completenessScore: 5,
+        },
+      ],
+      recentMessages: [
+        { authorId: "user_abc123", authorType: "user", contentMarkdown: "wtf. gpt down?" },
+        { authorId: "user_def456", authorType: "user", contentMarkdown: "nvm, fungerar nu" },
+      ],
+      replyTargets: [
+        {
+          quotedMessageId: "msg_buss_last",
+          conversationId: "conv_buss",
+          topicSummary: "Buss nio imorgon",
+          snippet: "Kan kika lite på vilka uteserveringar som kan verka rimliga",
+        },
+      ],
+      streamType: "dm",
+      category: "reply",
+    },
+    expectedOutput: {
+      // The quote-reply targets conv_buss; it must win over the more recent conv_gpt.
+      expectConversationId: "conv_buss",
+      minConfidence: 0.7,
+    },
+  },
+
+  // Continuity: a short acknowledgement from the OTHER participant continues the
+  // active exchange rather than spawning its own singleton conversation.
+  {
+    id: "continuity-short-ack-001",
+    name: "Continuity: short reply continues active back-and-forth",
+    input: {
+      newMessage: {
+        authorId: "user_def456",
+        authorType: "user",
+        contentMarkdown: ":fire: tokens",
+      },
+      activeConversations: [
+        {
+          id: "conv_exsules",
+          topicSummary: "exsules API/GraphQL",
+          messageCount: 6,
+          lastMessagePreview: "mer att AI blir förvirrad och slösar tokens lol",
+          participantIds: ["user_abc123", "user_def456"],
+          completenessScore: 3,
+        },
+      ],
+      recentMessages: [
+        {
+          authorId: "user_abc123",
+          authorType: "user",
+          contentMarkdown: "finns bara en remote, just nu dock",
+        },
+        {
+          authorId: "user_abc123",
+          authorType: "user",
+          contentMarkdown: "mer att AI blir förvirrad och slösar tokens lol",
+        },
+      ],
+      streamType: "dm",
+      category: "continuity",
+    },
+    expectedOutput: {
+      // ":fire: tokens" reacts to the "slösar tokens" line — stays in the exchange.
+      expectConversationId: "conv_exsules",
+      minConfidence: 0.6,
+    },
+  },
 ]
--- a/apps/backend/evals/suites/boundary-extraction/suite.ts
+++ b/apps/backend/evals/suites/boundary-extraction/suite.ts
@@ -93,6 +93,7 @@ function buildExtractionContext(input: BoundaryExtractionInput, workspaceId: str
       status: c.status ?? "active",
       contextMessageIds: [],
     })),
+    replyTargets: input.replyTargets,
     streamType: input.streamType || "scratchpad",
     workspaceId,
   }
--- a/apps/backend/evals/suites/boundary-extraction/types.ts
+++ b/apps/backend/evals/suites/boundary-extraction/types.ts
@@ -28,6 +28,18 @@ export interface EvalMessage {
 }
 
 /**
+ * One explicit quote-reply the new message makes, already resolved to the
+ * conversation that owns the quoted message (the production service does this
+ * resolution from `content_json`; evals provide it directly).
+ */
+export interface EvalReplyTarget {
+  quotedMessageId: string
+  conversationId: string
+  topicSummary: string | null
+  snippet: string
+}
+
+/**
  * Input for boundary extraction evaluation.
  */
 export interface BoundaryExtractionInput {
@@ -37,10 +49,12 @@ export interface BoundaryExtractionInput {
   recentMessages?: EvalMessage[]
   /** Active conversations in the stream */
   activeConversations?: EvalConversationSummary[]
+  /** Conversations the new message explicitly quote-replies (strong continuity signal) */
+  replyTargets?: EvalReplyTarget[]
   /** Stream type (channel, scratchpad, thread, dm) */
   streamType?: string
   /** Category for organizing test cases */
-  category?: "new-topic" | "continue-existing" | "topic-shift" | "resolution" | "ambiguous"
+  category?: "new-topic" | "continue-existing" | "topic-shift" | "resolution" | "ambiguous" | "reply" | "continuity"
 }
 
 /**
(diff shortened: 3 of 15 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (one non-blocking advisory)

UX and Mobile lenses were skipped: the only non-backend files are `packages/prosemirror/src/extractors.ts` + barrel — a pure JSON-tree utility with no UI surface, touch, or rendering.

Found 1 issue (advisory, not blocking):

1. `apps/backend/src/features/conversations/boundary-extraction-service.ts:555-557` — `resolveReplyTargets` fetches quoted messages with `MessageRepository.findByIdsInStreams(client, quotedMessageIds, [streamId])`, which scopes by `stream_id` + `deleted_at IS NULL` but carries no explicit `workspace_id` predicate (claude-md [INV-8, *Data Model and Persistence Safety*]: "must be queried/mutated with a workspace filter").
   ``https://github.com/threahq/threa/blob/3e6be81ee8db879fd82f4dc54b2b22cf75b5a1d7/apps/backend/src/features/conversations/boundary-extraction-service.ts#L555-L557``
   **Why it's non-blocking:** the message IDs come from untrusted `contentJson` (`quoteReply.attrs.messageId`), but `stream_id = ANY([streamId])` means a cross-workspace ID cannot match — the current stream belongs to this workspace, so isolation holds transitively. It is also strictly *tighter* than the already-accepted id-only lock query in the same file (L331-337, whose comment notes "no cross-tenant exposure to guard against here"), and CodeRabbit (which loads the INV-8 path rule) did not flag it. The agent's suggested swap to `findByIdsInWorkspace` would **regress** soft-delete filtering (that method lacks `deleted_at IS NULL`), so it's the wrong fix. **Recommendation:** either leave as-is, or add a one-line comment making the workspace-safety reasoning explicit (mirroring L331-337) to satisfy the letter of INV-8.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Forward-only (no backfill), same model/temp, reuse-not-duplicate all confirmed. The private `extractQuoteReplyMessageIds`/`walkJsonNodes` duplicate is removed in favor of the shared `@threa/prosemirror` helper (INV-35/37 satisfied).</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Cross-stream quotes correctly dropped; quoted messages with no primary conversation skipped; quoted conversation guaranteed into `validConvIds` via `mergeConversationsById` → `activeConversations`. The dedupe change in `collectQuoteReplyMessageIds` (new) vs the old non-deduped private helper does not alter the resolver's BFS (which dedupes via its own `visited` set); it only removes duplicate quoted-source blocks in `renderMessageWithQuoteContext` — a net improvement.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Candidate-conversation set and `replyTargets` traced Phase 1 → LLM → persist. `findPrimariesByMessageIds` is workspace-scoped; merging quoted conversations widens `validUpdateTargets` only to conversations in this stream/workspace, so no improper reassignment target is opened.</details>
<details><summary>📋 CLAUDE.md Compliance [1 advisory]</summary>INV-8 letter (above). INV-52 barrel imports, INV-54 (structural attrs not markdown), INV-25 (load-bearing comments only), INV-41 (no connection held across the AI call) all confirmed compliant.</details>
<details><summary>🏗️ Design [CLEAN]</summary>`resolveReplyTargets` issues two sequential batch queries (fetch messages → resolve primaries) — minimal for the dependency chain, not an N+1, inside the existing `withClient` before the AI call.</details>
<details><summary>🔒 Security [CLEAN]</summary>No high-confidence vulnerability. Untrusted `contentJson` message IDs are constrained by the stream predicate.</details>

---

PR #1017, opened 2026-06-19
feat(timeline): open at bottom with a jump-to-unread bar + progressive read + Escape

kristofferremback:
> [!IMPORTANT]
> **Direction update (dogfooding `693ac19`).** The original design landed you *on the first unread* (Slack's "open where you left off"). Testing on staging, we flipped to the other Slack mode: **open at the live bottom**, keep the unread preserved, and surface a top **"N new messages ↑"** button that jumps up to the "New" divider. The sections below describe the original landing; the current behavior is:
>
> - **Opens at the live bottom**, following the tail (the pre-existing cold-load behavior — `useTimelineScroll` is reverted to bottom-only).
> - **Read advances only through a *contiguous* run from where you left off** (`useLastSeenEvent`'s read frontier, pure `advanceFrontier` + unit-tested). Landing at the bottom leaves a gap above the read pointer, so it does **not** mark the unread read — the original "marks everything read" bug stays fixed. Scrolling up to the divider and reading down advances it; **Escape** marks all read.
> - **"N new messages ↑"** button at the top of the viewport (shown while unread sits above the fold) jumps to the divider and stops following the tail. The unread divider is a bookmark only (`scrollToUnread: false`).
> - Still in place from earlier rounds: `partial` mark-as-read (advance the pointer without optimistically zeroing the badge), the desktop **Escape** handler (mark-all-read + dismiss divider, with the tooltip-aware overlay guard), `useUnreadDivider.dismiss()`.

---

## Problem

Opening a stream marked **everything** read ~500ms after view, regardless of where the unread were. `useAutoMarkAsRead` (`apps/frontend/src/hooks/use-auto-mark-as-read.ts`) fired `markAsRead(streamId, events[events.length-1].id)` — the tail of the loaded window — on a debounce whenever the page was foregrounded and a count was elevated. So you'd open an unread channel, read the first screenful, and the rest below the fold would be silently marked read and lost. There was also no way to "escape" the unread state without scrolling.

This is the deferred follow-up to #1014 (which made the unread divider persistent). Two asks, both confirmed against Slack by the user:

1. **Progressive read** — don't advance read state past what the viewer has actually scrolled into view; unread below the fold stays unread until reached.
2. **Escape escapes the unread block** (desktop) — Slack's Esc-marks-channel-read: mark read, drop the divider, resume tailing.

Progressive read is only meaningful if you *land* on the first unread instead of the bottom — otherwise the bottom-most-seen row is the tail and everything is "seen". So landing-at-first-unread is the third, prerequisite piece.

## Solution

Three coupled changes: land the cold-load scroll on the first unread row, track the bottom-most row the viewer has actually seen and only mark up to it, and a desktop Escape that fully reads + resumes tail.

### How it works

**Land on first unread** (`use-timeline-scroll.ts`). The cold-load settle is generalized from `settleToBottom(maxMs)` to `settle(pin, maxMs)`. When the stream has an unread target, the initial-scroll layout effect pins the first-unread row near the top (`pinToIndex` → `scrollToIndex(idx, { align: "start", offset: -56 })`, the 56px revealing the divider + a sliver of context) and leaves `isFollowingTailRef=false` so new messages don't yank the reader down. The decision waits on `readStateResolved` so unread-vs-bottom is decided once, not corrected mid-settle; `stream-content.tsx` feeds the target index lazily via `initialUnreadIndexRef` (written after the divider computes, read in the layout effect) and gates on `lastReadEventId !== undefined`, with a 1000ms `readGateTimedOut` floor so a never-resolving read position can't hold the skeleton mask up. The divider hook's own `scrollIntoView` is now plain/thread-only (`scrollToUnread: !useVirtualized`) — the scroll engine owns the virtualized landing.

**Progressive mark** (`use-last-seen-event.ts`, new). `useLastSeenEvent` scans `[data-event-id]` rows in the scroll container on scroll (rAF-coalesced) and on content growth, and tracks the bottom-most row whose top has crossed the viewport bottom — excluding the band behind the floating composer (`--composer-height`). The seen index is monotonic and reset per stream. `useAutoMarkAsRead` marks up to that row, not the tail. When the bottom-most visible row is **not** the last rendered row (`atLastRow=false`), the mark is `partial`.

**Partial mark** (`use-unread-counts.ts`). A partial `markAsRead` advances only the read pointer (`membership.lastReadEventId`, the stream + membership mirrors in cache and IDB); it does **not** optimistically zero the badge. The server `stream:read` round-trip sets the true remaining count via the existing `applyStreamReadOrdinal`. A full mark is byte-for-byte the prior behavior.

**Escape** (`stream-content.tsx` + `use-unread-divider.ts`). A desktop, document-level keydown — active only while a divider is shown — marks the stream fully read (last loaded event), calls the new `dismiss()` on the divider (a latched divider has no clear-on-read), and `scrollToBottom({ force: true })` to resume tailing. Guarded against composer/editor/search focus and against open Radix overlays.

### Key design decisions

**1. Partial marks must not optimistically zero the badge — verified, load-bearing.** The unread system is ordinal-based and the read position MAX-merges (`apps/frontend/src/sync/unread-counters.ts` `applyStreamReadOrdinal`: `read = max(prevRead, lastReadOrdinal)`, where `prevRead = latest - unread`). If a mid-window mark forced `unread=0`, that implies `read=latest`, and a later `stream:read` for the mid ordinal can *never* lower it back — the badge would stickily over-clear. So mid-window marks skip the optimistic counter write and let the server round-trip own the count. The backend (`apps/backend/src/features/streams/service.ts:1678`) already computes `lastReadOrdinal` from the passed eventId and emits `stream:read`, so a positive remainder comes back. Reviewer-confirmed against both the reducer and the backend emission.

**2. Wait for the read position before deciding the landing.** Landing before `lastReadEventId` resolves would pick the bottom, and a late correction would fight the cold-load settle (the exact flake #1014's `readStateResolved` gate guards). Gating the scroll on the same signal makes the choice once. The 1000ms floor is insurance against a read position that never resolves (e.g. a non-member preview) hanging the mask.

**3. Suppress follow re-arm during the unread landing.** The settle's programmatic pins fire `scroll` → `handleScroll`; if the first-unread row is within the at-bottom band (a few unread), it would re-arm follow and the ResizeObserver would fight the pin and snap to the bottom. A `landedOnUnreadRef` keeps follow disarmed until a genuine user gesture clears it — so a near-bottom unread row still lands on the divider, and a real scroll-to-tail still re-arms normally. (Caught in self-review; see Design evolution.)

**4. Escape coordinates with overlays.** Radix listens for Escape in the capture phase and does not stop propagation, so a bubble-phase document handler would co-fire — closing a move dialog/dropdown *and* marking read. The handler bails when an open `[role=dialog|alertdialog|menu][data-state=open]` or `[data-radix-popper-content-wrapper]` is present. (Caught in self-review; see Design evolution.)

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/hooks/use-last-seen-event.ts` | Track the bottom-most seen row (monotonic, per stream); pure `pickBottomSeenId` + the scanning hook |
| `apps/frontend/src/hooks/use-last-seen-event.test.ts` | Unit tests for `pickBottomSeenId` |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-timeline-scroll.ts` | Generalize settle to `settle(pin, maxMs)`; add `pinToIndex`, `readStateResolved` gate, `getInitialScrollIndex`; land on first-unread without following tail; `landedOnUnreadRef` to suppress re-arm during the landing |
| `apps/frontend/src/hooks/use-auto-mark-as-read.ts` | Mark up to the seen event; `partial` option (with `lastMarkedPartialRef` so a partial→full transition at the same event re-fires) |
| `apps/frontend/src/hooks/use-unread-counts.ts` | `markAsRead(streamId, eventId, { partial })`; partial advances only the read pointer, never the optimistic counters |
| `apps/frontend/src/hooks/use-unread-divider.ts` | Add `dismiss()` to clear the latched divider for the session (Escape) |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Wire `useLastSeenEvent` + partial auto-mark; feed the unread landing target; `readGateTimedOut` floor; desktop Escape handler |
| `apps/frontend/src/components/timeline/event-list.tsx` | `findEventItemIndex` — item index by event id (mirrors `isFirstUnread`) for the landing target |
| `apps/frontend/src/hooks/index.ts` | Export `useLastSeenEvent` |
| `*.test.ts(x)` | Tests for partial mark (auto-mark + unread-counts), divider `dismiss`, `findEventItemIndex` |

## Out of scope (deliberate)

- **Escape marks up to the last *loaded* event**, not the absolute channel tail. If you're deep in history with newer events unloaded, those stay unread (Slack marks the whole channel). Left as a follow-up — flag if the full-channel semantics are wanted.
- **Mentions below the fold.** The server's `stream:read` zeroes mention/activity counts for the stream even on a partial read (existing backend semantic); a mid-fold mention is cleared. Not touched here.
- **Thread (plain-scroll) landing** still uses the divider's own `scrollIntoView`; only the virtualized timeline lands via the scroll engine. Progressive marking applies to both.

Two parallel reviewer agents over the diff turned up three things, all fixed in this branch:

- The read-state reviewer flagged a `partial`-vs-`lastEventId` timing skew (a scheduled partial mark could fire as full across an `atLastRow` flip, and a partial-then-tail-at-same-event cleared the badge only via the round-trip). Fixed by making the skip guard partial-aware (`lastMarkedPartialRef`) and adding `partial` to the effect deps, so a partial→full at the same event re-fires and the fire-time triple stays consistent.
- The scroll reviewer flagged the follow-re-arm tug-of-war (decision #3) and the Radix-Escape co-fire (decision #4); both fixed.

## Test plan

- [x] `apps/frontend` hooks + timeline suites — 769 pass (`bunx vitest run src/hooks src/components/timeline`)
- [x] New/updated unit tests: `pickBottomSeenId` (7), `findEventItemIndex` (3), partial mark in `use-auto-mark-as-read` (incl. partial→full re-fire) and `use-unread-counts` (badge stays, pointer advances), divider `dismiss`
- [x] `#1014` guardrails green: `use-unread-divider.test.ts`, broad sweep (859 across hooks/sync/sidebar earlier)
- [x] Full monorepo `tsc --noEmit` + `eslint` clean (pre-commit hook, exit 0)
- [ ] Browser E2E for the cold-load landing / progressive marking — **not run locally** (needs webServer + postgres + minio); relies on CI. The landing behavior in particular exercises real virtua measurement that jsdom can't reproduce.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FQs8HMgoPq3FonaHK64kaG)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1017"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784476518&installation_id=129946197&pr_number=1017&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1017&signature=9b9f595d550db92cde470635ffbf781fb68b4734808c59e81c8158b6f6687c3a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/event-list.test.ts
+++ b/apps/frontend/src/components/timeline/event-list.test.ts
@@ -5,6 +5,7 @@ import {
   filterVisibleItems,
   findFirstMessageId,
   findMessageItemIndex,
+  findEventItemIndex,
   getTimelineItemKey,
   groupTimelineItems,
   injectGapItems,
@@ -395,6 +396,34 @@ describe("findMessageItemIndex", () => {
       eventItem("evt_2", "msg_real"),
     ]
     expect(findMessageItemIndex(items, "msg_real")).toBe(1)
+  })
+})
+
+describe("findEventItemIndex", () => {
+  function eventItem(id: string): TimelineItem {
+    return { type: "event", event: createEvent({ id, sequence: id, eventType: "message_created", payload: {} }) }
+  }
+
+  it("returns the index of the item carrying the event id (the unread-divider anchor)", () => {
+    const items: TimelineItem[] = [eventItem("evt_1"), eventItem("evt_2"), eventItem("evt_3")]
+    expect(findEventItemIndex(items, "evt_2")).toBe(1)
+  })
+
+  it("matches a group on its first event, like the divider does", () => {
+    const items: TimelineItem[] = [
+      eventItem("evt_1"),
+      {
+        type: "session_group",
+        sessionId: "sess_1",
+        sessionVersion: 1,
+        events: [createEvent({ id: "evt_2", sequence: "2", eventType: "message_created", payload: {} })],
+      },
+    ]
+    expect(findEventItemIndex(items, "evt_2")).toBe(1)
+  })
+
+  it("returns -1 when the event is not in the window so the cold-load scroll falls back to the bottom", () => {
+    expect(findEventItemIndex([eventItem("evt_1")], "evt_gone")).toBe(-1)
   })
 })
 
--- a/apps/frontend/src/components/timeline/event-list.tsx
+++ b/apps/frontend/src/components/timeline/event-list.tsx
@@ -662,11 +662,12 @@ function TimelineItemContentImpl({ item, ctx, deferSecondaryHydration }: Timelin
   return (
     <>
       {showUnreadDivider && <UnreadDivider isDimmed={ctx.isDividerDimmed} />}
-      {/* The first-unread row gets extra top padding so the absolutely-positioned
-          divider has room to breathe above the message instead of crowding it.
-          Only wrap when the divider shows so every other row keeps its spacing
-          and DOM shape. */}
-      {showUnreadDivider ? <div className="pt-3">{rowContent}</div> : rowContent}
+      {/* The first-unread row reserves `pt-6` (24px) of top padding so the
+          absolutely-positioned divider, centered at 12px (`top-3`), gets equal
+          12px breathing room above and below the line — `pt-3` left the line
+          flush against the message top. Only wrap when the divider shows so
+          every other row keeps its spacing and DOM shape. */}
+      {showUnreadDivider ? <div className="pt-6">{rowContent}</div> : rowContent}
     </>
   )
 }
--- a/apps/frontend/src/components/timeline/event-list.tsx
+++ b/apps/frontend/src/components/timeline/event-list.tsx
@@ -286,6 +286,19 @@ export function findMessageItemIndex(items: TimelineItem[], messageId: string): 
   )
 }
 
+/**
+ * Index of the timeline item carrying `eventId` (the row, command group, or
+ * session group whose first event matches). Mirrors `isFirstUnread` so the
+ * scroll target resolves to the same item that renders the unread divider.
+ */
+export function findEventItemIndex(items: TimelineItem[], eventId: string): number {
+  return items.findIndex((item) => {
+    if (item.type === "event") return item.event.id === eventId
+    if (item.type === "command_group" || item.type === "session_group") return item.events[0]?.id === eventId
+    return false
+  })
+}
+
 /** Returns a stable key string for a timeline item */
 export function getTimelineItemKey(item: TimelineItem): string {
   switch (item.type) {
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -12,7 +12,10 @@ import {
   useStreamBootstrap,
   useWorkspaceUserId,
   useAutoMarkAsRead,
+  useLastSeenEvent,
+  useUnreadCounts,
   useUnreadDivider,
+  useIsMobile,
   useNewMessageIndicator,
   useAgentActivity,
   useAbortResearch,
@@ -61,6 +64,7 @@ import {
   itemDayStartMs,
   findFirstMessageId,
   findMessageItemIndex,
+  findEventItemIndex,
   getTimelineItemKey,
   filterVisibleItems,
   OLDER_SKELETON_ITEMS,
@@ -970,6 +974,25 @@ export function StreamContent({
   // (new message, link preview, virtua measuring) must not disarm follow.
   const userInteractedAtRef = useRef(0)
 
+  // Initial scroll target for the land-at-first-unread path. Updated each render
+  // once the unread divider has computed (below), read lazily by
+  // useTimelineScroll's cold-load layout effect. A ref because the divider is
+  // computed after this call; the getter is stable so it doesn't churn the
+  // effect.
+  const initialUnreadIndexRef = useRef<number | null>(null)
+  const getInitialScrollIndex = useCallback(() => initialUnreadIndexRef.current, [])
+
+  // The cold-load scroll waits for the read position so it lands on first-unread
+  // vs the bottom only once. This timeout is the floor: if the read position
+  // never resolves (e.g. a non-member preview with no membership row), land at
+  // the bottom anyway rather than hold the skeleton mask up forever.
+  const [readGateTimedOut, setReadGateTimedOut] = useState(false)
+  useEffect(() => {
+    setReadGateTimedOut(false)
+    const timer = setTimeout(() => setReadGateTimedOut(true), 1000)
+    return () => clearTimeout(timer)
+  }, [streamId])
+
   const {
     listRef,
     scrollerRef: virtualScrollerRef,
@@ -990,6 +1013,10 @@ export function StreamContent({
     skipInitialScroll,
     isJumpMode,
     userInteractedAtRef,
+    // Wait for the read position before the cold-load scroll decides
+    // unread-vs-bottom; only the virtualized timeline lands on first unread.
+    readStateResolved: !useVirtualized || lastReadEventId !== undefined || readGateTimedOut,
+    getInitialScrollIndex: useVirtualized ? getInitialScrollIndex : undefined,
   })
 
   // Scroll container element, owned by useTimelineScroll. Attached to the
@@ -1471,9 +1498,22 @@ export function StreamContent({
     clearSearch()
   }, [streamId, exitJumpMode, clearSearch])
 
-  // Auto-mark stream as read when viewing
-  const lastEventId = events.length > 0 ? events[events.length - 1].id : undefined
-  useAutoMarkAsRead(workspaceId, streamId, lastEventId, { enabled: !isDraft && !isLoading && !isJumpMode })
+  // Auto-mark stream as read when viewing. Read state never runs ahead of what
+  // the viewer has actually scrolled into view: mark up to the bottom-most seen
+  // row (Slack-style progressive read), not the tail of the loaded window. When
+  // the viewer is not at the live tail the mark is partial — the badge keeps the
+  // unread still below the fold (see useAutoMarkAsRead / markAsRead partial).
+  const autoMarkEnabled = !isDraft && !isLoading && !isJumpMode
+  const { lastSeenEventId, atLastRow } = useLastSeenEvent({
+    scrollContainerRef,
+    events: displayEvents,
+    streamId,
+    enabled: autoMarkEnabled,
+  })
+  useAutoMarkAsRead(workspaceId, streamId, lastSeenEventId, { enabled: autoMarkEnabled, partial: !atLastRow })
+
+  const isMobile = useIsMobile()
+  const { markAsRead } = useUnreadCounts(workspaceId)
 
   // Track live-arriving messages from other users for brief "new" indicator.
   const newMessageIds = useNewMessageIndicator(events, currentWorkspaceUserId ?? undefined, streamId, lastReadEventId)
@@ -1483,13 +1523,22 @@ export function StreamContent({
   // hide from this stream's render (e.g. thread membership rows) — otherwise
   // the divider can target an event id that never matches a rendered row and
   // silently fails to show.
-  const { dividerEventId, isDimmed: isDividerDimmed } = useUnreadDivider({
+  const {
+    firstUnreadEventId,
+    dividerEventId,
+    isDimmed: isDividerDimmed,
+    dismiss: dismissUnreadDivider,
+  } = useUnreadDivider({
     events: displayEvents,
     lastReadEventId,
     currentUserId: currentWorkspaceUserId ?? undefined,
     streamId,
     isLoading,
     highlightMessageId,
+    // The virtualized timeline lands on first unread through useTimelineScroll's
+    // cold-load settle (precise on an unmeasured list); only the plain thread
+    // scroller still needs the divider's own scrollIntoView.
+    scrollToUnread: !useVirtualized,
     // `lastReadEventId` is `string | null` once resolved; it is only `undefined`
     // while the read sources (stream row / membership) are still hydrating. Gate
     // on that so a not-yet-known read position can't be read as "all unread" and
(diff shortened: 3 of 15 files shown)


kristofferremback:
> ⤴️ **Superseded** by a newer review: https://github.com/threahq/threa/pull/1017#issuecomment-4754732607

### Code review
**Confidence: 5/7** — Good (2 real issues found and fixed in-review; rest advisory)

Two real issues surfaced and were **fixed in `8ccd92d`**; the remainder are advisory (verified, not blocking).

**Fixed in review:**

1. `apps/frontend/src/hooks/use-timeline-scroll.ts:295` — `scrollToBottom({ force })` armed follow but left `landedOnUnreadRef` set, so the scroll's own `handleScroll` re-disarmed follow. After Escape "resume tail" (and Jump-to-latest from an unread landing) the view reached the bottom but stopped following, so new messages no longer pinned. (bug) — now clears the flag on an explicit jump-to-tail.
   https://github.com/threahq/threa/blob/8ccd92dfcf91123a9c56570060d6d565296c932e/apps/frontend/src/hooks/use-timeline-scroll.ts#L292-L308

2. `apps/frontend/src/components/timeline/stream-content.tsx:1588` — the Escape `keydown` effect listed `events` in its deps, re-attaching the document listener on every live message while a divider showed. (reactivity) — now reads the last loaded event from a ref. Flagged by two lenses.
   https://github.com/threahq/threa/blob/8ccd92dfcf91123a9c56570060d6d565296c932e/apps/frontend/src/components/timeline/stream-content.tsx#L1564-L1598

**Advisory (not changed):**

3. `stream-content.tsx` — **Escape-marks-read has no in-product hint** (destructive, non-reversible). Slack shows a "Press Esc to mark as read" tooltip on the jump-to-latest button; this PR adds none. The behavior was explicitly requested (Slack parity), so this is a follow-up design call, not a defect. (ux-discoverability)

4. `use-last-seen-event.ts` — on a mobile **keyboard-open animation frame**, a scroll event can compute `viewportBottom` against the pre-keyboard layout and mark a band of rows "seen" eagerly. The read pointer is monotonic so it can't retract; the window is one frame and mostly hits "scrolled-up while the keyboard opens." Left as-is to keep scope tight. (mobile-regression)

5. `stream-content.tsx:1516` — Escape's full `markAsRead` comes from a second `useUnreadCounts(workspaceId)` call. This is idiomatic hook reuse, not the per-call service assembly INV-13 targets; cost is negligible. No change.

**Dismissed after verification (false positives):**

- "Radix popper guard is permanently suppressed / Escape double-fires with the emoji picker" — verified **no `forceMount`** on tooltip/hover-card/popover/dropdown, so `[data-radix-popper-content-wrapper]` is in the DOM **only while an overlay is open**. The guard suppresses Escape exactly when an overlay (incl. the open emoji picker) is up — correct, no double-fire.
- "`atLastRow` (DOM-last-row) vs `lastSeenEventId` (events-index) can diverge" — DOM render order follows the same ordered array, so they agree (second reviewer concurred, low confidence).
- "Thread plain-scroller bottom-then-jump snap (INV-21)" — pre-existing thread behavior; this diff only restricts the divider's `scrollToUnread` to `!useVirtualized` and doesn't change thread landing.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches the stated intent: scroll-to-first-unread, progressive partial read, Escape-to-read. No missing corresponding changes.</details>
<details><summary>🔍 Bugs [1 fixed]</summary>scrollToBottom/landedOnUnreadRef follow-resume (finding 1).</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Partial-read MAX-merge reasoning verified sound: skipping the optimistic zero on a mid-window mark is required, and the server `stream:read` reconciles the remaining count. No stuck/double-fire/wedge across open→partial→tail→full.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No clear introduced violations. INV-13 second-hook concern downgraded (advisory 5).</details>
<details><summary>🏗️ Design [1 advisory]</summary>Second useUnreadCounts call (advisory 5).</details>
<details><summary>🎛️ UX [1 fixed, 1 advisory]</summary>Escape resume-tail (fixed); Escape discoverability (advisory 3).</details>
<details><summary>📱 Mobile [1 advisory]</summary>Keyboard-open recompute race (advisory 4); Escape correctly desktop-gated.</details>

</details>

kristofferremback:
> [!NOTE]
> Superseded by a newer review: https://github.com/threahq/threa/issues/1017#issuecomment-4756993861


### Code review
**Confidence: 6/7** — Very Good (one minor issue, found and fixed in-review)

Re-review of the latest state (`55e5040`) after the earlier review rounds. One real issue surfaced and was **fixed in `2ab3dd3`**; everything else verified clean.

**Fixed in review:**

1. `apps/frontend/src/components/timeline/stream-content.tsx:1586` — the Escape overlay guard matched any `[data-radix-popper-content-wrapper]`, which Radix **tooltips** also render into. Tooltips are hover-triggered all over the message rows, so hovering one while pressing Escape silently swallowed the mark-read shortcut (and the prior comment's "never suppresses spuriously" was wrong for that case). Verified Radix Tooltip content carries `role="tooltip"`; now skips popper wrappers whose content is a tooltip. Dialogs/menus (by role) and real popovers (the reaction picker, which also `preventDefault`s) still suppress as before — no double-fire risk introduced. (ux-consistency, 85)
   https://github.com/threahq/threa/blob/2ab3dd362acbe119b31db2bac65ec6cd730a2f8d/apps/frontend/src/components/timeline/stream-content.tsx#L1580-L1596

**Verified and dismissed (no change):**

- *Correctness lens* raised five items but its own analysis exonerated each: the `useLastSeenEvent` cross-effect ordering on stream switch is guarded by the prefixed-ULID id-lookup miss + the per-stream `maxSeenIndex` reset; the `atLastRow`-vs-`lastSeenEventId` split is made correct by the `lastMarkedPartialRef` partial→full re-fire guard; the scroll-listener "leak" can't trigger given the keyed per-stream remount; the initial-scroll layout effect and the Escape-after-dismiss lifecycle are correct by design.
- *Spec & Design* and *Mobile*: no findings. The `readGateTimedOut` stale-state window is benign (the `null` target guard catches it); the settle-gate (`!virtualIsInitialSettling`) correctly stops marking rows seen behind the skeleton mask on touch too.
- The mobile keyboard-open one-frame recompute race and the Escape-discoverability hint remain the known advisories from the prior round (unchanged).

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches intent across all three behaviors.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No correctness defects at threshold.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Partial-read MAX-merge lifecycle re-verified sound; settle-gate doesn't permanently disable marking (re-enables on reveal).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No introduced violations.</details>
<details><summary>🏗️ Design [CLEAN]</summary>No concerns at threshold.</details>
<details><summary>🎛️ UX [1 fixed]</summary>Tooltip Escape-guard over-match (finding 1).</details>
<details><summary>📱 Mobile [CLEAN]</summary>Escape correctly desktop-gated; settle-gate + follow-suppression hold on touch.</details>

kristofferremback:
### Code review

Re-review of the **open-at-bottom** direction (HEAD `a0153d3`). The design is sound: the contiguous `advanceFrontier` engine, the visibility-gated divider fade, the partial-mark pointer lifecycle, and the Escape guard all hold up under scrutiny. Several flagged concerns dissolved on verification (see notes below). Three minor issues survived — two are comment hygiene, one is a small consistency gap.

Found 3 issues:

1. `apps/frontend/src/components/timeline/stream-content.tsx:1479-1491` — Two consecutive `Auto-mark stream as read when viewing.` comment headers describe the same code. The first block carries the **discarded** rationale ("mark up to the bottom-most seen row", "Held off until the cold-load settle reveals the rows…") from before the open-at-bottom pivot; the second block is the current one. (CLAUDE.md [INV-25, §7 Maintainability]: *"Comments explain why the current implementation works; they do not narrate the change history."* — delete the stale upper block.)
   https://github.com/threahq/threa/blob/a0153d3aa0e55d6ad3a34bb37c8fcbffdeddc9e0/apps/frontend/src/components/timeline/stream-content.tsx#L1479-L1491

2. `tests/browser/edit-last-message.spec.ts:159-164` — Comment asserts *"Cold load now lands on the first unread message (Slack-style), so … User A starts near the top"* — this is the reverted teleport behavior. The stream now opens at the live bottom, so the comment is stale and actively misleads the next reader into thinking the reversal was never applied (the test logic still passes either way via the `toPass` scroll-to-tail). (plan / INV-25: update the comment to the open-at-bottom precondition.)
   https://github.com/threahq/threa/blob/a0153d3aa0e55d6ad3a34bb37c8fcbffdeddc9e0/tests/browser/edit-last-message.spec.ts#L159-L164

3. `apps/frontend/src/components/timeline/stream-content.tsx:1901` — The "N new messages ↑" button's render condition (`unreadAboveViewport && unreadCount > 0 && !batchMode`) is not gated on `!isSearchOpen`, but the sibling Escape mark-read handler **is** (line 1546). So while the in-stream search overlay is open the jump button still shows and, on click, runs `disableAutoScroll()` + `scrollToIndex`, moving the underlying timeline out from under the active search-result navigation. Minor, but the two unread affordances should agree — add `!isSearchOpen` to the button condition for parity with Escape. (ux-consistency)
   https://github.com/threahq/threa/blob/a0153d3aa0e55d6ad3a34bb37c8fcbffdeddc9e0/apps/frontend/src/components/timeline/stream-content.tsx#L1901-L1901

---

Implementation matches the dogfooded direction: `scrollToUnread: false` (no teleport), contiguous-frontier read, top "N new ↑" jump, visibility-deferred divider fade, Escape = mark-all + dismiss + resume tail. Only issue #2 (stale test comment implying the old teleport behavior).
</details>

Traced frontier seeding/reset on stream switch, `useDividerDim` one-shot visibility latch + timer cleanup, and the Escape listener lifecycle — no wedging, retraction, or leaked timers/listeners found.
</details>

**Considered and dismissed:** the partial-mark / `activityCount` re-fire loop. `applyStreamReadOrdinal` (`apps/frontend/src/sync/unread-counters.ts:100`) zeroes `activityCounts[streamId]` on the `stream:read` socket event, and the 500 ms debounce cancels the pending re-mark before it fires when round-trip < debounce — so the partial path's skipped optimistic clear costs at most one extra mark on a >500 ms-latency link, not sustained hammering. Below threshold.
</details>

Issue #1 (INV-25 stale/duplicate comment). INV-21 (layout shift), INV-63 (silent success), INV-13 (construct-once) all checked clean for this diff.
</details>

`advanceFrontier`/`pickVisibleRange` are pure and unit-tested; `useDividerDim` is a clean one-shot. No config sprawl or parallel implementations.
</details>

Frontend read-tracking only; no new sinks.
</details>
<details><summary>🎛️ UX [1 issue]</summary>

Issue #3 (button not gated on `!isSearchOpen`).

**Considered and dismissed:** date-pill / search-bar overlap — the date pill is hidden while search is open (`!isSearchOpen`, line 2262) and the button's 56 px offset matches the established `top-14` loading-indicator position, clearing the `top-2` pill when search is closed. The earlier live "overlap" was a stale PWA bundle, now resolved. Also dismissed: Escape `overlayOwnsEscape` breadth (the guard is intentional — open poppers don't `stopPropagation`, so suppressing mark-read while one is open is correct) and the `ArrowUp` `aria-hidden` (the sibling "Jump to latest" `ArrowDown` omits it too; the button is text-labeled).
</details>

`useDividerDim`'s passive rAF scroll listener won't block momentum; the button is a standard `size="sm"` touch target. The `scrollToFirstUnread` `offset: -56` is a downward alignment nudge (divider stays visible), not a layout constraint.
</details>

PR #1018, opened 2026-06-19
fix(pwa): park new SW until reload + 404 missing assets, ending deploy-skew bricks

kristofferremback:
## Problem

A new build deploys; an already-open tab navigates to a route whose chunk hasn't loaded yet → `TypeError: Failed to fetch dynamically imported module: …/assets/workspace-layout-<hash>.js` → the tab bricks on the "Update needed" screen. Retention (#1007) reduced it but didn't end it. There are **two compounding root causes**:

**1. The SW swaps the precache out from under the running page.** The SW `skipWaiting()`s on install and `clients.claim()`s on activate (`apps/frontend/src/sw.ts`), and `useAppUpdate` nudges `registration.update()` on tab focus, a 5-minute poll, and every socket reconnect. So within seconds of a deploy the new SW seizes the still-running **old** page and `cleanupOutdatedCaches()` deletes the old precache. The page is now build N's JS controlled by build N+1's SW with N's precache gone — its next lazy `import()` of *its own* chunk misses the precache, falls through to the network, and 404s because build N is no longer the live deploy. The mechanism meant to keep users fresh is what breaks them. (The precache normally makes this impossible: it holds every chunk; we were deleting it mid-session.)

**2. A missing `/assets/*` is cached as immutable HTML, so the failure is permanent.** `public/_redirects` (`/* /index.html 200`) answers a missing chunk with `index.html` — 200, `text/html` — and `public/_headers` stamps `/assets/*` `immutable, max-age=1yr`. The bad HTML-as-JS response is pinned for a year in the browser HTTP cache **and at the CF edge**. `import()` keeps parsing HTML as a module; "Clear cache & reload" / `runSwRecovery` clears the SW + CacheStorage but can't evict the edge, so the tab stays bricked → the "Update needed / Visit /recover" screen (recovery exhausted its attempts).

## Solution

Remove the cause (the precache swap) and remove the permanence (the immutable-HTML poison).

### How it works

**1. Park the new SW until the user reloads (`sw.ts`, `use-app-update.ts`, `sw-messages.ts`).** Drop `skipWaiting()` on install — a new SW installs and stays in `waiting`. It activates only when the open page accepts the update toast and posts `SW_MSG_SKIP_WAITING`; the app reloads on the resulting `controllerchange`, landing on the new build's precached shell **and** chunks atomically. Each tab keeps its own build's worker + precache for its whole life, so it can always load its own chunks — version skew becomes structurally impossible for a given tab. `reloadForUpdate` first nudges `registration.update()` so a click that beats the install still finds the parked worker, with a `RELOAD_FALLBACK_TIMEOUT_MS` (3s) guard so a wedged worker can't strand Reload. Replaces the old `RELOAD_FRESH` network-first-navigation hack (removed). `clients.claim()` on activate is kept (correct on first install and at the user-driven swap).

**2. Missing `/assets/*` → real 404, never immutable HTML (`functions/assets/[[path]].ts`).** A Cloudflare Pages Function on `/assets/*` runs the normal asset pipeline via `next()`: a real asset passes through with its immutable headers intact; when the pipeline falls through to the HTML shell (asset gone), it returns `404` with `Cache-Control: no-store`. So any residual chunk miss (multi-tab skew, a user who ignores the toast for days, a mid-deploy edge propagation race) fails cleanly and transiently — recoverable on the next load instead of poisoning the browser/edge cache for a year.

**3. vitest split (`vite.config.ts`).** A jsdom `frontend` project and a `node` project, so the Pages Function and build-script tests run without the UI bootstrap (`./src/test/setup.ts` → `@/db`, DOM polyfills).

### Why not just retention or just client recovery

Retention (#1007) keeps old chunk *files* alive but does nothing for the precache swap (#1) or an edge that cached HTML during a propagation race (#2). Client recovery can't evict the CF edge, and it *unregisters the SW* — deleting the one precache that still had the old chunks — so it can make things worse. #1 stops the skew; #2 makes any leftover miss recoverable. Retention stays as defense-in-depth.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/functions/assets/[[path]].ts` | Pages Function: missing `/assets/*` → `404 no-store` instead of immutable HTML |
| `apps/frontend/functions/assets/[[path]].test.ts` | 3 tests: html fallback → 404 no-store; JS/CSS assets pass through |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/sw.ts` | Remove `skipWaiting()` on install; add `SW_MSG_SKIP_WAITING` handler; remove `serveNextNavFromNetwork`/`RELOAD_FRESH` machinery |
| `apps/frontend/src/hooks/use-app-update.ts` | `reloadForUpdate` does the skip-waiting → `controllerchange` → reload handshake (nudges `update()`, 3s fallback); drop `requestFreshNav` |
| `apps/frontend/src/lib/sw-messages.ts` | `SW_MSG_RELOAD_FRESH` → `SW_MSG_SKIP_WAITING` |
| `apps/frontend/src/hooks/use-app-update.test.ts` | Rewritten for the new handshake (no-waiting reload, activate-on-controllerchange, timeout fallback) |
| `apps/frontend/vite.config.ts` | vitest `node` project also covers `functions/**/*.test.ts` |

## Out of scope (deliberate)

- **Recovery reorder** (don't unregister the SW first — it's the precache that serves old chunks). Worth doing, but with #1+#2 in place recovery is rarely reached; separate follow-up to keep this PR focused.
- **`apps/backoffice`** has the same SW/`_headers` shape; left out to keep this scoped to the user-facing app.
- **Retention concurrency** (`cancel-in-progress` can skip the save step) — minor, defense-in-depth layer, separate change.

## Test plan

- [x] `vitest` full frontend suite — **2959 pass / 271 files** (both projects)
- [x] `use-app-update.test.ts` — 8 pass (handshake, nudge, controllerchange, timeout fallback)
- [x] `functions/assets/[[path]].test.ts` — 3 pass (404 no-store on html; JS/CSS passthrough)
- [x] `tsc --noEmit` frontend clean; pre-commit lint hook passed
- [ ] **CF integration not exercisable offline** — needs **staging verification**: (a) `wrangler pages deploy dist` (run with `workingDirectory: apps/frontend`) actually compiles `functions/` and auto-scopes `_routes.json` to `/assets/*`; (b) `next()` returns the SPA HTML fallback for a missing asset; (c) `_headers` does **not** re-stamp `immutable` over the Function's `no-store` 404. If (c) fails, the fallback is to serve `/assets/*` through the worker. Will verify on `staging.threa.io` before relying on it in prod.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CeKiRFBN6SWYHNq3S1zc1q)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1018"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784494865&installation_id=129946197&pr_number=1018&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1018&signature=a32719060d6a1eb261abebe477c5447d9f7bfe325c5d5057220361a8d883e6e4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/functions/assets/[[path]].test.ts
+++ b/apps/frontend/functions/assets/[[path]].test.ts
@@ -1,0 +1,49 @@
+import { describe, it, expect } from "vitest"
+import { onRequest } from "./[[path]]"
+
+function context(next: Response) {
+  return {
+    request: new Request("https://app.threa.io/assets/workspace-layout-AbC123.js"),
+    next: async () => next,
+  }
+}
+
+describe("assets onRequest", () => {
+  it("converts the SPA html fallback (missing chunk) into a no-store 404", async () => {
+    const fallback = new Response("<!doctype html><html></html>", {
+      status: 200,
+      headers: { "content-type": "text/html; charset=utf-8", "cache-control": "public, max-age=31536000, immutable" },
+    })
+
+    const res = await onRequest(context(fallback))
+
+    expect(res.status).toBe(404)
+    expect(res.headers.get("cache-control")).toBe("no-store")
+    expect(res.headers.get("content-type")).toContain("text/plain")
+  })
+
+  it("passes a real JS asset through untouched, immutable header intact", async () => {
+    const asset = new Response("export const x = 1", {
+      status: 200,
+      headers: { "content-type": "application/javascript", "cache-control": "public, max-age=31536000, immutable" },
+    })
+
+    const res = await onRequest(context(asset))
+
+    expect(res.status).toBe(200)
+    expect(res.headers.get("cache-control")).toContain("immutable")
+    expect(await res.text()).toBe("export const x = 1")
+  })
+
+  it("passes a real CSS asset through untouched", async () => {
+    const asset = new Response("body{}", {
+      status: 200,
+      headers: { "content-type": "text/css" },
+    })
+
+    const res = await onRequest(context(asset))
+
+    expect(res.status).toBe(200)
+    expect(res.headers.get("content-type")).toBe("text/css")
+  })
+})
--- a/apps/frontend/functions/assets/[[path]].ts
+++ b/apps/frontend/functions/assets/[[path]].ts
@@ -1,0 +1,42 @@
+/**
+ * Cloudflare Pages Function for /assets/* (auto-scoped: wrangler generates a
+ * _routes.json that only invokes this for asset paths; navigations bypass it).
+ *
+ * The problem it fixes: when a content-hashed chunk no longer exists on the live
+ * deployment, the SPA `_redirects` fallback (`/* /index.html 200`) answers the
+ * /assets/*.js request with index.html — HTML, status 200 — and `public/_headers`
+ * stamps /assets/* `immutable, max-age=1yr`. The browser (and the CF edge) then
+ * cache that HTML-as-JS for a year, so the tab's dynamic import() keeps parsing
+ * HTML as a module and stays bricked until the cache expires. Clearing the SW and
+ * CacheStorage can't evict the edge copy.
+ *
+ * Here we run the normal asset pipeline via next(): a real asset comes back with
+ * its own immutable headers untouched. Only when the pipeline falls through to
+ * the HTML shell (the asset is gone) do we replace it with a real 404 marked
+ * no-store, so import() fails cleanly and transiently — recoverable on the next
+ * load instead of permanently poisoned.
+ */
+interface AssetsFunctionContext {
+  request: Request
+  next: () => Promise<Response>
+}
+
+export async function onRequest(context: AssetsFunctionContext): Promise<Response> {
+  const response = await context.next()
+
+  // Real assets are application/javascript, text/css, image/*, font/* — never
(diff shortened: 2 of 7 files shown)


kristofferremback:
Closed because of #1019

PR #1026, opened 2026-06-20
fix(timeline): re-scan read frontier on content resize so short streams mark read

kristofferremback:
## Problem

After #1017 a stream opens at the live bottom and the read frontier
(`useLastSeenEvent`) advances only through the contiguous run of rows the viewer
actually has on screen. The frontier is recomputed in two places: a `scroll`
listener, and an effect keyed on `events.length` / `readIndex`. Neither fires
when row **geometry** changes without a scroll or a new event.

That gap is invisible on a long stream — you scroll, the listener fires, rows get
marked. But on a stream short enough to fit the viewport there is nothing to
scroll, so after the initial scan the frontier is frozen. Two things move the
rows right after that scan and neither re-scans:

1. the open-at-bottom **settle** finishing (the list pins to bottom behind the
   skeleton mask), and
2. **async embeds** — link/GitHub/Cloudflare cards and images — loading and
   growing the trailing rows into their final positions.

Result (reproduced on prod, `#gh-notifications`): the last messages sit fully
visible above the composer but stay unread, with no scroll gesture available to
nudge them into the seen band. The unread count never clears.

## Solution

Give the frontier a third trigger: observe the scrolling content box with a
`ResizeObserver` and re-run the same scan on any post-scan resize.

### How it works

- `useLastSeenEvent` gains an optional `contentRef` — the content box inside the
  scroller whose height tracks `scrollHeight`. In the scroll-listener effect it
  now also creates a `ResizeObserver(schedule)` and observes both the container
  (`el`, for viewport changes) and the content box (when distinct, for content
  growth). `schedule` is the existing rAF-debounced wrapper around `recompute`,
  so the resize path reuses the exact scan the scroll path uses — no second code
  path. The observer is disconnected in the effect cleanup alongside the scroll
  listener.
- `stream-content.tsx` passes `contentRef: useVirtualized ? virtualContentRef : undefined`.
  `virtualContentRef` is the `<div ref={contentRef}>` from `useTimelineScroll`
  that wraps the `Virtualizer` rows plus the `ComposerFooterSpacer`; when an
  embed grows a row, that box's height changes and the observer fires. The plain
  (thread) path keeps its content directly inside the scroller, so the hook's
  `content === el` guard falls back to observing the container alone.

The frontier math (`advanceFrontier`, `pickVisibleRange`) is unchanged — this
only changes *when* the scan runs, not what it concludes. Marking stays
conservative: a resize that doesn't bring a contiguous trailing row into view
leaves the frontier where it was.

## Key design decisions

**1. `ResizeObserver` on the content box, not a poll or a settle flag.** A resize
observer catches every cause at once — settle, embeds, image decode, font swap,
orientation — rather than wiring each event source in. Keying off an
`isInitialSettling` flag would fix the settle race but not the embeds that load
seconds later; those are the actual culprit in the screenshot.

**2. Reuse `schedule`, observe both nodes.** The observer callback is the same
rAF-coalesced `schedule` the scroll listener uses, so a burst of resizes
collapses to one scan per frame and the resize/scroll paths can't disagree.
Observing `el` as well keeps viewport-size changes (mobile keyboard, rotation)
covered without a separate observer.

**3. Conservative `typeof ResizeObserver !== "undefined"` guard.** Keeps the hook
safe under non-DOM test/SSR contexts; when absent the behavior is exactly the
prior scroll-only one.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-last-seen-event.ts` | Add optional `contentRef`; create/observe/disconnect a `ResizeObserver` in the scroll-listener effect so geometry changes re-run the scan. |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Pass `virtualContentRef` as `contentRef` on the virtualized path (plain path falls back to the container). |
| `apps/frontend/src/hooks/use-last-seen-event.test.ts` | Add a mounted-hook regression test for the resize re-scan. |

## Out of scope (deliberate)

- The plain (non-virtualized) thread path is not given a dedicated content box to
  observe — it has no single wrapper today, and its content also re-scans on the
  `events.length` path. If a short thread with late-loading embeds shows the same
  symptom, that's a follow-up.
- No change to the frontier semantics, the partial-mark lifecycle, or the divider
  — purely a re-scan trigger.

## Test plan

- [x] `bunx vitest run src/hooks/use-last-seen-event.test.ts` — 14 pass (includes the new resize regression test, which fails without the fix: with no observer created, the simulated resize is a no-op and the frontier stays at `e1`).
- [x] `bunx vitest run` on the neighboring suites (`use-unread-divider`, `stream-content`, `event-list`) — 93 pass.
- [x] `tsc --noEmit` (frontend `typecheck`) clean.
- [x] Pre-commit hook: eslint + full-monorepo `tsc --noEmit` clean.
- [ ] Not exercised in a real browser against a live embed-loading stream — the failure is timing/geometry-driven, so it's covered by the unit reproduction (stubbed `ResizeObserver` + mutable row rects) rather than an E2E.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01FQs8HMgoPq3FonaHK64kaG)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1026"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784547572&installation_id=129946197&pr_number=1026&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1026&signature=0de97840e07056c905da58e4a88a833b92db9643d9ad8ca60c0e12f1c2d73098"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -1484,6 +1484,10 @@ export function StreamContent({
   const autoMarkEnabled = !isDraft && !isLoading && !isJumpMode
   const { lastSeenEventId, atLastRow, unreadAboveViewport } = useLastSeenEvent({
     scrollContainerRef,
+    // Observe the virtualized content box so a settle/embed resize re-scans even
+    // when the stream is too short to scroll. The plain path's content lives in
+    // the scroller itself, so the hook's container fallback covers it.
+    contentRef: useVirtualized ? virtualContentRef : undefined,
     events: displayEvents,
     streamId,
     lastReadEventId,
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -1485,8 +1485,12 @@ export function StreamContent({
   const { lastSeenEventId, atLastRow, unreadAboveViewport } = useLastSeenEvent({
     scrollContainerRef,
     // Observe the virtualized content box so a settle/embed resize re-scans even
-    // when the stream is too short to scroll. The plain path's content lives in
-    // the scroller itself, so the hook's container fallback covers it.
+    // when the stream is too short to scroll. The plain (thread) path has no
+    // separate content box; the hook then observes the scroller itself, which
+    // catches viewport/keyboard resizes but NOT inner embed growth (a fixed-
+    // height overflow container's own box doesn't change as content grows) — a
+    // short thread ending in a late-loading card staying unread is a known
+    // follow-up, not covered here.
     contentRef: useVirtualized ? virtualContentRef : undefined,
     events: displayEvents,
     streamId,
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -1025,6 +1025,9 @@ export function StreamContent({
 
   // Unified API regardless of scroll mode
   const scrollContainerRef = useVirtualized ? virtuosoScrollerRef : plainScrollRef
+  // Content box for the plain (thread) scroller — its height tracks scrollHeight,
+  // so observing it (not the fixed h-full scroller) catches embed/image growth.
+  const plainContentRef = useRef<HTMLDivElement>(null)
   const isScrolledFarFromBottom = useVirtualized ? virtualIsScrolledFar : plainIsScrolledFar
   const scrollToBottom = useVirtualized ? virtualScrollToBottom : plainScrollToBottom
   const disableAutoScroll = useVirtualized ? virtualDisableAutoScroll : plainDisableAutoScroll
@@ -1484,14 +1487,12 @@ export function StreamContent({
   const autoMarkEnabled = !isDraft && !isLoading && !isJumpMode
   const { lastSeenEventId, atLastRow, unreadAboveViewport } = useLastSeenEvent({
     scrollContainerRef,
-    // Observe the virtualized content box so a settle/embed resize re-scans even
-    // when the stream is too short to scroll. The plain (thread) path has no
-    // separate content box; the hook then observes the scroller itself, which
-    // catches viewport/keyboard resizes but NOT inner embed growth (a fixed-
-    // height overflow container's own box doesn't change as content grows) — a
-    // short thread ending in a late-loading card staying unread is a known
-    // follow-up, not covered here.
-    contentRef: useVirtualized ? virtualContentRef : undefined,
+    // Observe the scrolling content box (its height tracks scrollHeight) so a
+    // settle / embed / image resize re-scans even when the stream is too short
+    // to scroll. Both paths pass one — the virtualized list's inner content div
+    // and the plain (thread) scroller's content wrapper. Observing the scroller
+    // itself wouldn't catch it: a fixed h-full box doesn't change as content grows.
+    contentRef: useVirtualized ? virtualContentRef : plainContentRef,
     events: displayEvents,
     streamId,
     lastReadEventId,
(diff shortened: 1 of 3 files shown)


kristofferremback:
### Code review

Tight, well-targeted fix. The `ResizeObserver` lifecycle is correct (one per effect run, `disconnect()` in cleanup, `streamId` in deps so it tears down on stream switch), there's no resize↔recompute feedback loop (React bails on unchanged primitive state; ≤2 cycles), scanning while the skeleton mask is up is safe (`rowEls.length === 0` early-returns), and the frontier still never retracts. INV-13/21/25/36 checked clean on the core change. One minor doc-accuracy issue surfaced (two lenses converged on the mechanism).

Found 1 issue:

1. `apps/frontend/src/components/timeline/stream-content.tsx:1487-1490` — The comment "The plain path's content lives in the scroller itself, so the hook's container fallback covers it" **overstates** the plain-path coverage. On the plain (thread) path `contentRef` is `undefined`, so the hook observes only `scrollContainerRef` — the `h-full overflow-y-auto` scroller. A `ResizeObserver` watches that element's own box, which is pinned by `h-full` and does **not** change when inner content grows (only `scrollHeight` does). So the plain path re-scans on viewport/keyboard resize but **not** on async embed growth — a short thread ending in a late-loading link/GitHub card keeps the exact "stuck unread" symptom this PR fixes for the virtualized path. The functional gap is a deliberate, documented follow-up (PR "Out of scope"), but the comment should state that limitation rather than claim coverage, or a future reader will trust it and assume threads are handled. (CLAUDE.md [INV-25, §7 Maintainability]: *"a stale comment is worse than none because the next reader … trusts it"* — here the comment is inaccurate from the start. Converged: UX 82, Mobile 50.)
   https://github.com/threahq/threa/blob/5abeec7aaecbf8f44fd7d1d341731304a4f47fd6/apps/frontend/src/components/timeline/stream-content.tsx#L1487-L1490

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Fix matches the stated intent (re-scan the read frontier on content-geometry changes so short, unscrollable streams mark trailing rows read). Cleanup, dep array, and the regression test are all present.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>RO created/disconnected correctly across effect deps `[enabled, streamId, recompute, scrollContainerRef, contentRef]`; `content !== el` guard avoids double-observe; `contentRef?.current ?? null` handles the undefined path. No leak, no wedge.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>No infinite resize↔recompute loop (primitive-state bailout caps it). Frontier never retracts (`advanceFrontier` holds or advances; external-read bump only moves forward). Scan during settle/skeleton is a no-op (no `[data-event-id]` rows yet).</details>
<details><summary>📋 CLAUDE.md Compliance [1 issue]</summary>Issue #1 (INV-25 inaccurate comment). INV-13 (observer constructed once per effect), INV-21 (RO reads geometry, never writes DOM), INV-36 (no speculative scope) all clean.</details>
<details><summary>🏗️ Design [CLEAN]</summary>RO reuses the existing rAF-coalesced `schedule`, so the resize and scroll paths share one scan and a burst collapses to one `recompute` per frame. Same per-row `getBoundingClientRect` cost as the existing scroll handler — bounded by the virtualized render window.</details>
<details><summary>🔒 Security [CLEAN]</summary>Frontend read-tracking only; no new sinks.</details>
<details><summary>🎛️ UX [1 issue]</summary>Issue #1 — plain/thread path with embeds remains stuck unread; the comment overstates coverage. Functional gap itself is documented out-of-scope.</details>
<details><summary>📱 Mobile [CLEAN]</summary>The container RO firing on keyboard/visual-viewport resize is beneficial (re-scan after the keyboard changes what's visible), and the `--composer-height` exclusion-band math is unaffected. Same plain-path embed gap as UX (scored 50, below threshold), folded into issue #1.</details>

PR #1030, opened 2026-06-20
fix(pi-remote): preserve unknown top-level fields in saveConfig

kristofferremback:
## Problem

`saveConfig()` in `extensions/pi-remote/src/threa-remote.ts` drops hand-edited top-level fields from `~/.pi/agent/threa-remote.json`. It builds the persisted object from in-memory config only (`{ ...config }`), then merges just `linkedSessions` and `streamCursors` from disk. Any other field — `defaultLabel`, `pollMs`, or unknown keys added by hand — is silently deleted on the next write.

## Solution

Extract the merge logic into `buildPersistedConfig()`, a pure function that starts from spread on-disk fields (so unknown/hand-edited keys survive) and overlays only *defined* values from the in-memory config (so managed fields like `baseUrl`/`apiKey` still win, but an unset optional like `config.defaultLabel = undefined` does not erase a hand-edited value on disk).

### How it works

- `buildPersistedConfig(inMemory, onDisk)` starts with `{ ...(onDisk ?? {}) }` — every field on disk is the baseline
- Iterates `Object.entries(inMemory)` and writes each non-`undefined` value onto the result — managed fields with a runtime value override disk, but unset optionals don't overwrite
- Deletes `enabled` (migrated to per-session link state) and `streamCursors` (merged below)
- Merges `linkedSessions` from both sides so concurrent Pi instances (INV-20) don't clobber each other's session links
- Conditionally writes `streamCursors` when either side has entries
- `saveConfig()` is simplified to a thin lock+write shell that delegates to `buildPersistedConfig`

### Key design decisions

**1. Spread on-disk first, overlay in-memory second** — the reverse of the old approach. This is the minimal pivot: any property in the on-disk JSON that the runtime doesn't know about (or whose value is unset) survives. The lock guard (INV-20) still covers the entire read-merge-write.

**2. Skip undefined values during overlay** — `for (const [key, value] of Object.entries(inMemory)) { if (value !== undefined) ... }`. Without this, `{ ...onDisk, ...config }` would set `defaultLabel: undefined` when the runtime hasn't loaded one, erasing the hand-edited value.

## Modified files

| File | Change |
| ---- | ------ |
| `extensions/pi-remote/src/threa-remote.ts` | Extract `buildPersistedConfig` (+40/-33) |
| `extensions/pi-remote/src/threa-remote.test.ts` | 6 new tests for field preservation (+101/-0) |

## Test plan

- [x] `bun test ./extensions/pi-remote/src/threa-remote.test.ts` — 91 pass, 164 expect calls

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1030"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784568492&installation_id=129946197&pr_number=1030&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1030&signature=cfb05f94084ae0d6fb4eb572758f87e8de69484de33b164c45f74555e27f3e46"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/extensions/pi-remote/src/threa-remote.test.ts
+++ b/extensions/pi-remote/src/threa-remote.test.ts
@@ -782,3 +782,104 @@ describe("defaultDisplayNameFor", () => {
     expect(__testing.defaultDisplayNameFor("/Users/kris/dev/personal/threa", "  Work Pi  ")).toBe("Work Pi - threa")
   })
 })
+
+describe("buildPersistedConfig", () => {
+  test("preserves hand-edited top-level fields from on-disk config not present in memory", () => {
+    const result = __testing.buildPersistedConfig(
+      {
+        baseUrl: "https://app.threa.io",
+        workspaceId: "ws_123",
+        apiKey: "threa_bk_test",
+        linkedSessions: { session_a: { linkId: "a", rootStreamId: "s1", activeStreamId: "s1", runtimeSessionId: "rs1", streamUrlPath: "/s1" } },
+      } as never,
+      {
+        defaultLabel: "coding",
+        pollMs: 5000,
+      } as never
+    )
+    // In-memory required fields present
+    expect(result.baseUrl).toBe("https://app.threa.io")
+    expect(result.workspaceId).toBe("ws_123")
+    // Hand-edited fields from on-disk absent from the in-memory config survive
+    expect(result.defaultLabel).toBe("coding")
+    expect(result.pollMs).toBe(5000)
+    // linkedSessions from in-memory still present
+    expect(result.linkedSessions).toMatchObject({ session_a: { linkId: "a" } })
+  })
+
+  test("in-memory defined fields override on-disk values", () => {
+    const result = __testing.buildPersistedConfig(
+      {
+        baseUrl: "https://staging.threa.io",
+        workspaceId: "ws_123",
+        apiKey: "threa_bk_test",
+      } as never,
+      {
+        baseUrl: "https://app.threa.io",
+      } as never
+    )
+    expect(result.baseUrl).toBe("https://staging.threa.io")
+  })
+
+  test("merges linkedSessions from both sides, in-memory keys win", () => {
+    const result = __testing.buildPersistedConfig(
+      {
+        baseUrl: "https://app.threa.io",
+        workspaceId: "ws_123",
+        apiKey: "threa_bk_test",
+        linkedSessions: {
+          session_b: { linkId: "b", rootStreamId: "s1", activeStreamId: "s1", runtimeSessionId: "rs2", streamUrlPath: "/s1" },
+        },
+      } as never,
+      {
+        linkedSessions: {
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: pi (current) | Reviewers: combined (spec+design, correctness+data-flow)

No issues found. Checked for bugs, data-flow lifecycle, CLAUDE.md compliance, plan adherence, design quality, and security on this backend-only diff.

**Lenses applied:**

- **Spec & Design** — CLAUDE.md audit: no violations introduced. The lock guard (INV-20) is preserved in the simplified `saveConfig()`; the merge logic is extracted into a pure function without weakening the lock scope. Code comments are concise and explain *why* the skip-undefined-overlay exists (INV-25). Historical context: #1000 established the linkedSessions merge and lock; #1012 added scratchpad labels. This change builds on both without regressing either.

- **Plan adherence** — the fix exactly matches the stated goal: `buildPersistedConfig` starts from spread on-disk fields so hand-edited keys survive, then overlays only defined in-memory values.

- **Correctness & Data Flow** — executed the primary merge sequences mentally:

  - *Hand-edited field preservation:* on-disk `defaultLabel: "coding"` + in-memory config with no `defaultLabel` key → `Object.entries` skips absent keys → on-disk value survives. Verified.
  - *In-memory override:* on-disk `baseUrl: "https://app.threa.io"` + in-memory `baseUrl: "https://staging.threa.io"` → for loop overlays the defined value → staging wins. Verified.
  - *Cross-instance linkedSessions:* on-disk has `session_b` from another Pi instance, in-memory has only `session_a`. Step 1 (spread onDisk) includes both; step 2 (for loop) overwrites with in-memory `session_a` only; step 3 (merge) restores on-disk `session_b` via `{ ...onDisk.linkedSessions, ...inMemory.linkedSessions }`. Verified.
  - *StreamCursors clean write:* neither side has cursors → merge condition false → no `streamCursors` key output. Verified.
  - *Global enabled stripped:* always removed regardless of source (onDisk or inMemory). Verified.

- **Security** — no user-input flow changes. The function operates on pre-validated `Config` objects. No injection vectors.

PR #1036, opened 2026-06-20
feat(drafts): synchronous localStorage staging so a fast type-then-reload never loses content

kristofferremback:
## Problem

The composer debounces its write to IndexedDB (`use-draft-message.ts`, `DEBOUNCE_MS = 500`). A keystroke does `setContent(newContent)` (React state) + `saveDraftDebounced(newContent)`, and the latter only reaches IDB after 500ms of quiet. Type, then reload inside that window, and the keystrokes are gone — they lived only in React state and the TipTap editor, both wiped by the reload. Losing in-progress user content is the cardinal sin the whole draft system is built to avoid ("duplicated drafts are acceptable; lost drafts are not"), and this was a real, easily-hit hole.

## Solution

Add a third persistence layer in front of IDB: a **synchronous `localStorage` staging buffer**, written on every change before the debounce arms, so the content is on disk before the browser can process a subsequent reload task. The three layers now each have their own cadence:

- **localStorage** — every change, synchronous (the crash/reload-safe buffer).
- **IndexedDB** — debounced (the existing 500ms), the local source of truth (unchanged).
- **server** — mirrored from IDB through the offline op queue (unchanged).

The framing rule for recovery is **content-diff, never a cross-device timestamp compare** — comparing this device's clock against a roamed draft's foreign clock would be unsafe, and cross-device conflicts already have a correct resolver (the `version`/split machinery).

### How it works

- **Stage on change.** `saveDraftDebounced` (`use-draft-message.ts`) calls `stageDraftContent(workspaceId, draftKey, contentJson)` synchronously before arming the timer. Key: `threa:draft-stage:{workspaceId}:{scope}`. The debounced `saveDraft` → IDB path is untouched.
- **Recover at load.** `reconcileStagedDrafts` (`draft-sync.ts`) runs in `CoordinatedLoadingProvider` **before** `seedDraftCacheFromIdb`, so a recovered body is already in `db.drafts` when the composer first reads — no empty flash, no second read path in the composer. It writes straight to IDB (not the in-memory cache) so it can't flip the draft cache "ready" with a partial set.
- **Content-diff, not timestamp.** For each staged entry, if its body equals the scope's loaded IDB draft, the last debounced flush already captured it → drop the buffer. If it differs, it's an un-flushed tail → write it into the draft (preserving `baseVersion` / attachments / context refs via `{ ...existing }`) and enqueue a push.
- **Cross-device safety via the existing split.** Recovery enqueues the push *before* the IDB write, so a sync bootstrap racing at boot sees the pending upsert and defers to the server split for that id (the same "unpushed edits win, server arbitrates" rule `applyDraftUpserted` already follows) instead of overwriting the tail with a newer server row. A draft that drifted on another device while the tab was gone thus resolves to two rows, never an overwrite.
- **Clear on every teardown.** `clearStagedDraft` is called synchronously, first thing, in `clearLoadedDraft` (discard), `resolveLoadedDraft` (send), `stashLoadedDraft`, `purgeScopeDrafts`, and `purgePlaintextScopeDrafts` — so a just-sent draft can't be recovered or re-staged by a racing flush (the same resurrection class the resolution guard closes). An emptied editor clears the key too (deletion propagation stays the debounce's job).

### Key design decisions

**1. Plaintext only — E2E drafts deliberately keep today's behavior (E2EE-4).** Staging the composer's `contentJson` writes plaintext to disk. The E2E draft design seals before anything touches disk and stores ciphertext only; there is no synchronous seal (WebCrypto is async). So `saveDraftDebounced` gates staging on `!e2eGateRef.current.enabled`, and `reconcileStagedDrafts` refuses to apply a staged body over a sealed row (`existing?.ciphertext != null`). The reload-loss window stays open for encrypted scratchpads — that is the correct trade per the project precedence (correctness/safety first; no-plaintext-at-rest outranks closing the gap for the minority of E2E scopes). `purgePlaintextScopeDrafts` also clears any stray staged plaintext on E2E composer mount as defense in depth.

**2. No clear-on-flush; content-diff reconcile instead.** The naive "flush localStorage→IDB then clear the key" is a race: a keystroke typed during the async IDB write would be cleared before its own debounce flushes it. Not clearing on flush and resolving staleness by content-diff at load is race-free and avoids comparing timestamps across devices.

**3. Reconcile before the cache seeds, writing IDB directly.** Recovering through the draft-store cache helpers would mark `hasSeededDraftCache` true with a partial set. Writing raw IDB rows before `seedDraftCacheFromIdb` keeps the cache-ready signal honest and gives the composer the content on first paint.

**4. Size cap + total try/catch.** A payload over 256KB is skipped (the debounce still carries it to IDB) so a pathological paste can't jank the synchronous write or exhaust the quota. Every `localStorage` access is wrapped — a quota error or private-mode block must never interrupt typing.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/lib/drafts/draft-staging.ts` | Synchronous `localStorage` staging buffer: `stageDraftContent` / `readStagedDraft` / `clearStagedDraft` / `listStagedDrafts`, plaintext-only, size-capped, fully guarded. |
| `apps/frontend/src/lib/drafts/draft-staging.test.ts` | Unit tests: round-trip, latest-wins, empty clears, oversize skip, corrupt entry, workspace scoping. |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-draft-message.ts` | Stage synchronously in `saveDraftDebounced` (plaintext gate); clear staging at every teardown (`clearLoadedDraft`, `resolveLoadedDraft`, `stashLoadedDraft`, `purgeScopeDrafts`, `purgePlaintextScopeDrafts`). |
| `apps/frontend/src/sync/draft-sync.ts` | `reconcileStagedDrafts` + `applyStagedDraft`: content-diff recovery into IDB, push-before-write for boot-race safety, refuse over sealed rows. |
| `apps/frontend/src/contexts/coordinated-loading-context.tsx` | Run `reconcileStagedDrafts` before `seedDraftCacheFromIdb`; never block the app on a failed recovery. |
| `apps/frontend/src/hooks/use-draft-message.test.ts` | Tests: plaintext stages synchronously before the debounce; E2E never stages; discard/send/stash clear the buffer. |
| `apps/frontend/src/sync/draft-sync.test.ts` | Tests: recover un-flushed tail (preserving baseVersion/attachments), drop already-flushed, create when absent, refuse over sealed, clear on empty. |
| `docs/features/architecture/drafts.md` | New "synchronous staging buffer" subsection + entry-point listing. |

## Out of scope (deliberate)

- **E2E drafts** — no synchronous seal exists, so the reload-loss window stays open for encrypted scopes (see decision 1).
- **Cross-tab live reactivity** on the staging key (the `storage` event) — durability is handled by the startup reconcile; the existing IDB/Dexie-liveQuery story covers same-origin tabs.
- **Deletion propagation via the buffer** — emptying the editor clears the key; propagating a delete remains the debounce's job.

## Test plan

- [x] `apps/frontend` full suite — 3026 pass (`bun run test`)
- [x] `draft-staging.test.ts` (8) + `draft-sync.test.ts` (57) + `use-draft-message.test.ts` (incl. 5 new staging cases) + `coordinated-loading-context.test.tsx` (24) — all pass
- [x] `tsc --noEmit` across the monorepo — clean
- [x] `eslint` + `prettier --check` on changed files — clean
- [ ] Playwright `test:e2e` — not run (needs the full backend/DB harness, unavailable in this environment); the change is covered by the unit/integration suites above

---

https://claude.ai/code/session_01Fge23FT6fenFa2wk867AaQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fge23FT6fenFa2wk867AaQ)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1036"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784576735&installation_id=129946197&pr_number=1036&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1036&signature=69f0a5ad712526133c37f2766a624f634c01b6609a03b79b488734ff6a6f8f35"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/contexts/coordinated-loading-context.tsx
+++ b/apps/frontend/src/contexts/coordinated-loading-context.tsx
@@ -17,6 +17,7 @@ import {
   useWorkspaceUsers,
 } from "@/stores/workspace-store"
 import { hasSeededDraftCache, seedDraftCacheFromIdb } from "@/stores/draft-store"
+import { reconcileStagedDrafts } from "@/sync/draft-sync"
 import { useSyncSnapshot, useSyncStatus } from "@/sync/sync-status"
 import { debugBootstrap, isBootstrapDebugEnabled } from "@/lib/bootstrap-debug"
 import {
@@ -141,9 +142,20 @@ export function CoordinatedLoadingProvider({ workspaceId, streamIds, children }:
 
   useEffect(() => {
     let cancelled = false
-    seedDraftCacheFromIdb(workspaceId).then(() => {
+    // Recover any synchronously-staged composer content into IDB BEFORE seeding
+    // the draft cache, so a draft typed-then-reloaded (before its debounce
+    // reached IDB) is already present when the composer first reads. A failed
+    // recovery must never block the app — seed regardless.
+    const run = async () => {
+      try {
+        await reconcileStagedDrafts(workspaceId)
+      } catch (err) {
+        console.error("Failed to recover staged drafts", err)
+      }
+      await seedDraftCacheFromIdb(workspaceId)
       if (!cancelled) setPrimedDraftWorkspaceId(workspaceId)
-    })
+    }
+    void run()
     return () => {
       cancelled = true
     }
--- a/apps/frontend/src/hooks/use-draft-message.test.ts
+++ b/apps/frontend/src/hooks/use-draft-message.test.ts
@@ -1,6 +1,14 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
 import { renderHook, act, waitFor } from "@testing-library/react"
-import { useDraftMessage, getDraftMessageKey, upsertLoadedDraft, resolveLoadedDraft } from "./use-draft-message"
+import {
+  useDraftMessage,
+  getDraftMessageKey,
+  upsertLoadedDraft,
+  resolveLoadedDraft,
+  clearLoadedDraft,
+  stashLoadedDraft,
+} from "./use-draft-message"
+import { readStagedDraft, stageDraftContent } from "@/lib/drafts/draft-staging"
 import { getScopeResolveSeq, resetDraftResolutionGuard } from "@/sync/draft-resolution-guard"
 import { ContextRefKinds, type JSONContent } from "@threa/types"
 import type { DraftContextRef } from "@/lib/context-bag/types"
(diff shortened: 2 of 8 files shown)


kristofferremback:
### Code review
**Confidence: 7/7** — Excellent (all findings fixed in this session, `a16d03f`)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3 (spec+design, correctness+data-flow, UX) — Mobile skipped (no mobile/layout/touch surface in the diff)

Reviewed the staging-buffer change across three lenses. Six issues surfaced (CodeRabbit independently flagged the top one); all are fixed in `a16d03f`, with two items deliberately left as documented trade-offs.

1. `apps/frontend/src/sync/draft-sync.ts` — `reconcileStagedDrafts` cleared the staging buffer in a `finally`, so a failed IDB write discarded the only un-flushed copy — the exact loss this feature exists to prevent. Now clears **only on success**; a transient failure keeps the buffer to retry next load. _(data-flow; CodeRabbit 🟠 Major — now ✅ resolved)_
2. `apps/frontend/src/hooks/use-draft-message.ts` — `saveDraftDebounced` gated staging on `e2eGateRef`, which trails a prop change by one effect: a one-frame window could stage E2E **plaintext to disk** (E2EE-4). Now gates on the render value `e2eEnabled` (re-binds synchronously); the seal at timer-fire still reads the ref. _(bug / security)_
3. `apps/frontend/src/hooks/use-draft-message.ts` — `restoreStashedDraftToComposer` didn't clear the buffer, so restoring a stash then reloading let the swapped-out draft's staged body overwrite the restored draft. Now clears on restore. _(state-lifecycle)_
4. `apps/frontend/src/lib/drafts/draft-staging.ts` — `listStagedDrafts` iterated a live `localStorage` by index; a concurrent cross-tab write could shift indices and skip an entry. Now snapshots the keys first. _(bug)_
5. `apps/frontend/src/lib/drafts/draft-staging.ts` — `MAX_STAGED_BYTES` measured UTF-16 code units, not bytes (~2× looser than documented). Renamed `MAX_STAGED_CHARS` with accurate accounting. _(design)_
6. `apps/frontend/src/sync/draft-sync.ts` — documented why the enqueue-before-write ordering is deliberate and its crash window benign (orphan op no-ops; the buffer, cleared only on success, retries). _(design / INV-25)_

**Deliberately not changed** (not introduced by this PR): the 256KB cap's silent non-staging (intentional perf guard — the debounce still persists to IDB; a "buffer momentarily absent" indicator would be noise per INV-63) and the E2E reload-loss gap (E2EE-4, no synchronous seal; a UI warning would be scope creep per INV-36).

Verification: full frontend suite **3028 pass**; `tsc` / ESLint / Prettier clean.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>All plan requirements implemented and verified against code: stage-on-change (plaintext-only), reconcile-before-cache-seed writing IDB directly, content-diff recovery, clear-on-teardown across all five paths, E2E exclusion.</details>
<details><summary>🔍 Bugs [2 fixed]</summary>#2 e2eGateRef one-frame stale (E2EE-4); #4 listStagedDrafts live-index iteration.</details>
<details><summary>🔁 Data Flow [2 fixed]</summary>#1 reconcile finally→loss; #3 restore buffer overwrite.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-58 / E2EE-4 honored (no plaintext at rest); INV-25 comment added for the deliberate ordering; INV-36 respected (no speculative UI).</details>
<details><summary>🏗️ Design [2 fixed]</summary>#5 MAX_STAGED_CHARS naming/accounting; #6 enqueue-before-write crash window documented.</details>
<details><summary>🔒 Security [CLEAN]</summary>E2E plaintext-at-rest window closed (#2); no other exposure.</details>
<details><summary>🎛️ UX [1 fixed, 2 deliberate]</summary>#3 restore overwrite fixed; oversize-paste signal + E2E gap left as documented trade-offs.</details>
<details><summary>📱 Mobile [N/A]</summary>No mobile / responsive / touch surface in the diff.</details>

---

PR #1037, opened 2026-06-21
fix(timeline): clear stuck unread divider on cold-loaded short streams (read-frontier late-mount)

kristofferremback:
## Summary

Fixes the bug where a stream that was **marked unread in a prior session** stays stuck with a red unread divider (and unread badge) after a cold app resume, when the whole conversation fits one screen — auto-read never fires and the divider never clears.

**The diagnosis changed after browser verification.** The original commits on this branch (`7a8cce1`, `8fa67e1`) treated the **focus gate** as the cause (relaxing `document.hasFocus()` on touch). Running the real stack + real virtualized timeline showed that is **not** the primary cause: even fully focused (`isFocused: true`, `isActive: true`, `unreadCount: 2`), `lastSeenEventId` **never emits**, so auto-read never fires regardless of focus.

### Root cause (primary)

`useLastSeenEvent`'s observer-attach effect runs when `autoMarkEnabled` flips true (`!isDraft && !isLoading && !isJumpMode`), but **virtua late-mounts its scroller via a ref callback AFTER that**. The effect captured a null element, armed no scroll listener / `ResizeObserver`, and never re-ran (a ref change does not re-run an effect). On a viewport-fitting stream there is no scroll to re-trigger a scan, so the read frontier never advanced → divider stuck. It's a mount-timing race, so it's platform-agnostic; it surfaced on mobile PWA cold-resume where the short-stream + timing conditions coincide.

This is the codebase's own "late-mount" lesson: `useTimelineScroll` already tracks the mounted scroller as **state** (`scrollerEl`) so *its* observers re-arm on mount. The read-frontier scanner was simply never given that signal.

**Fix:** expose `scrollerEl` from `useTimelineScroll` and feed it to `useLastSeenEvent`, whose attach effect now depends on it. The plain (thread) scroller mounts synchronously with its content, so it passes `null`.

### Secondary fix (focus gate, now done right)

After the primary fix, the focus gate is still a *real secondary* gate (verified: a visible-but-unfocused fine-pointer device stays blocked; a coarse phone clears). The provisional approach baked the relaxation into the **shared** `usePageActivity().isActive`, overloading a primitive that `socket-context`, `connection-status`, and `use-app-update` also read.

**Fix:** revert `use-page-activity` to the strict `isActive = isVisible && isFocused`, and compute the relaxed gate **locally** in `useAutoMarkAsRead`, relaxing focus only on a **phone-like** device — coarse pointer **and** a phone-width viewport. A coarse-but-wide device (iPad Split View / Stage Manager) keeps the focus gate, so working in the adjacent pane does not auto-read this one.

## Verification

Browser-verified end-to-end against the real stack + virtualized timeline (cold reload of a short, marked-unread stream):

| Scenario | `lastSeen` emits | gate | marks read |
|----------|:---:|:---:|:---:|
| focused (fine pointer) | yes (after fix) | active | yes |
| unfocused (fine pointer) | yes | blocked | no -> yes once focused |
| unfocused (coarse + narrow) | yes | relaxed | yes |

Before the primary fix, the focused case did **not** emit `lastSeen` — proving the focus relaxation alone would not have fixed the bug.

## Tests

- `use-last-seen-event.test.ts`: re-arms the scan when the scroller late-mounts (`scrollContainerEl` null → element).
- `use-auto-mark-as-read.test.ts`: phone-like (coarse + narrow) unfocused → marks; coarse-but-wide tablet → does not; hidden phone → does not; desktop unfocused → does not. (Hook spies replace the old `matchMedia` poke, since the device hooks now use module-scope subscriptions — INV-48.)

## File changes

**Primary (b):**
- `hooks/use-timeline-scroll.ts` — expose the mounted `scrollerEl` state.
- `hooks/use-last-seen-event.ts` — accept `scrollContainerEl`; attach effect depends on it.
- `components/timeline/stream-content.tsx` — pass the virtualized `scrollerEl` through.

**Secondary (a):**
- `hooks/use-page-activity.ts` — revert to strict `isActive` (net-zero vs `main`).
- `hooks/use-mobile.tsx` — add reactive `useIsCoarsePointer`.
- `hooks/use-auto-mark-as-read.ts` — localized phone-like focus relaxation.

### DIFF
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -981,6 +981,7 @@ export function StreamContent({
     listRef,
     scrollerRef: virtualScrollerRef,
     registerScroller: registerVirtualScroller,
+    scrollerEl: virtualScrollerEl,
     contentRef: virtualContentRef,
     shift,
     isScrolledFarFromBottom: virtualIsScrolledFar,
@@ -1489,6 +1490,12 @@ export function StreamContent({
   const autoMarkEnabled = !isDraft && !isLoading && !isJumpMode
   const { lastSeenEventId, atLastRow, unreadAboveViewport } = useLastSeenEvent({
     scrollContainerRef,
+    // The virtualized scroller late-mounts via a ref callback, AFTER
+    // `autoMarkEnabled` flips true — pass the mounted element so the read-frontier
+    // scan re-arms its observers once the scroller exists. The plain thread
+    // scroller mounts synchronously with the content, so it has no element to
+    // track (the ref is already live when enabled flips).
+    scrollContainerEl: useVirtualized ? virtualScrollerEl : null,
     // Observe the scrolling content box (its height tracks scrollHeight) so a
     // settle / embed / image resize re-scans even when the stream is too short
     // to scroll. Both paths pass one — the virtualized list's inner content div
--- a/apps/frontend/src/hooks/use-auto-mark-as-read.test.ts
+++ b/apps/frontend/src/hooks/use-auto-mark-as-read.test.ts
@@ -126,9 +126,9 @@ describe("useAutoMarkAsRead", () => {
     // user is clearly reading stays unread (a persisted mark-unread divider
     // stuck red across a cold app resume).
     hasFocus = false
-    const matchMediaSpy = vi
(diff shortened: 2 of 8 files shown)


kristofferremback:
@coderabbitai full review

PR #1039, opened 2026-06-21
feat(commands): pick a slash-command argument from a popover instead of typing it

kristofferremback:
## Problem

Slash commands already advertise their argument options to the client. For Pi's `/model`, the linked runtime sends its model list in `capabilities.modelSuggestions`, `availability.ts` injects it into `CommandInfo.args[name="model"].suggestions`, and it rides to the client in the stream bootstrap as `args[].suggestions` (`{ value, label?, description? }`). `/thinking` gets its levels the same way.

But the client never rendered those options. The only consumer of `CommandInfo.args` was the pass-through `args: cmd.args` in `use-command-suggestion.tsx`. Picking `/model` from the slash palette inserted a `/model` chip and left the user to type the value by hand — re-reading a markdown catalog Pi echoes back when the value is missing or ambiguous. The data was delivered and unused.

## Solution

When a command whose first argument carries `suggestions` is picked from the slash palette, open a mention-style popover at the caret listing those options. Picking one (click, or arrow + Enter) inserts the value as text after the command chip, so the message becomes `/model <value>` — which the existing send path already serializes via `serializeToMarkdown(doc)` (the same path `/invite <args>` uses). Typing after the command filters the list; Escape dismisses to free-form typing.

The gap was purely frontend: no backend, send-path, or type changes.

### How it works

1. `use-command-suggestion.tsx` fires a new `onCommandPicked(item)` after the command chip is inserted.
2. `rich-editor.tsx` checks `findPickableArg(item)` — the first arg with non-empty `suggestions`, and only for chip-inserting commands (client-action entries like `/memo`, `/giphy`, `/snippet` insert no chip, so they're excluded) — and calls `openArgPicker(arg)`.
3. `use-command-arg-picker.tsx` captures the caret position right after the inserted `/command ` (chip+space) as an anchor, then tracks the text typed between that anchor and the caret as the filter query (via `editor.on("update"/"selectionUpdate")`), closing when the caret leaves the argument region.
4. Selecting replaces the typed argument text with the chosen `value` and leaves the caret after it, so Enter sends `/command <value>`.
5. `command-arg-picker.tsx` renders the list by reusing the shared `SuggestionList` (the same primitive behind the @mention / #channel / /command popups), so positioning, keyboard nav, and a11y come for free.

### Key design decisions

**1. React-state picker, not a TipTap trigger.** The @/#/`/`/`:` popups are char-triggered `Suggestion` plugins. The option picker has no trigger character — it opens programmatically once a command is picked, after a space. So it can't be a `Suggestion` plugin; it lives in React state and anchors to the caret via `editor.view.coordsAtPos`.

**2. Keyboard routed through `editorProps.handleKeyDown`.** ProseMirror's `someProp` checks `editorProps` before plugin keymaps (this is why the existing "Enter to send" coexists with the `EditorBehaviors` keymap). Routing the picker's keys there lets it preempt send / edit-last-message / blur exactly as the suggestion plugins do for the @/slash popups — without a new extension or a change to `isSuggestionActive`. Keys the picker doesn't own (letters, Left/Right) fall through unchanged.

**3. Generic over the argument contract, not hard-coded to `/model`.** Anything with `args[].suggestions` gets the picker — `/model` and `/thinking` today, future commands for free. No command-name special-casing.

**4. Graceful fallback preserved.** When suggestions are empty (e.g. the runtime is offline) `findPickableArg` returns null and the old type-it-yourself flow is unchanged. Escape, or a query that matches nothing, returns the user to free-form typing.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/editor/triggers/command-arg-picker.tsx` | Presentational option list; reuses `SuggestionList` |
| `apps/frontend/src/components/editor/triggers/use-command-arg-picker.tsx` | Hook: state, caret-anchored query tracking, keyboard routing, value insertion; exports `findPickableArg` / `filterArgSuggestions` |
| `apps/frontend/src/components/editor/triggers/use-command-arg-picker.test.ts` | Unit tests for the two pure helpers |
| `tests/browser/command-arg-picker.spec.ts` | Browser E2E (injects a `/model` command with suggestions via bootstrap route interception) |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/editor/triggers/use-command-suggestion.tsx` | Adds `onCommandPicked`, fired after the chip is inserted |
| `apps/frontend/src/components/editor/rich-editor.tsx` | Wires the picker: `useCommandArgPicker`, the `onCommandPicked` handler, the keyboard branch in `editorProps.handleKeyDown`, and renders the portal |
| `apps/frontend/src/components/editor/triggers/index.ts` | Barrel exports for the two new modules |

## Out of scope (deliberate)

- No backend change. The options already exist on the wire; this only renders them. `/model`'s suggestions still require a linked Pi runtime advertising `modelSuggestions` — absent one, the command either doesn't surface or has no suggestions and the picker doesn't open.
- No auto-send on select. Selecting inserts the value and leaves the caret; the user presses Enter to send, matching the existing "type the command, press send" UX.

## Test plan

- [x] `bunx vitest run src/components/editor/triggers/ src/lib/commands.test.ts` — 50 pass (7 files; includes the 11 new arg-picker helper tests)
- [x] `bunx playwright test tests/browser/command-arg-picker.spec.ts` — 2 pass in real Chromium: open → live filter (`claude` hides GPT-5) → ArrowDown + Enter selects (preempting send) → value inserted as `/model anthropic/claude-sonnet-4`; and Escape dismisses, leaving `/model` for free-form typing
- [x] `tsc --noEmit` (frontend) clean; the commit's pre-commit hook also ran the monorepo typecheck, astro check, and API-docs check clean
- [x] `eslint` + `prettier --check` clean on all changed files

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1039"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784620363&installation_id=129946197&pr_number=1039&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1039&signature=9f30000837a548853674990cbf948f8f7b64998a0145c28b49e1b9acf1128b9f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
@coderabbitai review

PR #1041, opened 2026-06-21
feat(claude-code-channel): send interim messages + idle (not absolute) turn timeout

kristofferremback:
## Problem

A Claude Code channel turn (`extensions/claude-code-remote/`) could only ever emit **one** message — the single terminal `reply` that also closes the bot-invocation — and was force-closed at a hard **30-minute** wall regardless of whether Claude was still working.

The wall was `onReplyTimeout` in `channel-server.ts`: when the channel claims a message it sets `setTimeout(…, replyTimeoutMs)` (`config.ts` default `1_800_000`). The timer is **absolute from claim time** and nothing resets it, so a turn that runs longer than 30 minutes is reaped and the scratchpad gets `_Claude Code ended the turn without sending a reply._` — even mid-work. That's the "finished at exactly 30 minutes, flat" the report described. The claim lease itself is renewed continuously (`startRenewTimer`, every `CLAIM_TTL_SECONDS/3`), so the lease was never the problem — only this local timer.

Two limitations, one root cause: the turn model was rigidly one-invocation → one-`reply`, with a fixed deadline.

## Solution

Split outbound into two tools and turn the deadline into an *idle* timer instead of an absolute one.

- **`send(invocation_id, text)`** posts a progress/intermediate message and leaves the request open. Callable any number of times per turn.
- **`reply(invocation_id, text)`** is unchanged: posts the final message and closes the invocation.

Because the channel hears Claude only through its tool calls, each `send` doubles as a heartbeat that pushes the idle deadline out. The two features are the same fix: a long turn that streams progress can't be reaped, and the user sees it working instead of waiting on a silent 30 minutes.

No backend change — both wire paths already exist and are already used by this extension: `send` posts via `POST /streams/:id/messages` (the path the permission relay already uses; `sendMessageSchema` accepts `{ content, clientMessageId?, metadata? }`), and the silent-close path uses `complete`'s existing `noResponse: true`.

### How it works

1. **`send` tool** (`handleSend`) — looks up the in-flight invocation, runs the text through `uploadReplyAttachments` (so `THREA_ATTACH:` works in interim messages too), and `client.sendMessage(responseStreamId, …)`. The invocation stays in flight, so presence stays `busy`. Returns an error (like `reply`) if the id isn't open.
2. **Idle timeout** — `scheduleIdleTimeout` / `touchIdleTimeout` replace the inline `setTimeout`. `handleSend` calls `touchIdleTimeout` after a successful post; `handlePermissionRequest` resets the deadline of the turn whose `responseStreamId` matches the prompt's target stream (a pending tool approval is also a sign of life). Default raised 30 → 60 min: `idleTimeoutMs` / `THREA_IDLE_TIMEOUT_MS`, `3_600_000`, clamped to a 60s floor.
3. **Silent close when it already spoke** — `onReplyTimeout` now branches on `sentCount`: if the turn posted at least one interim message, it closes with `{ noResponse: true }` (no message) instead of the misleading "ended without a reply" notice; an entirely silent turn still gets the notice.
4. **Instructions** (`buildInstructions`, which land in Claude's system prompt) now describe both tools: `send` freely for progress, `reply` once to finish.

### Key design decisions

**1. Idle timer, not "raise the number" or "remove the timer."** The timeout exists for a real wedge — Claude ends a turn without `reply`, leaving presence `busy` and the next message queued behind it forever (`claimDrain` breaks while `inflight.size > 0`). Removing it reintroduces that wedge; a bigger absolute number just moves the guillotine. An idle timer keeps the wedge-recovery while never reaping an active turn. 60 min balances the two costs: long enough for a silent operation, short enough that a genuinely-forgotten `reply` unblocks the queue within an hour.

**2. `send` is additive; `reply` keeps its exact semantics.** Existing behaviour and the "always end with `reply`" instruction are preserved, so nothing regresses. If Claude `send`s its answer but forgets `reply`, decision #3 makes that benign — the content is already delivered and the turn closes quietly on idle.

**3. Stable per-send `clientMessageId` (`ccsend-<invocationId>-<seq>`), committed only on success.** `seq = sentCount + 1`; `entry.sentCount` advances only after the post resolves. A retry after a network failure reuses the same id (backend can dedupe) rather than double-posting.

**4. Config rename, no alias (INV-49).** `replyTimeoutMs` / `THREA_REPLY_TIMEOUT_MS` → `idleTimeoutMs` / `THREA_IDLE_TIMEOUT_MS`, since the semantics changed from "time to reply" to "time idle". This is a deliberate breaking rename: a user with the **old** env var set would fall back to the 60-min default until they update it. Acceptable for a weeks-old experimental extension with a single operator; flagged here rather than carrying a deprecated alias.

## Honest limit (documented)

A channel can't `send` a heartbeat *during* a single blocking tool call — Claude is awaiting the tool. So a 40-minute uninterrupted `Bash` run still needs the idle window to exceed it; `THREA_IDLE_TIMEOUT_MS` is the knob. This is called out in both the README and `docs/claude-code-channel.md`. (Minor: a `send` whose text is *only* a `THREA_ATTACH:` line inherits `uploadReplyAttachments`'s `"Done."` fallback for the empty body — harmless, not worth a branch.)

## Modified files

| File | Change |
| ---- | ------ |
| `extensions/claude-code-remote/src/channel-server.ts` | Add `send` tool + `handleSend`; `scheduleIdleTimeout`/`touchIdleTimeout` helpers; `Inflight.sentCount`; idle-reset on send + permission prompt; `onReplyTimeout` silent-close branch; two-tool instructions |
| `extensions/claude-code-remote/src/config.ts` | Rename `replyTimeoutMs`→`idleTimeoutMs`, `THREA_REPLY_TIMEOUT_MS`→`THREA_IDLE_TIMEOUT_MS`, default 30→60 min, doc the inactivity semantics |
| `extensions/claude-code-remote/src/channel-server.test.ts` | Tests for `send` (post + keep-open, unknown-invocation error), both `onReplyTimeout` branches, and the two-tool instructions |
| `extensions/claude-code-remote/README.md` | "Sending multiple messages per turn" section; config table + limitations updated |
| `docs/claude-code-channel.md` | "Outbound: send for progress, reply to finish" + "The idle-timeout safety net" rewrite; comparison-table row |

## Out of scope (deliberate)

- **No turn-end detection.** The Claude Code channel protocol sends the channel no "turn complete" / idle notification (confirmed against the channels reference) — the heartbeat *must* come from Claude's own tool calls, which is why the idle window is the backstop rather than a precise close.
- **Per-tool trace** stays a single "working" step — unchanged; still the standing parity gap with `pi-remote`.

## Test plan

- [x] `extensions/claude-code-remote` — `bun test`: 55 pass, 0 fail (5 new)
- [x] `extensions/claude-code-remote` — `bunx tsc --noEmit` clean
- [x] Pre-commit hooks ran repo-wide on the staged files: prettier, all-app eslint, `astro check`, dockerfile/migration checks, and full-monorepo `tsc --noEmit` — all green
- [ ] No live MCP harness drives the real Claude Code ↔ channel handshake; the new tests are unit-level against `handleSend`/`onReplyTimeout` via a fake `ThreaClient`, matching the existing test style for this extension

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1041"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784626464&installation_id=129946197&pr_number=1041&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1041&signature=ec4d914fd6ffa7f16e38e774b4b7365402fb4bf04da6a8c53574673d501859a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/claude-code-channel.md
+++ b/docs/claude-code-channel.md
@@ -76,15 +76,20 @@ A claim holds a lease (`claimTtlSeconds`, max 300) that expires if not renewed, 
 
 The channel handles one turn at a time. A message you send while Claude is still working waits in Threa until the current turn completes, then gets claimed and pushed. The one exception is a permission verdict (below), which is pulled through immediately so a blocked turn can continue.
 
-## Outbound: the reply tool completes the invocation
+## Outbound: send for progress, reply to finish
 
-The channel exposes one tool, `reply(invocation_id, text)`. Claude is told (via the server's `instructions`, which land in its system prompt) to call it exactly once per `<channel>` event. When Claude calls it:
+The channel exposes two output tools, and Claude is told (via the server's `instructions`, which land in its system prompt) how they differ:
 
-1. The channel looks up the in-flight invocation by `invocation_id`.
-2. It calls `POST /bot-invocations/:id/complete` with `finalMessageMarkdown: text`. That posts the reply into the scratchpad and closes the invocation.
-3. It clears the in-flight entry, and when nothing is in flight, flips presence back to `available`.
+- **`send(invocation_id, text)`** posts a progress or intermediate message and leaves the request open. Claude can call it any number of times during a turn. It posts via `POST /streams/:id/messages` (the same path the permission relay uses) — the invocation stays in flight, so presence stays `busy`.
+- **`reply(invocation_id, text)`** posts the final message and closes the request. Claude calls it once, last. It calls `POST /bot-invocations/:id/complete` with `finalMessageMarkdown: text`, which posts the reply into the scratchpad and closes the invocation; the channel then clears the in-flight entry and, when nothing is in flight, flips presence back to `available`.
 
-The `reply` tool is the only path back to Threa, which is why the instructions insist on it. As a safety net, an in-flight invocation that is never answered is force-closed after `replyTimeoutMs` (default 30 minutes) with a short notice, so a session can't get wedged in `busy` forever.
+Splitting the two means a long turn isn't silent until a single terminal reply — it can stream updates as it works and finish with a summary. This is the only path back to Threa, which is why the instructions insist on ending with `reply`.
+
+### The idle-timeout safety net
+
+A turn that goes silent without a `reply` would otherwise wedge the session in `busy` forever (the next message can't be claimed while one is in flight). So an in-flight invocation is force-closed after `idleTimeoutMs` (default 60 minutes, `THREA_IDLE_TIMEOUT_MS`) of **inactivity**. It's an _idle_ timer, not an absolute one: every `send` and every tool-approval prompt resets it, so an actively-working turn is never reaped — only one that has genuinely gone quiet. If the reaped turn had already `send`-ed something, it closes silently (`noResponse`); otherwise it posts a short "ended without a reply" notice.
+
+The one case the heartbeat can't cover is a single long-running tool call (a 40-minute test run): Claude is blocked awaiting the tool and can't `send` in the middle of it. The idle window must therefore exceed the longest single operation — raise `THREA_IDLE_TIMEOUT_MS` if a turn does heavy uninterrupted work. (This replaced an earlier fixed 30-minute reply timeout that force-closed long turns at exactly 30 minutes regardless of whether they were still working.)
 
 ## Attachments
 
@@ -117,7 +122,7 @@ The local terminal dialog stays open the whole time, so whichever answer arrives
 | ----------------------------------------- | --------------------------------------------------------- | -------------------------------------------------------------------- |
 | Where it runs                             | Inside Pi, via Pi's extension API                         | A subprocess Claude Code spawns over stdio                           |
 | How it gets a prompt                      | Claims invocation, injects via `pi.sendUserMessage`       | Claims invocation, pushes a `notifications/claude/channel` event     |
-| How it returns a reply                    | Hooks Pi's `agent_end`, posts the captured assistant text | Claude calls the `reply` tool, which completes the invocation        |
+| How it returns a reply                    | Hooks Pi's `agent_end`, posts the captured assistant text | Claude calls `send` for progress and `reply` to finish the turn      |
 | Trace detail                              | Rich per-tool steps (it hooks Pi's tool events)           | One "working" step (a channel can't see Claude's tool calls)         |
 | Session control (`/compact`, `/model`, …) | Yes                                                       | No (a channel can't drive Claude's session)                          |
 | Permission prompts                        | N/A (Pi runs the model in-process)                        | Relayed into the scratchpad as messages                              |
--- a/extensions/claude-code-remote/README.md
+++ b/extensions/claude-code-remote/README.md
@@ -136,7 +136,7 @@ The channel uploads the file (one per `THREA_ATTACH:` line; paths resolve agains
 | `THREA_DEFAULT_LABEL`      | `defaultLabel`     | (none)                 | Label applied to scratchpads this channel creates (only on first creation, not re-link) |
 | `THREA_PERMISSION_RELAY`   | `permissionRelay`  | `true`                 | Relay tool-approval prompts into the scratchpad                                         |
 | `THREA_POLL_MS`            | `pollMs`           | `3000`                 | Backstop claim poll (the socket pushes faster)                                          |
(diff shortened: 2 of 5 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (one real issue, fixed during review)
**Review models:** Orchestrator: Claude Opus 4.8 (1M context) | Reviewers: sonnet x2 (spec+design, correctness+data-flow)

Backend/extension-only diff (`extensions/claude-code-remote/`), so the UX and Mobile lenses were not run.

Found 1 issue — **fixed in `12b77fc2`** before this report:

1. `extensions/claude-code-remote/src/channel-server.ts` — `handleSend` (state-lifecycle, 80) — **fixed.** `handleSend` keeps the turn open across its `await uploadReplyAttachments` / `await sendMessage`, so — unlike `handleReply`, which clears its deadline up front specifically to dodge this — the idle timer could fire mid-post: `onReplyTimeout` removes the in-flight entry and completes the turn while the send is in flight. The original code then mutated the dead entry and returned a clean `"sent"`, misleading Claude into a follow-up `reply` that 404s. Low probability (the 60-min deadline must land during a post) but a real wedge. Fix: re-fetch the entry after the awaits; if it's gone, return an `isError` saying the message posted but the turn already closed for inactivity, instead of `"sent"`. Regression test added (`reports a closed turn when the idle timeout fires mid-send`).
   https://github.com/threahq/threa/blob/12b77fc22cfe72b0733c4d873f7bd2ad6d126ace/extensions/claude-code-remote/src/channel-server.ts#L435-L456

---

All three stated goals delivered and verified: (a) a `send` doesn't end the turn (`handleSend` never calls `complete`), (b) multiple messages per turn (callable repeatedly), (c) long-running sessions (idle timer resets on `send` + permission prompt). `send` tool is registered, routed, and documented in `buildInstructions`. The `replyTimeoutMs`→`idleTimeoutMs` rename is applied consistently across `config.ts`, `channel-server.ts`, and the README.
</details>
<details><summary>🔍 Bugs [1 found, fixed]</summary>

The mid-send idle-timeout race above. Timer lifecycle otherwise sound: `handleReply` (success + failure-re-arm), `onReplyTimeout`, and `shutdown` all clear the deadline before/around removing the entry, so no double-complete and no orphaned timers. The permission-relay `touchIdleTimeout` loop matches the right turn by `responseStreamId` (only one entry is ever in flight; a verdict invocation is `complete`d early and never added to `inflight`).
</details>

`Inflight.sentCount` is monotonic and committed only after a successful post (`seq = sentCount + 1`, assigned post-await), so a failed-then-retried `send` reuses the same `ccsend-<id>-<seq>` client id and the backend dedupes rather than double-posting. `onReplyTimeout` correctly branches on `sentCount` (silent `noResponse` vs. the notice). The re-arm path in `handleReply` carries `sentCount` and `prepared` forward.
</details>

INV-49 (no deprecated aliases after renames): the env-var rename ships with no alias — that's what the invariant *wants*, not a violation. INV-11 (no silent fallback): the `noResponse` silent close is a safe close after the turn verifiably communicated, not a masked misconfiguration. New comments (INV-25) are load-bearing (why the idle timer exists, why the silent close is correct).
</details>

`handleSend` mirrors `handleReply`'s established shape. One sub-threshold (~70) note: if `uploadReplyAttachments` succeeds but `sendMessage` fails, a `send` retry re-uploads and orphans the first attachment copy — `handleReply` avoids this by caching `prepared`, `handleSend` does not. Bounded (blob, not domain data) and rare; left as-is for now.
</details>

No injection/auth/secret/exposure issues. `clientMessageId` is built from a server-issued invocation id; env vars are trusted; no backend surface changed.
</details>

PR #1047, opened 2026-06-21
fix(sidebar): hide threads rooted in archived streams

kristofferremback:
## Problem

Archiving a stream only sets `archivedAt` on **that single row** — its thread descendants stay "active" in the DB. The sidebar's `processedStreams` filter only checked `stream.archivedAt` on each stream itself, so threads nested under an archived scratchpad/channel/top-level stream kept showing up in the Recent / Important / Everything-else buckets. The result was a sidebar cluttered with threads rooted in streams the user had already archived.

A frontend-only filter is not enough on its own: the workspace bootstrap excludes archived streams, so an archived root is never resident in the client cache on cold load, and `cleanupStaleEntities` prunes it on reconnect. The frontend therefore has no signal that a thread's root is archived in steady state — the filter's `archivedStreamIds` set is empty and the threads resurface. (Caught in self-review / unified review on the first push; see commit 2.)

## Solution

Two layers, each covering what the other can't:

1. **Backend (load-bearing):** `StreamRepository.listWithPreviews` — the workspace bootstrap's stream query — now excludes active threads whose `root_stream_id` references an archived stream (`EXISTS` subquery against the root). This covers cold load (threads never sent) and reconnect (threads not in the bootstrap set → pruned by `cleanupStaleEntities`). `listWithPreviews` is bootstrap-only, so no other surface is affected.
2. **Frontend (live-archive window):** the sidebar builds an `archivedStreamIds` set from cached streams and rejects any stream whose `rootStreamId` is archived. This hides threads in the transient window between the `stream:archived` socket event and the next reconnect, before the backend pruning takes effect. The visibility predicate is extracted into a pure, tested `isSidebarStreamVisible` helper.

### How it works

- Backend: `EXCLUDE_ARCHIVED_ROOT_THREADS` is AND-ed into every `listWithPreviews` WHERE branch — `AND NOT (s.type = 'thread' AND s.root_stream_id IS NOT NULL AND EXISTS (SELECT 1 FROM streams root WHERE root.id = s.root_stream_id AND root.archived_at IS NOT NULL))`. A thread's `root_stream_id` is its top-level non-thread ancestor (INV-62), so one lookup covers any nesting depth.
- Frontend: `isSidebarStreamVisible(stream, memberStreamIds, archivedStreamIds)` returns false when the stream is archived, when its `rootStreamId` is archived, or when it's a public stream the viewer isn't a member of. Non-public streams always appear (bootstrap already gates by access).

### Key design decisions

**Backend exclusion over "include archived streams in the bootstrap."** The alternative was to include archived streams in the bootstrap so the frontend `archivedStreamIds` set is populated. Rejected because `useWorkspaceStreams` has ~20 consumers (activity, memory, share picker, quick switcher, channel-link resolver, stream settings, …) — only some filter `archivedAt` today. Widening the cache contract from "active streams" to "all streams" risks surfacing archived streams in surfaces that don't filter. Excluding the specific threads under archived roots is surgical: `idbStreams` stays "active streams" (minus those threads), no other consumer sees a change, and the active roots/threads under them are untouched.

**Keep the frontend filter despite the backend fix.** The backend handles cold load + reconnect. The frontend filter is still needed for the live-archive window: when a root is archived via socket, its threads are already in IDB and won't be pruned until the next reconnect. `handleStreamArchived` keeps the root row in IDB (sets `archivedAt`), so `archivedStreamIds` is populated and the filter hides the threads immediately. Belt-and-suspenders, each covering a different timing.

**`root_stream_id` (top-level), not ancestor walk.** Archiving only ever marks the root row in practice, and the user's complaint is threads rooted in archived top-level streams. `root_stream_id` points at the nearest non-thread ancestor, so a single EXISTS covers any nesting depth without walking `parent_stream_id` chains. A thread whose *intermediate parent* is archived but whose *root* is active still shows — not the reported case, and consistent with "root-archived" semantics.

**Unarchive lifecycle.** Archiving a root prunes it (and now its threads) on reconnect; unarchiving re-adds the root via `handleStreamUnarchived`. The threads reappear on the next bootstrap. This matches the pre-existing behavior for the archived root itself (`handleStreamUnarchived` uses `db.streams.update`, a no-op on missing rows), so threads now follow the same restore-on-next-sync lifecycle as their root — consistent, not a new regression.

## Files changed

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/streams/repository.ts` | `listWithPreviews` excludes active threads whose `root_stream_id` is archived (all 4 WHERE branches); import `StreamTypes`. |
| `apps/backend/tests/e2e/sidebar-archived-roots.test.ts` | E2E: thread under archived root omitted from bootstrap, thread under active root kept. Verified it fails without the SQL fix. |
| `apps/frontend/src/components/layout/sidebar/utils.ts` | Add `isSidebarStreamVisible(stream, memberStreamIds, archivedStreamIds)` predicate; import `Visibilities`. |
| `apps/frontend/src/components/layout/sidebar/sidebar.tsx` | Compute `archivedStreamIds` memo; filter `processedStreams` through `isSidebarStreamVisible`; add the memo to the dep array. |
| `apps/frontend/src/components/layout/sidebar/utils.test.ts` | Cover the predicate: archived self, archived root (any depth), active root, public member/non-member, non-public. |

## Out of scope (deliberate)

- `handleStreamUnarchived` uses `db.streams.update` (a no-op when the row was pruned), so an unarchived root — and now its threads — only reappear on the next workspace bootstrap, not instantly. This is pre-existing behavior for the root; making unarchive restore immediately (via `put` + a thread refetch) is a separate follow-up.
- A thread whose intermediate parent is archived but whose root is active still shows. Not the reported case.
- The `GET /streams` list endpoint (`streamService.list`) is unchanged; it serves a different surface with its own `archiveStatus` filter.

## Test plan

- [x] E2E `bun test tests/e2e/sidebar-archived-roots.test.ts` — 1 pass (fails without the SQL fix, confirming it exercises the change)
- [x] E2E `tests/e2e/threads.test.ts` + `tests/e2e/search.test.ts` — 32 pass, no regressions
- [x] Frontend `vitest run src/components/layout/sidebar/` — 143 pass (12 files, 6 new)
- [x] Backend unit `bun test src/features/streams/{repository,service,handlers}.test.ts` — 73 pass
- [x] `tsc --noEmit` (backend + tests, frontend) clean
- [x] Pre-commit hooks (eslint + full monorepo typecheck + OpenAPI doc check) green on both commits
- [ ] Full e2e suite not run serially locally — hits the 300 req/60s rate limiter; CI runs it via `test-silent` with proper isolation. Targeted suites above cover the touched paths.

<details>

**Goal:** Threads rooted in archived streams (any nesting depth) must not appear in the sidebar — across cold load, reconnect, and live archive.

**What was built:**
- Backend: `listWithPreviews` (workspace bootstrap) excludes active threads whose `root_stream_id` is archived.
- Frontend: `isSidebarStreamVisible` predicate hides archived streams and threads whose `rootStreamId` is archived; `archivedStreamIds` memo feeds it.
- E2E + unit tests.

**Design decisions:** backend exclusion (surgical) over widening the bootstrap cache contract (risky for ~20 `useWorkspaceStreams` consumers); keep frontend filter for the live-archive window; resolve via `root_stream_id` (INV-62) not an ancestor walk.

**Design evolution (self-review):** First push was frontend-only. Unified review (confidence 2/7) caught that the frontend filter is ineffective in steady state — the workspace bootstrap excludes archived streams so `archivedStreamIds` is empty on cold load and after reconnect pruning. Fixed by adding the backend exclusion (commit 2), which is the load-bearing change; the frontend filter is retained only for the live-archive transient window.

**Schema changes:** none.

**What's NOT included:** instant restore on unarchive (pre-existing `db.streams.update` no-op issue); intermediate-parent-archived case; `GET /streams` endpoint changes.

**Status:** complete, both commits pushed, pre-commit hooks green.

</details>

---

🤖 _PR by [Pi](https://github.com/earendil-works/pi-coding-agent)_

### DIFF
--- a/apps/backend/src/features/streams/handlers.ts
+++ b/apps/backend/src/features/streams/handlers.ts
@@ -902,12 +902,20 @@ export function createStreamHandlers({
       // value. See `writeBootstrapEventsAndStream` in stream-sync.ts.
       const snapshotAt = new Date().toISOString()
 
-      const [members, botMemberIds, membership, latestSequence, activityCounts] = await Promise.all([
+      const [members, botMemberIds, membership, latestSequence, activityCounts, rootStream] = await Promise.all([
         streamService.getMembers(streamId),
         streamService.getBotMemberIds(workspaceId, streamId),
         streamService.getMembership(streamId, userId),
         eventService.getLatestSequence(streamId),
         activityService?.getUnreadCountsForStream(userId, workspaceId, streamId),
+        // For a thread, fetch the root so the bootstrap can surface
+        // `rootArchivedAt` — archiving marks only the root row, so the thread's
+        // own `archivedAt` can't tell the client it is sealed. No-op (null) for
+        // non-threads and threads whose root is missing (dangling root_stream_id
+        // is possible under INV-1's FK-less schema).
+        stream.type === StreamTypes.THREAD && stream.rootStreamId
+          ? streamService.getStreamById(stream.rootStreamId)
+          : Promise.resolve(null),
       ])
       const botRuntimePresences = await botRuntimeService.findLatestPresences({ workspaceId, botIds: botMemberIds })
       const commands = await commandAvailabilityService.listStreamCommands({ workspaceId, userId, streamId })
@@ -985,6 +993,7 @@ export function createStreamHandlers({
       res.json({
         data: {
           stream,
+          rootArchivedAt: rootStream?.archivedAt ?? null,
           events: eventsWithLinkPreviews,
           sharedMessages,
           members,
--- a/apps/backend/src/features/streams/repository.ts
+++ b/apps/backend/src/features/streams/repository.ts
@@ -333,6 +333,24 @@ export const StreamRepository = {
   },
 
   /**
+   * Active thread ids whose top-level root is `rootStreamId` (any nesting
+   * depth — `root_stream_id` points at the non-thread ancestor, INV-62).
+   * Used to route root lifecycle events (`stream:archived` / `stream:unarchived`)
+   * to descendant thread rooms so clients viewing a thread receive them and
+   * resolve the inherited archived state live, without a refresh.
+   */
+  async listThreadIdsByRoot(db: Querier, workspaceId: string, rootStreamId: string): Promise<string[]> {
+    const result = await db.query<{ id: string }>(
+      sql`SELECT id FROM streams
+          WHERE workspace_id = ${workspaceId}
+            AND root_stream_id = ${rootStreamId}
+            AND type = ${StreamTypes.THREAD}
+            AND archived_at IS NULL`
+    )
+    return result.rows.map((row) => row.id)
+  },
+
+  /**
    * List streams by a known set of IDs with optional filtering.
    * Used by the public API to fetch accessible stream details.
    */
--- a/apps/backend/src/features/streams/repository.ts
+++ b/apps/backend/src/features/streams/repository.ts
@@ -9,6 +9,7 @@ import type {
   ThreadSummary,
   E2eActor,
 } from "@threa/types"
+import { StreamTypes } from "@threa/types"
 import { parseArchiveStatusFilter, type ArchiveStatus } from "../../lib/sql-filters"
 
 export type { StreamType, Visibility, CompanionMode, MemoryMode, ArchiveStatus }
@@ -535,6 +536,21 @@ export const StreamRepository = {
 
     const { includeActive, includeArchived, filterAll } = parseArchiveStatusFilter(archiveStatus)
 
+    // A thread inherits its top-level ancestor via `root_stream_id` (INV-62).
+    // Archiving marks only the root row, so an active-only filter still returns
+    // a thread whose root scratchpad/channel was archived — and the sidebar
+    // (this bootstrap's consumer) would show it, since the frontend can't tell
+    // the root is archived once it's pruned from the client cache. Exclude
+    // active threads whose root is archived so they never enter the sidebar.
+    // Applies in every branch because `listWithPreviews` is workspace-bootstrap
+    // only, and a thread under an archived root is clutter in every view that
+    // reads the bootstrap streams.
+    const EXCLUDE_ARCHIVED_ROOT_THREADS = `AND NOT (
+        s.type = '${StreamTypes.THREAD}'
+        AND s.root_stream_id IS NOT NULL
+        AND EXISTS (SELECT 1 FROM streams root WHERE root.id = s.root_stream_id AND root.archived_at IS NOT NULL)
+      )`
+
     // CTE to get last message per stream. LEFT JOIN against
     // `e2e_streams` so the workspace bootstrap surfaces the E2E flag
     // inline — without it, cold-loaded sidebar rows have undefined
@@ -584,6 +600,7 @@ export const StreamRepository = {
                 AND s.type = ANY(${types})
                 AND (${filterAll} OR (${includeArchived} AND s.archived_at IS NOT NULL) OR (${!includeArchived} AND s.archived_at IS NULL))
                 AND (s.visibility = 'public' OR s.id = ANY(${userMembershipStreamIds}))
+                ${sql.raw(EXCLUDE_ARCHIVED_ROOT_THREADS)}
               ORDER BY COALESCE(lm.created_at, s.created_at) DESC`
         )
         return result.rows.map(mapRowToStreamWithPreview)
@@ -594,6 +611,7 @@ export const StreamRepository = {
             WHERE s.workspace_id = ${workspaceId}
               AND (${filterAll} OR (${includeArchived} AND s.archived_at IS NOT NULL) OR (${!includeArchived} AND s.archived_at IS NULL))
               AND (s.visibility = 'public' OR s.id = ANY(${userMembershipStreamIds}))
+              ${sql.raw(EXCLUDE_ARCHIVED_ROOT_THREADS)}
             ORDER BY COALESCE(lm.created_at, s.created_at) DESC`
       )
       return result.rows.map(mapRowToStreamWithPreview)
@@ -605,6 +623,7 @@ export const StreamRepository = {
             WHERE s.workspace_id = ${workspaceId}
               AND s.type = ANY(${types})
               AND (${filterAll} OR (${includeArchived} AND s.archived_at IS NOT NULL) OR (${!includeArchived} AND s.archived_at IS NULL))
+              ${sql.raw(EXCLUDE_ARCHIVED_ROOT_THREADS)}
             ORDER BY COALESCE(lm.created_at, s.created_at) DESC`
       )
       return result.rows.map(mapRowToStreamWithPreview)
@@ -614,6 +633,7 @@ export const StreamRepository = {
       sql`${sql.raw(SELECT_WITH_PREVIEW)}
           WHERE s.workspace_id = ${workspaceId}
             AND (${filterAll} OR (${includeArchived} AND s.archived_at IS NOT NULL) OR (${!includeArchived} AND s.archived_at IS NULL))
+            ${sql.raw(EXCLUDE_ARCHIVED_ROOT_THREADS)}
           ORDER BY COALESCE(lm.created_at, s.created_at) DESC`
     )
     return result.rows.map(mapRowToStreamWithPreview)
--- a/apps/backend/src/features/streams/service.test.ts
+++ b/apps/backend/src/features/streams/service.test.ts
@@ -334,6 +334,41 @@ describe("StreamService.resolveWritableMessageStream", () => {
     expect((error as HttpError).status).toBe(403)
     expect((error as HttpError).message).toBe("Not a member of this stream")
   })
+
+  test("should throw 403 when a thread's root stream is archived", async () => {
+    const getStreamByIdSpy = spyOn(service, "getStreamById")
+    // First call resolves the target thread (active itself); second call
+    // resolves its root, which is archived — the thread inherits the seal.
+    getStreamByIdSpy
+      .mockResolvedValueOnce({
+        id: "stream_thread",
+        workspaceId: "ws_1",
+        type: "thread",
+        rootStreamId: "stream_root",
+        archivedAt: null,
+      } as never)
+      .mockResolvedValueOnce({
+        id: "stream_root",
+        workspaceId: "ws_1",
+        type: "scratchpad",
+        archivedAt: new Date(),
+      } as never)
+    const isMemberSpy = spyOn(service, "isMember").mockResolvedValue(true)
+
+    const error = await service
+      .resolveWritableMessageStream({
+        workspaceId: "ws_1",
+        userId: "usr_1",
+        target: { streamId: "stream_thread" },
+      })
+      .catch((e) => e)
+
+    expect(error).toBeInstanceOf(HttpError)
+    expect((error as HttpError).status).toBe(403)
+    expect((error as HttpError).message).toBe("Cannot send messages to a thread under an archived stream")
+    // Membership is never checked once the root-archived seal is detected.
+    expect(isMemberSpy).not.toHaveBeenCalled()
+  })
 })
 
 describe("StreamService.findOrCreateDm", () => {
--- a/apps/backend/src/features/streams/service.ts
+++ b/apps/backend/src/features/streams/service.ts
@@ -302,6 +302,19 @@ export class StreamService {
       throw new HttpError("Cannot send messages to an archived stream", { status: 403 })
     }
 
+    // A thread inherits its lifecycle from its root (INV-62): archiving marks
+    // only the root row, so the thread itself stays "active" and would otherwise
+    // still accept writes. Reject when the thread's root is archived so an
+    // archived scratchpad/channel seals every nested thread too. The sidebar
+    // already hides these (listWithPreviews excludes them), but this is the
+    // authoritative gate covering deep links, the API, and move-to-thread.
+    if (stream.type === StreamTypes.THREAD && stream.rootStreamId) {
+      const root = await this.getStreamById(stream.rootStreamId)
+      if (root?.archivedAt) {
+        throw new HttpError("Cannot send messages to a thread under an archived stream", { status: 403 })
+      }
+    }
+
     const isMember = await this.isMember(stream.id, params.userId)
     if (!isMember) {
       throw new HttpError("Not a member of this stream", { status: 403 })
--- a/apps/backend/src/features/streams/service.ts
+++ b/apps/backend/src/features/streams/service.ts
@@ -743,10 +743,21 @@ export class StreamService {
           actorType: "user",
         })
 
+        // Route to the root's room AND its descendant thread rooms: a thread
+        // viewer only joins the thread's room, so without the thread ids in
+        // the payload it would never learn the root was archived and the
+        // composer would stay live until a refresh. Threads inherit access
+        // from the root (INV-62), so their rooms reach the same audience.
+        const threadStreamIds =
+          stream.type === StreamTypes.THREAD
+            ? []
+            : await StreamRepository.listThreadIdsByRoot(client, stream.workspaceId, stream.id)
+
         await OutboxRepository.insert(client, "stream:archived", {
           workspaceId: stream.workspaceId,
           streamId: stream.id,
           stream,
+          threadStreamIds,
         })
       }
       return stream
@@ -771,6 +782,10 @@ export class StreamService {
           workspaceId: stream.workspaceId,
           streamId: stream.id,
           stream,
+          threadStreamIds:
+            stream.type === StreamTypes.THREAD
+              ? []
+              : await StreamRepository.listThreadIdsByRoot(client, stream.workspaceId, stream.id),
         })
       }
       return stream
--- a/apps/backend/src/features/streams/service.ts
+++ b/apps/backend/src/features/streams/service.ts
@@ -732,7 +732,7 @@ export class StreamService {
       const stream = await StreamRepository.update(client, streamId, { archivedAt: new Date() })
       if (stream) {
         const evtId = eventId()
-        await StreamEventRepository.insert(client, {
+        const event = await StreamEventRepository.insert(client, {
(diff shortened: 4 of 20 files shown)


kristofferremback:
### Code review
**Confidence: 2/7** — Significant Concerns (blocking)
**Review models:** Orchestrator: opencode-go/glm-5.2 | Reviewers: inline (spec+design, correctness+data-flow, UX, mobile) — no subagent fan-out available in this harness; lenses run by the orchestrator against the diff + source

Found 1 issue:

1. `apps/frontend/src/components/layout/sidebar/sidebar.tsx:130-138` — **The fix doesn't achieve its goal: archived roots are not resident in IDB in steady state, so `archivedStreamIds` is empty and threads under archived roots still show.** The approach assumes an archived root stream is in `idbStreams` (so it can be collected into `archivedStreamIds` and matched against `stream.rootStreamId`). It isn't, in any persistent state:
   - **Cold load / fresh IDB:** the workspace bootstrap excludes archived streams. `workspaces.bootstrap` calls `streamService.listWithPreviews(workspaceId, userId)` with no `archiveStatus` (`apps/backend/src/features/workspaces/handlers.ts:172`), and `parseArchiveStatusFilter(undefined)` yields `includeActive=true, includeArchived=false` so the WHERE collapses to `archived_at IS NULL` (`apps/backend/src/features/streams/repository.ts:536,585`). So archived roots are never inserted into IDB on a fresh load; the active threads under them are. `archivedStreamIds` is empty → the root check never fires → the threads appear. This is exactly the reported "loads of threads rooted in archived streams" state.
   - **Live archive is defeated by the next reconnect:** `handleStreamArchived` does keep the row via `db.streams.update(payload.stream.id, { …payload.stream, _cachedAt: Date.now() })` (`apps/frontend/src/sync/workspace-sync.ts:738`), so right after archiving, `archivedStreamIds` does contain the root and the threads hide. But on the next reconnect (or any `applyReconnectBootstrapBatch`), `cleanupStaleEntities` → `deleteStale(db.streams, …, keepIds=bootstrapStreamIds, now)` (`workspace-sync.ts:2156,2242` + `deleteStale` impl) deletes every stream not in the active-only bootstrap set whose `_cachedAt < now`. The archived root is not in the active set and its `_cachedAt` predates the fetch, so it is **pruned from IDB**. `archivedStreamIds` becomes empty and the threads resurface. The fix only holds in the narrow window between the archive event and the next reconnect.
   - `applyWorkspaceBootstrap` pruning (`workspace-sync.ts:1935`) does the same when `fetchStartedAt` is supplied.

   A frontend-only fix cannot know which roots are archived on cold load — the backend simply doesn't send that information. The robust fix is backend-side: either include archived streams in the workspace bootstrap (the sidebar filter already excludes `archivedAt` rows, so they'd populate `archivedStreamIds` without showing themselves), or annotate thread streams with their root's archived status, or have `listWithPreviews` return archived roots that still own active threads. The PR's "Out of scope: no backend change" note explicitly rules out the only path that actually works, so the shipped change does not reliably hide the threads the user complained about. The 6 new unit tests for `isSidebarStreamVisible` all pass but only prove the predicate given a populated `archivedStreamIds` — they don't cover the integration reality that the set is empty in steady state, so the gap isn't caught.
   - bug: the predicate is correct in isolation but its input is empty in every persistent state, so the behavior the PR claims ("hide every nested thread under an archived root") is not delivered.
   - plan: the stated solution ("one lookup hides every nested thread") is only true when the archived root is resident in IDB, which the sync layer's stale-entity pruning and the active-only bootstrap prevent.
   https://github.com/threahq/threa/blob/65579dd5fc3492c94f6a64105c5862eaacb07e77/apps/frontend/src/components/layout/sidebar/sidebar.tsx#L129-L139
   https://github.com/threahq/threa/blob/65579dd5fc3492c94f6a64105c5862eaacb07e77/apps/frontend/src/sync/workspace-sync.ts#L736-L739
   https://github.com/threahq/threa/blob/65579dd5fc3492c94f6a64105c5862eaacb07e77/apps/frontend/src/sync/workspace-sync.ts#L2239-L2244
   https://github.com/threahq/threa/blob/65579dd5fc3492c94f6a64105c5862eaacb07e77/apps/backend/src/features/streams/repository.ts#L583-L587

---

The PR's stated goal — "hide threads rooted in archived streams … at any level of nesting" — is not met in steady state. The mechanism requires the archived root to be in `idbStreams`; the workspace bootstrap excludes archived streams and `cleanupStaleEntities`/`deleteStale` prune them on reconnect, so `archivedStreamIds` is empty in every persistent state and the threads reappear. See issue 1.

</details>

Issue 1 — the fix is ineffective for the reported symptom. Only the transient live-archive window (between the `stream:archived` event and the next reconnect) hides the threads; cold load and post-reconnect both re-surface them.

</details>
<details><summary>🔁 Data Flow [1 issue]</summary>

Traced the thread/root stream lifecycle: create (active, in IDB) → root archived live (`handleStreamArchived` → `db.streams.update`, row retained, `archivedStreamIds` populated, threads hidden ✓) → reconnect (`applyReconnectBootstrapBatch` → workspace bootstrap returns active-only set → `cleanupStaleEntities`/`deleteStale` prunes the archived root because `!keepIds.has(rootId) && _cachedAt < now`) → `idbStreams` loses the root → `archivedStreamIds` empty → threads shown again. The field that the fix depends on (an archived root row in IDB) is not preserved across the reconnect boundary.

</details>

No invariant violations introduced. The use of `rootStreamId` to resolve a thread's archived ancestor is consistent with INV-62 ("Stream access is inherited, never direct … visibility resolves through `root_stream_id` to the nearest non-thread ancestor"). The gap is a correctness bug, not an invariant breach. `Visibilities` remains used in `sidebar.tsx` (virtual DM draft), so no unused-import (INV-38) issue.

</details>

The extraction into `isSidebarStreamVisible` is clean, the memo dependency array correctly includes `archivedStreamIds`, and the O(n) set build + O(1) lookups are fine. The design flaw is upstream of the helper: the input set can't be populated by a frontend-only change.

</details>

No security implications. Hiding sidebar rows is a display concern; no auth/access path is changed.

</details>
<details><summary>🎛️ UX [1 issue]</summary>

Issue 1 (UX lens): the user's explicit expectation — archived-stream threads gone from the sidebar — is not delivered in steady state. After a reconnect or on a fresh device the threads return, so the sidebar clutter the user reported persists.

</details>

No mobile-specific regression; the filter runs identically on all viewports.

</details>

---

> **Superseded** by a later unified review on PR #1047 (the backend fix in commit `9e9844b4` resolved the finding below). See the newer `<!-- unified-review:superseded -->` comment.

<details><summary>📋 Superseded review (2/7 — Significant Concerns)</summary>



</details>

kristofferremback:
Addressed in commit 2 (`9e9844b4`): added the backend fix the review identified as necessary.

**What changed:** `StreamRepository.listWithPreviews` (the workspace bootstrap's stream query, bootstrap-only) now excludes active threads whose `root_stream_id` references an archived stream, via an `EXISTS` subquery against the root. This is the load-bearing change — it covers cold load (threads never sent) and reconnect (threads not in the bootstrap set → pruned by `cleanupStaleEntities`).

**Why not "include archived streams in the bootstrap":** `useWorkspaceStreams` has ~20 consumers (activity, memory, share picker, quick switcher, channel-link, stream settings…), only some filter `archivedAt` today. Widening the cache contract to "all streams" risks surfacing archived streams elsewhere. Excluding the specific threads under archived roots is surgical — `idbStreams` stays "active streams" (minus those threads), no other consumer sees a change.

**Frontend filter retained** for the live-archive window (between the `stream:archived` socket event and the next reconnect), which the backend pruning doesn't cover until reconnect.

**Verified:** new E2E test `sidebar-archived-roots.test.ts` passes with the fix and fails without it (confirmed by stashing the SQL change); threads + search e2e suites still green (32 pass); frontend sidebar tests 143 pass; backend typecheck clean.

PR description updated with the design decisions and the self-review evolution note.

kristofferremback:
> [!NOTE]
> This review was superseded by a [newer review](https://github.com/threahq/threa/pull/1047#issuecomment-4775808287).

<details><summary>Superseded review (click to expand)</summary>


### Code review
**Review models:** Orchestrator: opencode-go/glm-5.2 | Reviewers: inline (spec+design, correctness+data-flow, UX, mobile) — no subagent fan-out available in this harness; lenses run by the orchestrator against the diff + source. Supersedes the prior 2/7 review (comment 4763473719), whose single finding is now resolved by the backend exclusion added in commit `9e9844b4`.

No issues survived the ≥80 threshold.

The prior review's blocking finding — that the frontend-only filter was ineffective in steady state because the workspace bootstrap excludes archived streams (so `archivedStreamIds` is empty on cold load and after reconnect pruning) — is resolved. Verified against source:
- `StreamRepository.listWithPreviews` is **workspace-bootstrap-only** (`workspaces/handlers.ts:172` is the sole caller via `streamService.listWithPreviews`), so the new `EXCLUDE_ARCHIVED_ROOT_THREADS` predicate affects no other surface.
- The `EXISTS (SELECT 1 FROM streams root WHERE root.id = s.root_stream_id AND root.archived_at IS NOT NULL)` subquery is supported by `idx_streams_root` on `root_stream_id` (`20251210155323_core_schema.sql:73`) and the `streams.id` PK; no seq scan.
- The predicate is applied to all four WHERE branches of `listWithPreviews`, so it holds regardless of the `types` / `userMembershipStreamIds` / `archiveStatus` filter combination the bootstrap uses (it passes none → the final bare branch, which now carries the exclusion).
- The new E2E test (`sidebar-archived-roots.test.ts`) fails without the SQL change and passes with it (confirmed by the author stashing the fix), so the integration reality the prior review flagged — that `archivedStreamIds` is empty in steady state — is now covered at the boundary that matters.
- The frontend `isSidebarStreamVisible` filter is retained and is still correct for the live-archive window (between the `stream:archived` socket event and the next reconnect): `handleStreamArchived` keeps the root row in IDB via `db.streams.update` (`workspace-sync.ts:738`), so `archivedStreamIds` populates and threads hide immediately, before the backend pruning takes effect on reconnect.

---

The PR's goal — "hide threads rooted in archived streams, at any nesting depth" — is now met across cold load, reconnect, and live archive. The two-layer design (backend exclusion + frontend filter) matches the PR description's stated rationale. No missing corresponding changes: the backend change is at the query the sidebar reads, the frontend change is at the sidebar reader, and the E2E test pins the contract at the HTTP boundary.

</details>

No logic errors. The `EXCLUDE_ARCHIVED_ROOT_THREADS` fragment is `AND NOT (...)`, correctly excluding only threads (type = `'thread'`) with a non-null `root_stream_id` whose root is archived; active roots, threads under active roots, and archived roots themselves (already excluded by the active-only filter) are unaffected. The `StreamTypes.THREAD` constant interpolates to the string `'thread'` matching the DB `type` column. `sql.raw` is appropriate — the fragment is a static constant with no user input.

</details>

Traced the thread/root lifecycle end to end:
- **Cold load:** workspace bootstrap query excludes archived roots (active-only) AND now excludes threads whose root is archived → neither enters `idbStreams`. `archivedStreamIds` is empty, but there are no threads-under-archived-roots to filter either. Sidebar clean. ✓
- **Live archive:** `stream:archived` → `handleStreamArchived` keeps root in IDB (sets `archivedAt`), `archivedStreamIds` populates, frontend filter hides threads immediately. The workspace-bootstrap React Query cache drops the root (separate store), but `idbStreams` retains it for the filter. ✓
- **Reconnect:** `cleanupStaleEntities`/`deleteStale` prunes the archived root from IDB (not in the active-only bootstrap set), so `archivedStreamIds` empties — but the backend exclusion now keeps the threads out of the bootstrap set too, so they're pruned from IDB as well (not in `keepIds`). Sidebar stays clean. ✓ This is the state the prior review correctly identified as broken; the backend fix closes it.
- **Unarchive:** `stream:unarchived` sends only the root; `handleStreamUnarchived` does `db.streams.update` on the root (no-op if pruned). Root + threads reappear on the next workspace bootstrap. This is pre-existing behavior for the archived root itself, now extended consistently to its threads. Documented as out of scope, not a regression.

No field is dropped or raced across these states.

</details>

No invariant violations. The use of `root_stream_id` to resolve a thread's archived ancestor is consistent with INV-62 ("Stream access is inherited, never direct … visibility resolves through `root_stream_id` to the nearest non-thread ancestor"). The query is workspace-scoped (`s.workspace_id = ${workspaceId}`) per INV-8. No schema change (INV-17). No new `USE_STUB_AUTH`-style silent fallback. The retained frontend filter does not duplicate the backend logic — each covers a distinct timing window, which is the " colocate shared behavior on one path" exception rather than a parallel implementation (INV-35).

</details>

The choice to exclude the specific threads rather than widen the bootstrap to include archived streams is the right call given ~20 `useWorkspaceStreams` consumers (activity, memory, share picker, quick switcher, channel-link, stream settings, …), only some of which filter `archivedAt` today. Widening the cache contract would risk surfacing archived streams in those surfaces; the surgical exclusion keeps `idbStreams` as "active streams" (minus threads under archived roots) with no contract change for other consumers. The `EXCLUDE_ARCHIVED_ROOT_THREADS` constant is defined once and applied in every branch — no drift across the four WHERE clauses. The EXISTS subquery is correlated on the PK and a partial index, so the cost is one indexed lookup per thread row, acceptable for a bootstrap query that already runs per workspace connect.

</details>

No security implications. `sql.raw` wraps a fully static string constant with no user-controlled interpolation (`StreamTypes.THREAD` is a compile-time constant). The predicate does not change access semantics — it only filters which already-accessible streams the bootstrap returns; `checkStreamAccess` / `listAccessibleStreamIds` (INV-62) are untouched. No injection, no auth bypass.

</details>

The user's reported symptom — sidebar cluttered with threads rooted in archived streams — is now resolved in every state: cold load, reconnect, and live archive. No new friction introduced. The out-of-scope unarchive-restore delay (threads reappear on next bootstrap rather than instantly) matches the pre-existing root behavior and is not a new inconsistency a user would notice in normal use.

</details>

No mobile-specific impact; the filter and query are viewport-independent.

</details>

</details>

kristofferremback:
### Code review
**Confidence: 5/7** — Good

Found 1 issue:

1. `apps/frontend/src/components/timeline/system-event.tsx:21-24` — `stream_archived` / `stream_unarchived` now render as first-class timeline rows on **any** stream type (this PR adds them to `event-item.tsx`'s render switch; previously they fell through unrendered). The system message is hardcoded "This thread has been sealed in the labyrinth" / "…restored from the labyrinth", but the archived row lands on the **root** stream's own timeline — so archiving a channel or scratchpad now shows a row calling it "this thread". The composer's `disabledReason` already distinguishes ("The stream this thread belongs to has been archived…"), so the timeline copy is inconsistent with it. (ux-consistency: mislabels non-thread streams as threads.) Vary the copy by stream type, or use a stream-type-neutral phrasing.
   https://github.com/threahq/threa/blob/6dd6e7bcac8f081c0d6125bd88c73db054491f54/apps/frontend/src/components/timeline/system-event.tsx#L21-L24

---

The branch is broader than the PR body (which describes only the sidebar/bootstrap fix). All 5 commits' intents are covered and internally consistent: event-type promotion → `TIMELINE_BROADCAST_EVENT_TYPES`; thread-room routing → delivery-groups + `listThreadIdsByRoot` + payload `threadStreamIds`; composer block → `resolveWritableMessageStream` + `rootArchivedAt` bootstrap field; bootstrap exclusion → `listWithPreviews`; sidebar filter → `isSidebarStreamVisible`. API→frontend wiring is complete (`rootArchivedAt` typed in `StreamBootstrap` and consumed in `stream-content.tsx`). Consider refreshing the PR body, which predates commits 1–3.
</details>

No correctness issues ≥80. Verified the live-seal path: `stream:archived` is routed to descendant thread rooms, and `handleStreamArchived` (registered on the workspace socket, `workspace-sync.ts:1680`) updates the root's IDB row, so `stream-content.tsx`'s `rootFromCache` re-renders and seals the composer without a refresh. `handleAppendEvent`'s `payload.streamId !== streamId` guard correctly drops the root-scoped row from the thread's own timeline (no cross-stream append).
</details>

INV-61 contiguity holds: `StreamEventRepository.insert` allocates a dense `broadcastSequence` via `isTimelineBroadcastEventType` (same constant), the event is inserted only on the root stream, thread viewers never append it, and `computeTimelineHoles` skips `broadcastSequence == null` rows — so pre-promotion `stream_archived` rows (null slot) create no phantom holes. The two-source root-archived resolution (IDB-first, bootstrap fallback, deliberately not merged with `??`) correctly avoids the stale-after-unarchive collapse.
</details>

No introduced violations. INV-62 (root_stream_id inheritance), INV-4/INV-7 (outbox + event row written in the same `withTransaction`), INV-20 (race-safe archive/unarchive), INV-31 (broadcast flag derived from the constant), INV-51/52 (`listThreadIdsByRoot` colocated in the streams repo) all satisfied.
</details>
<details><summary>🏗️ Design [1 non-blocking note]</summary>

`repository.ts:85-89` — `EXCLUDE_ARCHIVED_ROOT_THREADS` is AND-ed into **all 4** `listWithPreviews` branches, including the `includeArchived` / `filterAll` ones. Not an active bug: the only caller (`workspaces/handlers.ts:172`) passes no filter → active-only, and the PR scopes this as bootstrap-only. But the signature accepts `archiveStatus:["archived"]`, and a future "archived streams" view would silently lose threads-under-archived-roots with no error. Cheap defense-in-depth: gate the exclusion behind the active-only branches (or guard it on `!includeArchived`). Left as a note since there's no current trigger (INV-36).
</details>

No high-confidence vulnerabilities. The send-block (`resolveWritableMessageStream`) is the authoritative server-side gate covering deep links / API / move-to-thread; the frontend composer-hide is cosmetic only.
</details>
<details><summary>🎛️ UX [1 issue]</summary>

See finding 1 (thread-vs-stream copy mislabel). Otherwise the disabled-composer banner explains the sealed state in place, and the sidebar filter is consistent.
</details>

New timeline rows render via `SystemEvent` (`py-2 px-3 sm:px-6 text-center`, no tap targets, no layout shift). The root-archived composer path reuses the existing `FloatingComposerShell` disabled banner, which publishes `--composer-height`, so scroll padding adjusts — no new mobile regression.
</details>

PR #1056, opened 2026-06-22
feat(frontend): sidebar hint for streams with an unsent (loaded) draft

kristofferremback:
## Problem

A draft you start in a stream's composer is kept (loaded) when you navigate away, but nothing in the sidebar tells you it's there — you rediscover an unsent message only by reopening the stream or digging through the Drafts explorer. A *stashed* draft is different: you parked that one deliberately, so it shouldn't produce the same nudge.

## Solution

Show a small pencil hint on a sidebar stream row when its composer holds a loaded, non-empty draft — the "stepped away without sending or stashing" case. The loaded-vs-stashed line is the one the draft system already draws: a draft is *loaded* when its scope's `composerLoaded` pointer references it, *stashed* when it doesn't. `useAllDrafts` already exposes that as `isStashed`, so the hint reuses it instead of re-deriving the pointer rule.

### How it works

- `streamIdsWithLoadedDraft(drafts)` (`apps/frontend/src/hooks/use-all-drafts.ts`) is a pure transform over the existing `UnifiedDraft[]`: keep rows that are `!isStashed` with a resolved `streamId`, drop `thread` and `scratchpad` types (a thread reply isn't the stream's own composer draft; a scratchpad is itself a draft). Empty drafts are already filtered upstream by `useAllDrafts` — sealed E2E rows count there on the strength of their `ciphertext` — so a returned id is real, unsent content.
- The sidebar builder (`sidebar.tsx`) already calls `useAllDrafts(workspaceId)`, so the set comes from data on hand — no new store subscription. It's derived via a stable signature string (`[...set].sort().join(",")`) so the resulting `Set` keeps its identity across draft edits that don't change membership, and the heavy `processedStreams` map doesn't rebuild on every keystroke.
- Each row gets a `hasLoadedDraft` flag on `StreamItemData`, attached in the same map that sets `nameDecrypting` — so it reaches the leaf as plain data with no new prop drilling through `sections.tsx` / `sidebar-stream-list.tsx`.
- `StreamItem` renders `<DraftIndicator />` in the existing right-side cluster (next to the label dots and the mention `@`), gated on `stream.hasLoadedDraft && !isActive`.

### Key design decisions

**1. Suppress on the active stream** — the open stream's composer already shows the draft, so a sidebar dot there is noise (and the framing was "stepped away from the stream"). The gate is `!isActive` on the row; navigating away flips `isActive` and the hint appears.

**2. Stashed drafts never qualify** — routing through `isStashed` rather than "does any row exist for this scope" keeps a parked draft silent. This is the explicit requirement; the test `does not flag a stream whose only draft for the scope is stashed` pins it.

**3. Attach a flag, don't drill a callback** — `getUnreadCount`/`getMentionCount` are threaded as callbacks because section headers also aggregate them (`sumUnread`). The draft hint is row-only, so it rides `StreamItemData` like `nameDecrypting`: fewer touched files, one computation, and no per-row store subscription (which would fan out N Dexie live queries on the drafts table).

**4. Subtle, matches the existing vocabulary** — `DraftIndicator` mirrors `MentionIndicator`: a `lucide-react` `Pencil` at `h-3 w-3` in `text-muted-foreground` with a right-side tooltip ("Unsent draft"). `Pencil` is already the app's edit/draft glyph (`stream.tsx`, `saved-item.tsx`). It's conditionally rendered, so it reserves no space and causes no layout shift (INV-21); the tooltip is portal-based.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/draft-indicator.tsx` | Sidebar pencil + "Unsent draft" tooltip; accessible (`role="img"`, `aria-label`). |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-all-drafts.ts` | Add pure `streamIdsWithLoadedDraft(drafts)` helper (loaded, non-thread, non-scratchpad). |
| `apps/frontend/src/hooks/index.ts` | Re-export the helper from the barrel. |
| `apps/frontend/src/components/layout/sidebar/sidebar.tsx` | Derive `streamsWithLoadedDraft` (stable signature) and set `hasLoadedDraft` per row. |
| `apps/frontend/src/components/layout/sidebar/types.ts` | Add `hasLoadedDraft?: boolean` to `StreamItemData`. |
| `apps/frontend/src/components/layout/sidebar/stream-item.tsx` | Render `<DraftIndicator />` gated on `hasLoadedDraft && !isActive`. |
| `apps/frontend/src/hooks/use-all-drafts.test.ts` | 4 unit tests for `streamIdsWithLoadedDraft`. |
| `apps/frontend/src/components/layout/sidebar/stream-item.test.tsx` | 3 render tests (shows when loaded, hidden on active, hidden when absent). |

## Out of scope (deliberate)

- **Scratchpad rows** — scratchpads render via `ScratchpadItem` and are themselves drafts, so they get no "unsent draft" hint (and scratchpad-typed drafts are excluded from the set).
- **Thread drafts** — a loaded draft in a thread does not light up its parent channel row; only a stream's own `stream:{id}` composer draft counts.
- **Virtual (not-yet-created) DM drafts** — `virtualDmStreams` aren't built through the `processedStreams` map, so a draft typed into a never-opened DM won't flag until the DM exists. Rare; left for a follow-up.
- **Section-level aggregation** — collapsed section headers don't summarize "has drafts"; the hint is per visible row.

## Test plan

- [x] `vitest run` on the two affected files — 16 pass (7 new: 4 helper + 3 render)
- [x] `apps/frontend` `tsc --noEmit` — clean
- [x] `eslint` on all 8 changed files — clean
- [x] Full pre-commit hook (lint + typecheck across all workspaces, migration/docs checks) — passed
- [ ] No manual in-browser pass — verified via component tests (mounts a real `StreamItem` through the shared `TooltipProvider`) rather than a running app

<details>

**Goal** — Surface a sidebar hint on streams the user stepped away from with an unsent draft, excluding stashed drafts.

**What was built** — A pure `streamIdsWithLoadedDraft` selector over `useAllDrafts` output; a `hasLoadedDraft` flag on `StreamItemData` set in the sidebar builder; a `DraftIndicator` leaf rendered by `StreamItem` when loaded and not active.

**Design decisions** — Reuse `isStashed` from `useAllDrafts` for the loaded/stashed line; attach a flag rather than drill a callback (row-only signal); stable signature to avoid `processedStreams` churn; suppress on the active stream; match `MentionIndicator`'s visual register (INV-21-safe).

**What's NOT included** — Scratchpad / thread / virtual-DM flagging, section aggregation (see Out of scope).

**Status** — Implemented, tested (16 pass), typecheck/lint clean, committed `9700a09`, pushed.

</details>

---

https://claude.ai/code/session_01Kaj8XexYUDbF9poxQ5tbuV

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kaj8XexYUDbF9poxQ5tbuV)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1056"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784756045&installation_id=129946197&pr_number=1056&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1056&signature=2832a7d53e2447ddec8114586f93dcd468ebc8491a425fbe2e0de42e082e8821"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/draft-indicator.tsx
+++ b/apps/frontend/src/components/draft-indicator.tsx
@@ -1,0 +1,32 @@
+import { Pencil } from "lucide-react"
+import { cn } from "@/lib/utils"
+import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip"
+
+interface DraftIndicatorProps {
+  className?: string
+}
+
+/**
+ * Sidebar hint that a stream has an unsent draft loaded into its composer — the
+ * user navigated away without sending or stashing it. The caller decides when to
+ * render it; a stashed draft holds no composer pointer for its scope, so it never
+ * qualifies (that's the loaded-vs-stashed distinction this hint draws).
+ */
+export function DraftIndicator({ className }: DraftIndicatorProps) {
+  return (
+    <Tooltip>
+      <TooltipTrigger asChild>
+        <span
+          role="img"
+          aria-label="Unsent draft"
+          className={cn("flex-shrink-0 text-muted-foreground cursor-default", className)}
+        >
+          <Pencil className="h-3 w-3" />
+        </span>
+      </TooltipTrigger>
+      <TooltipContent side="right" className="text-xs">
+        Unsent draft
+      </TooltipContent>
+    </Tooltip>
+  )
+}
--- a/apps/frontend/src/components/layout/sidebar/sidebar.tsx
+++ b/apps/frontend/src/components/layout/sidebar/sidebar.tsx
@@ -10,6 +10,7 @@ import { useSealedNamePendingResolver } from "@/hooks/use-decrypted-stream-name"
 import {
   useActivityCounts,
   useAllDrafts,
+  streamIdsWithLoadedDraft,
   createDmDraftId,
   useDraftScratchpads,
   useLiveSavedCount,
@@ -124,6 +125,19 @@ export function Sidebar({ workspaceId }: SidebarProps) {
   const mutedStreamIdSet = useMemo(() => new Set(unreadState?.mutedStreamIds ?? []), [unreadState?.mutedStreamIds])
   const dmPeerByStreamId = useMemo(() => new Map(idbDmPeers.map((peer) => [peer.streamId, peer.userId])), [idbDmPeers])
 
+  // Streams the user stepped away from with an unsent (loaded, non-stashed)
+  // draft, surfaced as a per-row hint. The signature string keeps the derived
+  // Set referentially stable across draft edits that don't change membership, so
+  // the heavy `processedStreams` map below doesn't rebuild on every keystroke.
+  const loadedDraftStreamIdSignature = useMemo(
+    () => [...streamIdsWithLoadedDraft(allDrafts)].sort().join(","),
+    [allDrafts]
+  )
+  const streamsWithLoadedDraft = useMemo(
+    () => new Set(loadedDraftStreamIdSignature ? loadedDraftStreamIdSignature.split(",") : []),
+    [loadedDraftStreamIdSignature]
+  )
+
   const processedStreams = useMemo(() => {
     return idbStreams
       .filter((stream) => {
@@ -157,6 +171,7 @@ export function Sidebar({ workspaceId }: SidebarProps) {
           section,
           dmPeerUserId,
           nameDecrypting: isSealedNamePending(streamWithPreview),
+          hasLoadedDraft: streamsWithLoadedDraft.has(stream.id),
         }
       })
   }, [
@@ -172,6 +187,7 @@ export function Sidebar({ workspaceId }: SidebarProps) {
     unreadState,
     workspaceId,
     isSealedNamePending,
+    streamsWithLoadedDraft,
   ])
 
   // System streams are auto-created infrastructure — don't count toward "has content"
--- a/apps/frontend/src/components/layout/sidebar/stream-item.test.tsx
+++ b/apps/frontend/src/components/layout/sidebar/stream-item.test.tsx
@@ -265,6 +265,57 @@ describe("StreamItem", () => {
     expect(screen.getByRole("button", { name: "Settings" })).toBeInTheDocument()
   })
 
+  it("shows an unsent-draft hint on a stream with a loaded draft", () => {
+    const stream = createStream({ hasLoadedDraft: true })
+
+    renderWithRouter(
+      <StreamItem
+        workspaceId="workspace_1"
+        stream={stream}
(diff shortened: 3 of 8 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: Opus 4.8 | Reviewers: Sonnet x4 (spec+design, correctness+data-flow, UX, mobile)

No issues found. Checked for bugs, data-flow lifecycle, CLAUDE.md compliance, plan adherence, design quality, security, UX, and mobile/responsive.

**Verified clean on the points most likely to bite:**
- **Loaded vs. stashed** — `streamIdsWithLoadedDraft` gates on `isStashed`, so a parked (stashed) draft never lights the hint; the `stream-item` test pins the stashed-only case.
- **No re-render storm** — the signature→Set two-step keeps `streamsWithLoadedDraft` referentially stable when membership is unchanged, so `processedStreams` doesn't rebuild on every (debounced) keystroke; an empty signature yields an empty Set (no `[""]` entry).
- **No false positives** — thread and scratchpad drafts are excluded by `type`; virtual-DM synthetic ids can't match real stream rows.
- **Mobile parity** — `DraftIndicator` mirrors `MentionIndicator` exactly; the non-interactive `TooltipTrigger` span doesn't intercept the row's touch/long-press handlers (those live on the `Link`).

**One candidate raised and dismissed:** an INV-21 "layout shift" flag on the conditionally-rendered pencil (UX lens 80, Spec lens 55). Dismissed on verification — the indicator follows the exact data-driven conditional-render pattern of its cluster siblings `MentionIndicator` and `StreamLabelDots`; INV-21 targets hover/tooltip/popover reflow, and the tooltip here is a Radix portal that causes no shift. Reserving a permanent slot would contradict the siblings and add whitespace to every row.

---

PR #1054, opened 2026-06-22
feat(board): board view design + read-only workspace board (slice 1)

kristofferremback:
**Design docs:** [`docs/board-view-design.md`](../blob/claude/facebook-workspace-feature-kjvrg6/docs/board-view-design.md) · [`docs/nonlinear-stream-views-exploration.md`](../blob/claude/facebook-workspace-feature-kjvrg6/docs/nonlinear-stream-views-exploration.md)

## Problem

Threa has exactly one way to look at a stream: the chronological timeline. It answers "what is being said, in order" but not "what matters in here, right now" — Tuesday's decision is buried under today's chatter. Separately, the **conversations** primitive (AI-clustered topics per stream, with status/completeness/staleness) is computed for every stream but only ever surfaced inside one stream's optional overlay (`convView`/`convOverlay`); there is no cross-stream view of it, and `ConversationRepository.findByWorkspace` had **no callers and no HTTP route**.

This PR is the design of record for a second, co-equal way to interact with streams — the **board** — plus its first implementation slice.

## Solution

A cross-stream **board**: a wall of conversations rendered as cards, ordered by recent activity, scoped to what the viewer can read. This slice is read-only — the foundation the authored-post path (slice 2) and lenses/scope filter (slice 3) build on. The design (and the path that got here: conversation-as-post → workspace-board-as-entrypoint → authored-posts-seed-conversations, validated by a read-only prod floor measurement) is in `docs/board-view-design.md`.

### How it works

1. **Backend feed** — `ConversationRepository.findByWorkspaceForViewer` (`apps/backend/src/features/conversations/repository.ts`) selects a workspace's conversations ordered by `last_activity_at DESC`, access-filtered **in SQL** via the canonical `streamAccessPredicateSql(workspaceId, userId, "conversations.stream_id")` (INV-62) so the `LIMIT` counts only readable rows, and excludes empty resolved shells (`cardinality(message_ids) > 0`). It uses `composeSql` (the predicate is a nested fragment) and pre-resolves `SELECT_FIELDS` through `sql` so the raw column list inlines rather than being parametrized.
2. **Service + handler** — `ConversationService.listByWorkspace(workspaceId, userId, options)` maps `addStalenessFields` over the rows; the `listByWorkspace` handler validates the query with the existing `listConversationsSchema` (Zod, INV-55) and returns `{ conversations }`. Access is enforced in-query, so unlike `listByStream` the handler does **not** gate on a single stream.
[...]

PR #1063, opened 2026-06-23
feat(memos): cross-conversation + cross-language memo dedup, extraction model upgrade, companion default-off

kristofferremback:
## Problem

GAM's extracted memos and AI-clustered conversations are low quality, and the newly merged board (#1054) makes that quality load-bearing — its "post" *is* a conversation, and its Decisions lens *is* a memo join. Verified against production read-only data on the highest-volume stream in the workspace (the human DM `stream_01KJMS776MNP2Q382MJ639Y2JD`, 3,637 messages, 454 conversations, 175 memos):

- **Cross-language duplicate memos.** A bilingual (SV/EN) stream stored the same knowledge twice — e.g. the "don't roll your own auth" lesson captured three times (1 EN, 2 SV), the stacked-PR merge-order fact twice. Dedup only ran *within* a single conversation (the revision path, `findActiveBySourceConversation`), so the same fact recurring across conversations produced one memo each. The memorizer wrote each memo in its conversation's language, and `text-embedding-3-small` aligns weakly across languages, so even a vector check couldn't catch the EN/SV pair.
- **Spec-violating, fragment, and catch-all titles** on the board cards. `gpt-5.4-nano` ignored the topic-summary spec at scale: 42% of titles ran 9+ words with framing ("Discussion about…"), tone-labels ("Casual chat about…"), and language-labels ("Swedish banter about…") the prompt explicitly forbids. The largest conversation was a 247-message catch-all titled *"General chat or reaction message"* — a vague title acts as an attractor that absorbs every later ambiguous message.
- **No author-type discipline for AI chats.** Companion scratchpads (chats with an AI persona) default `memoryMode: auto`, so GAM mines the model's own turns as if they were organizational decisions.

(Aside that motivated the dig: the board design doc's floor measurement had mislabeled this DM as *"an AI-persona DM"*. It is a human↔human DM — `Pierre Boberg` is a `usr_`, not a persona/bot, and `resolveActorType` has no inference path to "bot". The mislabel, not the code, was treating a person as a bot.)

## Solution

Four independent changes, each mapped to a finding. Memo dedup + canonical language are the core; the model upgrade and the companion-default are the adjacent wins the user signed off on ("lean towards both upgrading", "include the title fix", "good to extract from bot chats, but off by default").

### How it works

1. **Model upgrade + title tightening** (`83e24b9`). `MEMO_CLASSIFIER_MODEL_ID`, `MEMO_MEMORIZER_MODEL_ID` (`memos/config.ts`) and `BOUNDARY_EXTRACTION_MODEL_ID` (`conversations/boundary-extraction/config.ts`) move `gpt-5.4-nano → gpt-5.4-mini` — the tier `model-reference.md` names for "memo generation where quality justifies the cost over nano". The boundary "Naming new conversations" rules gain an explicit ban on vague catch-all titles ("General chat", "Reaction message", "Random") that name nothing and become attractors, plus a hard "never exceed 5 words". `StaticConfigResolver` reads these constants, so the three asserting tests (`config-resolver.test.ts`, `eval-config-resolver.test.ts`) updated in lockstep.

2. **Cross-conversation dedup gate** (`ede8cc3`, hardened in `4f209bf`). New `MemoRepository.findNearDuplicate` — a stream-scoped pgvector cosine search (`m.embedding <=> $emb::vector < maxDistance`, mirroring `semanticSearch`) that excludes memos already attached to the current conversation (`source_conversation_id IS DISTINCT FROM …`), so the per-conversation revision path keeps owning same-conversation repeats. The dedup runs **inside the insert transaction**, under a per-stream advisory lock (`pg_advisory_xact_lock(hashtext('memo-batch:'+streamId))`): a candidate within `MEMO_DEDUP_DISTANCE` (0.15 cosine) of an existing active memo is dropped before insert. Because uncommitted rows are visible within the same transaction, the in-transaction check sees both committed memos from other batches *and* survivors inserted earlier in this batch — one mechanism covers cross-batch and in-batch dedup, and the advisory lock serializes concurrent same-stream batches (INV-20). Drops are counted (`memosDeduped`) and logged.

3. **Canonical memo language** (`1f9f310`, refined in `4f209bf`). New `memoLanguage` workspace setting (`DEFAULT_WORKSPACE_SETTINGS`, default `null`), updatable through the existing settings API (`updateWorkspaceSettingsSchema` + `flattenUpdates`). The memorizer system prompt's rule 5 is built by `memoLanguageRule(memoLanguage)`: when set, "WRITE EVERY MEMO IN <language>… keep proper nouns verbatim"; when null, the prior "write in the conversation's language". `MemoService` resolves the effective language from the workspace override (`WorkspaceSettingsRepository.findOverrides`), falling back to the participants' most common `locale`, resolved through `Intl.DisplayNames` to a language *name* (`sv-SE → "Swedish"`) so the prompt never reads "WRITE EVERY MEMO IN sv-SE". One language per stream both stops the EN/SV pairs at the source and makes the stream-scoped embedding dedup in (2) reliable.

4. **Companion scratchpads default memory off** (`3d0bb0e`). `createScratchpadInTransaction` computes `hasCompanion` from `companionMode`/`companionPersonaId` and defaults `memoryMode` to `off` for companion scratchpads (`auto` for plain ones); an explicit `memoryMode` still wins. Extracting from AI chats becomes opt-in. Bot-runtime streams already defaulted off (`bot-runtimes/service.ts:268`).

### Key design decisions

**1. Dedup is authoritative and race-safe inside the transaction** — `MEMO_DEDUP_DISTANCE = 0.15` is a conservative start: tighter errs toward keeping near-duplicates, looser risks merging genuinely distinct facts (e.g. an auth *decision* vs. an auth *vuln*). It is eval-tunable, and this PR does not include a live eval run to tune it (see Out of scope).

**2. Canonical language defaults to participant locale (resolved to a name), not English** — forcing English would silently rewrite a Swedish-first workspace's knowledge base. Defaulting to the participants' most common locale keeps each stream in its own language while still being *one* language per stream (which is all the stream-scoped dedup needs); the explicit setting is the global override. Null stays "follow the conversation" so nothing regresses for workspaces that never set it.

**3. Drop, don't merge provenance** — a deduped memo's second conversation does not get its `source_conversation_id` linked to the surviving memo. Merging provenance would need an update path and a multi-valued source link; deferred deliberately (INV-36) rather than half-built.

**4. Companion gate lives on the one creation path** — `createScratchpadInTransaction`, not sprinkled across callers (INV-29/43). DMs/channels/threads don't carry a companion here, so they're unaffected; the human Pierre DM (`companion_mode: off`) keeps extracting.

### Design evolution (self-review)

A 2-agent self-review (spec+design, correctness+data-flow) caught two issues, both fixed in `4f209bf`:

- **The dedup gate was not race-safe.** The first version read `findNearDuplicate` from the pool *outside* the insert transaction, so two concurrent `processBatch` runs for the same stream — reachable because the check worker re-dispatches a per-stream job each cycle with no in-flight lock — could both pass the read and insert the duplicate the gate exists to stop (the earlier claim that batches were "serialized under the pending-item lock" was wrong; no such lock exists). Fixed by moving the dedup into the insert transaction under a per-stream `pg_advisory_xact_lock`, which also let the in-memory `cosineDistance` in-batch pre-check be deleted (same-transaction row visibility subsumes it).
- **A BCP-47 locale tag leaked into the prompt** ("WRITE EVERY MEMO IN sv-SE") on the default-language path; now resolved to a language name via `Intl.DisplayNames`.

Two reviewer findings were dismissed as verified false positives: the message-sourced UNION branch of `findNearDuplicate` correctly omits the conversation-exclude guard (message memos have `source_conversation_id IS NULL`), and the "explicit null setting" path is the intended `auto` behavior (a null override is deleted, never stored).

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/memos/config.ts` | `gpt-5.4-mini` model ids; `MEMO_DEDUP_DISTANCE`; rule-5 `memoLanguageRule` + `getMemorizerSystemPrompt(timezone, memoLanguage)` |
| `apps/backend/src/features/memos/repository.ts` | new `findNearDuplicate` (stream-scoped cosine dedup query) |
| `apps/backend/src/features/memos/service.ts` | in-transaction dedup gate under a per-stream advisory lock; `memoLanguage` resolution (`mostCommon` locale → `localeToLanguageName`, settings override); `memosDeduped` counter |
| `apps/backend/src/features/memos/memorizer.ts` | `memoLanguage` on `MemorizerContext`, passed to the system prompt |
| `apps/backend/src/features/conversations/boundary-extraction/config.ts` | `gpt-5.4-mini`; ban catch-all titles + 5-word cap |
| `apps/backend/src/features/streams/service.ts` | companion scratchpad defaults `memoryMode: off` |
| `apps/backend/src/features/workspace-settings/{handlers,service}.ts` | `memoLanguage` in update schema + `flattenUpdates` |
| `packages/types/src/workspace-settings.ts` | `memoLanguage` on `WorkspaceSettings` / defaults / update input |
| `apps/backend/src/features/memos/{service,memorizer}.test.ts` | dedup-drop + advisory-lock tests; canonical-language prompt test |
| `apps/backend/src/features/streams/service.test.ts` | companion-default-off / plain-auto / explicit-wins tests |
| `apps/backend/src/lib/ai/config-resolver.test.ts`, `apps/backend/evals/framework/eval-config-resolver.test.ts` | model-id assertions updated to `gpt-5.4-mini` |

## Out of scope (deliberate)

- **No live eval run.** The model upgrade follows `model-reference.md`, but `MEMO_DEDUP_DISTANCE` and title quality aren't validated against a real sample here. The follow-up is an eval suite calling the production memorizer/classifier entry points (INV-44/45) over the Pierre-DM sample to tune the threshold and confirm the upgrade.
- **No frontend control** for `memoLanguage` yet — backend setting + API only.
- **Boundary-extractor context window untouched** — unbounded thread inclusion (`boundary-extraction-service.ts:87-89`) and no re-titling as a conversation grows (an early "Ye" stays the card name) are real but separate from memo quality; left for a follow-up.
- **Provenance merge on dedup** — deferred (decision 3).

## Test plan

- [x] `bun test src/features/memos src/features/conversations src/features/streams src/features/workspace-settings src/lib/ai` — 176 pass (post-hardening run; full affected set green)
- [x] New: dedup gate drops a cross-conversation duplicate; advisory lock acquired in the save transaction; canonical-language prompt branch; companion scratchpad defaults off (+ plain auto, + explicit wins)
- [x] `tsc --noEmit` + lint across the monorepo clean (pre-commit hook, all 5 commits)
- [x] Pre-PR self-review (2 reviewer agents): 2 findings fixed (TOCTOU dedup race, locale-tag-in-prompt), 2 dismissed as verified false positives
- [ ] No live LLM eval / integration run — dedup threshold and title quality not yet measured on real data (see Out of scope)

---

### DIFF
--- a/apps/backend/evals/framework/eval-config-resolver.test.ts
+++ b/apps/backend/evals/framework/eval-config-resolver.test.ts
@@ -13,7 +13,7 @@ describe("EvalConfigResolver", () => {
     const resolver = createEvalConfigResolver({ base })
 
     const config = await resolver.resolve(COMPONENT_PATHS.BOUNDARY_EXTRACTION)
-    expect(config.modelId).toBe("openrouter:openai/gpt-5.4-nano")
+    expect(config.modelId).toBe("openrouter:openai/gpt-5.4-mini")
     expect(config.temperature).toBe(0.2)
   })
 
--- a/apps/backend/src/features/conversations/boundary-extraction/config.ts
+++ b/apps/backend/src/features/conversations/boundary-extraction/config.ts
@@ -3,7 +3,7 @@
 import { z } from "zod"
 import { CONVERSATION_STATUSES } from "@threa/types"
 
-export const BOUNDARY_EXTRACTION_MODEL_ID = "openrouter:openai/gpt-5.4-nano"
+export const BOUNDARY_EXTRACTION_MODEL_ID = "openrouter:openai/gpt-5.4-mini"
 
 /** Low temperature for classification consistency. */
 export const BOUNDARY_EXTRACTION_TEMPERATURE = 0.2
@@ -75,8 +75,9 @@ Use it whenever the new message gives you evidence the prior placement was wrong
 - confidence: 0.0 to 1.0 confidence in this classification overall.
 
 ## Naming new conversations
-When you set newConversationTopic, write a short title of 2-5 words that names the topic itself:
-- Lead with the subject. Do NOT add framing like "Discussion about", "Chat about", "Conversation regarding", "Thoughts on", "Questions about", and do NOT describe the tone ("Casual chat", "Quick question"). That a conversation discusses something is already implied — name the thing, not the act of discussing it.
+When you set newConversationTopic, write a short title of 2-5 words that names the topic itself. Never exceed 5 words.
+- Lead with the subject. Do NOT add framing like "Discussion about", "Chat about", "Conversation regarding", "Thoughts on", "Questions about", and do NOT describe the tone ("Casual chat", "Quick question", "Banter about"). That a conversation discusses something is already implied — name the thing, not the act of discussing it.
+- Never use a vague catch-all label as a title: "General chat", "Reaction message", "Random", "Misc", "Off-topic", and the like name nothing and become a magnet that wrongly absorbs later messages. Always name the concrete subject the messages are actually about; if a short opener has no subject of its own, name what it is reacting to.
 - Do NOT state which language the conversation is in (never write "in Swedish", "auf Deutsch", etc.); that label is noise next to the conversation.
 - Write the title in the dominant language of the conversation, not English by default. If the participants are talking in Swedish, the title is in Swedish; if in Japanese, in Japanese. When the messages mix languages, follow the language the topic is actually discussed in and reuse the participants' own phrasing.
 - Keep names, products, technical terms, and other proper nouns exactly as they appear in the conversation. Never translate, localize, or re-spell them — carry the participants' own words into the title verbatim.
--- a/apps/backend/src/features/memos/config.ts
+++ b/apps/backend/src/features/memos/config.ts
@@ -7,9 +7,9 @@ import { z } from "zod"
 import { KNOWLEDGE_TYPES, type KnowledgeType, type StreamType } from "@threa/types"
 import { formatDate } from "../../lib/temporal"
 
-export const MEMO_CLASSIFIER_MODEL_ID = "openrouter:openai/gpt-5.4-nano"
+export const MEMO_CLASSIFIER_MODEL_ID = "openrouter:openai/gpt-5.4-mini"
 
-export const MEMO_MEMORIZER_MODEL_ID = "openrouter:openai/gpt-5.4-nano"
+export const MEMO_MEMORIZER_MODEL_ID = "openrouter:openai/gpt-5.4-mini"
 
 export const MEMO_TEMPERATURES = {
   classification: 0.1,
--- a/apps/backend/src/features/memos/config.ts
+++ b/apps/backend/src/features/memos/config.ts
@@ -234,7 +234,7 @@ How to write memos:
 2. EXTRACT, DON'T SUMMARIZE. Capture the durable conclusion — the decision, the answer, the fact, the procedure that was worked out — not a play-by-play of the discussion. A memo is what someone would want to recall in six months, never a transcript of who said what.
 3. BE TERSE. An abstract is a few sentences at most. If it reads like a recap of the conversation, rewrite it down to the bare conclusion.
 4. OMIT THE FORGETTABLE. Greetings, status pings, back-and-forth, and unresolved tangents produce no memo. Returning very few memos — or only the one thing that actually mattered — is correct and expected.
-5. WRITE IN THE CONVERSATION'S LANGUAGE. Use the same language the participants used. Do NOT translate (e.g. a Swedish conversation produces Swedish memos).
+{{MEMO_LANGUAGE_RULE}}
 6. BE FACTUAL. State the knowledge directly. No meta-commentary like "this memo captures..." or "the team discussed...".
 7. Use consistent vocabulary with prior memos when the same concept reappears.
 8. RESOLVE PRONOUNS when possible - If you can determine who "he/she/they" refers to from the conversation, use their actual name. If unclear (e.g., conversation continues from offline), leave the pronoun. When in doubt, preserve the original wording.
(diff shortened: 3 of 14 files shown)


kristofferremback:
### Code review
**Confidence: 6/7** — Very Good (2 issues found, both fixed in 4f209bf)

Both reviewers independently surfaced the same two issues; both are now fixed in `4f209bf`.

1. `apps/backend/src/features/memos/service.ts` — **TOCTOU race on the dedup gate (INV-20).** The original gate read `findNearDuplicate` from `this.pool` *outside* the insert transaction, so two concurrent `processBatch` calls for the same stream (the check worker re-dispatches a per-stream job each cycle with no in-flight lock) could both pass the read before either committed, re-inserting the duplicate the gate exists to prevent. **Fix:** the authoritative dedup now runs *inside* the insert transaction under a per-stream `pg_advisory_xact_lock(hashtext('memo-batch:'+streamId))`; same-transaction row visibility means it also subsumes the prior in-memory in-batch check, so the `cosineDistance` helper and the staged pre-check were removed.
   https://github.com/threahq/threa/blob/4f209bfdffc349a023ff78fc38e5feabfdc0f30a/apps/backend/src/features/memos/service.ts#L450-L478

2. `apps/backend/src/features/memos/service.ts` — **Locale tag leaked into the prompt.** The default-language fallback returned a BCP-47 tag (`sv-SE`), injected verbatim as `"WRITE EVERY MEMO IN sv-SE"`. **Fix:** `localeToLanguageName` resolves it via `Intl.DisplayNames` to a language name (`Swedish`) before it reaches the memorizer; the admin-set string path is unchanged.
   https://github.com/threahq/threa/blob/4f209bfdffc349a023ff78fc38e5feabfdc0f30a/apps/backend/src/features/memos/service.ts#L36-L49

Dismissed (verified false positives): the second `findNearDuplicate` UNION branch omitting the `excludeConversationId` guard is correct by design — message-sourced memos have `source_conversation_id IS NULL`, so the guard is a no-op there and they can never be the current conversation's memos. The "explicit `null` clears the setting but the heuristic still fires" concern is the intended `auto` behavior (a null override is deleted, not stored).

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>All four planned changes present and fully wired (type → default → update-input → handler schema → flatten → service read; model constants → asserting tests).</details>
<details><summary>🔁 Data Flow [2 issues — fixed]</summary>TOCTOU dedup race (fixed via in-transaction advisory lock); locale→language resolution (fixed).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-16 (gpt-5.4-mini is in model-reference.md), INV-20 (now race-safe), INV-47 (language ternary flattened during the fix), INV-44/33 (config co-located) all hold.</details>
<details><summary>🏗️ Design [CLEAN]</summary>Dedup reuses the existing `semanticSearch` vector pattern; no config sprawl; the per-candidate dedup query is a small per-stream indexed scan.</details>
<details><summary>🔒 Security [CLEAN]</summary>Embedding literal is built from an internal `number[]`, not user input — no injection surface.</details>

---

PR #1080, opened 2026-06-25
feat(board): authored posts via a conversation directive on the standard send

kristofferremback:
**Design doc:** [`docs/board-view-design.md`](../blob/main/docs/board-view-design.md) §"authored posts seed conversations" (slice 1b) · no Linear ticket

> **Note:** the design changed mid-PR. This was originally a dedicated `POST /board/posts` endpoint backed by an `is_authored_boundary` boolean; it's now an extension to the **standard message send** via a three-state `conversation` directive. The history is in the commits; this description reflects the final state on `d4e3fa6c`.

## Problem

The board (the workspace-wide, non-chronological feed behind the `board-view` flag) could only show **derived** conversations — topics the AI clusters out of timeline messages asynchronously. That makes the board only as good as the extractor, and a just-typed topic appears after a worker round-trip, not instantly.

Conversation creation only existed inside the boundary-extraction worker. There was no way for a sender to say "this message starts a new topic," and nothing stopped the async worker from later re-clustering or moving a human-declared post.

## Solution

Extend the normal message send so a message can **declare its conversation** instead of leaving the async extractor to infer it. A board "New post" is then just a standard send that declares a new conversation — created **synchronously, in the send's own transaction**, so the board reflects the topic the instant the request returns. The declaration is also a lock: the async extractor never re-clusters or reassigns a declared message.

This generalizes cleanly to the planned follow-up ("attach a message to an existing conversation") with no new column and no boolean.

### How it works

1. **`messages.conversation_intent`** (`TEXT`, migration `20260625120000_message_conversation_intent.sql`, which also drops the abandoned `is_authored_boundary`): `NULL` → inferred (extractor clusters it, unchanged); `'new'` → the send minted a fresh conversation seeded with this message; `'existing'` → the send attached it to a caller-named conversation. It's the provenance/lock signal — the membership itself lives on `conversations.message_ids`.

2. **The `conversation` directive** (`@threa/types` `ConversationDirective`) on `CreateMessageInput*` — a discriminated union `{ intent: "new" } | { intent: "existing"; conversationId }`. The id is a **sibling** of the discriminant, so `existing` without a non-empty id is a validation error, distinct from `new`. The server mints the id for `new` (no new id-generation site). A type-level conformance check in `handlers.ts` locks the runtime Zod schema to the published type so they can't drift (INV-31).

3. **Synchronous assignment via an injected `ConversationAssigner`** (`features/conversations/conversation-assigner.ts`, wired in `server.ts`) — `EventService.createMessageInTransaction` calls it inside the send's transaction, so messaging stays decoupled from conversation internals (INV-12/13). `new` = mint + seed + bump; `existing` = read the target **workspace-scoped and row-locked** (`findByIdForUpdate`, INV-8 + INV-20), validate same stream, attach, reactivate-if-resolved, bump. Both emit `conversation:created`/`updated` + `message_assigned` through the outbox (INV-4/7).

4. **The extractor skips declared messages** (`boundary-extraction-service.ts`): `processMessage` short-circuits when `conversationIntent !== null` (returns the conversation it already owns, no LLM call), and the reassignment loop skips any candidate that's declared. The human declaration wins over the async pass (INV-20).

5. **Agent replies join conversations too.** Invoking or DMing an agent is conversing with it, so the boundary-extraction dispatch now runs for **any actor** (user *or* agent), not just users. A deterministic `assignAgentReply` (no LLM) joins the stream's most-recently-active conversation, or mints one if there is none yet. It shares the `emitAssignmentEvents` helper with the declared-send path so both emit identical events (INV-35/37). Covers the case where a companion's reply to a post now shows up in the post's conversation card.

6. **Frontend board "New post"** (`components/board/board-composer.tsx`, mounted in `board.tsx`) — a collapsed "Write a post…" affordance expanding to a target picker + the shared `MessageComposer`. Targets: **New scratchpad** (companion AI) / **New quick note** (plain) / separator / **channels + DMs**. Existing scratchpads, threads, system, archived, and E2E streams are excluded.
   - **Existing channel/DM** → standard `messageService.create` with `conversation: { intent: "new" }` and a `clientMessageId` (idempotent retry). Reflects on the board immediately via the synchronous directive + refetch.
   - **New scratchpad / quick note** → reuses the existing draft-scratchpad + promote-on-send components (`createScratchpad` + `queueDraftMessage` → `promoteDraft`). The server scratchpad row is born **with** its first message — never created empty — and the durable queue keys the send by clientId. The extractor mints the first conversation (eventually consistent on the board, which has no live subscription yet).

### Key design decisions

**1. A directive, not a boolean or a dedicated endpoint.** A single-purpose `is_authored_boundary` flag and a board-only endpoint wouldn't generalize. The three-state `conversation_intent` + the `{ intent }` union extend the *standard* send and already cover the "attach to existing" follow-up.

**2. Synchronous, in-transaction.** The whole point of declaring is that the board reflects the topic instantly and deterministically, and the extractor leaves it locked. The injected `ConversationAssigner` keeps `EventService` free of conversation internals.

**3. New scratchpads via draft-promotion, not eager create.** Creating the server stream up-front and then sending leaves an empty scratchpad if the send fails. Reusing the sidebar's draft + promote-on-send flow means zero empty scratchpads and an idempotent send — and no parallel implementation to drift (INV-35/37).

## Files

**New**

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260625120000_message_conversation_intent.sql` | Drop `is_authored_boundary`; add `conversation_intent TEXT` |
| `apps/backend/src/features/conversations/conversation-assigner.ts` | In-transaction declared-conversation assignment (`new`/`existing`) |
| `apps/backend/src/features/conversations/assignment-events.ts` | Shared `emitAssignmentEvents` (declared-send + agent-reply) |
| `apps/frontend/src/components/board/board-composer.tsx` | "New post" composer + `isPostableStream` predicate |
| `apps/frontend/src/components/board/board-composer.test.tsx` | `isPostableStream` unit tests |

**Modified (backend)**

| File | Change |
| ---- | ------ |
| `features/messaging/repository.ts` | `conversationIntent` on `Message`/insert/row/SELECT |
| `features/messaging/event-service.ts` | `conversation` directive on `CreateMessageParams`; calls the assigner in-txn |
| `features/messaging/handlers.ts` | `conversationDirectiveSchema` + INV-31 conformance lock |
| `features/messaging/index.ts` | Export `ConversationAssigner` type |
| `features/conversations/boundary-extraction-service.ts` | Skip declared messages; deterministic `assignAgentReply` |
| `features/conversations/boundary-extraction-outbox-handler.ts` | Dispatch for any actor (user or agent) |
| `features/conversations/repository.ts` | `findByIdForUpdate` (workspace-scoped + `FOR UPDATE`) |
| `features/conversations/index.ts` | Export `conversationAssigner` |
| `server.ts` | Inject `conversationAssigner` into `EventService` |

**Modified (types + frontend)**

| File | Change |
| ---- | ------ |
| `packages/types/src/{api,constants,domain,index}.ts` | `ConversationDirective`, `CONVERSATION_INTENTS`/`ConversationIntent`, non-optional `Message.conversationIntent` |
| `apps/frontend/src/hooks/use-conversations.ts` | `useCreateBoardPost` (existing-stream directive send + new-scratchpad draft-promotion) |
| `apps/frontend/src/pages/board.tsx` | Mount `BoardComposer` |

**Test fixtures** — `conversationIntent: null` added to `Message` literals in `evals/.../suite.ts`, `agents/{conversation-summary-service,quote-resolver}.test.ts`, `conversations/.../llm-extractor.test.ts`, `saved-messages/service.test.ts`, `lib/ai/message-formatter.test.ts`.

## Out of scope (deliberate)

- **Full board realtime** — workspace-room `conversation:*` delivery / conversations on the IDB sync engine. The board still refreshes on open/reconnect; new-scratchpad posts surface there eventually-consistently.
- **Reply-from-board / write-interactivity** — deferred; reply accrual rides the existing extractor + agent-reply assignment.
- **E2E authored posts** — the picker excludes E2E streams (no client-side sealing here).

## Test plan

- [x] Backend `bun test src/features/messaging` (101) + `src/features/conversations` (48) pass, incl. the directive-schema tests and assigner paths
- [x] Frontend `vitest run src/hooks/use-conversations.test.tsx` (6) + `board-composer.test.tsx` (4) pass
- [x] Full-monorepo `tsc --noEmit`, eslint, astro check, and OpenAPI spec check pass (pre-commit hook)
- [ ] **Assigner + extractor-skip + agent-reply not live-DB integration tested** — DB-bound, and the conversations feature has no service-level live-DB harness; logic is typechecked and unit-covered at the schema/hook layer. Worth an integration test when a DB harness is handy.

---

### DIFF
--- a/apps/backend/evals/suites/boundary-extraction/suite.ts
+++ b/apps/backend/evals/suites/boundary-extraction/suite.ts
@@ -68,6 +68,7 @@ function toMessage(evalMsg: EvalMessage, streamId: string, sequence: number = 1)
     replyCount: 0,
     reactions: {},
     metadata: {},
+    isAuthoredBoundary: false,
     clientMessageId: null,
     sentVia: null,
     editedAt: null,
--- a/apps/backend/evals/suites/boundary-extraction/suite.ts
+++ b/apps/backend/evals/suites/boundary-extraction/suite.ts
@@ -68,7 +68,7 @@ function toMessage(evalMsg: EvalMessage, streamId: string, sequence: number = 1)
     replyCount: 0,
     reactions: {},
     metadata: {},
-    isAuthoredBoundary: false,
+    conversationIntent: null,
     clientMessageId: null,
     sentVia: null,
     editedAt: null,
--- a/apps/backend/src/db/migrations/20260624120000_message_authored_boundary.sql
+++ b/apps/backend/src/db/migrations/20260624120000_message_authored_boundary.sql
@@ -1,7 +1,0 @@
--- Authored board posts declare a human boundary the AI must not re-cluster.
--- A board "New post" creates a message + a conversation seeded with it, in one
--- transaction; this flag marks that message so the async boundary-extraction
--- worker skips it (never reassigns it, never re-clusters it). It is a durable
--- structural property of the message, not transient workflow state — the human
--- assignment wins over the async AI pass (INV-20).
-ALTER TABLE messages ADD COLUMN is_authored_boundary BOOLEAN NOT NULL DEFAULT FALSE;
--- a/apps/backend/src/db/migrations/20260624120000_message_authored_boundary.sql
+++ b/apps/backend/src/db/migrations/20260624120000_message_authored_boundary.sql
@@ -1,0 +1,7 @@
+-- Authored board posts declare a human boundary the AI must not re-cluster.
+-- A board "New post" creates a message + a conversation seeded with it, in one
+-- transaction; this flag marks that message so the async boundary-extraction
+-- worker skips it (never reassigns it, never re-clusters it). It is a durable
+-- structural property of the message, not transient workflow state — the human
+-- assignment wins over the async AI pass (INV-20).
+ALTER TABLE messages ADD COLUMN is_authored_boundary BOOLEAN NOT NULL DEFAULT FALSE;
--- a/apps/backend/src/db/migrations/20260625120000_message_conversation_intent.sql
+++ b/apps/backend/src/db/migrations/20260625120000_message_conversation_intent.sql
@@ -1,0 +1,11 @@
+-- A message can declare which conversation it belongs to, instead of leaving
+-- the async boundary-extractor to infer it. `conversation_intent` records that
+-- declaration so the extractor skips (and never re-clusters) a human-declared
+-- message:
+--   NULL       → inferred: the extractor clusters it (default, unchanged).
+--   'new'      → the send minted a fresh conversation seeded with this message.
+--   'existing' → the send attached this message to a caller-named conversation.
+-- The conversation membership itself lives on `conversations.message_ids`; this
+-- column is the provenance/lock signal, not a second copy of the id.
+ALTER TABLE messages DROP COLUMN IF EXISTS is_authored_boundary;
+ALTER TABLE messages ADD COLUMN conversation_intent TEXT;
--- a/apps/backend/src/features/agents/conversation-summary-service.test.ts
+++ b/apps/backend/src/features/agents/conversation-summary-service.test.ts
@@ -17,6 +17,7 @@ function makeMessage(sequence: bigint, content: string): Message {
     replyCount: 0,
     reactions: {},
     metadata: {},
+    isAuthoredBoundary: false,
     clientMessageId: null,
     sentVia: null,
     editedAt: null,
--- a/apps/backend/src/features/agents/conversation-summary-service.test.ts
+++ b/apps/backend/src/features/agents/conversation-summary-service.test.ts
@@ -17,7 +17,7 @@ function makeMessage(sequence: bigint, content: string): Message {
     replyCount: 0,
     reactions: {},
     metadata: {},
-    isAuthoredBoundary: false,
+    conversationIntent: null,
     clientMessageId: null,
     sentVia: null,
     editedAt: null,
--- a/apps/backend/src/features/agents/quote-resolver.test.ts
+++ b/apps/backend/src/features/agents/quote-resolver.test.ts
@@ -42,6 +42,7 @@ function createMessage(overrides: Partial<Message> & { id: string }): Message {
     replyCount: 0,
     reactions: {},
     metadata: {},
+    isAuthoredBoundary: false,
     clientMessageId: null,
     sentVia: null,
     editedAt: null,
--- a/apps/backend/src/features/agents/quote-resolver.test.ts
+++ b/apps/backend/src/features/agents/quote-resolver.test.ts
@@ -42,7 +42,7 @@ function createMessage(overrides: Partial<Message> & { id: string }): Message {
     replyCount: 0,
     reactions: {},
     metadata: {},
-    isAuthoredBoundary: false,
+    conversationIntent: null,
     clientMessageId: null,
     sentVia: null,
     editedAt: null,
--- a/apps/backend/src/features/conversations/assignment-events.ts
+++ b/apps/backend/src/features/conversations/assignment-events.ts
@@ -1,0 +1,58 @@
+import type { PoolClient } from "pg"
+import { StreamTypes } from "@threa/types"
+import { ConversationRepository, type Conversation } from "./repository"
+import { StreamRepository } from "../streams"
+import { MessageRepository, type Message } from "../messaging"
+import { OutboxRepository } from "../../lib/outbox"
+import { addStalenessFields } from "./staleness"
+
+/**
+ * Emit the conversation aggregate + per-message membership events for a single
+ * deterministic assignment (a declared send or an agent reply), mirroring the
+ * boundary extractor's persist phase (INV-4/7). This is the one place the
+ * declared-send and agent-reply paths share so their event shape can't drift
+ * (INV-35/37).
+ *
+ * Re-reads the conversation so the payload reflects the membership/activity
+ * writes that just happened, and resolves a thread's `parentStreamId` so the
+ * parent channel's subscribers receive the update too. Returns the refreshed
+ * conversation for callers that surface it.
+ */
+export async function emitAssignmentEvents(
+  client: PoolClient,
+  params: { workspaceId: string; message: Message; conversationId: string; created: boolean; reason: string }
+): Promise<Conversation> {
+  const { workspaceId, message, conversationId, created, reason } = params
+
+  let parentStreamId: string | undefined
+  const stream = await StreamRepository.findById(client, message.streamId)
+  if (stream?.type === StreamTypes.THREAD && stream.parentMessageId) {
+    const parentMessage = await MessageRepository.findById(client, stream.parentMessageId)
+    parentStreamId = parentMessage?.streamId
+  }
+
+  const [refreshed] = await ConversationRepository.findByIds(client, workspaceId, [conversationId])
+  if (!refreshed) {
+    // The row we just wrote is gone under our own transaction — fail loud (INV-11).
+    throw new Error(`Conversation ${conversationId} vanished during assignment`)
+  }
+
+  await OutboxRepository.insert(client, created ? "conversation:created" : "conversation:updated", {
+    workspaceId,
+    streamId: message.streamId,
+    conversationId: refreshed.id,
+    conversation: addStalenessFields(refreshed),
+    parentStreamId,
+  })
+  await OutboxRepository.insert(client, "conversation:message_assigned", {
+    workspaceId,
+    streamId: message.streamId,
+    parentStreamId,
+    messageId: message.id,
+    conversationId: refreshed.id,
+    isPrimary: true,
+    reason,
+  })
+
+  return refreshed
+}
--- a/apps/backend/src/features/conversations/boundary-extraction-outbox-handler.ts
+++ b/apps/backend/src/features/conversations/boundary-extraction-outbox-handler.ts
@@ -1,6 +1,5 @@
 import type { Pool } from "pg"
 import { parseMessagePayload } from "../../lib/outbox"
-import { AuthorTypes } from "@threa/types"
 import { logger } from "../../lib/logger"
 import { JobQueues } from "../../lib/queue"
 import type { QueueManager } from "../../lib/queue"
@@ -10,7 +9,10 @@ import { E2eStreamsRepository } from "../e2e-streams"
 export type BoundaryExtractionHandlerConfig = DebouncedOutboxHandlerConfig
 
 /**
- * Dispatches boundary-extraction jobs for user messages arriving via the outbox.
+ * Dispatches boundary-extraction jobs for messages arriving via the outbox.
+ * User messages are LLM-clustered; agent (persona/bot) replies are assigned
+ * deterministically to the conversation they reply within (handled in
+ * `BoundaryExtractionService.processMessage`) — both flow through this dispatch.
  */
 export class BoundaryExtractionHandler extends DebouncedOutboxHandler {
   private readonly jobQueue: QueueManager
@@ -39,7 +41,10 @@ export class BoundaryExtractionHandler extends DebouncedOutboxHandler {
       return
     }
 
-    if (messageEvent.actorType !== AuthorTypes.USER) {
+    // Every message — user or agent (persona/bot) — belongs to a conversation.
+    // User messages are LLM-clustered; agent replies are assigned
+    // deterministically in processMessage. Skip only malformed actorless events.
+    if (!messageEvent.actorId) {
       return
     }
 
--- a/apps/backend/src/features/conversations/boundary-extraction-service.ts
+++ b/apps/backend/src/features/conversations/boundary-extraction-service.ts
@@ -65,6 +65,22 @@ export class BoundaryExtractionService {
         return { message: null, stream: null, extractionContextBase: null }
       }
 
+      // Authored board posts declare their conversation synchronously at
+      // creation. The message:created outbox still fires for them, but the async
+      // pass must never re-cluster or move a human-declared boundary (INV-20) —
+      // short-circuit with the conversation it already owns.
+      if (message.isAuthoredBoundary) {
+        const authoredPrimary = await ConversationRepository.findPrimaryByMessageId(client, workspaceId, message.id)
+        return {
+          message,
+          stream,
+          extractionContextBase: null,
+          authoredSkip: true,
+          authoredPrimary,
+          validUpdateTargets: new Set<string>(),
+        }
+      }
+
       if (stream.type === StreamTypes.SCRATCHPAD) {
         const existingConversations = await ConversationRepository.findByStream(client, stream.id)
         return {
@@ -167,6 +183,11 @@ export class BoundaryExtractionService {
     if (!fetchedData.message || !fetchedData.stream) {
       logger.warn({ messageId, streamId }, "Message or stream not found for boundary extraction")
       return null
+    }
+
+    if (fetchedData.authoredSkip) {
+      logger.debug({ messageId, streamId }, "Skipping boundary extraction for authored board post")
+      return fetchedData.authoredPrimary ?? null
     }
 
     const {
@@ -389,6 +410,13 @@ export class BoundaryExtractionService {
             reassignedMessage = m
             messagesById.set(r.messageId, m)
           }
+        }
+
+        // An authored board post is a human-declared boundary: the async pass
+        // never moves it out of the conversation it seeded (INV-20).
+        if (reassignedMessage?.isAuthoredBoundary) {
+          logger.warn({ messageId: r.messageId }, "Skipping reassignment of an authored board post")
+          continue
         }
 
         await ConversationRepository.removePrimaryMessage(client, workspaceId, fromConvId, r.messageId)
--- a/apps/backend/src/features/conversations/boundary-extraction-service.ts
+++ b/apps/backend/src/features/conversations/boundary-extraction-service.ts
@@ -2,7 +2,7 @@ import type { Pool, PoolClient } from "pg"
 import { sql, withTransaction, withClient } from "../../db"
 import { ConversationRepository, type Conversation } from "./repository"
 import { MessageRepository, type Message } from "../messaging"
-import { StreamRepository } from "../streams"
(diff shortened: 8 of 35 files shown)


kristofferremback:
### Code review

No bugs, data-flow, security, CLAUDE.md, or mobile issues found. Four UX-polish findings (all on the new composer) plus two low-severity design notes. None are blocking — the feature works; these are refinements.

Found 4 issues:

1. `apps/frontend/src/components/board/board-composer.tsx:168-170` — `autoFocus` lands in the message editor while the placeholder reads "Pick where to post, then write…". Focus and the instructed first action point in opposite directions; a keyboard user must Shift-Tab to reach the target Select. For a surface that *requires* a target before submit, focus should open on the Select (or the editor should only autofocus once a target is chosen). (ux-consistency)
   https://github.com/threahq/threa/blob/d1c2cf58a8eaa994b864cec997bc3f2dad412c58/apps/frontend/src/components/board/board-composer.tsx#L167-L171

2. `apps/frontend/src/components/board/board-composer.tsx:90` — Split persistence: the draft *text* survives collapse/refresh (IDB via `useDraftComposer`), but the selected target is local `useState("")` that unmounts on close. Re-opening reloads the text yet resets the picker to "Post to…", so the user must re-pick their destination with no signal that the rest of their draft is intact. (ux-affordance)
   https://github.com/threahq/threa/blob/d1c2cf58a8eaa994b864cec997bc3f2dad412c58/apps/frontend/src/components/board/board-composer.tsx#L89-L91

3. `apps/frontend/src/components/board/board-composer.tsx:148` — Cancel (X) collapses via `onClose` without clearing the draft, so a half-written post lives on behind the collapsed "Write a post…" button with no indicator that content is waiting. Other composers in the app are always-visible in their stream; this is the only one that hides a live draft. Consider a subtle "draft" affordance on the collapsed button, or confirm-on-discard. (ux-affordance)
   https://github.com/threahq/threa/blob/d1c2cf58a8eaa994b864cec997bc3f2dad412c58/apps/frontend/src/components/board/board-composer.tsx#L147-L150

4. `apps/frontend/src/components/board/board-composer.tsx:138-139` — "New scratchpad" / "New quick note" sit in a "Post to…" list of destinations, so they read as existing places rather than create-new actions. The labels were your explicit call, so this is just a nudge: a leading `+` icon (or a small "Creates a new scratchpad" caption) would disambiguate create-vs-post-to-existing. (ux-discoverability)
   https://github.com/threahq/threa/blob/d1c2cf58a8eaa994b864cec997bc3f2dad412c58/apps/frontend/src/components/board/board-composer.tsx#L137-L140

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>API↔frontend↔types all line up; the discriminated `target` union is consistent across Zod body, `CreateAuthoredPostParams`, `CreateBoardPostParams`, and the composer. Deliberate scope deferrals (no cross-client liveness, no thread pre-creation) are honored, not gaps.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Correctness agent stepped through the create-post transaction, the extractor skip, and the newScratchpad path — no surviving issues. The `clientMessageId` idempotency fast-path doesn't apply (createAuthoredPost doesn't pass one); concurrent double-submit is standard no-client-id behavior, not a regression, and the composer disables submit while pending.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Optimistic prepend dedupes by `conversation.id` across all cached `workspaceList` pages (correct TanStack v5 prefix match); the seeded conversation has `cardinality(message_ids)=1` so it passes the board feed's non-empty filter; no self-echo double-insert (board reads optimistically, not from the per-stream socket room).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>Outbox-in-transaction (INV-4/7), markdown derived server-side (INV-58), Zod-validated body (INV-55), silent success (INV-63), access via `validateStreamAccess` for stream targets (INV-62) all hold. The `confidence: 1` insert flagged by a reviewer is a false positive — `confidence` is a pre-existing `conversations` column (`REAL NOT NULL DEFAULT 0.5`).</details>
<details><summary>🏗️ Design [2 minor notes]</summary>

- `apps/backend/src/features/conversations/boundary-extraction-service.ts` (authored-skip Phase-1 return) — **not a live bug.** The `if (!message || !stream) return` guard runs before the `authoredSkip` check; on the authored path both are present so it passes harmlessly. It's a load-bearing ordering: if `authoredSkip` were ever hoisted above message/stream hydration, the guard would swallow a valid authored message into a `null` return. Worth a one-line comment pinning the order, or a guard test.
- `apps/frontend/src/components/board/board-composer.tsx:143` — `resolveStreamName(...) ?? "Untitled stream"` can show a DM whose peer hasn't hydrated into `dmPeers` yet as "Untitled stream" in the picker. Minor (the user picks from a short list); `streamLabel`/the "generic" context placeholder reads better than a hand-rolled "Untitled stream" tail.
- `ConversationService → StreamService` cross-feature dependency is the intended compose-via-`Querier` pattern (barrel-imported, INV-52 clean) — no action.
</details>
<details><summary>🔒 Security [CLEAN]</summary>The newScratchpad target correctly skips `validateStreamAccess` because the author owns the scratchpad it creates in the same transaction; stream targets are access-checked. No injection/auth-bypass/exposure surface introduced.</details>
<details><summary>🎛️ UX [4 issues]</summary>See numbered list above.</details>
<details><summary>📱 Mobile [CLEAN]</summary>Select+Cancel row fits ~320px without overflow; the inline-mounted shared composer's keyboard choreography is fine at the top of the scroll column; 32px Cancel target is slightly under the 44px guideline but minor (below threshold).</details>

---

kristofferremback:
### Code review (deep pass — concurrency & conversation foundations)
**Confidence: 4/7** — Acceptable (foundations sound in the primary flows; several real, fixable robustness gaps)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x4 (concurrency+invariants deep-dive, correctness+data-flow, spec+design+CLAUDE.md, UX+mobile)

Verified each high-impact claim against the code. **Two reviewer findings were false positives and dropped:** a "TOCTOU → null conversation" (the `message:created` outbox is inserted *inside* createAuthoredPost's transaction, so the boundary-extract job can't run until commit, by which point the conversation exists) and a lock-ordering deadlock (the reviewer retracted it on analysis). The foundation is correct for the **primary** flows — companion scratchpad (one conversation) and channel @mention (agent replies in a thread). The issues below are mostly multi-conversation/edge and robustness.

## Real issues (ranked)

1. `apps/backend/src/features/conversations/handlers.ts:49-51` — **Empty `contentJson` is accepted by the API.** `content: z.array(z.any())` has no minimum, so `{ type: "doc", content: [] }` passes Zod → `deriveContentMarkdown` returns `""` → a conversation seeded with a blank post. The frontend guards via `composer.canSend`, but a direct API call doesn't. Fix: `.min(1)` on `content` (or a `.refine`). (bug)
   https://github.com/threahq/threa/blob/3b4d4ba2ded47d1f40e6ef8f4ecf883f47d7daa2/apps/backend/src/features/conversations/handlers.ts#L48-L53

2. `apps/backend/src/features/conversations/boundary-extraction-service.ts` (`assignAgentReply`, ~L591) — **An inline agent reply joins the *most-recently-active* conversation, not necessarily the one it's replying to.** For the common case (scratchpad = one conversation; @mention → fresh thread) this is correct. But in a multi-conversation surface — notably an **AI-persona DM** (prod has one DM with 450+ conversations) where the agent replies *inline* — if the user fired two messages into different conversations before the reply lands, the reply attaches to whichever was bumped last. Deterministic signal would be the trigger message's conversation (the most-recent *user* message in the stream), not "active[0]". (concurrency/data-flow)
   ``https://github.com/threahq/threa/blob/3b4d4ba2ded47d1f40e6ef8f4ecf883f47d7daa2/apps/backend/src/features/conversations/boundary-extraction-service.ts#L578-L600``

3. `apps/backend/src/features/conversations/service.ts:124-202` — **`createAuthoredPost` has no idempotency key.** `createMessageInTransaction` supports `clientMessageId` dedup, but createAuthoredPost passes none, so a retry / lost-response / double-tap creates a duplicate message + conversation — and for a **new scratchpad**, a duplicate *scratchpad* (two near-identical cards). Frontend only guards via `isPending`. Fix: thread a `clientMessageId` through (and consider a uniqueness guard on the scratchpad create). (bug)
   https://github.com/threahq/threa/blob/3b4d4ba2ded47d1f40e6ef8f4ecf883f47d7daa2/apps/backend/src/features/conversations/service.ts#L124-L150

4. `apps/frontend/src/pages/board.tsx:91-95` — **A post to a brand-new scratchpad renders as "Unknown stream" with the wrong glyph.** The optimistic card's label comes from `resolveStreamName(conversation.streamId, …)` and `streamById.get(streamId)?.type`, but the just-created stream isn't in the workspace cache yet → `"Unknown stream"` + fallback icon until the next bootstrap. The header *link* still works (it uses the server `streamId`). Fix: have `createBoardPost` return the created stream so `onSuccess` can seed the streams cache (`{ ...stream, lastMessagePreview: null }`, per the root CLAUDE.md note). (ux-feedback / data-flow)
   https://github.com/threahq/threa/blob/3b4d4ba2ded47d1f40e6ef8f4ecf883f47d7daa2/apps/frontend/src/pages/board.tsx#L87-L96

5. `apps/backend/src/features/conversations/service.ts:195-199` — **Post-commit hydration that throws turns a committed post into a 500.** `hydrateBoardMessages` runs after the transaction commits (correct, INV-41); if it throws (e.g. an attachment/link-preview read fails), the message + conversation are already durable but the caller gets an error and may retry → duplicate (compounds #3). Fix: wrap the post-commit hydration in try/catch, returning `openingMessage: null` on failure and logging. (bug)
   https://github.com/threahq/threa/blob/3b4d4ba2ded47d1f40e6ef8f4ecf883f47d7daa2/apps/backend/src/features/conversations/service.ts#L194-L202

6. `apps/backend/src/features/conversations/boundary-extraction-service.ts` (`assignAgentReply`) — **A scratchpad whose only conversation is RESOLVED gets a *second* conversation when the agent replies.** `findActiveByStream` excludes resolved, so an agent reply mints a new one → two conversations in one scratchpad. (The existing scratchpad user-message path has the same shape, so this is consistent, not a new regression — but worth a guard: reuse the most-recent conversation regardless of status and `reactivateIfResolved`.) (invariant)
   ``https://github.com/threahq/threa/blob/3b4d4ba2ded47d1f40e6ef8f4ecf883f47d7daa2/apps/backend/src/features/conversations/boundary-extraction-service.ts#L578-L600``

## Watchlist (real, lower priority)

- **Dispatch fan-out:** every persona/bot `message:created` now dispatches a BOUNDARY_EXTRACT job + 2 row locks + 2 outbox events. Cheap per-message and processed individually (verified: the debounced handler loops `for (const event of events) await processEvent(event)`, so no sibling-drop), but high-volume API bots scale the boundary-extraction queue by *agent* traffic now, not just user traffic. Monitor. (`boundary-extraction-outbox-handler.ts:45-47`)
- **INV-8 letter:** `findActiveByStream`/`findByStream` filter by `stream_id` only, no `workspace_id`. Low real risk (stream ids are workspace-unique) and **pre-existing** (the scratchpad path already calls `findByStream(client, stream.id)` unscoped) — not introduced here, but the broader gap is worth closing.
- **`ConversationService` → `EventService` + `StreamService` coupling:** crosses feature boundaries to orchestrate message + scratchpad + conversation in one transaction. It's the price of atomicity (INV-7) and uses barrel imports + `Querier`-composing methods, so defensible — but if you'd rather keep stream lifecycle in `StreamService`, a `createScratchpadWithPost` there is the alternative.
- **@mention thread dual-reference:** the @mention message is a member of its channel conversation *and* the rendered origin of the agent's thread conversation → two board cards reference it. Display nuance, not corruption.
- **Minor:** `topicSummary: trimmedTitle || undefined` drops a literal title `"0"` (use `!== "" ?`); the migration comment is a bit narrative (INV-25); "New scratchpad" gives no signal it summons an AI (the four UX findings from the prior review still stand).

---
<details><summary>✅ Verified safe (checked, not issues)</summary>

- Outbox events in createAuthoredPost & assignAgentReply commit in the same transaction as the writes (INV-4/7). ✓
- assignAgentReply is idempotent on re-delivery (message-row `FOR UPDATE` + `findPrimaryByMessageId` early-return). ✓
- Two agent replies racing in a fresh thread serialize on the stream `FOR UPDATE` → only one conversation minted. ✓
- `createScratchpadInTransaction` is purely transactional (all `OutboxRepository.insert(db, …)`, no out-of-txn socket emit) → clean rollback. ✓
- The authored-boundary extractor skip + reassignment protection are correct for the primary paths. ✓
</details>

---

kristofferremback:
### Code review — conversation-directive rework
**Confidence: 4/7** — Acceptable (sound design; a few real, fixable issues — one of which regresses a claimed benefit)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x3 (concurrency+assigner+extractor deep-dive, spec+design+CLAUDE.md, frontend correctness+data-flow)

The refactor is clean — declared/inferred split, synchronous in-transaction assignment via an injected `ConversationAssigner` (messaging stays decoupled), no board endpoint. Verified each high-impact claim; **one reviewer finding was a false positive and dropped** (infinite-query "refetches only page 1" — TanStack v5 refetches all loaded pages on invalidation, so the board doesn't truncate).

## Real issues

1. `apps/frontend/src/hooks/use-conversations.ts:128-133` — **Board posts carry no `clientMessageId`, so there's no duplicate-on-retry protection.** The whole point of routing through the standard send was to inherit its idempotency (`createMessageInTransaction`'s `clientMessageId` fast-path) — but the hook omits the key, so `MessageRepository.insert` runs without `ON CONFLICT` and a double-submit / retry creates two messages **and** two conversations (the assigner mints a fresh `conversationId()` each time). The button's `isPending` guard narrows the window but isn't a safety net. **This regresses the idempotency I claimed the reuse gave us — my error.** Fix: mint a stable `clientMessageId` per compose attempt (in the composer, regenerated after success) and pass it through. (bug / data-flow)
   https://github.com/threahq/threa/blob/66daf7d3b022a8c17afc2f6aa453f901f8da7943/apps/frontend/src/hooks/use-conversations.ts#L127-L134

2. `apps/frontend/src/hooks/use-conversations.ts:122-126` — **Orphan scratchpad on partial failure, and retry mints another.** For a `newScratchpad` target the hook does two steps: `createStream` then `messageService.create`. If the send fails after the stream is created, an empty scratchpad is left in the sidebar (createStream's optimistic insert isn't rolled back), and retrying creates a *second* scratchpad before the next send attempt. The draft is preserved (good), but each failed retry leaks another empty scratchpad. Options: clean up the stream on send-failure, or reuse a stable draft-scratchpad id across retries, or accept it (an empty scratchpad is openable) — but the retry-multiplies behavior should be addressed. (state-lifecycle)
   https://github.com/threahq/threa/blob/66daf7d3b022a8c17afc2f6aa453f901f8da7943/apps/frontend/src/hooks/use-conversations.ts#L121-L134

3. `packages/types/src/api.ts:344` + `apps/backend/src/features/messaging/handlers.ts:30` — **`ConversationDirective` has two sources of truth** (a hand-written union in `api.ts`, a Zod `discriminatedUnion` in `handlers.ts`), neither derived from the other → drift risk if the `existing` branch grows a field (INV-31). Fix: define the Zod schema once (e.g. in `@threa/types` next to `CONVERSATION_INTENTS`) and `z.infer` the type, or export `z.infer<typeof conversationDirectiveSchema>`. (claude-md INV-31)
   https://github.com/threahq/threa/blob/66daf7d3b022a8c17afc2f6aa453f901f8da7943/packages/types/src/api.ts#L344-L344

4. `apps/backend/src/features/conversations/conversation-assigner.ts:88-90` — **The `existing` path's `findById` isn't workspace-scoped** (`WHERE id = $1` only); the only tenant guard is the app-level `target.workspaceId !== workspaceId` check on the next line (INV-8). Not exploitable via the public API today (the guard runs before any write, and there's no UI for `existing` yet), but it deviates from every other conversation lookup. Fix: use the workspace-scoped `findByIds`, or add a `workspaceId` filter to `findById`. (claude-md INV-8 / security — defense-in-depth)
   https://github.com/threahq/threa/blob/66daf7d3b022a8c17afc2f6aa453f901f8da7943/apps/backend/src/features/conversations/conversation-assigner.ts#L88-L90

## Watchlist (real, lower priority)

- **`emitAssignment` duplicates the agent-reply emit logic** (`conversation-assigner.ts` ↔ `boundary-extraction-service.ts` `assignAgentReply`): both fetch the stream, compute thread `parentStreamId`, re-read the conversation, and emit `conversation:created/updated` + `message_assigned`. Extract a shared helper before a third copy appears (INV-35/37).
- **`existing` path takes no row lock** before `addPrimaryMessage` + `reactivateIfResolved` (INV-20). Harmless today (idempotent array ops; at worst a duplicate `conversation:updated`), but the `assignAgentReply` path locks the stream and this one doesn't — inconsistent.
- **`Message.conversationIntent` is optional (`?`) in `domain.ts:461`** but non-optional `| null` in the backend repo — the wire always carries it, so tighten to non-optional `| null` to avoid `undefined`-vs-`null` ambiguity for consumers.
- **`emitAssignment` re-fetches the stream** the send transaction already loaded (`createMessageInTransaction` has it in scope) — minor redundant SELECT on every board post.
- **Declared messages still enqueue a boundary-extract job** that immediately short-circuits — harmless queue noise; gating the dispatch would require surfacing `conversation_intent` in the `message:created` payload (probably not worth the coupling).
- The dispatch gate's `!actorId` lost its explicit "user-only" semantics — fine today (persona/bot handled by the agent-reply path) but worth a comment so a future actor-having message type isn't silently assigned a conversation.

## Verified safe
- Declared assignment commits in the message transaction (INV-4/7); the extractor's `declaredSkip` correctly short-circuits because the outbox isn't visible until commit, by which point the conversation + primary exist. ✓
- `new` always minting a fresh conversation (even in a scratchpad) is intended (authored = new topic). ✓
- `DuplicateMessageError` rollback path unwinds the assigner's writes too (same transaction). ✓

---

PR #1085, opened 2026-06-25
fix(timeline): cold-load stability — coalesce catch-up + hydrate secondary content eagerly off-screen

kristofferremback:
Work toward one invariant for the stream timeline: **content outside the viewport must never move what's inside it**, and previews/media should be laid out off-screen so they're already there when scrolled to (no growth-on-arrival). Two pieces land here; a third (the cold-load follow-disarm) is pending real-browser verification.

## 1. Coalesce the catch-up replay into one append

Opening a long stream cold paged the tail forward in heavy steps as the workspace **sync catch-up** replayed missed messages — each replayed message writes to `db.events` per entry, and the timeline's `useStreamEvents` `useLiveQuery` re-runs on every write. Every other IDB-backed reactive surface is coalesced through the apply window (`useBatchedValue`, #1021); the timeline was the omission.

Fix: wrap the rendered window in `useBatchedValue` in `useEvents`. While a catch-up holds the apply window open, the timeline stays on the cached tail and settles **once** when it closes. Outside a catch-up window it's a pass-through (live delivery / optimistic sends unaffected).

## 2. Hydrate secondary content eagerly, off-screen

The virtualized timeline (`TimelineMessageList`) staggered secondary-content hydration (link previews, attachment images, embeds) after reveal, releasing rows in batches **bottom-up**. So off-screen rows above the viewport hydrated **last** — they grew under the user as they scrolled up into history, and a search/jump near not-yet-hydrated rows drifted a beat later as content filled in.

Fix: hydrate every mounted row (the whole virtua buffer, on- and off-screen) eagerly, so each reaches its final height before it scrolls into view. Growable media already reserves its box up front — images via `inlineImageBox`, link-preview cards at fixed `h-32`/`h-16` — so eager hydration adds no in-viewport layout shift. (`EventList`, the non-virtualized path, only defers on the pre-`ready` coordinated-load phase, no per-row wave — unaffected.)

## Not yet in this PR — pending real-browser verification

The cold-load "settles ~2 pages up with Jump-to-latest showing" (auto-follow getting disarmed) couldn't be root-caused from static reading — the `handleScroll` model is explicitly defended against the obvious triggers ("content growth never lowers scrollTop"). Confirming the actual trigger needs observing it in a running browser, which isn't available in this environment (the E2E stack is Docker-gated). That part will be verified and fixed against a real app session and added here.

## Modified / new files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-events.ts` | Wrap the rendered window in `useBatchedValue` so catch-up replay coalesces into one append |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Drop the bottom-up hydration stagger in `TimelineMessageList`; hydrate every buffered row eagerly |
| `apps/frontend/src/hooks/use-events.batch.test.tsx` | New: end-to-end proof the timeline window holds steady during a catch-up replay and settles once on close |

## Test plan

- [x] `apps/frontend` timeline + hooks + stores suites — 342 timeline tests + the new batch test pass
- [x] Monorepo lint + typecheck (pre-commit over all apps/packages) — clean
- [ ] Cold-load + search-jump verified in a real browser — **pending** (Docker-gated E2E here; to be done against a live app session)

---

### DIFF
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -469,7 +469,6 @@ export function StreamContent({
     events,
     holes,
     isLoading,
-    bootstrapFetching,
     isConfirmedEmpty,
     error,
     pagedSharedMessages,
@@ -1003,17 +1002,6 @@ export function StreamContent({
   // (new message, link preview, virtua measuring) must not disarm follow.
   const userInteractedAtRef = useRef(0)
 
-  // Hold the cold-load reveal while the stream's tail is still converging: the
-  // bootstrap that establishes the true newest page is still fetching. (A
-  // concurrent sync catch-up replaying into this stream is detected inside the
-  // scroll hook via the apply window — this ref covers the bootstrap fetch the
-  // apply window does not wrap.) Without it the settle reveals the stale cached
-  // tail and then visibly jumps forward as bootstrap/catch-up land newer
-  // messages. A ref so the settle's rAF loop reads it live without re-running
-  // the layout effect.
-  const shouldHoldRevealRef = useRef(false)
-  shouldHoldRevealRef.current = bootstrapFetching
-
   const {
     listRef,
     scrollerRef: virtualScrollerRef,
@@ -1035,7 +1023,6 @@ export function StreamContent({
     skipInitialScroll,
     isJumpMode,
     userInteractedAtRef,
-    shouldHoldRevealRef,
   })
 
   // Scroll container element, owned by useTimelineScroll. Attached to the
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -469,6 +469,7 @@ export function StreamContent({
     events,
     holes,
     isLoading,
+    bootstrapFetching,
     isConfirmedEmpty,
     error,
     pagedSharedMessages,
@@ -1002,6 +1003,17 @@ export function StreamContent({
   // (new message, link preview, virtua measuring) must not disarm follow.
   const userInteractedAtRef = useRef(0)
 
+  // Hold the cold-load reveal while the stream's tail is still converging: the
+  // bootstrap that establishes the true newest page is still fetching. (A
+  // concurrent sync catch-up replaying into this stream is detected inside the
+  // scroll hook via the apply window — this ref covers the bootstrap fetch the
+  // apply window does not wrap.) Without it the settle reveals the stale cached
+  // tail and then visibly jumps forward as bootstrap/catch-up land newer
+  // messages. A ref so the settle's rAF loop reads it live without re-running
+  // the layout effect.
+  const shouldHoldRevealRef = useRef(false)
+  shouldHoldRevealRef.current = bootstrapFetching
+
   const {
     listRef,
     scrollerRef: virtualScrollerRef,
@@ -1023,6 +1035,7 @@ export function StreamContent({
     skipInitialScroll,
     isJumpMode,
     userInteractedAtRef,
+    shouldHoldRevealRef,
   })
 
   // Scroll container element, owned by useTimelineScroll. Attached to the
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -25,7 +25,7 @@ import {
   streamKeys,
   workspaceKeys,
 } from "@/hooks"
-import { useSocket, useCoordinatedLoading } from "@/contexts"
+import { useSocket } from "@/contexts"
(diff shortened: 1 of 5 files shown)


kristofferremback:
### Code review

Found 2 issues — one real residual gap (not a regression), one minor:

1. `apps/frontend/src/hooks/use-events.ts:293-302` — **Socket-connecting race leaves a residual cold-load jump.** On the very first open after an app restart the socket is still `"connecting"`, and the bootstrap query is gated on `!!socket` ([`use-streams.ts:107`](https://github.com/threahq/threa/blob/c4e327913758f3d0974989ad3d195540aed395ce/apps/frontend/src/hooks/use-streams.ts#L107)) — so `fetchStatus` is `"idle"`, `bootstrapFetching` is `false`, and no apply-window is open yet. `held` is therefore `false`, and the settle converges + reveals the **stale cached tail** in ~3 frames (~50ms) before the socket connects. When it then connects (~100–500ms later) the bootstrap fires and newer messages land → one visible cached→authoritative jump, in the exact scenario the PR targets. (ux-feedback) This is **not a regression** — it collapses the old multi-step walk to at most one jump — but the stated "no jumping" goal isn't fully met in the connect sub-window. A clean close: OR `useSocketStatus() === "connecting"` into the hold. It's offline-safe — a real offline cold start transitions `connecting → reconnecting` quickly (`socket-context.tsx`), and the 5s backstop bounds any hold regardless.
   https://github.com/threahq/threa/blob/c4e327913758f3d0974989ad3d195540aed395ce/apps/frontend/src/hooks/use-events.ts#L293-L302

2. `apps/frontend/src/hooks/use-timeline-scroll.ts:24` — **Backstop cap widened 2000ms → 5000ms also widens the no-hold worst case.** The old 2000ms hard cap on `settleToBottom` is now `SETTLE_BACKSTOP_MS = 5000`. For the non-held path (offline / up-to-date / no apply window), reveal still fires on height convergence in ~3 frames, so this is harmless in practice — but if virtua measurement genuinely never stabilizes for 3 consecutive frames (pathological reflow, very heavy first paint), worst-case blank time grows from 2s to 5s. Low impact given convergence normally fires fast; flagging because the cap change is silent for the unrelated no-hold path. (design)
   https://github.com/threahq/threa/blob/c4e327913758f3d0974989ad3d195540aed395ce/apps/frontend/src/hooks/use-timeline-scroll.ts#L15-L24

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>The diff fully implements the described fix: `bootstrapFetching` exposed, threaded through `shouldHoldRevealRef`, consumed live in the settle loop, backstop applied. No missing corresponding changes.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Correctness agent stepped the cold-load sequence and stream-switch: the backstop fires correctly even while held (`performance.now() - start >= maxMs` runs after the held branch), the ref is read live each frame (stable `useRef` identity, no stale closure), `resetKey` re-seeds `isInitialSettling`/`didInitialScrollRef` on stream switch, and the offline path (`fetchStatus: idle`) skips the hold. No wedge/strand path found.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>`isInitialSettling` and the two hold signals traced through cold-load, revisit (`staleTime: Infinity` → no refetch → instant reveal), and stream-switch lifecycles without stranding or premature reveal — except the socket-connecting window in issue 1.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No INV violations introduced. The new refs are component-scoped (not module singletons, INV-9 ok); the apply-window global is the existing approved one. Comments are load-bearing intent, not change-narration (INV-25 ok). No layout shift introduced (INV-21).</details>
<details><summary>🏗️ Design [1 minor]</summary>Issue 2 (backstop widening). Otherwise sound: imperative `isApplyWindowOpen()` poll in the rAF loop avoids a subscription + extra `StreamContent` renders, and the ref-written-during-render pattern matches the repo's own `useBatchedValue`.</details>
<details><summary>🔒 Security [CLEAN]</summary>Scroll/reveal UI logic only; no input handling, no data exposure.</details>
<details><summary>🎛️ UX [1 issue]</summary>Issue 1 (socket-connecting residual jump). The core tradeoff — brief skeleton instead of instant-but-stale-then-jumping — is aligned with the user's stated priority ("content should settle before showing… no jumping"). Deep-link/draft exemptions (`skipInitialScroll`) are correct.</details>
<details><summary>📱 Mobile [CLEAN]</summary>The longer (≤5s) pin window doesn't fight the keyboard (`pinToBottom` idempotent, ResizeObserver/visualViewport paths unchanged), and the settle still aborts+reveals on the first `touchmove`/`pointerdown` (overlay is `pointer-events-none`), so a touch user is never trapped behind the skeleton.</details>

---

PR #1100, opened 2026-06-27
feat(board): put conversations on the sync-engine rails for live updates

kristofferremback:
**Design:** [`docs/board-view-design.md`](https://github.com/threahq/threa/blob/main/docs/board-view-design.md) — "Realtime / sync model — the board must ride the sync engine, not React-Query"

## Problem

The board (PR #1080/#1098) was a refetch-on-open React-Query list with **no socket subscription**. `useWorkspaceConversations` was pure `useInfiniteQuery` with `refetchOnMount`/`refetchOnReconnect` and nothing else — so a reply only surfaced after a manual refresh. The design doc names the exact dogfooding failure:

> Kris typed a direct reply in a stream, opened the board, and his activity wasn't there — a refresh fixed it. For a surface meant to be a co-equal way to interact (not just a read-only digest), seconds-late is disqualifying.

Two stacked causes, both confirmed in code: conversations lived on a different data-plane than messages (not in IDB at all, so no optimism and no live updates), and `conversation:*` events were delivered to per-stream rooms only (`delivery-groups.ts`), so the workspace board never received them.

## Solution

Put conversations on the same rails the timeline rides — IDB store + gate-applied events + optimistic writes — instead of a parallel React-Query path.

### How it works

1. **`conversations` IDB store (Dexie v36).** `apps/frontend/src/db/database.ts` adds `CachedBoardPost` (the wire `BoardPost` + `id` = conversation id, `workspaceId`, `_lastActivityMs`, `_cachedAt`, `_status?`) and the store `"id, workspaceId, [workspaceId+_lastActivityMs], _cachedAt"`. Cleared by `clearAllCachedData`.
2. **The board reads reactively from IDB.** `board-store.ts`'s `useBoardPosts` runs a `[workspaceId+_lastActivityMs]` range scan in `.reverse()` (newest activity first) via `useLiveQuery`, so the feed re-sorts in place the instant a row changes — no refetch.
3. **The query becomes the fetch/seed engine.** `useWorkspaceConversations` keeps the keyset `useInfiniteQuery` but an effect writes every loaded page into IDB via `seedBoardPosts` (`bulkPut`, reconciles in place, never drops rows a page omits). `board.tsx` renders from `useBoardPosts` and uses the query only for pagination / loading / error.
[...]

PR #1104, opened 2026-06-28
feat(board): stable view with pending-update pill, not live re-sort

kristofferremback:
## Problem

PR #1100 put the board's conversations on the sync-engine rails (IDB + gate-applied events + optimistic writes) so cards update live. Dogfooding the consequence: a card you're half-way through reading **jumps out of view** the moment it — or anything above it — gets activity, because the feed re-sorts on every `lastActivityAt` bump. For a surface whose whole job is "get an overview," motion-under-the-eye is the wrong default ("don't move shit on me").

This is the same class of defect INV-61 forbids in the timeline (off-screen content must never shift on-screen content), one projection over: the timeline's rule is about *insertion*; the board needs it for *ordering*.

`docs/board-view-design.md` § "Stable view + pending updates" designed the fix; this PR builds the v1.

## Solution

Hold the order the viewer is looking at **frozen**, accumulate live changes out of band, and reveal them on demand via an X-style "N new" pill — the X / Slack / iMessage pattern. The `conversations` IDB store stays live and activity-sorted (the sync engine keeps it true); a new projection layer sits between it and render. None of #1100's data plane changes — only the projection and the dropped optimistic bump-to-top.

### How it works

1. **`useStableBoardView`** (`hooks/use-stable-board-view.ts`) wraps the live `useBoardPosts` feed and holds a **committed snapshot** — a frozen ordered list of conversation ids plus each id's commit-time activity (`activityById`). Render walks the committed order; each card reads its *content* reactively from IDB (a reply body fills in place) but its *position* never moves while committed.

2. **`reconcileStableView`** (pure, unit-tested) folds the live feed into the committed view by the committed window's frozen lower bound (`floor` = min `activityById`):
   - a live id **at/above the floor** that isn't committed is a fresh arrival → **buffered**, surfaced as the pill count;
   - a live id **below the floor** is older content paged in by "Load more" → **appended below the frozen window**, no pill (it lands below the viewport, shifting nothing on-screen).

3. **Commit triggers** — the pill (`revealNew`: set `scrollTop = 0`, then `commit()` so the new cards flow in where the viewer can see them) and remount (navigating away/back resets to a fresh snapshot). The committed snapshot re-snapshots the live order and prunes retained content to the new set.

4. **The viewer's own post** arms a one-shot `revealNext`; the next live change carrying a fresh arrival commits instead of buffering, so your own authored post surfaces rather than hiding behind its own "1 new" pill. Wired through `BoardComposer`'s new `onPosted` callback.

5. **Dropped the rendered bump-to-top** — `optimisticBoardReply` still raises `_lastActivityMs` as IDB truth (a fresh snapshot lands the card at the top, and the live feed re-sorts), but the committed view ignores it, so the viewer's own reply updates **in place** instead of chasing a moving card. Only the doc comment changed; the write is unchanged.

6. **`useBoardScrollAnchor`** (`hooks/use-board-scroll-anchor.ts`) keeps the topmost visible card's exact screen offset stable across above-fold async reflow — avatars / link previews / images *above* the viewport loading after layout and growing the region. A `ResizeObserver` on the content re-pins the anchor card by adding the above-anchor height delta to `scrollTop`; disabled within `ANCHOR_DISABLE_PX` (4px) of the top so a pill-click reveal lets new cards flow in rather than holding the old first card. This is the board's hand-rolled equivalent of the timeline's `virtua` `shift` (`use-timeline-scroll.ts`); the board is top-anchored and non-virtualized, so it can't reuse `shift` directly.

A vanished committed card (delete / lost access) keeps rendering from last-known content (a retained map) until the next commit drops it — a removal never shifts the rows below it (edge case 2). Recency-section grouping reads `activityById` (commit-time), not live `lastActivityAt`, so a bumped card keeps its section as well as its position and the sections stay monotonic with the frozen order.

### Key design decisions

**1. Focused v1; virtua and the backend paging-key deferred.** The doc's full model is large. This ships the committed-view projection + pill + commit triggers + a hand-rolled top scroll-anchor, and defers (a) adopting `virtua` for the board — it stays non-virtualized, justified by the floor measurement of ~430 active cards; (b) a backend stable-paging-key (edge 1's full fix). The client keyset cursor is captured per page at fetch time, so the re-return/skip window is narrow and self-heals via socket→IDB; the `reconcileStableView` floor-split handles "Load more" correctly without it. (The AskUserQuestion to confirm this scope couldn't be delivered — permission stream closed — so I took the recommended option, which matches the doc's own v1 decisions.)

**2. Snapshot order frozen; content reactive.** Two orderings, one truth: the IDB store is the live activity-sorted truth; the committed view is a deliberately-frozen projection of it. A card's *position* is the snapshot's, its *content* is live. This is what lets a reply body fill in place without the card moving.

**3. Pill counts new conversations only; seen-card bumps stay silent (edge 3).** A seen card getting activity stays put with no pill increment — otherwise "12 new" when nothing is actually new to look at. The optional in-place "updated" dot for seen-card activity is deferred.

**4. v1 freezes the whole snapshot as "seen" (edge 6).** The faithful intersection-observer "only what scrolled past is seen" is deferred as much harder; freezing the whole committed snapshot matches X and is the doc's own v1 call.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/hooks/use-stable-board-view.ts` | The committed-view projection over the live IDB feed: `reconcileStableView` (pure fold), `useStableBoardView` (frozen order + buffer/pill + commit/revealNext), `BoardViewPost` (re-exported post type so the page need not import `@/db`, INV-15). |
| `apps/frontend/src/hooks/use-board-scroll-anchor.ts` | `ResizeObserver` top scroll-anchor: holds the topmost visible card's offset across above-fold reflow; disabled near the top. |
| `apps/frontend/src/hooks/use-stable-board-view.test.tsx` | 11 tests — `reconcileStableView` (commit, buffer, bump-stays-put, paginate-append, mixed split) and the hook (loading→empty, freeze+pill+commit, revealNext, vanished-card retention, workspace reset). |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/pages/board.tsx` | Render from `useStableBoardView` instead of raw `useBoardPosts`; "N new" pill over the feed; resolve the Radix viewport and attach the anchor; `data-board-card` per card; group by `activityById`; wire `revealNext` to the composer. |
| `apps/frontend/src/components/board/board-composer.tsx` | Optional `onPosted` callback, fired after a successful post (reveals the viewer's own card instead of pilling it). |
| `apps/frontend/src/stores/board-store.ts` | `optimisticBoardReply` doc comment only — the `_lastActivityMs` bump is now IDB truth that the committed view ignores; the write is unchanged. |
| `apps/frontend/src/sync/workspace-sync.test.ts` | Make a pre-existing flaky test deterministic: poll for the async IDB-merge `invalidateQueries` instead of a single `setTimeout(0)` (the handler awaits a Dexie transaction before invalidating). |
| `docs/board-view-design.md` | Mark the stable-view section v1-shipped; add a "What shipped" subsection and the deferred follow-ups. |

## Out of scope (deliberate)

- **`virtua` adoption / board virtualization** — deferred; board stays non-virtualized (~430 active cards per the floor measurement).
- **Backend stable-paging-key** (edge 1's full fix) — client keyset cursors are good enough for v1; re-return/skip self-heals via socket→IDB.
- **Intersection-observer "only what scrolled past is seen"** (edge 6) and the **per-card "updated" dot** (edge 3) — both deferred per the doc.
- **No schema / migration change** — frontend-only projection layer over the existing `conversations` IDB store from #1100.

## Test plan

- [x] `src/hooks/use-stable-board-view.test.tsx` — 11 pass
- [x] `src/pages/board.test.tsx`, `src/stores/board-store.test.ts`, `src/components/board`, `src/sync/workspace-sync.test.ts` — 77 pass together (incl. the de-flaked sync test)
- [x] `tsc --noEmit` (frontend) clean; eslint clean on all touched files (incl. INV-15 `@/db`-in-page check, resolved via the `BoardViewPost` re-export)
- [ ] **Scroll-anchor async-reflow compensation not integration-tested** — it's DOM-layout behavior jsdom can't exercise (`ResizeObserver` is a no-op polyfill, `getBoundingClientRect` returns zeros). The hard case (above-fold avatar/link-preview load while reading) needs a real device; flagged for manual `/verify` before merge.
- [ ] **Multi-agent `/code-review` loop** — running post-create (requested); findings will be pushed as follow-up commits and tallied here.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01LJf9rGx2FN1tRxCiN3f38f)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1104"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785218258&installation_id=129946197&pr_number=1104&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1104&signature=dc7a087edfc81ba4fba9dbb401d81d56b953bc761a89037ddc74bf72b4ae5a19"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/board/board-composer.tsx
+++ b/apps/frontend/src/components/board/board-composer.tsx
@@ -82,7 +82,7 @@ function targetForValue(value: string): BoardPostTarget | null {
  * composer. The composer closes on a successful post and the feed refreshes to
  * show it, so the result is visible without a success toast (INV-63).
  */
-export function BoardComposer({ workspaceId }: { workspaceId: string }) {
+export function BoardComposer({ workspaceId, onPosted }: { workspaceId: string; onPosted?: () => void }) {
   const [open, setOpen] = useState(false)
 
   if (!open) {
@@ -98,10 +98,18 @@ export function BoardComposer({ workspaceId }: { workspaceId: string }) {
     )
   }
 
-  return <BoardComposerForm workspaceId={workspaceId} onClose={() => setOpen(false)} />
+  return <BoardComposerForm workspaceId={workspaceId} onClose={() => setOpen(false)} onPosted={onPosted} />
 }
 
-function BoardComposerForm({ workspaceId, onClose }: { workspaceId: string; onClose: () => void }) {
+function BoardComposerForm({
+  workspaceId,
+  onClose,
+  onPosted,
+}: {
+  workspaceId: string
+  onClose: () => void
+  onPosted?: () => void
+}) {
   const streams = useWorkspaceStreams(workspaceId)
   const users = useWorkspaceUsers(workspaceId)
   const dmPeers = useWorkspaceDmPeers(workspaceId)
@@ -165,6 +173,9 @@ function BoardComposerForm({ workspaceId, onClose }: { workspaceId: string; onCl
       composer.clearAttachments()
       writeStoredTarget(workspaceId, "")
       onClose()
+      // Reveal the just-posted card rather than letting it wait behind its own
+      // "N new" pill — the viewer's own action should surface immediately.
+      onPosted?.()
     } catch {
       toast.error("Couldn't post to the board. Please try again.")
     } finally {
--- a/apps/frontend/src/hooks/use-board-scroll-anchor.ts
+++ b/apps/frontend/src/hooks/use-board-scroll-anchor.ts
@@ -1,0 +1,77 @@
+import { useEffect, useRef } from "react"
+
+/** Below this scrollTop the anchor is disabled, so prepended cards flow in at the
+ *  top (the pill-click reveal) instead of being held off-screen above the fold. */
+const ANCHOR_DISABLE_PX = 4
+/** Sub-pixel layout jitter shouldn't trigger a scroll correction. */
+const MIN_COMPENSATION_PX = 0.5
+
+/** Attribute carrying a card's conversation id, read to re-find the anchor card. */
+export const BOARD_CARD_ATTR = "data-board-card"
+
+interface Anchor {
+  id: string
+  /** The card's top offset within the viewport at the moment it was measured. */
+  offset: number
+}
+
+/** The topmost card still touching the viewport, and its offset within it. */
+function measureAnchor(viewport: HTMLElement): Anchor | null {
+  const cards = viewport.querySelectorAll<HTMLElement>(`[${BOARD_CARD_ATTR}]`)
+  const viewportTop = viewport.getBoundingClientRect().top
+  for (const card of cards) {
+    const top = card.getBoundingClientRect().top - viewportTop
+    if (top + card.offsetHeight > 0) {
+      const id = card.getAttribute(BOARD_CARD_ATTR)
+      return id ? { id, offset: top } : null
+    }
+  }
+  return null
+}
+
+/**
+ * Formalize INV-61 for the top-anchored board feed: the topmost visible card
+ * keeps its exact screen offset across any mutation. The hard case is async
+ * reflow — avatars, link previews and images *above* the viewport loading after
+ * the initial layout and growing the region — which would push the card the
+ * viewer is reading downward. A `ResizeObserver` on the content catches every
+ * such growth (the timeline gets the same guarantee from `virtua`'s `shift`; a
+ * non-virtualized board hand-rolls it) and re-pins the anchor card by adding the
+ * above-anchor height delta to `scrollTop`.
+ *
+ * Disabled near the top so a pill-click reveal (scroll to top, then commit) lets
+ * the new cards flow in rather than holding the old first card in place.
+ */
+export function useBoardScrollAnchor(viewport: HTMLElement | null): void {
+  const anchorRef = useRef<Anchor | null>(null)
+
+  useEffect(() => {
+    if (!viewport) return
+
+    // Re-measure the anchor on every user scroll; our own corrections write
+    // scrollTop and re-measure to the same offset, so they don't drift it.
+    const onScroll = () => {
+      anchorRef.current = measureAnchor(viewport)
+    }
+    onScroll()
+    viewport.addEventListener("scroll", onScroll, { passive: true })
+
+    const content = viewport.firstElementChild
+    const observer = new ResizeObserver(() => {
+      if (viewport.scrollTop <= ANCHOR_DISABLE_PX) return
+      const anchor = anchorRef.current
+      if (!anchor) return
+      const card = viewport.querySelector<HTMLElement>(`[${BOARD_CARD_ATTR}="${CSS.escape(anchor.id)}"]`)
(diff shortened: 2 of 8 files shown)


kristofferremback:
### Code review

No issues remain. Checked for bugs, data-flow lifecycle, CLAUDE.md compliance, plan adherence, design quality, security, UX, and mobile across **4 review rounds** — every finding from rounds 1–3 was fixed and re-verified clean.

<details><summary>📋 Review loop — findings & fixes</summary>

Spec & design and correctness & data-flow came back **clean every round** (the render-time stable-view derivation, the `revealNext` timer lifecycle, `reconcileStableView` floor math, and the anchor's settle-timer were each traced for loops/leaks/wedges and held up). All surviving findings were UX/mobile and were fixed:

**Round 1** → fixed in `4f323a0`
- `mobile-touch`: "N new" pill below the 44px tap-target minimum → `min-h-11`.
- `ux-affordance`: pill overlapped the composer at the top → (initially) scroll-gated.
- `ux-consistency`: `revealNext` armed indefinitely, could fire on a foreign post → bounded with an 8s auto-disarm timer.
- `accessibility`: pill missing `aria-label` / icon not `aria-hidden` → added.

**Round 2** → fixed in `097dc47`
- `ux-discoverability`: the scroll-gate hid new-content affordance at the top → replaced with a sticky pill (always reachable, no overlap).
- `accessibility`: no `aria-live` for the count → added an `aria-live="polite"` status region.
- `mobile-touch`: anchor wrote `scrollTop` mid-iOS-momentum (judder) → defer compensation during touch/momentum, re-pin once on settle (desktop unchanged).

**Round 3** → fixed in `e96642a`
- `ux-affordance` + `mobile-touch` (both lenses): the zero-height (`h-0`) pill hung over the first card and stole its taps → gave the sticky row real height so it sits in-flow; the only cost is a banner-style shift at the very top (anchor compensates while scrolled), which both reviewers endorsed.

**Round 4** → clean (UX + mobile, the lenses the round-3 layout change touched).

</details>

---

PR #1112, opened 2026-06-29
fix(board): board replies always land flat in the conversation

kristofferremback:
## Problem

Replying to a board card whose conversation was a **single message in a channel or DM** didn't show the reply under that card. `planBoardReply` (in `apps/frontend/src/hooks/use-conversations.ts`) routed a lone channel/DM root to a brand-new thread (`{ kind: "newThread" }`): the send promoted a draft thread off the opening message, so the reply landed in its own conversation and surfaced as a *separate* board card (initially a collapsed "1 more message") rather than in place. On device this read as a dropped send — you replied, and nothing appeared on the card you replied from. The eager-optimistic work in #1111 made every other board reply appear instantly in place; the lone-root case was the one path that didn't.

## Solution

Board replies always join the conversation flat. A reply from the board feed is a reply *into the conversation* — which #1110 framed as a soft thread — while starting a strict thread stays the timeline's gesture (`stream-panel.tsx`), not the board's. Concretely, `useReplyToBoardPost` now always sends with `conversation: { intent: "existing", conversationId }` into the conversation's own stream, so the eager optimistic event renders under the card that produced it. A lone channel/DM root now gains an in-place reply instead of sprouting a thread card.

This removes the only branch in the board reply path, so the routing helper and its supporting state collapse out:

- Deleted `planBoardReply` / `BoardReplyPlan` and the `newThread` queue branch (the `createDraftPanelId` + `streamCreation: { type: THREAD, … }` send). `useReplyToBoardPost` now returns `void` instead of `{ plan }`.
- Dropped the now-unused inputs `openingMessageId`, `hostStreamType`, `messageCount` from `ReplyToBoardPostInput`, and the `hostStreamType` prop threaded `board-card.tsx → BoardReplyComposer`.
- Removed the composer's transient `"Posted to a new thread"` resting note (the `"posted"` `RestingState`, its 4s fade timer, and the `posted` plumbing through `close`). It existed only because a new-thread reply had no in-place echo; every reply now has one, so the visible reply is the confirmation (INV-63 — success is silent).

### Key design decision

**Board joins the conversation; the timeline threads** — chosen over keeping the lone-root→thread split (Kris's call this session). The two reply surfaces map to the two shapes: replying from the timeline opens a thread via `stream-panel.tsx`; replying from the board joins the conversation. This reverses the earlier "a bare channel/DM root should thread, not sprout a top-level reply" ruling — the accepted tradeoff is that a reply to a lone channel/DM root card now appears as a top-level message in that channel's timeline. The board surface stays direct (reply lands where you replied), which is the behavior the device test demanded.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-conversations.ts` | `useReplyToBoardPost` always sends `intoConversation`; removed `planBoardReply`/`BoardReplyPlan`, the `newThread` branch, the `createDraftPanelId` import, and three now-dead `ReplyToBoardPostInput` fields. |
| `apps/frontend/src/components/board/board-reply-composer.tsx` | Removed the `"posted"` resting state, its label, its fade timer, and the `hostStreamType` prop; `handleSubmit` no longer reads a `plan`. |
| `apps/frontend/src/components/board/board-card.tsx` | Stopped passing `hostStreamType` to `BoardReplyComposer`. |
| `apps/frontend/src/hooks/use-conversations.test.tsx` | Dropped the `planBoardReply` suite; `useReplyToBoardPost` tests now assert flat routing, including the lone-channel-root → `intoConversation` (no thread) case. |

## Out of scope (deliberate)

- The timeline thread-reply path (`stream-panel.tsx`) is untouched — threading from the timeline still works exactly as before.
- Bounding/collapsing tall board cards and making E2E scratchpads first-class on the board are separate follow-ups, not in this PR.

## Test plan

- [x] `apps/frontend` `vitest run` over the affected files — `use-conversations.test.tsx`, `board.test.tsx`, `use-board-card-messages.test.tsx`, `board-composer.test.tsx`: 39 pass
- [x] Full pre-commit suite on commit: all-app ESLint, monorepo `tsc --noEmit` (incl. frontend app + test tsconfigs), `check:migrations`, OpenAPI `--check`, astro check — all clean
- [ ] Not manually re-verified on device this session — the behavior change (lone-root reply now lands in place) is the exact fix requested from the prior device test; covered by the hook unit test rather than a fresh device run

---

---
_Generated by [Claude Code](https://claude.ai/code/session_019bL9QBbQjeSr9p4JzwvwbF)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1112"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785332947&installation_id=129946197&pr_number=1112&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1112&signature=3ef59b7e2023f34f6242e4d68149e734396308ecaa4ac946ad26ebbc92343e7a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/board/board-card.tsx
+++ b/apps/frontend/src/components/board/board-card.tsx
@@ -158,7 +158,7 @@ export function BoardCard({ workspaceId, post, contextLabel, streamType }: Board
         )}
       </div>
 
-      <BoardReplyComposer workspaceId={workspaceId} post={post} hostStreamType={streamType} />
+      <BoardReplyComposer workspaceId={workspaceId} post={post} />
     </div>
   )
 }
--- a/apps/frontend/src/components/board/board-reply-composer.tsx
+++ b/apps/frontend/src/components/board/board-reply-composer.tsx
@@ -21,25 +21,17 @@ import type { BoardPost, JSONContent } from "@threa/types"
 const COMPOSER_POPOVER_SELECTOR = '[role="listbox"],[data-radix-popper-content-wrapper],[role="dialog"]'
 
 // Resting-affordance state. `draft` signals a persisted-but-collapsed reply (so
-// the user knows reopening restores it); `posted` is a transient confirmation
-// that a new-thread reply went through — it stands in for the in-place echo a
-// same-conversation reply gets, since a new-thread reply lands in its own
-// conversation and only surfaces on the next board load.
-type RestingState = "idle" | "draft" | "posted"
+// the user knows reopening restores it). A sent reply needs no resting note: it
+// lands in place under the card, so the visible echo is the confirmation.
+type RestingState = "idle" | "draft"
 const RESTING_LABEL: Record<RestingState, string> = {
   idle: "Write a reply…",
   draft: "Continue reply…",
-  posted: "Posted to a new thread",
 }
-// How long the transient "posted" confirmation stays before the affordance
-// returns to its normal invitation.
-const POSTED_NOTE_MS = 4000
 
 interface BoardReplyComposerProps {
   workspaceId: string
   post: BoardPost
-  /** Host stream type of the post's conversation, selecting the reply routing. */
-  hostStreamType: string | undefined
 }
 
 /**
@@ -50,12 +42,11 @@ interface BoardReplyComposerProps {
  * mirrors the composer's own container — same radius, border, surface, padding,
  * and placeholder — and the composer mounts already open (`initialMobileChromeOpen`)
  * so the tap lands straight in the toolbar view instead of stepping through the
- * mobile compacted phase. Routing — channel → thread, everything else → the
- * conversation — lives in {@link useReplyToBoardPost}.
+ * mobile compacted phase. The send always lands flat in the conversation (see
+ * {@link useReplyToBoardPost}), so the reply shows in place under this card.
  *
  * The resting affordance carries state (best-effort, in-session): it flags a
- * persisted draft after an Escape ("Continue reply…") and a transient "Posted to
- * a new thread" confirmation after a new-thread send, so a collapse never reads
+ * persisted draft after an Escape ("Continue reply…") so a collapse never reads
  * as a no-op or an accidental discard.
  */
 export function BoardReplyComposer(props: BoardReplyComposerProps) {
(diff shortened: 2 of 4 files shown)


kristofferremback:
### Code review

No issues found. Checked for CLAUDE.md compliance, plan/intent adherence, design quality, bugs, data-flow lifecycle, security, UX, and mobile/responsive behavior.

Notes from the pass:
- **Dead-symbol cleanup is complete (INV-38).** `planBoardReply`, `BoardReplyPlan`, `POSTED_NOTE_MS`, the `"posted"` `RestingState`, the `createDraftPanelId` import, and the three dropped `ReplyToBoardPostInput` fields (`openingMessageId`, `hostStreamType`, `messageCount`) have no remaining references anywhere in the frontend source or tests. The `hostStreamType` prop drop propagates cleanly `BoardCard → BoardReplyComposer → BoardReplyComposerForm`.
- **No orphaned timer/leak.** The removed `useEffect` was the sole owner of the `"posted"` fade `setTimeout` (cleared in its own cleanup). `useEffect`/`useCallback`/`useMemo` imports remain in use.
- **Optimistic data-flow preserved.** The reply is tagged `{ intent: "existing", conversationId }`, exactly what `useBoardCardMessages` reads to render it in place; `attachmentIds`/`attachments` forwarding is unchanged.
- **INV-63 alignment.** Removing the "Posted to a new thread" note is correct now that every reply has an in-place echo — the visible reply is the confirmation. The previously-documented INV-21 layout-shift concern is eliminated along with the state that created it (the remaining `idle`/`draft` labels share identical DOM and the shorter strings can't regress truncation).

---

kristofferremback:
Closing in favor of the right design. The flat-reply approach here fixes the "dropped send" feel but flattens lone channel posts into the channel timeline, which interleaves conversations (`conv-1, conv-1, conv-2, conv-1…`) — the exact mess Threa avoids. The agreed direction instead: a board reply to a **lone channel message converts it into a thread** (root stays in the channel, back-and-forth lives in the thread) and retires the now-redundant lone conversation so it surfaces as one coherent card. Existing conversations (DMs, established channel conversations, threads) keep replying flat where they live. New work will land in a fresh PR.

---

PR #1113, opened 2026-06-29
feat(board): convert a lone channel/DM post into a thread on reply

kristofferremback:
## Problem

Replying to a board post from the feed should keep the parent stream's top level clean. The earlier behavior (`planBoardReply`'s `newThread` branch, since reverted) threaded a lone channel/DM post but left the original one-message conversation behind, so the board showed **two cards** — the post and the thread — and the reply landed on the second, reading as a dropped send. Making board replies flat instead (PR #1112, closed) avoided the duplicate but flattened lone posts into the channel timeline, interleaving unrelated conversations (`conv-1, conv-1, conv-2, conv-1…`) — the "Slack where information dies" mess Threa exists to avoid.

The right shape (Kris's call): a board reply joins the conversation, but a **lone** post (a single message, no back-and-forth yet) in a channel or DM has no established shape, so the reply should **convert it into a thread** — opener stays in the parent stream as the thread's root, the exchange lives in the thread, the parent stream's top level stays one-opener-per-post. The blocker was the conversation model: a thread is a new stream, so the reply's conversation is necessarily a *different* row from the source, and there was **no mechanism to retire the source** so the board wouldn't double-card.

## Solution

A new `threadFromMessage` conversation directive that, in the reply's send transaction, mints the thread's conversation seeded with the reply **and retires the lone source** — by emptying it, not by inventing a new "hidden" status. An empty conversation already drops off the board (`findByWorkspaceForViewer` filters `cardinality(message_ids) > 0`), so emptying the source is the retirement.

### How it works

1. **Routing** (`apps/frontend/src/hooks/use-conversations.ts`, `planBoardReply`): a lone (`messageCount <= 1`, has an opener) post in a `CHANNEL` or `DM` → `{ kind: "convertToThread" }`; everything else → `intoConversation`. DMs and channels route identically (a fresh DM exchange started from the board threads too); established conversations, thread cards, and scratchpads reply flat.
2. **Send** (`useReplyToBoardPost`): the `convertToThread` branch promotes a draft thread off the opener (`createDraftPanelId` + `streamCreation: { type: THREAD, parentStreamId, parentMessageId }`, the same path as the timeline's thread reply) and carries `conversation: { intent: "threadFromMessage", sourceConversationId }`. The durable queue's post-promotion plaintext send already forwards `conversation` (`use-message-queue.ts:248`), so no queue change was needed.
3. **Assign + retire** (`apps/backend/src/features/conversations/conversation-assigner.ts`): the `threadFromMessage` branch shares the `new` mint+seed (a fresh conversation in the thread stream, seeded with the reply, `conversation:created`), then locks the source (`findByIdForUpdate`, INV-8/20) and — only if it's still lone and lives in a different stream than the reply — removes its message(s) (`removePrimaryMessage`), `resolveIfEmpty`, and emits `conversation:updated` for it via `emitConversationRetired` (routed by the source's own stream, INV-62). The opener is **not moved** — it stays in the parent stream as the thread's `parentMessageId`, referenced not relocated, so no channel broadcast slot is vacated (simpler than `moveMessagesToThread`, INV-61).
4. **Board drop** (`apps/frontend/src/stores/board-store.ts`, `mergeBoardConversation`): an incoming aggregate with `messageIds.length === 0` deletes the row instead of upserting, mirroring the server filter. Returns `true` even with no cached row so the `conversation:updated` handler doesn't try to hydrate a card that shouldn't exist. This also fixes a latent gap: reassignment already emits `conversation:updated` for an emptied source (`service.ts:300-310`) but the board kept the stale card.

### Key design decisions

**1. Retire by emptying, not by a new status.** Conversation `status` is `active | stalled | resolved`; `resolved` already means agent-invisible + "not a current topic" (`conversation-highlight.ts`, `cross-surface-stitch.ts`), and the board shows all statuses — so overloading `resolved` would both pollute that meaning and fail to hide the card. There is no soft-delete on conversations. Emptiness is the existing, proven board-retirement mechanism (it's how reassignment retires a conversation), so the directive reuses it.

**2. A new directive variant, not `moveMessagesToThread` or a second REST call.** `moveMessagesToThread` *relocates* messages out of the channel (vacates broadcast slots, needs a validation lease, moves the root) — wrong shape, since the opener must stay as the thread parent. A separate resolve call after the send isn't atomic: a thread-created-but-source-not-retired crash re-introduces the duplicate. Folding the retire into the assigner (which already runs in the send transaction) makes the conversion all-or-nothing (INV-20).

**3. Retirement is idempotent and defensive.** Unlike `existing` (which 400s on a stale id), a missing/foreign `sourceConversationId` is a no-op so a retried send can't fail; a non-lone source (gained a message in a race) or a source in the reply's own stream is left intact rather than gutted.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/constants.ts` | Add `threadFromMessage` to `CONVERSATION_INTENTS` / `ConversationIntents`. |
| `packages/types/src/api.ts` | Add the `{ intent: "threadFromMessage"; sourceConversationId }` variant to `ConversationDirective`. |
| `apps/backend/src/features/messaging/handlers.ts` | Add the variant to the Zod directive union (the compile-time type↔schema lock enforces parity). |
| `apps/backend/src/features/conversations/conversation-assigner.ts` | `threadFromMessage` branch: mint+seed the thread conversation, then empty + retire the lone source. |
| `apps/backend/src/features/conversations/assignment-events.ts` | New `emitConversationRetired` — `conversation:updated` for the emptied source, routed by its own stream. |
| `apps/frontend/src/hooks/use-conversations.ts` | `planBoardReply` → `convertToThread`; `useReplyToBoardPost` carries the `threadFromMessage` directive. |
| `apps/frontend/src/stores/board-store.ts` | `mergeBoardConversation` deletes an emptied conversation instead of upserting. |
| `apps/frontend/src/components/board/board-reply-composer.tsx` | Rename the plan kind in the resting-note selector (`convertToThread`). |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/features/conversations/conversation-assigner.test.ts` | Unit-tests the `threadFromMessage` mint+retire, non-lone/foreign/same-stream guards (scoped `spyOn`, INV-48). |

## Out of scope (deliberate)

- **Seamless optimistic swap.** Between send and echo, the reply lands in the thread (not shown under the source card yet); the card swaps to the thread on echo, bridged by the "Posted to a new thread" note. No dropped-send feel (the card never vanishes), but optimistically tagging the reply onto the source so the card content is unchanged through the swap is a fast follow, not this cut — it's the exact reconciliation spot prior optimistic-swap bugs lived in.
- **Private-source live retirement for non-members.** A public-channel source retires live for other board viewers (`conversation:updated` reaches the workspace room); a private channel/DM source relies on the members' own stream-scoped echo + board refetch. Matches how board conversation events already deliver.
- **Established conversations are not retroactively threaded** — only lone posts convert; an established channel/DM conversation replies flat where it lives.

## Test plan

- [x] Backend `bun test src/features/conversations src/features/messaging` — 147 pass (incl. 4 new assigner tests: mint+retire, non-lone intact, idempotent-missing, same-stream guard)
- [x] Frontend `vitest run` — `use-conversations.test.tsx` (routing: lone channel→thread, lone DM→thread, established→flat, thread/scratchpad→flat, deleted-opener→flat), `board-store.test.ts` (delete-on-empty + uncached-empty handled), board render suites
- [x] Full pre-commit suite: all-app ESLint, monorepo `tsc --noEmit` (incl. frontend app + test tsconfigs), `check:migrations`, OpenAPI `--check`, astro check — clean
- [ ] Not device-tested this session — staging label added for manual verification of the card swap on a real board

---

---
_Generated by [Claude Code](https://claude.ai/code/session_019bL9QBbQjeSr9p4JzwvwbF)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1113"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785343970&installation_id=129946197&pr_number=1113&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1113&signature=3581a3c3f526b956ef7bda4bb1160b7f3ab2a7e67aaae4f479cee354a689d49c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/conversations/assignment-events.ts
+++ b/apps/backend/src/features/conversations/assignment-events.ts
@@ -71,7 +71,12 @@ export async function emitConversationRetired(
   const stream = await StreamRepository.findById(client, streamId)
   const { parentStreamId, streamVisibility } = await resolveConversationDelivery(client, stream)
   const [refreshed] = await ConversationRepository.findByIds(client, workspaceId, [conversationId])
-  if (!refreshed) return
+  if (!refreshed) {
+    // The row we just locked + emptied in this same transaction is gone — an
+    // invariant break (INV-11). Throwing rolls back the whole send rather than
+    // silently dropping the board-removal event and leaving a stale source card.
+    throw new Error(`Conversation ${conversationId} vanished during retirement`)
+  }
   await OutboxRepository.insert(client, "conversation:updated", {
     workspaceId,
     streamId,
--- a/apps/backend/src/features/conversations/assignment-events.ts
+++ b/apps/backend/src/features/conversations/assignment-events.ts
@@ -53,3 +53,31 @@ export async function emitAssignmentEvents(
 
   return refreshed
 }
+
+/**
+ * Emit the aggregate update for a conversation that just lost its last message
+ * (e.g. a lone source conversation retired when its message was threaded off).
+ * Mirrors the reassignment path's emptied-source emit (`service.ts`): the board
+ * drops a now-empty conversation (its `cardinality > 0` filter on the server,
+ * the delete-on-empty merge on the client), so this is the signal that retires
+ * the card. Routed by the conversation's OWN stream (INV-62) — `streamId` here
+ * is the source's parent stream, not the thread the reply landed in.
+ */
+export async function emitConversationRetired(
+  client: PoolClient,
+  params: { workspaceId: string; conversationId: string; streamId: string }
+): Promise<void> {
+  const { workspaceId, conversationId, streamId } = params
+  const stream = await StreamRepository.findById(client, streamId)
+  const { parentStreamId, streamVisibility } = await resolveConversationDelivery(client, stream)
+  const [refreshed] = await ConversationRepository.findByIds(client, workspaceId, [conversationId])
+  if (!refreshed) return
+  await OutboxRepository.insert(client, "conversation:updated", {
+    workspaceId,
+    streamId,
+    conversationId,
+    conversation: addStalenessFields(refreshed),
+    parentStreamId,
+    streamVisibility,
+  })
+}
--- a/apps/backend/src/features/conversations/conversation-assigner.test.ts
+++ b/apps/backend/src/features/conversations/conversation-assigner.test.ts
@@ -43,6 +43,13 @@ describe("conversationAssigner — threadFromMessage", () => {
     removePrimaryMessage = spyOn(ConversationRepository, "removePrimaryMessage").mockResolvedValue(undefined as never)
     resolveIfEmpty = spyOn(ConversationRepository, "resolveIfEmpty").mockResolvedValue(undefined as never)
     emitRetired = spyOn(assignmentEvents, "emitConversationRetired").mockResolvedValue(undefined as never)
+    // The thread (the reply's stream `thr_1`) hangs off `msg_open` in `chan_1` —
+    // so the only retire-able source is the lone conversation owning that message.
+    spyOn(streams.StreamRepository, "findById").mockResolvedValue({
+      id: "thr_1",
+      parentStreamId: "chan_1",
+      parentMessageId: "msg_open",
+    } as never)
     // Default: the actor can reach the source stream (the legit board-reply case).
     checkStreamAccess = spyOn(streams, "checkStreamAccess").mockResolvedValue({ id: "chan_1" } as never)
   })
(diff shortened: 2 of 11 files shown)


kristofferremback:
### Code review
**Confidence: 7/7** — Excellent (final state, after fixes)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x2 final pass (spec+design, correctness+data-flow) — UX + mobile ran clean in the first pass with no user-facing changes since.

No issues remain on the final commit (`ba17024`). Checked CLAUDE.md compliance, plan adherence, design, bugs, data-flow lifecycle, and security.

This review looped — the findings raised during it were all fixed in-branch:

- **Source retirement was not access-gated** (own review, first pass) → added `checkStreamAccess(source.streamId, actor)` before retiring (`17d7eb9`).
- **Retirement could target any accessible lone conversation, not the replied-to one** (CodeRabbit, Critical) → bound the retire to the thread's actual parent: `source.streamId === thread.parentStreamId && source.messageIds[0] === thread.parentMessageId` (`ba17024`). A crafted `sourceConversationId` for an unrelated conversation now fails the parent-identity check (covered by a dedicated test).
- **Silent no-op if the locked source vanished on refresh** (CodeRabbit, Major, INV-11) → `emitConversationRetired` now throws, rolling back the send rather than dropping the board-removal event (`ba17024`).
- **`mergeBoardConversation` threw on `conversation:*` payloads omitting `messageIds`** (caught by CI — `workspace-sync.test.ts`) → guard on an explicit empty array; a partial payload falls through to upsert (`ba17024`).

Final-pass verification highlights:
- Parent-binding fully closes the cross-stream arbitrary-retire path; `parentMessageId` is set at thread-creation (which itself requires access to the parent stream), so it isn't attacker-controllable into a stream the actor can't reach.
- The new throw is unreachable on a normal flow — the row is locked + UPDATE-only (never deleted) within the same transaction.
- Idempotent on retry: a re-run finds `messageIds.length === 0` and skips; sends dedupe on `clientMessageId` before the assigner anyway.

---

PR #1114, opened 2026-06-29
feat(board): show convert-to-thread reply in place during the swap

kristofferremback:
## Problem

Replying from the board to a lone channel/DM post converts it into a thread (PR #1113): the opener stays in the parent stream as the thread root, the reply goes into a new thread, and the lone source conversation is retired so the board shows one card. The reply is queued as a thread-draft send (`useReplyToBoardPost` → `convertToThread`), so its optimistic `message_created` event is written under the **thread-draft stream** (`streamId: panelId`), never under the source card's stream.

`useBoardCardMessages` reads a card's replies from its own stream rail (`useStreamRail(post.conversation.streamId)`), so the source card never saw the in-flight reply. From send until the server echo swapped the source card for the thread card, the source card showed only the bare opener — a "did my reply send?" window. PR #1113 papered over it with a transient "Posted to a new thread" note on the composer.

## Solution

Surface the in-flight reply on the source card by reading the **pending send row**, keyed to the source conversation it threads off. Each `threadFromMessage` send carries a `sourceConversationId` directive (`db.pendingMessages[].conversation`), so the source card can render the reply in place — opener + reply — from send through draft promotion, then drop it as the pending row is deleted on send success and the `conversation:*` echo hands the card over to the thread.

Read-side only: no change to the send path, the queue drain, or draft promotion. The optimistic event stays off the source channel's timeline (it remains on the thread-draft stream), so the channel doesn't flash an interleaved flat reply.

### How it works

1. **`usePendingThreadConversions(workspaceId)`** — one shared `liveQuery` per workspace (the same INV-9 carve-out the rail registry already takes), ref-counted across every card. It scans `db.pendingMessages` (the in-flight outbox, tiny) for rows whose `conversation.intent === "threadFromMessage"`, then `bulkGet`s their optimistic events by `clientId` and projects each into a `RenderableMessage` (`eventToRenderable`, so attachments carry through). Result is a `Map<sourceConversationId, RenderableMessage[]>`.
2. **Early return** — when no conversion is in flight (`pendings.length === 0`) the query returns the shared `EMPTY_CONVERSIONS` map *before* touching `db.events`, so in the common case it observes only `pendingMessages` (rarely written) and never re-runs on ordinary message writes.
3. **Fold into the card** — `useBoardCardMessages` merges the conversion replies into `pendingReplies` (deduped against the rail's tagged rows and the server `messageIds`), so the existing card render path (`BoardCard` → `displayedReplies`) shows them in place with no card-level change.
4. **Clears in step with the swap** — the drain deletes the pending row on send success (`db.pendingMessages.delete`), which is when the `conversation:updated` (source emptied → card removed) and `conversation:created` (thread card) echoes land.

### Key design decisions

**1. Read the pending send, not the optimistic event tag.** The reply's optimistic event lives on the thread-draft stream and is moved to the real thread stream at promotion, so tagging it with the source conversation (as flat `existing` replies do) would either pollute the source channel timeline or carry the source tag onto the real thread reply (wrong — it belongs to the thread). The pending row already holds the `sourceConversationId` link and is naturally scoped to the in-flight window, so it's the clean key.

**2. Cover the round-trip, accept the inherent card swap.** Continuity spans the long part of the wait — the send round trip plus draft promotion. The final source-card → thread-card transition at echo is a different conversation row (stream-scoped), so it's a remove+add, not an update; that swap is unchanged from #1113. This masks the reply, it doesn't merge the two rows.

**3. Keep the "Posted to a new thread" note.** With the reply now visible in place the note no longer covers an invisible reply, but it still names the routing outcome the reply alone doesn't convey — that this lone card is becoming a thread. Comments updated to reflect that.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-board-card-messages.ts` | Add `usePendingThreadConversions` (shared workspace liveQuery over `pendingMessages`, INV-9 carve-out) and `buildPendingConversions`; fold conversion replies into `pendingReplies`; extend `__clearBoardRailRegistry` to drain the new registry. |
| `apps/frontend/src/hooks/use-board-card-messages.test.tsx` | 3 cases: surfaces the convert-to-thread reply on the lone source card; clears it when the pending row is deleted on send success; doesn't leak onto a non-source card. Plus a `seedPendingConversion` helper and `pendingMessages` cleanup in `beforeEach`. |
| `apps/frontend/src/components/board/board-reply-composer.tsx` | Update two now-stale comments (the reply *is* shown in place); note text unchanged. |

## Out of scope (deliberate)

- The source-card → thread-card row swap itself is not merged into a single in-place update (different stream-scoped conversation rows). Unchanged from #1113.
- The other board-reply-routing follow-ups (Ariadne context-reply, tall-card bounding, E2E scratchpads on the board) are separate.

## Test plan

- [x] `vitest run use-board-card-messages.test.tsx` — 15 pass (3 new)
- [x] `vitest run` board + sync suites (`board-composer`, `board-store`, `workspace-sync`) — 66 pass
- [x] `tsc --noEmit -p tsconfig.json` (frontend) — clean
- [x] `eslint` on the 3 changed files — clean
- [x] Pre-commit hook (full monorepo typecheck + OpenAPI `--check` + prettier) — passed
- [ ] No manual board run in this session — behavior is covered by the hook tests (optimistic insert → pending-row delete → no-leak); the live source→thread visual swap is unchanged from #1113.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01XBw6m1RdSddjY4oMN6zFmv)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1114"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785354092&installation_id=129946197&pr_number=1114&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1114&signature=c086a6fd0f63abf04e781e3e0148d3e4c77c596fb4e33ca51bc33a883ad4145f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/board/board-card.tsx
+++ b/apps/frontend/src/components/board/board-card.tsx
@@ -20,6 +20,11 @@ interface BoardCardProps {
   contextLabel: string
   /** Resolved stream type, selecting the context glyph. */
   streamType: string | undefined
+  /** Called when a reply here converts this lone post into a thread. The reply
+   *  retires this card and mints a new thread card, a fresh board arrival that
+   *  would otherwise wait behind the "N new" pill — arming the stable view's
+   *  reveal lets the thread card take this card's place in line instead. */
+  onConvertedToThread: () => void
 }
 
 const TYPE_GLYPH: Record<string, LucideIcon> = {
@@ -36,7 +41,7 @@ const TYPE_GLYPH: Record<string, LucideIcon> = {
  * conversation (no reply indentation); a collapsed "N more messages" gap fills
  * in on click. The stream it lives in is the header locator, not a topic line.
  */
-export function BoardCard({ workspaceId, post, contextLabel, streamType }: BoardCardProps) {
+export function BoardCard({ workspaceId, post, contextLabel, streamType, onConvertedToThread }: BoardCardProps) {
   const { conversation } = post
   const { getActorName } = useActors(workspaceId)
   const currentUserId = useWorkspaceUserId(workspaceId)
@@ -158,7 +163,12 @@ export function BoardCard({ workspaceId, post, contextLabel, streamType }: Board
         )}
       </div>
 
-      <BoardReplyComposer workspaceId={workspaceId} post={post} hostStreamType={streamType} />
+      <BoardReplyComposer
+        workspaceId={workspaceId}
+        post={post}
+        hostStreamType={streamType}
+        onConvertedToThread={onConvertedToThread}
+      />
     </div>
   )
 }
--- a/apps/frontend/src/components/board/board-reply-composer.tsx
+++ b/apps/frontend/src/components/board/board-reply-composer.tsx
@@ -22,9 +22,11 @@ const COMPOSER_POPOVER_SELECTOR = '[role="listbox"],[data-radix-popper-content-w
 
 // Resting-affordance state. `draft` signals a persisted-but-collapsed reply (so
 // the user knows reopening restores it); `posted` is a transient confirmation
-// that a convert-to-thread reply went through — it bridges the brief swap where
-// this lone card retires and the thread card takes its place on the server echo,
-// since (unlike a same-conversation reply) the reply isn't shown in place here.
+// that a convert-to-thread reply went through. The reply itself shows in place on
+// the card (its in-flight send surfaces under this lone source card until the
+// thread card takes over on echo — see `usePendingThreadConversions`), so this
+// note isn't covering an invisible reply; it names the routing outcome the reply
+// alone doesn't convey — that this lone card is becoming a thread.
 type RestingState = "idle" | "draft" | "posted"
 const RESTING_LABEL: Record<RestingState, string> = {
   idle: "Write a reply…",
@@ -199,11 +201,11 @@ function BoardReplyComposerForm({
     try {
       // Eager + offline-first: the reply is enqueued as an optimistic event and
       // the send drains in the background, so there's no created message to await
-      // here — only the resolved plan, which selects the resting note. An
-      // `intoConversation` reply rides the card's own rail and shows in place; a
-      // `convertToThread` reply lands in a thread off the opener (and retires this
-      // lone card, which swaps to the thread on echo), so the transient "Posted to
-      // a new thread" note bridges the brief swap.
+      // here — only the resolved plan, which selects the resting note. Both routes
+      // show the reply in place: an `intoConversation` reply rides the card's own
+      // rail, and a `convertToThread` reply surfaces from its in-flight send under
+      // this lone card until the thread card takes over on echo. The transient
+      // "Posted to a new thread" note only names the convert-to-thread routing.
       const { plan } = await reply.mutateAsync({
         conversation: post.conversation,
         openingMessageId: post.openingMessage?.id ?? null,
--- a/apps/frontend/src/components/board/board-reply-composer.tsx
+++ b/apps/frontend/src/components/board/board-reply-composer.tsx
@@ -42,6 +42,9 @@ interface BoardReplyComposerProps {
   post: BoardPost
   /** Host stream type of the post's conversation, selecting the reply routing. */
   hostStreamType: string | undefined
+  /** Invoked once a reply converts this lone post into a thread, so the board can
+   *  reveal the resulting thread card in place instead of behind the "N new" pill. */
+  onConvertedToThread: () => void
 }
 
 /**
@@ -118,6 +121,7 @@ function BoardReplyComposerForm({
   workspaceId,
   post,
   hostStreamType,
+  onConvertedToThread,
   onClose,
 }: BoardReplyComposerProps & {
   onClose: (opts?: { refocus?: boolean; hadContent?: boolean; posted?: boolean }) => void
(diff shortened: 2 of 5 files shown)


kristofferremback:
### Code review
**Confidence: 7/7** — Excellent (after fixes)

Reviewed at `0124421`; fixes pushed in `14c10c5`. Four lenses ran (frontend diff). Two comment-accuracy findings fixed; one UX finding dismissed as pre-existing. No correctness, data-flow, security, or mobile issues. CodeRabbit's own review returned no actionable comments.

Found 3 issues (2 fixed, 1 dismissed):

1. `apps/frontend/src/hooks/use-board-card-messages.ts` — `claude-md` (INV-25): the doc comment's "Without this … a dead 'did my reply send?' window" leaned on change-narration. **Fixed in `14c10c5`** — kept the structural why (the optimistic event lands on the thread-draft stream, never the source card's rail), dropped the counterfactual.
2. `apps/frontend/src/hooks/use-board-card-messages.ts` — `design`: the reactivity comment misattributed what spares the `db.events` firehose. **Fixed in `14c10c5`** — it's the early-return guard (skip `db.events` when nothing is converting), not bulkGet's key-scoped observation; outbox writes do re-run the query, which is fine at that table's size. _(Flagged independently by the spec and correctness lenses.)_
3. `apps/frontend/src/components/board/board-reply-composer.tsx` — `ux-affordance`: the "Posted to a new thread" note offers no navigation to the new thread. **Dismissed** — the note's behavior is unchanged by this PR (it predates it, from #1113), and adding a link to the transient resting label is constrained by INV-21 (the composer already avoids an icon/link in the "posted" state to prevent layout shift). A pre-existing product choice, not a regression this diff introduces.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Read-side only as intended — no write-path / queue-drain / promotion changes; the optimistic event stays off the source channel timeline (it remains on the thread-draft stream). The kept "Posted to a new thread" note matches the stated intent.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No logic, null-hazard, stale-closure, or listener-leak issues. Ref-counted teardown and `useSyncExternalStore` snapshot identity (stable `EMPTY_CONVERSIONS` reference; atomic Map replacement) verified.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Traced send → optimistic event on thread-draft stream + pending row → source card shows reply → promotion moves the event's streamId (still resolved by `clientId`) → send-success deletes the pending row in step with the `conversation:*` echo. No double-render (the thread card is keyed by a different conversation id), no blink-out across the window, two pending replies to the same source card accumulate + sort correctly, and `bulkGet` index alignment with `pendings` holds.</details>
<details><summary>📋 CLAUDE.md Compliance [1 fixed]</summary>INV-25 comment narration — fixed in `14c10c5`. The INV-9 carve-out for the shared `pendingConversionsRegistry` mirrors the rail registry's documented exception and is justified the same way (one liveQuery per workspace, ref-counted, teardown on last unmount).</details>
<details><summary>🏗️ Design [1 fixed]</summary>Reactivity-comment precision — fixed in `14c10c5`. The liveQuery early-returns before touching `db.events` when no conversion is in flight, so it does not observe the message firehose in the common case.</details>
<details><summary>🔒 Security [CLEAN]</summary>Read-side IDB only; no injection / auth / XSS surface introduced.</details>
<details><summary>🎛️ UX [1 dismissed]</summary>Note-navigation affordance — pre-existing, out of scope (see #3). Render parity confirmed: the in-place reply uses the same `RenderableMessage` shape and `pendingReplies` render path as a confirmed reply, and stays visible across send-retry backoff (the pending row isn't deleted on failure).</details>
<details><summary>📱 Mobile [CLEAN]</summary>No CSS, breakpoint, touch, or viewport changes; surfacing an extra pending row on the otherwise-unchanged card introduces no mobile-specific regression (tall-card bounding is a separate follow-up, not amplified here).</details>

---

kristofferremback:
Closing — superseded by a correction to the conversation model, not a rebase.

This PR's machinery (`usePendingThreadConversions`, the pending-row bridge, `revealNext`-on-convert) exists to mask the flicker of a **cross-row card swap**: convert-to-thread retires the source conversation and mints a new thread conversation, so the board removes one card and adds another. That swap only exists because the design assumed a conversation lives in a single stream.

It doesn't. A conversation spans a root stream and its threads (this is already the documented intent — `docs/board-view-design.md`), and must — a top-level discussion continuing in a thread, or a thread answered at the top level, are the same conversation. Under the correct model, convert-to-thread keeps **one** conversation spanning the source root + the new thread (one-root invariant: a conversation never crosses roots, which keeps access a single root-membership check). The conversation id is stable, the board card updates in place, and there's no swap to mask — so this PR's compensating machinery is unnecessary.

Outcome of the discussion is a design-doc update capturing the corrected model (one-root conversations, convert-to-thread without retire/mint, recency-biased continuation, multi-stream board rendering), then a unified implementation. The failure-mode mapping here wasn't wasted — it's what surfaced the wrong assumption.

---

PR #1116, opened 2026-06-30
feat(link-previews): sync-render the in-app message preview card from baked per-viewer data

kristofferremback:
## Problem

The in-app **message** link preview card (`InAppLinkPreviewCard`) resolves its data asynchronously through `useResolvedInAppLink` — a permission-checked round-trip per card. To avoid a layout jump when the skeleton swaps for the resolved card, #1108 reserved a fixed footprint (`CARD_BODY_MESSAGE="min-h-[5.5rem]"`, `CARD_BODY_MEMO="min-h-[7rem]"`) on every tier. The timeline's scroll anchor no-ops a mid-list row that grows after first paint (`use-timeline-scroll`, `overflow-anchor: none`), so the reserve is load-bearing for INV-21.

The cost the user flagged: a short (1-line) message card carries reserved bottom padding it doesn't use. That was accepted in #1108 as a deliberate zero-shift trade-off, with sync-rendering noted as the follow-up that removes it.

## Solution

Bake the resolved, access-tiered in-app data into the message payload so the card paints once at its natural height — the way web (GitHub/Linear) cards already ship their `previewData`. The card then renders synchronously: no skeleton, no second round-trip, no reserve.

The constraint that shapes where this can happen: the resolve is **per-viewer**. `accessTier` (`full`/`private`) depends on the viewer's access to the linked stream, and a DM's `recipientName` excludes the viewer. The shared live `link_preview:ready` broadcast has no single viewer to resolve against — baking `full` content there would leak a private linked message's content to a recipient who can't see it. So data is baked **only where a `userId` is in hand**: the history / catch-up REST assembly. That is exactly the surface where the mid-list INV-21 shift mattered; live messages (received at the bottom) keep the async resolve and its reserve.

### How it works

1. **Type** — `LinkPreviewSummary` gains optional `inAppData?: InAppLinkPreviewData` (`packages/types/src/domain.ts`), the resolved access-tiered payload.
2. **Backend resolve (per-viewer)** — `LinkPreviewService.getPreviewsForMessages` takes a `userId` and, for each completed in-app preview row, calls the existing `resolveInAppTarget(workspaceId, userId, contentType, row)` and assigns the result to `summary.inAppData`. Resolves across the page are collected and run under one `Promise.all` rather than serializing per message. Non-in-app (web) previews get no `inAppData`. Only caller is `enrichEventsWithLinkPreviews` (`streams/handlers.ts`), which already has the viewer's `userId`.
3. **Override propagation** — `applyLinkPreviewStateToEvents` overrides a `message_created` event's `payload.linkPreviews` with the enriched map version only when it differs from the stored payload. The stored payload has no `inAppData`, so without comparing it the enriched copy looked equal and the override was skipped. `areLinkPreviewArraysEqual` now compares `inAppData` (`isInAppDataEqual`, reference-equal fast path then `JSON.stringify`), so the difference registers and the baked data survives.
4. **Frontend sync render** — `InAppLinkPreviewCard` reads `preview.inAppData`. When present it renders `ResolvedInAppLink` directly and disables the resolve query (`hydrate && !baked`); when absent it keeps the async path (skeleton → card). A `reserve` boolean threads down to `MessageLinkCard` / `MemoLinkCard` / `MinimalCard`: `true` on the async path (skeleton preceded it — keep the fixed footprint), `false` on the baked path (sizes to content). The body class becomes `reserve ? CARD_BODY_MESSAGE : undefined`.

### Key design decisions

**1. Bake only at the per-viewer boundary, never on the broadcast.** The resolve is access-tiered per viewer; the live `link_preview:ready` event is shared. Putting resolved `full` content on the broadcast would leak private linked-message content to non-members (INV-8 ownership / the resolve's whole reason for being permission-checked). History/catch-up fetches carry a `userId` and assemble per-viewer, so that is the only safe place to bake. Live messages stay async — no regression, and they land at the bottom where the mid-list anchor problem doesn't apply.

**2. Keep the reserve on the async path; drop it only when baked.** A given card mount is either baked (history) or async (live) and never switches in place, so there's no in-session reserve→natural transition to cause a shift. Dropping the reserve unconditionally would reintroduce the skeleton→card jump on the live path that #1108's reserve exists to prevent. Threading a `reserve` flag keeps INV-21 intact on the async path and removes the padding everywhere the data is known up front.

**3. Compare `inAppData` in the array-equality check.** The alternative — always override when the map has an entry — would defeat the no-op short-circuit that keeps `enrichEventsWithLinkPreviews` from rebuilding unchanged events. Comparing the field is the minimal change that lets the baked data through without weakening the existing guard.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/domain.ts` | Add `inAppData?: InAppLinkPreviewData` to `LinkPreviewSummary` with a docblock on the per-viewer baking contract |
| `apps/backend/src/features/link-previews/service.ts` | `getPreviewsForMessages(workspaceId, userId, messageIds)` resolves in-app previews per viewer (concurrent across the page) and bakes `inAppData`; import `isInAppLinkContentType` |
| `apps/backend/src/features/streams/handlers.ts` | Pass `userId` to `getPreviewsForMessages`; `areLinkPreviewArraysEqual` compares `inAppData` via new `isInAppDataEqual` |
| `apps/frontend/src/components/timeline/in-app-link-preview-card.tsx` | Render synchronously when `preview.inAppData` is present (no skeleton, query disabled); thread a `reserve` flag so baked cards use natural height while the async path keeps the fixed footprint |
| `apps/backend/src/features/link-previews/service.test.ts` | Tests: per-viewer `inAppData` baked on in-app previews but not web; private tier baked when the viewer lacks access |
| `apps/backend/src/features/streams/handlers.test.ts` | Test: override applies when only `inAppData` differs (equality-propagation guard) |
| `apps/frontend/src/components/timeline/in-app-link-preview-card.test.tsx` | Test: baked `inAppData` renders on first paint with no resolve call |

## Out of scope (deliberate)

- **The singular `GET .../link-previews` endpoint** (`getForMessage` → `getPreviewsForMessage`) is unchanged. It feeds dismiss-state hydration, not the card's data (the card reads `inAppData` off the event-payload previews), so it doesn't need the per-viewer resolve.
- **Live broadcast messages** keep the async resolve + reserve by design (see decision 1). On reload they're re-fetched through the history path and become baked.
- **Backend `contentPreview` mid-markdown-link truncation** (a separate observed bug where the snippet is cut inside `[label](url)`) is not addressed here.

## Test plan

- [x] Backend `bun test src/features/link-previews` — 154 pass
- [x] Backend `bun test src/features/streams/handlers.test.ts` — 11 pass
- [x] Frontend `vitest run` over `components/timeline`, `components/message`, `sync/stream-sync` — 389 pass (incl. the new card test)
- [x] Frontend `vitest run` over `link-preview-list`, `composer-link-previews`, `in-app-link-inline`, `in-app-link-block` — 17 pass
- [x] `tsc --noEmit` across `packages/types`, `apps/backend`, `apps/frontend` — clean
- [x] ESLint clean on all touched files; Prettier applied
- [ ] No browser/visual harness run here — the zero-shift behavior on the async path is unchanged (same reserve constants) and asserted only structurally, not pixel-measured in this PR

---

---
_Generated by [Claude Code](https://claude.ai/code/session_0135TgSPHM4BoyZpPMfBHTQK)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1116"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785390077&installation_id=129946197&pr_number=1116&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1116&signature=5c5872a23a4e44430a41c43ccc9b986dea5ad0d98aab04cf63695213f687868a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/link-previews/service.test.ts
+++ b/apps/backend/src/features/link-previews/service.test.ts
@@ -405,6 +405,28 @@ describe("LinkPreviewService.getPreviewsForMessages", () => {
     expect(previews.every((p) => p.inAppData?.kind === "memo")).toBe(true)
     expect(accessibleSpy).toHaveBeenCalledTimes(1)
   })
+
+  test("a failed memo-access lookup leaves memo previews unbaked without failing the page", async () => {
+    spyOn(LinkPreviewRepository, "findByMessageIds").mockResolvedValue(
+      new Map([
+        [
+          "msg_1",
+          [
+            makePreview({ id: "lp_memo", contentType: "memo_link", targetStreamId: null, targetMemoId: "memo_a" }),
+            makePreview({ id: "lp_stream", contentType: "stream_link", targetStreamId: "stream_ok" }),
+          ],
+        ],
+      ]) as never
+    )
+    spyOn(SearchRepository, "getAccessibleStreamsWithMembers").mockRejectedValue(new Error("db down"))
+    const service = makeService({ tryAccess: async () => makeStream({ id: "stream_ok", slug: "general" }) }, {})
+
+    const previews = (await service.getPreviewsForMessages(WORKSPACE_ID, VIEWER_ID, ["msg_1"])).get("msg_1")!
+
+    // The non-memo preview still bakes; the memo one is left unbaked, no throw.
+    expect(previews.find((p) => p.id === "lp_stream")!.inAppData).toMatchObject({ kind: "stream", accessTier: "full" })
+    expect(previews.find((p) => p.id === "lp_memo")!.inAppData).toBeUndefined()
+  })
 })
 
 describe("LinkPreviewService.resolveInAppLinkByUrl", () => {
--- a/apps/backend/src/features/link-previews/service.test.ts
+++ b/apps/backend/src/features/link-previews/service.test.ts
@@ -348,6 +348,63 @@ describe("LinkPreviewService.getPreviewsForMessages", () => {
 
     expect(result.get("msg_1")![0].inAppData).toEqual({ kind: "stream", accessTier: "private" })
   })
+
+  test("a single resolve failure leaves that preview unbaked without failing the page", async () => {
+    spyOn(LinkPreviewRepository, "findByMessageIds").mockResolvedValue(
+      new Map([
+        [
+          "msg_1",
+          [
+            makePreview({ id: "lp_ok", contentType: "stream_link", targetStreamId: "stream_ok" }),
+            makePreview({ id: "lp_bad", contentType: "stream_link", targetStreamId: "stream_err" }),
+          ],
+        ],
+      ]) as never
+    )
+    const service = makeService(
+      {
+        tryAccess: (async (streamId: string) => {
+          if (streamId === "stream_err") throw new Error("boom")
+          return makeStream({ id: "stream_ok", slug: "general" })
+        }) as never,
+      },
+      {}
+    )
+
+    const previews = (await service.getPreviewsForMessages(WORKSPACE_ID, VIEWER_ID, ["msg_1"])).get("msg_1")!
+
+    expect(previews.find((p) => p.id === "lp_ok")!.inAppData).toMatchObject({ kind: "stream", accessTier: "full" })
+    expect(previews.find((p) => p.id === "lp_bad")!.inAppData).toBeUndefined()
+  })
+
+  test("resolves the viewer's accessible streams once for multiple memo previews", async () => {
+    spyOn(LinkPreviewRepository, "findByMessageIds").mockResolvedValue(
+      new Map([
+        [
+          "msg_1",
+          [
+            makePreview({ id: "lp_a", contentType: "memo_link", targetStreamId: null, targetMemoId: "memo_a" }),
+            makePreview({ id: "lp_b", contentType: "memo_link", targetStreamId: null, targetMemoId: "memo_b" }),
+          ],
+        ],
+      ]) as never
+    )
+    const accessibleSpy = spyOn(SearchRepository, "getAccessibleStreamsWithMembers").mockResolvedValue(["stream_1"])
+    const service = makeService(
+      {},
+      {
+        getById: (async (_ws: string, memoId: string) => ({
+          memo: { title: memoId, abstract: "abstract", knowledgeType: "decision" },
+          sourceStream: { id: "stream_1", type: "channel", name: "eng" },
+        })) as never,
+      }
+    )
+
+    const previews = (await service.getPreviewsForMessages(WORKSPACE_ID, VIEWER_ID, ["msg_1"])).get("msg_1")!
+
+    expect(previews.every((p) => p.inAppData?.kind === "memo")).toBe(true)
+    expect(accessibleSpy).toHaveBeenCalledTimes(1)
+  })
 })
 
 describe("LinkPreviewService.resolveInAppLinkByUrl", () => {
--- a/apps/backend/src/features/link-previews/service.test.ts
+++ b/apps/backend/src/features/link-previews/service.test.ts
@@ -292,6 +292,64 @@ describe("LinkPreviewService.resolveInAppLink", () => {
   })
 })
 
+describe("LinkPreviewService.getPreviewsForMessages", () => {
+  test("bakes per-viewer in-app data onto in-app previews but not web previews", async () => {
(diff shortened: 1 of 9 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x4 (spec+design, correctness+data-flow, UX, mobile), 2 rounds

Reviewed across two rounds (the second after fixes). Four issues found, **all fixed and verified clean** in the current diff (`5c927c3`):

1. `apps/frontend/src/sync/stream-sync.ts` — `handleLinkPreviewReady` replaced `linkPreviews` wholesale, so a live `link_preview:ready` broadcast (which can't carry per-viewer `inAppData`) wiped data the history fetch had baked, reverting the card to the async path. **Fixed** (`e66c44a`): `preserveBakedInAppData` carries baked `inAppData` forward by preview id when the broadcast lacks it; a re-extracted preview (edit) gets a fresh id and is correctly not matched.
2. `apps/backend/src/features/link-previews/service.ts` — `getPreviewsForMessages` re-ran `resolveUserAccessibleStreamIds` once per memo preview and fanned out every resolve concurrently. **Fixed** (`e66c44a`): accessible streams resolved once per page and shared; fan-out bounded with `pLimit(8)`.
3. `apps/backend/src/features/link-previews/service.ts` — the optional bake used `Promise.all`, so one `resolveInAppTarget` rejection rejected the whole history/catch-up enrichment (also raised by CodeRabbit). **Fixed** (`e66c44a`): per-item try/catch + `logger.warn`, leaving that preview unbaked while the rest complete.
4. `apps/backend/src/features/link-previews/service.ts` — the `pending`/`result` shared-summary aliasing that lets the fan-out write `inAppData` through into the returned map was load-bearing but unannotated. **Fixed** (`5c927c3`): clarifying comment so a future copy-on-write refactor can't silently drop the bake.

Security: confirmed no cross-viewer path — `inAppData` is only set on per-viewer history/catch-up objects, never on any broadcast (`toLinkPreviewSummary` omits it), and the client merge only carries the same client's own previously-baked data.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Per-viewer-only baking enforced; no broadcast path carries inAppData. Fixes match stated intent without scope creep.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Index pairing in the batch loop correct; `[] ?? query` semantics correct (explicit empty array yields the private tier without re-querying); per-item catch can't abort the page.</details>
<details><summary>🔁 Data Flow [CLEAN — 1 fixed]</summary>Baked `inAppData` now survives a live `link_preview:ready` via merge-by-id (finding 1).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-21 reserve preserved on the async path; INV-25 comments load-bearing after finding 4.</details>
<details><summary>🏗️ Design [CLEAN — 2 fixed]</summary>Memo accessible-streams dedup + bounded fan-out (finding 2); shared-aliasing annotated (finding 4).</details>
<details><summary>🔒 Security [CLEAN]</summary>No cross-viewer leak of per-viewer access-tiered data.</details>
<details><summary>🎛️ UX [CLEAN]</summary>No new layout-shift scenario; a mount is baked-or-async and never switches in place; dismiss/hover/link affordances unchanged on the sync path.</details>
<details><summary>📱 Mobile [CLEAN]</summary>No `isMobile` branches; dropping the min-h reserve on baked cards only shrinks height (strictly better for scroll stability); `max-w-md` unchanged.</details>

---

PR #1120, opened 2026-06-30
fix(link-previews): strip markdown before truncating preview snippets

kristofferremback:
## Problem

In-app link preview cards bake a short text snippet per viewer — message `contentPreview`, stream `description`, memo `abstract`. All three were built by slicing the **raw markdown** at `CONTENT_PREVIEW_MAX_LENGTH` (200) chars in `apps/backend/src/features/link-previews/service.ts`.

When the 200-char cut lands inside a markdown link, it drops the closing `](url)`. The frontend strips markdown at render (`stripMarkdownToInline` in `in-app-link-preview-card.tsx`, per INV-60), but its link regex `\[([^\]]+)\]\([^)]+\)` needs the closing `](url)` to match. With it gone, the strip can't repair the half-link and the literal `[Kristoffer Östlu…` ships to the user as characters.

The same slice-then-strip pattern sat on all three fields, so any preview whose link straddled the boundary leaked literal syntax.

## Solution

Strip the markdown server-side **before** truncating, so the cut always lands on plain text. A new module-level helper does it once and the three sites call it:

```ts
function buildPreviewSnippet(markdown: string): string {
  const text = stripMarkdownToInline(markdown)
  return text.length > CONTENT_PREVIEW_MAX_LENGTH ? text.slice(0, CONTENT_PREVIEW_MAX_LENGTH) + "…" : text
}
```

`stripMarkdownToInline` is the shared stripper in `packages/types/src/markdown-strip.ts` — the same one the frontend uses, documented for "single-line preview surfaces … where the source text is markdown but the surface only shows a flattened snippet". It strips syntax and collapses `\n+`→space, so the snippet is exactly the flattened single line the card shows, and the 200-char cut counts displayed characters. The frontend still runs `stripMarkdownToInline` at render; on already-clean text that is idempotent (verified: `stripMarkdown(stripMarkdown(x)) === stripMarkdown(x)`), so nothing regresses — the boundary cut is simply no longer destructive.

This mirrors how push-notification bodies already strip server-side (`push/service.ts`, via the same helper): when there is no render step that can repair a snippet, the strip has to happen before the text is bounded.

### Key design decisions

**1. Strip then truncate, not truncate then strip** — the failure mode is order-dependent. Truncating first can sever a token (`](url)`, a fenced block) that the stripper then can't recognize; stripping first removes all syntax so the remaining cut is over plain text only. The handover flagged "truncate after stripping markdown, or strip server-side" — this does both at once.

**2. Fix all three fields, not just the message** — the handover named only `contentPreview`, but `description` and `abstract` share the identical slice-then-frontend-strip path (the card runs `stripMarkdownToInline` on all three) and so carry the identical latent bug. One helper covers them.

**3. Ship stripped text on the wire for these snippet fields** — a narrow, deliberate deviation from INV-60's "backend sends raw markdown, frontend strips at render". These are derived *preview* snippets, not canonical content, and the bounding (200-char cut) is what makes raw markdown unsafe here. The canonical message/stream/memo content is untouched and still travels as raw markdown.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/link-previews/service.ts` | Add `buildPreviewSnippet` (strip-then-truncate); route message `contentPreview`, stream `description`, memo `abstract` through it; import `stripMarkdownToInline` from `@threa/types` |
| `apps/backend/src/features/link-previews/service.test.ts` | Add a test: a markdown link spanning the 200-char boundary yields a snippet that keeps the label, contains no `](` / `https://`, and stays ≤ 201 chars |

## Out of scope (deliberate)

- **Sync-render for stream/memo cards** — message cards were the only ones needing synchronous render (#1116); stream/memo cards keep the async path. Not touched (INV-36).
- **Multi-byte / emoji-shortcode boundary splits** — `.slice` on UTF-16 code units can still split a surrogate pair or a `:shortcode:`. Pre-existing (the old slice had the same property) and not amplified by this change, so left alone.

## Test plan

- [x] `bun test src/features/link-previews/service.test.ts` — 20 pass (19 existing + 1 new boundary test)
- [x] `tsc --noEmit` (backend) — clean
- [x] `eslint` + `prettier --check` on both changed files — clean
- [x] Verified the shared stripper's edge behavior directly: `""` → `""` (no crash), idempotent on stripped text, `[label](url) tail` → `label tail`
- [x] Pre-PR review (2 reviewer agents, sonnet — spec+design, correctness+data-flow): 2 findings fixed — (a) use `stripMarkdownToInline` over `stripMarkdown` to flatten newlines and match the push precedent; (b) tightened the helper comment to a load-bearing ordering constraint per INV-25. Correctness/data-flow pass was clean (empty-string, idempotency, no-expansion, surrogate-split-pre-existing all confirmed).

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01UwN5gAzDhNgBfKgedrPCP9)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1120"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785410159&installation_id=129946197&pr_number=1120&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1120&signature=868d23841ef746452a9f9376139b44c0cdedc0c27c70d4ff139ec0f383cf7227"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/link-previews/service.test.ts
+++ b/apps/backend/src/features/link-previews/service.test.ts
@@ -210,6 +210,40 @@ describe("LinkPreviewService.resolveInAppLink", () => {
     })
   })
 
+  test("strips markdown before truncating so a link cut at the boundary never leaks literal syntax", async () => {
+    // A link whose markdown spans the snippet cut: a raw slice would drop the
+    // closing `](url)` and ship the literal "[label…". Stripping first leaves
+    // clean text to truncate, so no bracket or url leaks to the card.
+    const label = "Kristoffer Östlund"
+    const longContent = `[${label}](https://app.threa.io/w/ws_self/s/stream_1?m=msg_2) ${"x".repeat(300)}`
+    spyOn(LinkPreviewRepository, "findById").mockResolvedValue(
+      makePreview({
+        contentType: "message_link",
+        targetMessageId: "msg_1",
+        url: "https://app.threa.io/w/ws_self/s/stream_1?m=msg_1",
+      })
+    )
+    spyOn(MessageRepository, "findById").mockResolvedValue({
+      id: "msg_1",
+      streamId: "stream_1",
+      authorType: "user",
+      authorId: "user_author",
+      contentMarkdown: longContent,
+      deletedAt: null,
+    } as never)
+    spyOn(UserRepository, "findById").mockResolvedValue({ name: "Author", avatarUrl: null } as never)
+    const service = makeService({ tryAccess: async () => makeStream({ slug: "general" }) }, {})
+
+    const result = await service.resolveInAppLink(WORKSPACE_ID, VIEWER_ID, "lp_1")
+
+    const preview = (result as { contentPreview?: string }).contentPreview ?? ""
+    expect(preview.startsWith(`${label} `)).toBe(true)
+    expect(preview).not.toContain("](")
+    expect(preview).not.toContain("https://")
+    // Truncated to the 200-char cap plus the single ellipsis, not the full tail.
+    expect(preview.length).toBeLessThanOrEqual(201)
+  })
+
   test("a DM message resolves the non-author participant as the recipient", async () => {
     spyOn(LinkPreviewRepository, "findById").mockResolvedValue(
       makePreview({
--- a/apps/backend/src/features/link-previews/service.ts
+++ b/apps/backend/src/features/link-previews/service.ts
@@ -9,7 +9,7 @@ import type {
   LinkPreviewSummary,
   MessageLinkPreviewData,
 } from "@threa/types"
-import { getAvatarUrl, isInAppLinkContentType } from "@threa/types"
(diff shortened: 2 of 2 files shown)


kristofferremback:
@coderabbitai review

---

PR #1119, opened 2026-06-30
feat(board): open a conversation in the side panel (Mechanism B)

kristofferremback:
## Problem

The board shipped (flag-gated `board-view`) and a conversation can now span its
root + threads under one root (#1117). But a conversation could only be read as a
**board card** — a windowed preview (opening + 3 recent + "N more"). There was no
way to open one conversation on its own, read it in full, and reply scoped to it.
`docs/board-view-design.md` §"Conversations as soft threads" calls this **Mechanism
B**: a conversation as a first-class side panel, peer to a thread.

The side-panel infra is already generic (`panel-context.tsx` routes `?panel=`,
`ThreadPanelSlot` + `usePanelLayout` host arbitrary content), but every panel id
was a stream id or a `draft:` id, and only `stream.tsx` mounted the panel.

## Solution

A `conv:<id>` panel kind that renders a conversation as a projection — read
flattened-chronological across its root + threads off the same `db.events` rail the
board card and timeline ride, replied to through the recency-biased board-reply
path. No stream is mutated; access stays the conversation's **single root check**
(INV-62) — the panel adds no per-message-stream gating, which the one-root invariant
makes redundant.

### How it works

1. **Panel kind + dispatcher.** `panel-context.tsx` gains `conv:<id>` helpers
   (`isConversationPanel` / `parseConversationPanel` / `createConversationPanelId`,
   mirroring the existing `draft:` helpers). A new `PanelHost`
   (`components/layout/panel-host.tsx`) picks `ConversationPanel` vs `StreamPanel`
   by panel kind; both `stream.tsx` and `board.tsx` host the panel through it, so
   either surface opens either kind. `board.tsx` grows the two-column layout
   (`usePanelLayout` + `ThreadPanelSlot`) it didn't have, with the mobile
   full-screen takeover `stream.tsx` already uses.

2. **The panel body reuses the card's projection.** `ConversationPanel` calls
   `useBoardCardMessages` (the #1117 cross-stream rail + the convert-to-thread
   pending-reply bridge) for the live opening + replies, and backfills the full
   ordered list via `getBoardMessages` — the panel is permanently "expanded", so it
   runs the same merge the board card runs on expand. The scoped reply reuses
   `BoardReplyComposer` (directive `existing`, or `threadFromMessage` for a lone
   post), targeting the conversation's most-recently-active stream
   (`displayedReplies.at(-1)?.streamId ?? conversation.streamId`).

3. **Sourcing the post by id.** `useConversationBoardPost` prefers the reactive
[...]

PR #1121, opened 2026-06-30
feat(conversations): bidirectional message↔conversation navigation

kristofferremback:
**Design:** `docs/board-view-design.md` §"Conversations as soft threads — panel + provenance + attached context" (Mechanisms A/B). Follows PR #1119 (conversation side panel, Mechanism B).

## Problem

After #1119 a conversation opens in the side panel (`?panel=conv:<id>`), but the two surfaces — the stream timeline and the conversation surfaces (board card, conversation panel) — were one-way streets. From a conversation row you could only reach the message's home stream via the timestamp permalink (easy to miss); from a stream message there was no way at all to jump to the conversation it belongs to. The data to link them already exists client-side (each out-of-stream row carries its own origin `streamId`; the per-stream conversation list carries `messageId → conversationId`), it just wasn't surfaced as an action.

## Solution

Two per-message context-menu actions, one in each direction, both routed through the shared `messageActions` registry so the desktop dropdown (`MessageContextMenu`) and the mobile drawer (`MessageActionDrawer`) pick them up with no per-surface wiring.

### How it works

**Direction A — conversation/board → stream ("View in channel/thread/DM/scratchpad").** A new `viewInStream: { href; label }` field on `MessageActionContext`, set only by `MessageItem` (the component that renders a message *outside* its home stream — board card, conversation panel, label page). The `view-in-stream` action is gated on it (`when: (ctx) => !!ctx.viewInStream`) and navigates via `getHref` to the existing message permalink `/w/:wid/s/:streamId?m=:messageId` — the same URL the row's timestamp already used, so no new routing. Because the in-stream timeline never sets `viewInStream`, the action never appears there (you're already in the stream). The label noun is stream-type aware (`viewInStreamLabel`), read from `useStreamFromStore(streamId)`, defaulting to "channel" when the row's stream isn't cached.

**Direction B — stream → conversation ("Show in conversation").** A new `onShowInConversation` callback on `MessageActionContext`, wired in `SentMessageEvent` to `openPanel(createConversationPanelId(conversationId))`. The message's primary conversation is resolved through a new `MessageConversationProvider` / `useMessageConversationId(messageId)` context, built from the same `buildConversationOverlayModel(...).conversationIdByMessageId` membership the overlay uses — so the two never disagree about which conversation owns a message. The action is gated on a known conversation, so it's absent for messages with none.

To make Direction B work *without* the user first toggling the overlay on (`?convOverlay=on`), `stream-content` now fetches the per-stream conversation list always for channels/DMs (`useConversations(..., { enabled: supportsConversationOverlay })`) and builds the map once. This call shares the query key with the overlay's own `useConversations`, so the two dedupe to one request when both are live, and reuses that hook's socket subscription + reconnect invalidation (INV-53).

### Key design decisions

**1. Gate by context field, not by surface flag.** Each action keys off a context field that exactly one side sets (`viewInStream` ← `MessageItem`; `onShowInConversation` ← `SentMessageEvent`). Both directions live in the one `messageActions` array shared by the stream timeline and the board/conversation surfaces, yet each appears only where it makes sense — no `if (surface === ...)` branching, and the mobile drawer inherits both for free.

**2. Always-fetch the conversation list for channels/DMs.** Reading cache-only (no new fetch) would mean the action only appears after the user opened the overlay — almost never — making it dead weight. Resolving lazily on click risks a menu item that no-ops. Always-fetching the small per-stream list on open is the only option that makes the action reliably present; it's deduped with the overlay's fetch and reuses its INV-53 wiring, so the cost is one small request per channel/DM open and no new sync path.

**3. Reuse the overlay membership model, don't re-derive.** `useMessageConversationId` is backed by `buildConversationOverlayModel`, the same builder the overlay uses, so "Show in conversation" and the overlay's per-row coloring/reassignment can never point at different conversations for the same message.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/timeline/conversation-overlay/message-conversation-context.tsx` | `MessageConversationProvider` + `useMessageConversationId` — per-stream primary `messageId → conversationId` membership, available regardless of overlay state. |
| `apps/frontend/src/components/timeline/conversation-overlay/message-conversation-context.test.tsx` | Hook coverage: known message → id, unknown message → null, no provider → null. |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/message-actions.ts` | Add `viewInStream` + `onShowInConversation` to `MessageActionContext`; add `view-in-stream` (getHref nav) and `show-in-conversation` (callback) actions. |
| `apps/frontend/src/components/message/message-item.tsx` | Set `viewInStream` on the menu context (stream-type-aware label via `useStreamFromStore`). |
| `apps/frontend/src/components/timeline/message-event.tsx` | `SentMessageEvent`: resolve the message's conversation via `useMessageConversationId`, wire `onShowInConversation` → `openPanel(createConversationPanelId(...))`. |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Always-on per-stream conversation fetch for channels/DMs; build the membership map; wrap the timeline in `MessageConversationProvider`. (Large diff is mechanical re-indentation from the one added wrapper level — `git diff -w` shows 18 real lines.) |
| `apps/frontend/src/components/timeline/message-actions.test.ts` | Coverage for both new actions (visibility gating, href, callback). |

## Out of scope (deliberate)

- **Direction B is channel/DM-only.** A thread's messages resolve membership through their root, which isn't in the thread's own `listByStream` result, so "Show in conversation" doesn't appear on thread rows — matching where the conversation overlay itself applies. Thread support needs a root-scoped lookup; left for a follow-up.
- **FU2** (full message actions on board/conversation rows + surface-aware copy-link that yields a `?panel=conv:` URL + conversation-level copy-link / mod+Shift+L) is the next build, not this PR.

## Test plan

- [x] `apps/frontend` `vitest run` over `message-actions.test.ts` + `message-conversation-context.test.tsx` + `message-context-menu.test.tsx` — 77 pass
- [x] `apps/frontend` `vitest run` over `timeline/`, `message/`, `board/` + the `no-happy-path-success-toast` guard — 353 pass
- [x] `tsc --noEmit` (frontend) clean; eslint clean on all changed files; prettier formatted
- [x] Self-review (claim-verification pass): every named symbol/file/URL confirmed against the diff; permalink URL matches the existing timestamp link; `createConversationPanelId` confirmed exported from the `@/contexts` barrel
- [ ] No manual browser pass this session — behavior verified via unit/integration tests and the shared-registry wiring (both actions render through the same `MessageContextMenu` / `MessageActionDrawer` paths already exercised by existing tests)

---

---
_Generated by [Claude Code](https://claude.ai/code/session_013F3MgsttAHoufXppr2Lmb7)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1121"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785430416&installation_id=129946197&pr_number=1121&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1121&signature=06510fdeb3133e802bc7b85b0dd54faa8c0c5a535a904a481430d351dc24c601"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/message/message-item.tsx
+++ b/apps/frontend/src/components/message/message-item.tsx
@@ -25,8 +25,21 @@ import { useUserProfile } from "@/components/user-profile"
 import { useFormattedDate } from "@/hooks/use-formatted-date"
 import { useTouchCapable } from "@/hooks/use-touch-capable"
 import { useLongPress } from "@/hooks/use-long-press"
+import { useStreamFromStore } from "@/stores/stream-store"
 import { STREAM_ICONS } from "@/lib/streams"
 import { cn } from "@/lib/utils"
+
+/** Stream-type noun for the "View in …" jump action. Defaults to "channel"
+ * (the common case) when the row's stream isn't cached to read its type. */
+const STREAM_NOUN: Record<string, string> = {
+  channel: "channel",
+  thread: "thread",
+  dm: "DM",
+  scratchpad: "scratchpad",
+}
+function viewInStreamLabel(type: string | undefined): string {
+  return `View in ${(type && STREAM_NOUN[type]) || "channel"}`
+}
 
 /** Same-author messages within this window collapse into a continuation (no
  * repeated header) — matches the timeline's grouping. */
@@ -111,12 +124,14 @@ export function MessageItem({
   const interactiveName = message.authorType === "user" && Boolean(message.authorId)
   const attachments = message.attachments ?? []
   const linkPreviews = message.linkPreviews ?? []
+  // The row's own stream — only to pick the "View in channel/thread/…" noun.
+  const rowStream = useStreamFromStore(streamId)
 
-  // A deliberately small action set for surfaces with no thread context: file
-  // under a label, copy the content, copy a link. `isThreadParent: true`
-  // suppresses "Reply in thread" (its only gate) rather than point it at a
-  // thread we don't have here; the timestamp is the way back into the stream.
-  // No `currentUserId`, so the owner-only edit/delete actions stay hidden.
+  // A deliberately small action set for surfaces with no thread context: jump
+  // to where the message lives, file under a label, copy the content, copy a
+  // link. `isThreadParent: true` suppresses "Reply in thread" (its only gate)
+  // rather than point it at a thread we don't have here. No `currentUserId`,
+  // so the owner-only edit/delete actions stay hidden.
   const menuContext: MessageActionContext = {
     contentMarkdown: message.contentMarkdown,
     actorType: message.authorType,
@@ -126,6 +141,10 @@ export function MessageItem({
     messageId: message.id,
     workspaceId,
     streamId,
+    viewInStream: {
+      href: `/w/${workspaceId}/s/${streamId}?m=${message.id}`,
+      label: viewInStreamLabel(rowStream?.type),
+    },
     onLabelMessage: () => setLabelPickerOpen(true),
   }
 
--- a/apps/frontend/src/components/message/message-item.tsx
+++ b/apps/frontend/src/components/message/message-item.tsx
@@ -26,19 +26,15 @@ import { useFormattedDate } from "@/hooks/use-formatted-date"
 import { useTouchCapable } from "@/hooks/use-touch-capable"
 import { useLongPress } from "@/hooks/use-long-press"
 import { useStreamFromStore } from "@/stores/stream-store"
-import { STREAM_ICONS } from "@/lib/streams"
+import { STREAM_ICONS, streamFallbackLabel } from "@/lib/streams"
 import { cn } from "@/lib/utils"
 
-/** Stream-type noun for the "View in …" jump action. Defaults to "channel"
- * (the common case) when the row's stream isn't cached to read its type. */
-const STREAM_NOUN: Record<string, string> = {
-  channel: "channel",
-  thread: "thread",
-  dm: "DM",
-  scratchpad: "scratchpad",
-}
-function viewInStreamLabel(type: string | undefined): string {
-  return `View in ${(type && STREAM_NOUN[type]) || "channel"}`
+/** Label for the "View in …" jump action. Uses the shared stream-type noun
+ * ("channel"/"thread"/"DM"/"scratchpad") when the row's stream is cached; a
+ * type-neutral phrase otherwise, so an uncached thread row never mislabels as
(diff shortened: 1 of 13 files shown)


kristofferremback:
### Code review

No issues found. Round 1 (4 lenses on `73b6a06`) surfaced two ≥80 findings, both now fixed in `ca4bc43`:

1. **Reactivity** — `useConversations` returned a fresh `[]` default each render while disabled/loading, so the `buildConversationOverlayModel` memo in `stream-content` rebuilt the message→conversation `Map` every render and re-rendered every timeline row (worst on threads/scratchpads, where the feature is inactive). Fixed with a shared module-level `EMPTY_CONVERSATIONS` constant so the disabled/loading query keeps one identity.
2. **Label** — the "View in …" label used a hand-rolled noun map defaulting to "channel" (a thread row mislabeled as "View in channel" before its stream hydrated), against the frontend "one place turns a stream into a label" rule. Fixed by reusing the shared `streamFallbackLabel(type, "noun")` and a type-neutral "Go to message" fallback.

Round 2 (spec+design and correctness re-run on `ca4bc43`) confirmed both fixes hold: stable array identity, no shared-constant mutation aliasing in the overlay/socket paths, and the reassignment path stays reactive end-to-end (`conversation:updated` → cache → memo → `useMessageConversationId` → action). Below-threshold round-1 notes (timestamp/menu permalink duplication, transient action presence during load, first-menu ordering, the type-required inner guard) were not actionable.

Checked for bugs, data-flow lifecycle, CLAUDE.md compliance, plan adherence, design quality, security, UX, and mobile.

---

PR #1130, opened 2026-07-01
feat(board): quote a text selection from board/conversation message rows

kristofferremback:
## Problem

The stream timeline lets you select text in a message and quote just that snippet (`TextSelectionQuote` → the floating "Quote" button → the composer's `quoteReply` node). The shared out-of-stream row (`MessageItem`, rendered by the board card, conversation panel, and label page) shipped the menu "Quote reply" and the mobile swipe-to-quote in #1126 — but both quote the *whole* message. Selecting a phrase on a board/conversation row did nothing. This is the next item on the board/conversation action-parity line (following #1124's copy-link + react and #1126's quote reply).

## Solution

Reuse the stream's `TextSelectionQuote` on the shared row rather than build a second selection mechanism (INV-35). Two things stood in the way and this PR removes them: the shared `MessageItem` didn't expose the DOM hooks the component reads, and the component assumed a single stream and a single instance — neither holds on the board, where each card is its own `QuoteReplyProvider` and a conversation spans its root plus threads.

### How it works

1. **`MessageItem` exposes the DOM contract `TextSelectionQuote` reads.** Both the head and continuation returns now carry `data-message-id`, `data-stream-id`, `data-author-name`, `data-author-id`, `data-actor-type` on the row container, and the content column gets the `message-content` class so the component's `.message-content .markdown-content` selector matches. `data-stream-id` is the row's own `streamId` (already per-message — `message.streamId ?? <card stream>`), not the card's.

2. **`TextSelectionQuote` gains container scoping.** New optional `containerRef` prop: when set, a selection whose message element isn't inside that container is ignored. The stream mounts one instance for the whole timeline and passes no ref (unchanged). The board mounts one instance *per card* — without scoping, every card's instance would react to a selection in any card and route it to its own provider's composer (the last-mounted would win). Scoping makes each card handle only its own rows.

3. **`TextSelectionQuote` reads the row's own stream.** It now takes `data-stream-id` off the selected message element and falls back to the `streamId` prop when absent. A conversation row in a thread quotes against the thread it lives in, not the card's anchor stream. The stream timeline sets no `data-stream-id`, so it keeps using the prop — behavior there is byte-for-byte unchanged.

4. **Mount inside the per-conversation provider.** `board-card.tsx` and `conversation-panel.tsx` (`ConversationPanelBody`) each render `<TextSelectionQuote>` inside their existing `QuoteReplyProvider`, scoped via a ref to the card root / the panel's message list. The quote node itself, the trimming, and the composer auto-open path are all the #1126 machinery — this only adds the selection entry point.

### Key design decisions

**1. Container scoping over a board-global provider.** The per-card provider is deliberate (from #1126 — a board-global provider collides on `handlerRef`, letting the last composer win). Since the provider is per-card, the selection detector must be per-card too. A `containerRef` gate is the minimal way to keep each instance's reach to its own rows; the stream, being single-provider, opts out by omitting the ref.

**2. `data-stream-id` on the row, prop as fallback.** A conversation is one root spanning many streams, so the correct quote target is the message's own stream, which only the row knows. Reading it from the DOM (where `TextSelectionQuote` already pulls author metadata) keeps one code path for both surfaces and leaves the single-stream timeline untouched via the prop fallback — no change to `message-event.tsx`.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/text-selection-quote.tsx` | Optional `containerRef` scoping; read the row's `data-stream-id` with the `streamId` prop as fallback |
| `apps/frontend/src/components/message/message-item.tsx` | Emit `data-message-id`/`data-stream-id`/author data attrs and the `message-content` class on both row shapes |
| `apps/frontend/src/components/board/board-card.tsx` | Mount `TextSelectionQuote` inside the card's `QuoteReplyProvider`, scoped to the card root |
| `apps/frontend/src/components/conversations/conversation-panel.tsx` | Mount `TextSelectionQuote` inside the panel's `QuoteReplyProvider`, scoped to the message list |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/timeline/text-selection-quote.test.tsx` | Behavior test: a mocked selection routes with the row's `data-stream-id` (not the anchor) and its snippet through the scoped provider; a selection outside the container is ignored |

## Out of scope (deliberate)

- **Touch selection.** `TextSelectionQuote` is mouse-only (touch rows are `select-none`; swipe-to-quote already covers touch). Unchanged.
- **Label page.** No `QuoteReplyProvider` there, so no instance is mounted and the gesture stays off — same gating as the menu action.
- Remaining action-parity follow-ups (Discuss-with-Ariadne span, Save/Reminder conversation origin, Reply-in-thread, Edit, Delete, Share) are separate items.

## Test plan

- [x] `vitest run text-selection-quote.test.tsx` — 2 pass (snippet routes with row stream id; out-of-container selection ignored)
- [x] `vitest run conversation-panel / board-composer / quote-reply-context / message` — 16 pass (existing menu + swipe coverage intact)
- [x] `vitest run message-input / message-event` — 41 pass (shared `TextSelectionQuote` change doesn't regress the stream)
- [x] `tsc --noEmit` (frontend) clean
- [x] eslint + prettier clean on the five files
- [x] Full pre-commit hook (monorepo lint + typecheck + migrations + OpenAPI) passed on commit

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rhstbkbdrm9AUWboj4ETj7)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1130"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785533592&installation_id=129946197&pr_number=1130&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1130&signature=9da32b2f68fbceb7cbd594ee8f6716e818ba28dcdde5bc04e9a3bc3147c453bd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/board/board-card.tsx
+++ b/apps/frontend/src/components/board/board-card.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react"
+import { useRef, useState } from "react"
 import { Link } from "react-router-dom"
 import { useQuery } from "@tanstack/react-query"
 import { Hash, FileEdit, User, MessageSquareText, ChevronDown, PanelRight, type LucideIcon } from "lucide-react"
@@ -8,6 +8,7 @@ import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip
 import { MessageItem, isContinuation, type RenderableMessage } from "@/components/message/message-item"
 import { BoardReplyComposer } from "@/components/board/board-reply-composer"
 import { QuoteReplyProvider } from "@/components/timeline/quote-reply-context"
+import { TextSelectionQuote } from "@/components/timeline/text-selection-quote"
 import { useActors } from "@/hooks"
 import { useWorkspaceUserId } from "@/hooks/use-workspaces"
 import { useConversationService, usePanel, createConversationPanelId } from "@/contexts"
@@ -46,6 +47,9 @@ export function BoardCard({ workspaceId, post, contextLabel, streamType }: Board
   const conversationService = useConversationService()
   const { openPanel } = usePanel()
   const [expanded, setExpanded] = useState(false)
+  // Scopes text-selection quoting to this card so a selection routes into this
+  // card's composer, not a sibling card's provider.
+  const cardRef = useRef<HTMLDivElement>(null)
 
   // Bodies ride the same `db.events` rail the timeline does — live and
   // offline-first — with the cached server projection as the cold-start fallback.
@@ -125,7 +129,9 @@ export function BoardCard({ workspaceId, post, contextLabel, streamType }: Board
     // Scope quote reply to this card: a message row's "Quote reply" routes into
     // this card's own reply composer, not another card's.
     <QuoteReplyProvider>
-      <div className="rounded-xl border bg-card p-3 sm:p-4">
+      {/* Desktop text-selection → floating "Quote" button, scoped to this card. */}
+      <TextSelectionQuote streamId={streamId} containerRef={cardRef} />
+      <div ref={cardRef} className="rounded-xl border bg-card p-3 sm:p-4">
         <div className="flex items-center gap-1.5 text-xs text-muted-foreground">
           {/* The stream the post lives in — a real link back into it (INV-40). */}
           <Link
--- a/apps/frontend/src/components/conversations/conversation-panel.tsx
+++ b/apps/frontend/src/components/conversations/conversation-panel.tsx
@@ -18,6 +18,7 @@ import { MessageItem, isContinuation, type RenderableMessage } from "@/component
 import { RelativeTime } from "@/components/relative-time"
 import { BoardReplyComposer } from "@/components/board/board-reply-composer"
 import { QuoteReplyProvider } from "@/components/timeline/quote-reply-context"
+import { TextSelectionQuote } from "@/components/timeline/text-selection-quote"
 import { SidebarToggle } from "@/components/layout"
 import { useActors } from "@/hooks"
 import { useWorkspaceUserId } from "@/hooks/use-workspaces"
@@ -266,10 +267,15 @@ function ConversationPanelBody({ workspaceId, post, hostStreamType }: Conversati
   // opener lives in the parent stream).
   const lastActiveStreamId = displayedReplies.at(-1)?.streamId ?? conversation.streamId
 
+  // Scopes text-selection quoting to this panel's message list.
+  const listRef = useRef<HTMLDivElement>(null)
+
   return (
     // Quote reply from a row routes into this conversation's reply composer.
     <QuoteReplyProvider>
-      <div className="min-h-0 flex-1 overflow-y-auto px-4 py-3 [&>*:first-child]:mt-0">
+      {/* Desktop text-selection → floating "Quote" button, scoped to this list. */}
+      <TextSelectionQuote streamId={conversation.streamId} containerRef={listRef} />
+      <div ref={listRef} className="min-h-0 flex-1 overflow-y-auto px-4 py-3 [&>*:first-child]:mt-0">
         {all.map((message, i) => (
           <MessageItem
             key={message.id}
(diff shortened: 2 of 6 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet x4 (spec+design, correctness+data-flow, UX, mobile) → sonnet x2 re-review after fixes

Ran the full lens set over the diff, then fixed the two findings that scored ≥80 and re-reviewed the fix commit clean. No issues survive on `cf6aa79`.

**Fixed during review (commit `cf6aa79`):**

1. `apps/frontend/src/components/timeline/stream-content.tsx:1893` — the stream timeline mounted an **unscoped** `TextSelectionQuote`. Once `MessageItem` gained `data-message-id` + `.message-content`, that instance also matched the rows a conversation side panel (`PanelHost` → `ConversationPanel`) renders beside the timeline on the stream page — so selecting text in the panel fired a second "Quote" button and routed the quote to the stream's own composer. Fixed by scoping the stream instance to its message-list container (`containerRef={quoteScopeRef}`), the same mechanism the board/panel instances already use. _(bug, 85)_
2. `apps/frontend/src/components/timeline/text-selection-quote.tsx` — the board mounts one instance per card, each on a `document` `selectionchange` listener; every instance did `sel.toString()` + two `closest()` DOM walks on each tick before the container gate discarded all but one (O(N-cards) per tick during a drag). Fixed by moving the `containerRef.current.contains(range.startContainer)` gate ahead of that work so non-owning instances bail after one `contains()` call. _(performance, 80)_

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Reuses the stream's `TextSelectionQuote` (INV-35) rather than a second mechanism; the two new mount sites and the `MessageItem` DOM contract match the intent.</details>
<details><summary>🔍 Bugs [1 fixed]</summary>Unscoped stream-timeline detector matching panel rows — fixed (see #1 above). Gate reorder verified behaviorally equivalent (container is always an ancestor of the message row, so gating on `range.startContainer` vs the message element is identical); ref-before-attach window is a non-issue (React attaches refs during commit, before the listener effect runs, and the test suite exercises this ordering).</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Quote payload (messageId, streamId, authorName/Id, actorType, snippet) round-trips DOM → `triggerQuoteReply` → provider handler; per-message `data-stream-id` correctly points a thread row's quote at its own stream.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-21 satisfied (portal button is `fixed`, out of flow — no layout shift). The `message-content` class matches the compact-mode `gap` CSS rule but is inert on the non-flex block div. New comments explain non-obvious scoping/ordering (INV-25 ok).</details>
<details><summary>🏗️ Design [CLEAN]</summary>Fix reuses the established `containerRef` scoping — no parallel mechanism.</details>
<details><summary>🔒 Security [CLEAN]</summary>No untrusted-HTML paths; React-safe.</details>
<details><summary>🎛️ UX [CLEAN]</summary>Consistent across board card / conversation panel / timeline; per-card scoping prevents overlapping buttons. (Portal viewport-clamping for a selection near the top edge was raised at 50 — below threshold and unchanged from the pre-existing stream behavior; noted as possible future polish.)</details>
<details><summary>📱 Mobile [CLEAN]</summary>Strictly mouse-gated (`useInputMode() === "mouse"`); the existing swipe-to-quote and long-press-drawer touch handlers on `MessageItem` are untouched.</details>

---

PR #1138, opened 2026-07-02
feat(agents): schedule_follow_up tool + follow-up infra (roadmap 1.1)

kristofferremback:
**Roadmap:** [`docs/plans/ariadne-collaborator-roadmap.md`](../blob/main/docs/plans/ariadne-collaborator-roadmap.md) — step **1.1** (Phase 1, the pathfinder durable-write tool).

## Problem

Ariadne has ephemeral in-product writes (`send_message`, `react_to_message`) but no tool that creates durable state — she can't say "I'll check back on this tomorrow" and actually come back. The cheapest team-member behavior is missing, and with it the template for every later durable-write tool (delegations, `save_memo`, brief updates).

The obvious shortcut — reuse `scheduled_messages` — is wrong by construction: a scheduled message is user-authored (`ScheduleParams.userId`) and fires by creating a real `AuthorTypes.USER` message (`finalizeSendInTx` → `EventService.createMessage`). A persona must never author as the user, and a synthetic user message would pollute the timeline. Follow-ups need their own substrate whose firing wakes the *persona*, not the user.

## Solution

A dedicated `agent_follow_ups` tracking table (INV-57) whose firing enqueues a `PERSONA_AGENT` job directly — the queue's `processAfter` is the reusable primitive, not the scheduled-messages table. The `schedule_follow_up` tool ships with its whole substrate (schema, service, firing worker) in one unit; the turn it fires runs in a follow-up (step 1.2), the timeline card in 1.3, workspace-tunable cap in 1.4.

### How it works

1. **Schema** — `agent_follow_ups` (`agfu_` ULID, workspace-scoped, no FKs/enums per INV-1/3). Partial index `(workspace_id, stream_id) WHERE status = 'pending'` backs the cap count. `source_conversation_id` (nullable) anchors the follow-up to the trigger's primary conversation for later board-lens visibility.
2. **Create** (`AgentFollowUpService.schedule`) — one transaction: a per-`(workspace, stream)` `pg_advisory_xact_lock`, then `insertIfUnderCap` (a `WHERE (SELECT count(*) … pending) < limit` guarded insert — the lock makes the count exact, INV-20), then `enqueueQueuedJob(AGENT_FOLLOW_UP_FIRE, processAfter = scheduledFor)` and write the queue id back. Cap resolved through `resolveFollowUpLimit()` (returns `DEFAULT_MAX_PENDING_FOLLOW_UPS = 10` today; the seam 1.4 replaces).
3. **Fire** (`AgentFollowUpService.fire`, driven by the `agent.follow_up_fire` worker) — CAS `pending → fired` (`markFired`), and on success enqueue the `PERSONA_AGENT` job **in the same transaction** (INV-7) so a fired row always has its turn queued. `messageId` is synthetic (`followup_<id>`) — the already-supported catch-up path from `checkForUnseenMessages`. An `onDLQ` hook marks the row `failed` if the fire job exhausts retries.
4. **Cancel** (`AgentFollowUpService.cancel`) — CAS `pending → cancelled`; only when the CAS wins does it tombstone the fire queue row. A fired-first race makes cancel return `null` (reported to the caller); a cancelled-first race makes the fire worker's CAS return `null` and no-op (queue delivery can't be revoked). Both guard on `status = 'pending'`, so exactly one wins.
5. **Tool** (`schedule_follow_up`) — `{ note, scheduledFor }`, future-dated and ≤ 30 days out (`MAX_FOLLOW_UP_HORIZON_DAYS`). The result carries the resolved limit + current pending count so the model self-regulates; `cap_reached` tells it to schedule fewer. It also returns `scheduledForLocal` — the time rendered in the user's timezone — and the promptBlock tells the model to quote that, never raw UTC (see decision 5). Future/horizon validation grounds on the turn's injected `currentTime` (deterministic in evals; warns loudly per INV-11 and falls back to wall-clock if that time is unparseable). Bound to the persona/session/stream in `persona-agent.ts` (identity fixed at bind time, like `react_to_message`); the `source_conversation_id` anchor reuses the canonical `resolveEligibleConversation` (trigger-preferred, window-fallback) so it isn't null in the common case where the trigger isn't classified yet.

### Key design decisions

**1. Follow-ups are not scheduled messages** — separate table + `PERSONA_AGENT` firing, per the design ruling above. A persona firing a user-authored message is the failure mode this avoids.

**2. Privacy category `messaging`, not a workspace-write grant** — scheduling a follow-up is the agent planning its own return, in-product participation like replying, not data egress or a workspace read. Its abuse bound is the per-stream pending cap, not a privacy grant. (`tool-privacy.ts`; the `satisfies Record<AgentToolName, …>` forces categorization.)

**3. Fire carries `followUpId`, not a `trigger` union (deviation from the step's Shape)** — the roadmap specced `trigger: { kind: 'follow_up', followUpId }`, but that changes `PersonaAgentJobData.trigger` from a string to a union and touches every existing `trigger === MENTION` reader. Instead an additive optional `followUpId` field is emitted; the CAS + enqueue go through `enqueueQueuedJob` in one transaction rather than a post-commit `jobQueue.send`. Until 1.2 reads `followUpId`, a fired follow-up runs as a companion-mode catch-up turn. Recorded in the roadmap step.

**4. Cap seam, no per-stream column yet** — `resolveFollowUpLimit()` exists so 1.4's workspace override slots in without touching the cap check; the per-stream column is deferred until a real need (INV-36).

**5. Confirm times in the user's timezone, not UTC** — observed on staging: a follow-up fired at the correct instant, but Ariadne told the user the UTC wall-clock (she was parroting the result's `toISOString()`). The instant was never wrong; only the phrasing was. Fix is presentational — the tool hands the model a pre-rendered local string (`scheduledForLocal`) and instructs it to use that — rather than changing the shared temporal prompt for every turn.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260702134214_agent_follow_ups.sql` | `agent_follow_ups` table + pending partial index |
| `apps/backend/src/features/agents/follow-up-repository.ts` | Row mapping + CAS transitions + count-guarded insert + cap lock |
| `apps/backend/src/features/agents/follow-up-service.ts` | schedule / cancel / fire lifecycle; cap seam |
| `apps/backend/src/features/agents/follow-up-worker.ts` | `agent.follow_up_fire` job handler |
| `apps/backend/src/features/agents/tools/schedule-follow-up-tool.ts` | The tool (Zod input, horizon check, local-time echo, promptBlock) |
| `*.test.ts` (repository / service / tool) | CAS + scoping, cap + fire/cancel races, validation, tz echo |

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/constants.ts` | `schedule_follow_up` tool name; `FollowUpStatuses` |
| `packages/types/src/tool-privacy.ts` (+ test) | Categorize the tool as `messaging` |
| `packages/backend-common/src/id.ts` (+ index) | `agentFollowUp{,Queue}Id` generators |
| `apps/backend/src/lib/queue/job-queue.ts` (+ index) | `AGENT_FOLLOW_UP_FIRE` queue + job data; `followUpId` on `PersonaAgentJobData` |
| `apps/backend/src/features/agents/config.ts` | `DEFAULT_MAX_PENDING_FOLLOW_UPS`, `MAX_FOLLOW_UP_HORIZON_DAYS` |
| `apps/backend/src/features/agents/tools/{tool-deps,index}.ts` | `FollowUpToolDeps` callback + barrel |
| `apps/backend/src/features/agents/companion/tool-set.ts` | Gate + build the tool on the live turn (passes timezone/currentTime) |
| `apps/backend/src/features/agents/persona-agent.ts` | Bind `followUpDeps` with canonical conversation anchor |
| `apps/backend/src/features/agents/{index,built-in-agents}.ts` | Barrel exports; Ariadne `enabledTools` |
| `apps/backend/src/server.ts` | Construct service, wire `scheduleFollowUp`, register worker + DLQ hook |
| `docs/plans/ariadne-collaborator-roadmap.md` | Check step 1.1; record the deviation; backlog enclave parity |

## Out of scope (deliberate)

- **The fired turn's context/prompt** ("why you woke up") — step 1.2. Today a fired follow-up runs a plain companion catch-up turn with no self-awareness: on staging it either declined ("I already have a follow-up scheduled… I'll ping you then") or, given "check in in two minutes", re-read that as a fresh instruction and scheduled another follow-up every cycle (a loop). 1.2 fixes this — the fired-turn prompt will state "this IS your follow-up firing; do the check-in and don't re-schedule unless there's genuinely new future work."
- **Plaintext streams only; no enclave/E2E support.** The tool is wired into the plaintext companion toolset; the enclave's `buildEnclaveTools` deliberately omits it (it needs a backend callback the enclave can't make with zero plaintext egress), so E2E scratchpads silently don't offer it — clean degrade, no leak. Full parity (a sealed `note` decrypted only in the enclave + enclave fire dispatch) is tracked in the roadmap backlog.
- **Timeline card + cancel affordance** — step 1.3 (no outbox event here). The `cancel` service method is present as substrate but has no HTTP/UI surface yet.
- **Workspace-tunable cap** — step 1.4 (the `resolveFollowUpLimit()` seam is the only forward investment).

## Design evolution

Pre-PR self-review (2 reviewer agents) — all fixed:

- **Cap wasn't race-safe.** The count-guarded insert has no row to lock, so two concurrent creates could each read the pre-insert count and both land. The first version's comment claimed this "can't race" because "one session runs per stream" — false: distinct personas can run concurrent sessions in one stream, and the cap counts by `stream_id`. Fixed with a per-`(workspace, stream)` `pg_advisory_xact_lock` around the count+insert; comment corrected.
- **Tool copy over-promised.** The description/promptBlock told the model follow-ups are "visible to the people in this stream and can be cancelled by them" — neither surface exists until 1.3. Softened.
- **`failed` status was unreachable.** Wired an `onDLQ` hook on `agent.follow_up_fire` → `markFailed`.
- **Dead code removed** (INV-38/36): unused `findByIdScoped`, the unused `jobQueue` service dep, and the speculative `idx_agent_follow_ups_queue_ref` index.

`cancel`/`markCancelled` are kept: they are 1.1's specified cancel substrate (the "cancel wins races cleanly" done-when) and are exercised by tests; 1.3 wires the HTTP/UI surface.

Subsequent rounds (after the PR opened):

- **`source_conversation_id` reused the canonical resolver** (`/code-review`, INV-35): the first cut hand-rolled a prefer-only `findPrimaryByMessageId`, dropping the recency fallback so the anchor was null in the common case (trigger not yet classified). Now calls `resolveEligibleConversation`. A follow-up 2-agent review returned 7/7.
- **Timezone confirmation** (staging feedback): the tool returns `scheduledForLocal` and the promptBlock tells the model to quote it, so Ariadne stops repeating the UTC wall-clock. Validation grounds on the injected `currentTime`.
- **INV-11 loud fallback** (CodeRabbit): warn when an injected `currentTime` is unparseable instead of silently degrading to wall-clock.
- **Rebased onto `origin/main`** to resolve a `server.ts` conflict (main's backfill workers now sit alongside the follow-up worker).

## Test plan

- [x] `bun test` follow-up repository/service/tool + tool-privacy — 21 backend (repo 8, service 6, tool 7) + 13 types pass
- [x] `bun run --cwd apps/backend typecheck` (main + `tsconfig.tests.json`) clean; types + backend-common clean
- [x] Pre-commit hook (full monorepo lint + typecheck + prettier + OpenAPI check) passed on every commit
- [x] Pre-PR review + a second `/code-review` pass: **7/7**, all findings fixed (cap race + false comment, over-promising copy, unreachable `failed`, dead code, resolver reuse); CodeRabbit's follow-up nits (test coverage, redundant re-export, INV-11 warn) addressed
- [x] Manual staging exercise (pr-1138): scheduling fires at the correct instant; time now shown in the user's local zone. Two known 1.1-only rough edges (decline / re-schedule loop on fire) are the 1.2 gap, called out in Out of scope.
- [ ] No live-DB harness here, so CAS races are covered via stubbed repo/querier like their siblings (scheduled-messages) rather than integration-tested.

---


kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: Claude Sonnet ×2 (spec+design, correctness+data-flow)

No issues found. Checked for bugs, data-flow lifecycle, CLAUDE.md compliance, plan adherence, design quality, and security on this backend-only diff.

One issue surfaced in an earlier round and was fixed in `7c426cf`: the `source_conversation_id` anchor hand-rolled a prefer-only lookup instead of reusing the canonical `resolveEligibleConversation` (INV-35), which dropped the recency fallback and left the anchor null in the common case (trigger not yet classified). It now reuses the resolver with the context window. Re-review confirmed the fix and found nothing further.

---

PR #1140, opened 2026-07-02
feat: in-stream conversation provenance chip for topic revivals

kristofferremback:
**Design of record:** `docs/board-view-design.md` §"Conversations as soft threads", bullet **A. On-message provenance indicator (read)**, and Open-decision #1 (indicator trigger).

## Problem

A conversation is a *projected* grouping — a filter over a stream's flat messages — so a topic's messages are scattered through the timeline rather than contiguous like a thread. When someone replies to an old topic after unrelated chatter ("we talked lunch, then a bug, the demo, and now someone out of a meeting replies *Pizza*"), the reply lands at the tail (sequence-appended, INV-61) detached from the lunch exchange three hours up. In the flat timeline that reads as a non-sequitur. The board and the conversation panel are already coherent by construction; the flat timeline is the one broken read, and the fix is to make it self-explaining at that one spot — not to move anything (Kris's standing rule: never mutate/reorder a stream to fix alignment).

## Solution

A quiet on-message provenance chip. When a message *revives* a scattered conversation, it renders `⧉ continues <topic> · 3h ago` above the message body, linking to that conversation's side panel (`?panel=conv:<id>`, Mechanism B, already shipped). The reply stays exactly where it is — this is a pure read annotation.

### How it works

1. **Revival detection** — `annotateConversationRevivals` (`event-list.tsx`) walks the timeline and stamps a `revival` annotation on any message that *block-starts* a conversation (its conversation differs from the previous message row) whose conversation has already appeared earlier in the timeline. Because a block start means the previous row is a different conversation, an earlier member is necessarily separated from this one — that separation is the non-sequitur the chip explains. A conversation's first appearance is a fresh topic, not a revival, so it carries no chip; a contiguous run carries no chip. `previousActivityAt` anchors the "· 3h ago" to the prior member row's time (the topic's last activity), not the reviving message.

2. **Always-on membership** — detection reads `conversationIdByMessageId` (`buildMessageConversationMap`), the always-on `messageId → conversationId` map that already backs the per-message "Show in conversation" action. No dependence on the conversation overlay being painted (`convOverlay=on`) — the handover's stated obstacle ("membership annotated only when the overlay is on") was already solved by that provider, so the chip rides it.

3. **Flat-timeline only** — `annotateConversationRevivals` runs in the `timelineItems` memo (`stream-content.tsx`) gated on `supportsConversationOverlay` (`!isDraft && (channel || dm)`), so it never runs for threads or drafts. Thread replies are contiguous by construction; nothing reads as a revival there.

4. **Render** — `revival` threads through the existing `TimelineItem → EventItem → MessageEvent → SentMessageEvent` path as a `preface` slot rendered above the header row, and is compared in `timelineItemEqual` so only the reviving row re-renders on a new message. The chip is a `<Link to={getPanelUrl(createConversationPanelId(id))}>` (URL-driven, INV-40/INV-59).

### Key design decisions

**1. Trigger only on revivals (Open-decision #1, on the recorded lean).** A loud "continues X" fires only when the prior member is old/non-adjacent; an ordinary sequential topic switch or a fresh conversation gets nothing. In-window detection is conservative by construction — a block start with an earlier member can never be a false positive, and the worst case is a missing chip when the prior member is scrolled out of the loaded window (a false negative, not a wrong one).

**2. Annotation on `TimelineItem`, not a React context.** The always-on membership for "Show in conversation" uses a context (`MessageConversationProvider`), but that map is conversation-event-driven (rarely changes). Revival status is order-driven — it changes on every new message. A revival *context* would change identity every message and re-render **every** message row (context consumers bypass the row memo), regressing this heavily-memoized timeline. Stamping `revival` onto the item and folding it into `timelineItemEqual` keeps the per-row memo intact: appending a message doesn't change the revival status of rows above it, so only the new row re-renders.

**3. `Layers` glyph, neutral/muted, not gold (frontend-design pass).** The chip uses `Layers` — the conversation identity across the overlay (`conversation-overlay.tsx`) — rather than `CornerDownRight`, which `moved-from-indicator.tsx` already reserves as the "moved here" identity; reusing it would blur two meanings. It stays neutral/muted rather than the thread affordance's `primary` gold: a conversation is a quiet "soft thread", not Ariadne's structural gold line. Topic label + `·` + terse `RelativeTime` mirror the overlay's `topicSummary || fallback` rendering. Static footprint (INV-21) — hover only shifts color/underline.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/event-list.tsx` | Add `annotateConversationRevivals` (pure, exported); add `revival?` to the `event` `TimelineItem`, its `timelineItemEqual` comparison, and pass-through to `EventItem` |
| `apps/frontend/src/components/timeline/conversation-overlay/model.ts` | Add the `ConversationRevival` type (conversationId, topicSummary, previousActivityAt) |
| `apps/frontend/src/components/timeline/stream-content.tsx` | Build `conversationsById`; run `annotateConversationRevivals` in the `timelineItems` memo, gated on `supportsConversationOverlay` |
| `apps/frontend/src/components/timeline/event-item.tsx` | Accept + forward `revival` to `MessageEvent` |
| `apps/frontend/src/components/timeline/message-event.tsx` | `ConversationProvenanceChip` component + `preface` slot on `MessageLayout`, above the header row; thread `revival` through `MessageEvent` → `SentMessageEvent` |
| `apps/frontend/src/components/timeline/conversation-overlay/model.test.ts` | 3 unit tests for `annotateConversationRevivals` (revival anchored to prior time; no chip on contiguous run / first appearance; unassigned + non-message items don't break the run) |
| `apps/frontend/src/components/timeline/message-event.test.tsx` | 2 render tests (chip links to `conv:<id>`; absent on non-revivals) |

## Out of scope (deliberate)

- **The "subtle tick" on ordinary sequential switches** (doc: "at most a subtle tick") — the doc frames it as optional/minimal, so v1 ships only the revival chip (INV-36).
- **Mechanism C** (`conversationReference` PM node — author-side "attach context") and the overlay demotion / lenses+scope tracks — separate follow-ups per the design doc.
- **Cross-window revival** — when the prior member of a revived topic is not in the loaded page, no chip renders (see decision #1). Detecting it from the conversation's own `messageIds` rather than in-window order is a deliberate non-goal here.

## Test plan

- [x] `apps/frontend` `vitest run` on the touched suites — `model.test.ts`, `event-list.test.ts`, `message-event.test.tsx`, `event-item.test.tsx`: **98 pass**
- [x] `tsc --noEmit -p apps/frontend/tsconfig.json` clean
- [x] Full-monorepo `tsc` via the pre-commit hook on both commits — clean
- [ ] Not driven in a live browser — the render path is covered by the two component tests (real `MessageEvent` mount asserting the chip's `href`), not a full-app screenshot

---

https://claude.ai/code/session_014tNf68pV5HB8B4UVNyEZaQ

---
_Generated by [Claude Code](https://claude.ai/code/session_014tNf68pV5HB8B4UVNyEZaQ)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1140"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785602566&installation_id=129946197&pr_number=1140&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1140&signature=6b4ec19024c352c03a0dea36a14bfc2a964398a5725e438bd0c0110e37fcccf1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/conversation-overlay/model.test.ts
+++ b/apps/frontend/src/components/timeline/conversation-overlay/model.test.ts
@@ -260,6 +260,19 @@ describe("annotateConversationRevivals", () => {
     expect(revivals(items, membership)).toEqual([null, null, null])
   })
 
+  it("does not manufacture a revival when only an unassigned aside separates two members", () => {
+    // a1 → x1(unassigned) → a2, all with no *other* conversation between. A lone
+    // unclustered message is not a topic switch, so a2 must not get a chip.
+    const items = [
+      msg("a1", "2026-06-01T00:00:00.000Z"),
+      msg("x1", "2026-06-01T00:01:00.000Z"),
+      msg("a2", "2026-06-01T00:02:00.000Z"),
+    ]
+    const membership = { a1: "conv_a", a2: "conv_a" }
+
+    expect(revivals(items, membership)).toEqual([null, null, null])
+  })
+
   it("ignores unassigned rows and does not break the run across non-message items", () => {
     const gap: TimelineItem = { type: "gap", afterEventId: "evt_a1", missingCount: 1 }
     const items = [
--- a/apps/frontend/src/components/timeline/conversation-overlay/model.test.ts
+++ b/apps/frontend/src/components/timeline/conversation-overlay/model.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from "vitest"
 import type { ConversationWithStaleness, StreamEvent } from "@threa/types"
-import { annotateConversationRows, type TimelineItem } from "../event-list"
+import { annotateConversationRows, annotateConversationRevivals, type TimelineItem } from "../event-list"
 import {
   buildConversationOverlayModel,
   buildMessageConversationMap,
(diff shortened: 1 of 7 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet (spec+design, correctness+data-flow, UX, mobile) across 5 rounds

No issues remain. The revival-detection logic and the chip's placement were both driven to a clean state over five review rounds; the final pass surfaced nothing.

**What the rounds found and fixed** (each verified against the code, fixed, and re-reviewed):

1. **False revival on an unassigned aside** (correctness) — an unassigned message between two members of the same conversation reset the run tracker and produced a spurious chip. Fixed: only a *different real conversation* breaks the run; added a regression test.
2. **Chip placement** — iterated to resolve competing constraints:
   - a preface line above the body shifted content on async membership load (INV-21);
   - an animated grow-in (to soften the shift) jittered virtua's scroll on mobile and read as "loading in late" — worse;
   - a header-inline placement fixed the vertical shift but clipped on narrow phones (non-wrapping header inside an `overflow-x-hidden` container) and dropped the chip on continuation rows.
   - **Final:** the chip renders in the message **footer**, on its own full-width line, mirroring `ThreadSlot` (which the design doc names as the affordance to mirror). Truncates cleanly on mobile, shows on both head and continuation rows, no animation — appears on load like `LabelStack`/link previews already do.
3. **Touch target** — the footer nav link now carries `min-h-[26px]`, matching the reaction-pill floor on the same surface.
4. **Comment hygiene** — trimmed change-narration from the chip docstring (INV-25).

Verified clean in the final pass: revival detection (block-start / first-appearance / unassigned-aside / non-message-item edges), `timelineItemEqual` memo comparison of `revival`, always-on membership gating (channels/DMs, independent of the overlay), `topicSummary` rendered as escaped text (no injection), INV-18 module-level components, no dead code.

---
<details><summary>🔁 Correctness &amp; data flow — CLEAN</summary>`annotateConversationRevivals` traced against block-start/first-appearance/unassigned/interleaving cases; `previousActivityAt` tracks the most-recent prior occurrence; memo invalidates correctly on revival appear/disappear.</details>
<details><summary>📋 CLAUDE.md — CLEAN</summary>INV-18 (module-level chip), INV-21 (footer appears like existing async adornments; no hover shift), INV-25 (docstring narration trimmed), INV-40/59 (URL-driven `&lt;Link&gt;`), INV-36 (revival-only, no speculative subtle-tick), INV-60 (topicSummary rendered like the overlay).</details>
<details><summary>🏗️ Design — CLEAN</summary>Annotation-on-`TimelineItem` (not a context) keeps the per-row memo intact; footer placement mirrors `ThreadSlot` per the design doc.</details>
<details><summary>🔒 Security — CLEAN</summary>`topicSummary` is JSX text / aria-label only (auto-escaped); href built from ids, not summary text.</details>
<details><summary>🎛️ UX — CLEAN</summary>Reads as a footer provenance affordance; "· time" is the topic's activity, same shape as `ThreadSlot`'s "N replies · time"; keyboard/aria fine; batch-mode inert.</details>
<details><summary>📱 Mobile — CLEAN</summary>Own full-width line truncates a long topic; icon/time `shrink-0`; ≥26px tap target; no gesture conflict with row long-press/swipe.</details>

---

PR #1142, opened 2026-07-02
feat(agents): follow-up turn invocation (roadmap 1.2)

kristofferremback:
Roadmap: [`docs/plans/ariadne-collaborator-roadmap.md`](docs/plans/ariadne-collaborator-roadmap.md) step **1.2** (follows #1138 which shipped step 1.1).

## Problem

Step 1.1 shipped the `schedule_follow_up` tool and its firing substrate: a fired follow-up CASes `pending → fired` and enqueues a `PERSONA_AGENT` job carrying an additive `followUpId` (`follow-up-service.ts` `enqueuePersonaTurn`). But nothing **read** `followUpId`. The job ran as a plain companion-mode catch-up turn on a synthetic `messageId` (`followup_<id>`) with no idea *why* it woke up. Two bugs followed, both observed live on staging:

1. **The fired turn declines the check-in.** With no context that it *is* the reminder, Ariadne reads the stream, sees she has a pending follow-up, and says "I already have a follow-up scheduled… I'll ping you then" — not knowing this turn is that ping.
2. **Infinite re-scheduling.** Given "check in in two minutes", the fired turn re-reads the note as a fresh instruction and calls `schedule_follow_up` again — every cycle, forever, until the user says stop. The tool is bound on the fired turn (the `followUpDeps` closure only needs the `scheduleFollowUp` dep, which is present), so nothing stopped it.

Both stem from the same root: the synthetic-messageId catch-up turn has no follow-up context.

## Solution

Read `followUpId` on the turn, load the follow-up row, and inject a system-prompt section that tells the persona the turn **is** the scheduled check-in firing now — act on it, don't re-schedule the same note, and stay silent if there's nothing to add.

### How it works

1. **Threading** — `followUpId` flows job → `createPersonaAgentWorker` (`persona-agent-worker.ts`) → `PersonaAgentInput` → `PersonaAgent.run`. Additive field, no `trigger` enum (see decisions).
2. **Row load** — `run()` loads the follow-up through a new `loadFollowUp` dep seam (`AgentFollowUpService.getById` → `AgentFollowUpRepository.findById`, a workspace-scoped single-query read, INV-30), bound in `server.ts` next to `scheduleFollowUp`. If the row is gone or the seam is unbound (test harnesses), it logs and degrades to the old plain catch-up rather than failing.
3. **Prompt section** — the loaded `{ note, scheduledFor }` threads through `buildAgentContext` (`ContextParams.followUp`) into `buildSystemPrompt`, which emits `## Scheduled follow-up firing now`: it quotes the note and the scheduled time rendered in the user's timezone (reusing `formatCurrentTime` from `lib/temporal`, INV-42), states "This IS that reminder firing" (kills bug 1), forbids re-scheduling the same note while allowing a follow-up for genuinely new future work (kills bug 2), and points at `keep_response` for silent completion.
4. **Silent completion** — follow-up turns set `allowNoMessageOutput` on the `TurnRequest` (previously supersede-only). That exposes the runtime's `keep_response` tool and makes a no-message turn return gracefully with `noMessageReason` instead of the runtime auto-committing the model's last plain-text draft as filler or throwing "completed without sending a message".

### Key design decisions

**1. `followUpId` presence, not a `follow_up` trigger enum** — the roadmap's original Shape called for a `follow_up` trigger variant. Shipped instead as the additive `followUpId`, consistent with 1.1's own recorded deviation: a `trigger` union would touch every `trigger === MENTION` reader across the agent. Presence of `followUpId` (and a successfully loaded row) is the variant signal.

**2. No migration** — the Shape flagged a possible nullable `trigger_message_id` migration. Not needed: the synthetic `followup_<id>` is a non-null string that already serves as `trigger_message_id`, so `withCompanionSession`'s `findByTriggerMessage` dedup (idempotency keyed on the synthetic id) works unchanged.

**3. Silent completion reuses the supersede machinery** — rather than a new "keep quiet" path, follow-up turns flip the same `allowNoMessageOutput` flag the supersede rerun uses, so `keep_response` and the graceful no-message return are the exact same runtime code path (`agent-runtime.ts`). The prompt does the steering; no runtime change.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/agents/persona-agent.ts` | Load the follow-up row via `loadFollowUp` when `followUpId` is set; thread `followUp` into `buildAgentContext`; set `allowNoMessageOutput` for follow-up turns; add `followUpId` to `PersonaAgentInput` and `loadFollowUp` to `PersonaAgentDeps`. |
| `apps/backend/src/features/agents/persona-agent-worker.ts` | Extract `followUpId` from job data and pass to `agent.run`. |
| `apps/backend/src/features/agents/companion/context.ts` | Add `ContextParams.followUp`; pass it through `composeSystemPrompt` → `buildSystemPrompt`. |
| `apps/backend/src/features/agents/companion/prompt/system-prompt.ts` | New `## Scheduled follow-up firing now` section; import `formatCurrentTime` for user-tz rendering. |
| `apps/backend/src/features/agents/follow-up-service.ts` | `getById` read seam. |
| `apps/backend/src/features/agents/follow-up-repository.ts` | `findById` (workspace-scoped SELECT). |
| `apps/backend/src/server.ts` | Bind `loadFollowUp: agentFollowUpService.getById`. |
| `apps/backend/src/lib/queue/job-queue.ts` | Update the `followUpId` doc comment (now read, not just emitted). |
| `docs/plans/ariadne-collaborator-roadmap.md` | Check 1.2; record the three shipped deviations. |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/features/agents/persona-agent-worker.test.ts` | Worker threads `followUpId` into `agent.run`; leaves it undefined for a normal companion job. |

## Out of scope (deliberate)

- **Follow-up visibility (timeline card + cancel affordance)** is roadmap **1.3** — this PR is the invocation/context turn only.
- **Enclave/E2E parity** — the fired turn runs the plaintext companion path; the enclave still omits the tool (backlogged in the roadmap; needs a sealed note + enclave fire dispatch).
- The `## Scheduled follow-up firing now` continuation prompt in the runtime (`prepareConversationForModel`) still uses supersede-flavored wording when `allowNoMessageOutput` is set; it's only reached when the last history message is assistant/system and is harmless here — left untouched rather than branching the runtime for a prompt nicety.

## Test plan

- [x] `bun test apps/backend/src/features/agents/` — 338 pass (includes the new/updated cases below)
- [x] `buildSystemPrompt` follow-up section — present with note + user-tz time + both bug-guards (`This IS that reminder firing`, `Do NOT schedule another follow-up`) + `keep_response`; absent for a normal turn
- [x] `AgentFollowUpService.getById` — reads workspace-scoped via `findById`; null when gone
- [x] `AgentFollowUpRepository.findById` — SELECT scoped to `(id, workspace_id)`, maps the row; null on no match
- [x] `createPersonaAgentWorker` — forwards `followUpId`; undefined for a normal job
- [x] `bun run --cwd apps/backend typecheck` — clean (app + tests tsconfig)
- [x] Pre-commit hook (full monorepo lint + typecheck + prettier + OpenAPI check) — green
- [ ] Live fired-job → turn is exercised on staging, not in an automated integration test — the two bugs live in the DB+queue+AI stack this repo has no live-DB harness for; the seams (prompt content, flag, worker threading) are unit-tested individually instead.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01J2SHDkYnQWRXFb45PCNur8)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1142"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785620860&installation_id=129946197&pr_number=1142&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1142&signature=30b7ed28dcb55fa673d823b6a35a97e06fe58d55936e77217820750514f268d8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/agents/companion/context.ts
+++ b/apps/backend/src/features/agents/companion/context.ts
@@ -46,6 +46,13 @@ export interface ContextParams {
   policy: ContextWindowPolicy
   /** Invocation time override for deterministic evals/tests. Production uses Date.now(). */
   currentTime?: Date
+  /**
+   * Present when this turn is a fired follow-up (roadmap 1.2): the note the
+   * follow-up carried and when it was scheduled. Drives the "Scheduled follow-up
+   * firing now" prompt section so the turn knows it IS the check-in, not a fresh
+   * request to schedule one.
+   */
+  followUp?: { note: string; scheduledFor: Date }
 }
 
 export interface AgentContext {
@@ -102,7 +109,7 @@ async function resolveScratchpadCustomPrompt(
  */
 export async function buildAgentContext(deps: ContextDeps, params: ContextParams): Promise<AgentContext> {
   const { db, userPreferencesService, conversationSummaryService } = deps
-  const { workspaceId, streamId, stream, messageId, persona, trigger, policy, currentTime } = params
+  const { workspaceId, streamId, stream, messageId, persona, trigger, policy, currentTime, followUp } = params
 
   const triggerMessage = await MessageRepository.findById(db, messageId)
   const invokingUserId = triggerMessage?.authorType === AuthorTypes.USER ? triggerMessage.authorId : undefined
@@ -374,7 +381,8 @@ export async function buildAgentContext(deps: ContextDeps, params: ContextParams
       rollingConversationSummary,
       tools,
       conversationTopic,
-      spawnedFromContext
+      spawnedFromContext,
+      followUp
     )
     if (turnDigestBlock) {
       systemPrompt += `\n\n${turnDigestBlock}`
--- a/apps/backend/src/features/agents/companion/prompt/system-prompt.test.ts
+++ b/apps/backend/src/features/agents/companion/prompt/system-prompt.test.ts
@@ -145,6 +145,58 @@ describe("buildSystemPrompt", () => {
     expect(blank).not.toContain("## Discussion This Thread Was Spawned From")
   })
 
+  test("injects the scheduled-follow-up section when the turn is a fired follow-up", () => {
+    const prompt = buildSystemPrompt(
+      persona,
+      {
+        ...scratchpadContext,
+        temporal: {
+          currentTime: "2026-07-03T09:00:00.000Z",
+          timezone: "UTC",
+          utcOffset: "UTC+0",
+          dateFormat: "YYYY-MM-DD",
+          timeFormat: "24h",
+        },
+      },
+      null,
+      undefined,
+      undefined,
+      null,
+      [],
+      null,
+      null,
+      { note: "check whether the deploy went out", scheduledFor: new Date("2026-07-03T09:00:00.000Z") }
+    )
+
+    expect(prompt).toContain("## Scheduled follow-up firing now")
+    // Carries the note and the scheduled time rendered in the user's timezone.
+    expect(prompt).toContain("check whether the deploy went out")
+    expect(prompt).toContain("2026-07-03 09:00")
+    // The two staging-bug guards: act now (don't decline) and don't re-schedule.
+    expect(prompt).toContain("This IS that reminder firing")
+    expect(prompt).toContain("Do NOT schedule another follow-up")
+    expect(prompt).toContain("keep_response")
+  })
+
+  test("omits the scheduled-follow-up section for a normal turn", () => {
+    const provided = buildSystemPrompt(persona, scratchpadContext, null, undefined, undefined, null, [], null, null)
+    const explicitNull = buildSystemPrompt(
+      persona,
+      scratchpadContext,
+      null,
+      undefined,
+      undefined,
+      null,
+      [],
+      null,
+      null,
+      null
+    )
+
+    expect(provided).not.toContain("## Scheduled follow-up firing now")
+    expect(explicitNull).not.toContain("## Scheduled follow-up firing now")
+  })
+
   test("web search recency guidance references Current Time when the tool is temporally grounded", () => {
     const prompt = buildSystemPrompt(
       persona,
--- a/apps/backend/src/features/agents/companion/prompt/system-prompt.ts
+++ b/apps/backend/src/features/agents/companion/prompt/system-prompt.ts
@@ -1,6 +1,6 @@
 import { AgentTriggers, StreamTypes } from "@threa/types"
 import { buildToolPromptSections, formatConversationMemoryForPrompt, type AgentTool } from "@threa/agent-runtime"
-import { buildTemporalPromptSection } from "../../../../lib/temporal"
+import { buildTemporalPromptSection, formatCurrentTime } from "../../../../lib/temporal"
 import type { Persona } from "../../persona-repository"
 import type { StreamContext } from "../../context-builder"
 import { WORKSPACE_RESEARCH_TOOL_NAME } from "../../tools"
@@ -25,7 +25,8 @@ export function buildSystemPrompt(
   rollingConversationSummary?: string | null,
   tools: AgentTool[] = [],
   conversationTopic?: string | null,
-  spawnedFromContext?: string | null
+  spawnedFromContext?: string | null,
+  followUp?: { note: string; scheduledFor: Date } | null
 ): string {
   if (!persona.systemPrompt) {
     throw new Error(`Persona "${persona.name}" (${persona.id}) has no system prompt configured`)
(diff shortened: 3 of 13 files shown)


kristofferremback:
### Code review

No issues found. Checked for bugs, data-flow lifecycle, CLAUDE.md compliance, plan adherence, design quality, and security on this backend-only diff (UX and Mobile lenses skipped — no frontend changes).

Verified in particular:
- `AgentFollowUpRepository.findById` is workspace-scoped (`WHERE id = … AND workspace_id = …`) — INV-8; `getById` reads via `pool` directly — INV-30; `loadFollowUp` runs with no DB connection held across it — INV-41.
- The missing-row path logs a warning and degrades to a plain catch-up rather than failing silently — not an INV-11 violation (and unreachable in practice: follow-up rows are status-transitioned, never deleted, and the seam is always bound in `server.ts`).
- `isSupersedeRerun` and `isFollowUp` are mutually exclusive by construction (fired-follow-up jobs never set `supersedesSessionId`), so a follow-up turn cannot enter the supersede reconciliation branch; `allowNoMessageOutput` is a plain pass-through to the runtime.
- `scheduledFor` arrives as a `Date` end-to-end; `formatCurrentTime` is non-throwing `Intl` formatting with a correct ISO fallback when `context.temporal` is absent.
- `formatCurrentTime` reuse is the shared prompt-rendering helper, not a parallel implementation of the tool's own en-US confirmation formatter — INV-35 fine.
- `followup_${id}` messageId and `followUpId` derive from the same id, so they can't diverge; `followUpId` threads end-to-end with no dropped field.

---

kristofferremback:
Assessment of the "missing abstraction" concern (Kris's question; roadmap changes in #1145):

**This PR is right and should land as-is** — it fixes two live staging bugs, the deviations are recorded, and folding a refactor in would break the one-logical-unit rule. The fresh-session-per-firing design is also right: sessions here are per-trigger execution records (three-phase, minutes-bounded, context rebuilt per turn) and the stream is the durable conversation — *resuming* the scheduling session would fight the supersede machinery, the one-running-session-per-stream index, and the proposed INV-64.

**But the concern is well-founded — the hint-injection mechanism doesn't scale, and this PR is where it becomes visible:**

1. `PersonaAgentInput` now carries **four orthogonal optional variant signals** — `trigger?: MENTION`, `supersedesSessionId?`, `rerunContext?` (`persona-agent.ts:138-149`), and this PR's `followUpId?` — each with scattered if-branches, a bespoke prompt section, and repurposed flags. This PR flips `allowNoMessageOutput` (previously supersede-only) and its own out-of-scope section notes the runtime continuation prompt now shows *supersede-flavored wording on follow-up turns*. Design decision #1 deferred the trigger union because it "would touch every `trigger === MENTION` reader" — that's the debt announcing itself, and Phase 8's ambient turns are invocation kind #5.
   → **Roadmap step 1.5 (new, #1145)**: discriminated `purpose` union resolved in one place into per-purpose prompt sections, *derived* runtime flags (fixes the wording leak mechanically), and context-assembly hints. Hard prerequisite for Phase 8.

2. **The fired turn has purpose but no episodic memory.** `agent_follow_ups.session_id` (NOT NULL since #1138) records the creating session, but the turn injects only `{note, scheduledFor}` — the "this IS that reminder" hint is compensating for the absence of the creating session's episode summary. That artifact is exactly roadmap 3.1.
   → **3.1 extended + pulled forward (#1145)**: fired turns load the creating session's episode summary via `session_id` — hint carries purpose, summary carries substance.

3. One watch-item on this PR's bug-2 fix: it's **prompt-only**, and prompt guards regress silently on model changes. The service layer already has the pending cap; a cheap mechanical backstop (fired turn may not re-schedule the same note, or same-note dedup within a window) would make the infinite-loop fix hold regardless of model. Fine as a follow-up — flagging so it isn't mistaken for a durable guarantee.

---

PR #1141, opened 2026-07-02
feat(saved): remember the conversation a message was saved from

kristofferremback:
## Problem

A saved message always deep-linked to its home-stream permalink — `saved-item.tsx` hardcoded `href = /w/${workspaceId}/s/${saved.streamId}?m=${messageId}`. That's wrong for a message saved from a **conversation** (the board card / conversation panel): a conversation spans one root stream plus its threads (INV-62), so landing the user on a single stream drops the conversation framing they saved from. The same stream-only assumption was baked into every reminder surface — the Activity-feed row (`activity-item.tsx`) and the web-push service-worker click (`sw.ts`).

This continues the board/conversation action-parity line (#1124 copy-link+react → … → #1137 discuss-with-Ariadne): the conversation surfaces already carry `conversationId`, but there was no way to Save/remind from them and nothing persisted the origin.

## Solution

Persist the conversation a message was saved from, and route every surface that resurfaces it through a shared conversation-panel link builder. A save from a conversation surface records `conversation_id`; the Saved list, the Activity-feed reminder, and the web-push reminder then deep-link into the conversation panel (`?panel=conv:<id>&m=<msgId>`) instead of the bare stream permalink. Stream-origin saves and standalone to-dos are untouched — they keep the stream link.

### How it works

1. **Persist (backend).** Append-only migration `20260702120000_saved_messages_conversation_origin.sql` adds a nullable `conversation_id` to `saved_messages` (no FK per INV-1; not indexed — it's read only as part of a row already fetched by id/list). `SavedMessagesRepository.upsert` writes it and, on the conflict path, sets `conversation_id = EXCLUDED.conversation_id` so a re-save adopts the current surface's origin (latest-write-wins, matching the existing `remind_at = EXCLUDED.remind_at`). `insertStandalone` leaves it null.

2. **Validate the client hint (INV-62).** `SavedMessagesService.save` treats the incoming `conversationId` as a distrusted display hint (same stance as #1137). `resolveConversationOrigin` loads the conversation workspace-scoped (`ConversationRepository.findByIds`) and keeps it **only** when `conversation.streamId === accessStreamId` — the message's access root (`stream.rootStreamId ?? stream.id`). A missing or mismatched hint drops to null; it never throws, so a stale hint degrades to the stream link rather than blocking the save. The handler's Zod schema accepts `conversationId` only alongside `messageId`.

[...]

PR #1137, opened 2026-07-02
feat(conversations): Discuss with Ariadne from a conversation seeds its full span

kristofferremback:
**Design refs:** `docs/board-view-design.md` §"Conversations span streams within one root", INV-62 (access is a single root check). Continues the board/conversation action-parity line (#1124 → #1126 → #1130 → #1134 → #1136).

## Problem

"Discuss with Ariadne" seeds a private scratchpad whose AI context is the thing you started from. Started from a stream, that's a windowed slice of the stream (`ThreadResolver`, ~50 messages around the anchor). But a **conversation** — the AI-clustered topic that spans a root stream and its threads (one root, `board-view-design.md`) — is a *subset* of its root channel, not a whole stream. Wiring the board card / conversation panel through the existing `thread` ref would have pulled ~50 recent *channel* messages (unrelated chatter) and silently dropped the conversation's cross-stream thread replies. There was no ref kind that resolves "this conversation's own messages", so the action was simply absent on the conversation surfaces — the next gap in the action-parity line.

## Solution

Add a `conversation` context-ref kind that the backend resolves from the conversation's `message_ids`, so Ariadne sees the topic across root + threads and nothing else. The frontend action on the shared `MessageItem` row targets it; the backend precompute/resolve path already in place (`stream:created` outbox → `context_summaries`) warms it like any other ref.

### How it works

1. **Type spine** (`packages/types/src/context-bag.ts`) — `ContextRefKinds.CONVERSATION`; `ContextRef` becomes a discriminated union of `ThreadContextRef | ConversationContextRef`. The conversation branch carries `conversationId` + `streamId` (the root, for access display / source enrichment) + optional `originMessageId` (focal). The wire mirrors (`StreamContextRef`, `EnrichedContextRef`) gain `kind: ContextRefKind` + `conversationId`.
2. **`ConversationResolver`** (new) — loads the conversation, fetches its member messages via `MessageRepository.findByIdsInWorkspace`, drops soft-deleted rows, orders **by wall-clock** (per-stream `sequence` isn't comparable across streams), windows to `CONVERSATION_WINDOW_TOTAL = 50` centered on the focal, and marks the focal for the renderer's before/focal/after split. `canonicalKey` is `conversation:<id>` so the summary cache is conversation-scoped.
[...]

PR #1134, opened 2026-07-02
feat(board): edit your own message from board/conversation rows

kristofferremback:
## Problem

"Edit message" was reachable only from the in-stream timeline (`MessageEvent`). The shared out-of-stream row `MessageItem` — used by the board card, the conversation panel, and the label page — carried `currentUserId` for reaction toggling but supplied no `onEdit` handler, so the `edit-message` entry in the shared action registry (`message-actions.ts`) stayed hidden everywhere `MessageItem` renders. A user could react to and quote their own message from a conversation but not edit it without first jumping into its home stream.

This continues the board/conversation action-parity line (#1124 copy-link + react, #1126 quote reply, #1130 quote-with-snippet).

## Solution

Host the timeline's existing `MessageEditForm` on the shared row rather than build a second editor (INV-35). The form is already self-contained — it owns its save/offline path (`messageService.update` → `enqueueOperation` fallback), its query invalidation, and both surfaces (desktop inline editor, touch bottom-sheet drawer). `MessageItem` only had to wire the edit state and feed the form its inputs.

### How it works

1. **Gate unchanged.** `MessageItem` now sets `onEdit: startEditing` in its `MessageActionContext`. The existing `edit-message` `when` gate (`actorType === "user" && authorId === currentUserId && e2eEnabled !== true && !!onEdit`) does the rest, so the action shows only on the author's own user-authored rows.
2. **Surface latch.** `startEditing` latches `editingSurfaceTouch` from `useInputMode()` at edit start (mirrors `MessageEvent`) — reading input mode mid-edit would remount the form and drop unsaved text. Desktop (`inlineEditing = isEditing && !editingSurfaceTouch`) replaces the row body with the inline editor; touch renders the form's own `Drawer` from `overlays` while the row stays put beneath it.
3. **Initial doc from markdown.** Board/conversation payloads (`BoardPostMessage`) carry only `contentMarkdown` on the wire (INV-58), so `editInitialContentJson` parses it back with `parseMarkdown`. A no-op edit is still detected because `MessageEditForm`'s baseline re-serializes that same doc.
4. **Clear-to-delete.** `MessageEditForm` calls `onDelete` when submitted empty (its intrinsic delete path). `MessageItem` routes that through the same `DeleteMessageDialog` + `messageService.delete` (with `enqueueOperation` offline fallback) the timeline uses — not a standalone "Delete message" menu entry.
[...]

PR #1143, opened 2026-07-03
feat(sharing): conversation-aware back-link for shared messages

kristofferremback:
Continues the board/conversation action-parity line (#1124 copy-link+react → #1126/#1130 quote → #1134 edit → #1136 delete → #1137 discuss → #1141 save/reminder deep-linking) with **#4 Share**. Design basis: `docs/board-view-design.md` §"Conversations span streams within one root"; access model INV-62; deep-link precedent `buildConversationPanelPath` (`apps/frontend/src/lib/stream-links.ts`, added in #1141).

## Problem

Two gaps on the conversation surfaces (board card / conversation panel / label page), all rendered through `MessageItem`:

1. **No Share action.** The in-stream row (`MessageEvent`) has offered the cross-stream Share picker for a while, but `MessageItem` never wired `onShare`, so a message viewed on a board card or in the conversation panel had no way to be shared into another stream.

2. **The shared pointer's back-link is always the stream permalink.** When message M is shared into stream T, the resulting card in T links back to M via `/w/:ws/s/:streamId?m=:messageId` (`shared-message-block.tsx`). A conversation spans its root + threads, so a message shared *from* a conversation loses that context on the round-trip back — the reader lands in the raw home stream rather than the conversation they were looking at.

## Solution

Wire the cross-stream Share picker onto `MessageItem` (parity with `MessageEvent`), and carry the source conversation on the pointer so its back-link reopens the message in that conversation's side panel instead of the stream permalink. The conversation origin rides as an **optional third segment** on the pointer wire format, mirroring the `quote:` optional-segment pattern, so every existing two-segment pointer round-trips unchanged.

### How it works

1. **Wire format (`packages/prosemirror/src/pointer-urls.ts`, `markdown.ts`).** `SharedMessageHref` gains optional `conversationId`. `buildSharedMessageHref` emits the legacy two-segment `shared-message:streamId/messageId` when there's no origin and appends `/conversationId` only when present; `parseSharedMessageHref` reads `parts[2] || undefined`. The markdown line serializer/`parseSharedMessageLine` regex gain an optional `(?:\/([\w-]+))?` segment. A two-segment (legacy / in-stream) link parses with `conversationId` undefined.

2. **Node attr (`shared-message-extension.ts`).** `SharedMessageAttrs` + the TipTap node gain `conversationId` (`data-conversation-id`, `default: null`). `renderHTML` omits the attribute entirely when absent, so an in-stream share's HTML/markdown stays the legacy shape. The attr round-trips through copy/paste (composer insertion in `message-input.tsx` already spreads the whole attrs object).

3. **Back-link (`lib/markdown/components.tsx`, `shared-message-block.tsx`).** `findSharedMessageInChildren` threads `conversationId` from the parsed href to `SharedMessagePointerBlock`, which links to `buildConversationPanelPath(workspaceId, conversationId, messageId)` when present, else the stream permalink as before.

4. **Share surface (`components/message/message-item.tsx`).** `onShare` opens `ShareMessageModal` with `attrs` that include the row's `conversationId` prop (set by the board card / conversation panel; undefined on the label page → plain stream pointer). `MessageEvent`'s share attrs are unchanged — in-stream shares stay two-segment.

### Key design decisions

**1. Conversation-aware pointer, not just the Share affordance.** The user chose the fuller scope ("do it in full, with backwards compatibility") over shipping only the picker with a stream-permalink back-link. So the pointer itself carries the origin.

**2. `conversationId` is a display/nav hint, not an access vector — so no backend change.** The back-link is built client-side from the node attr / parsed markdown and is resolved under the *clicking viewer's own* access when followed: the conversation panel fetches the conversation workspace-scoped and self-guards, so a stale or tampered id at worst opens an empty/denied panel for the tamperer. Content hydration stays `messageId`-keyed with the existing `shared_messages` grant checks (`hydrateSharedMessageIds`, `ShareService.validateAndRecordShares` — both untouched). This is the same treatment the pointer's cached `authorName` already gets (trusted display data), distinct from `streamId`, which *is* validated server-side because it gates access. Given that, a backend body-rewrite/validation pass (as #1141 did for the persisted `saved_messages.conversation_id`, which is the *saver's own* return path) would add a conversation lookup to the hot render/ingest path purely for a cosmetic back-link — disproportionate (INV-36).

**3. E2E rows on `MessageItem` don't offer Share.** There's no plaintext-decrypt path on this surface (that lives on the in-stream row, which passes `sourcePlaintext`), and a ciphertext pointer share is rejected server-side (`E2E_SHARING_NOT_ALLOWED`). Gated on `rowStream?.e2eEnabled === true` → route E2E sharing through the timeline instead. Board/conversation surfaces are non-E2E anyway (GAM can't run over ciphertext); the only E2E-capable `MessageItem` surface is the label page.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/prosemirror/src/pointer-urls.ts` | `SharedMessageHref.conversationId?`; build appends optional 3rd segment; parse reads it |
| `packages/prosemirror/src/markdown.ts` | `sharedMessage` serializer passes `conversationId`; `parseSharedMessageLine` regex + return gain the optional segment |
| `apps/frontend/src/components/editor/shared-message-extension.ts` | `SharedMessageAttrs.conversationId?` + node attr (`data-conversation-id`, omitted when null) |
| `apps/frontend/src/lib/markdown/components.tsx` | `findSharedMessageInChildren` threads `conversationId` to the pointer block |
| `apps/frontend/src/lib/markdown/shared-message-block.tsx` | conversation-aware back-link via `buildConversationPanelPath`, else stream permalink |
| `apps/frontend/src/components/message/message-item.tsx` | wire `onShare` → `ShareMessageModal` with `conversationId` in attrs; gated off E2E rows |

## Test files

| File | Change |
| ---- | ------ |
| `packages/prosemirror/src/pointer-urls.test.ts` | round-trip: two-segment, three-segment, legacy-parse, reject bad href |
| `packages/prosemirror/src/markdown.test.ts` | round-trip a conversation-origin `sharedMessage` node losslessly |
| `apps/frontend/src/lib/markdown/shared-message-block.test.tsx` | two-segment → stream permalink; three-segment → conversation panel back-link |
| `apps/frontend/src/components/conversations/conversation-panel.test.tsx` | end-to-end: share a panel row → handoff carries `conversationId` |

## Out of scope (deliberate)

- **The `truncated` deep-chain placeholder** (`shared-messages/card-body.tsx`) still links to the stream permalink. The hydration payload (`HydratedSharedMessage`) doesn't carry `conversationId`, and `truncated` is a pathological-depth guard (`MAX_HYDRATION_DEPTH`), so threading the field through the recursive hydrator isn't worth it.
- **Public developer markdown reference** (`apps/public-site/.../markdown.astro`) keeps documenting the two-segment form. The third segment is an internal frontend origin hint that external API authors have no conversation context to set; two-segment stays canonical for external authoring.
- **No backend change** — see design decision #2.

## Test plan

- [x] `bun test` in `packages/prosemirror` — 135 pass (incl. new pointer-urls + markdown cases)
- [x] Frontend `vitest run` over `lib/markdown`, `components/editor`, `components/share`, `components/shared-messages`, `stores/share-handoff-store`, `components/conversations/conversation-panel` — all green (conversation-panel 15 pass incl. new share test; shared-message-block 8 pass)
- [x] `apps/backend` `bun test src/features/messaging/sharing` — 52 pass (grant-recording path unaffected by the wire-format change)
- [x] `tsc --noEmit` clean in `apps/frontend` and `packages/prosemirror`; full-monorepo pre-commit tsc (app + test), astro check, and OpenAPI `--check` all passed on commit
- [x] eslint + prettier clean on all changed files
- [x] Self-review (claim-verification pass): every named symbol/path/behavioral claim above checked against the diff; no findings

---
_Generated by [Claude Code](https://claude.ai/code/session_01FJugH9pBZBv9y8Dh5dNQMg)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1143"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785651355&installation_id=129946197&pr_number=1143&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1143&signature=f83f1561acd2e89e99bf66d7d21b2b53bb7be3d3a56f42cb801c4c34087ab186"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/conversations/conversation-panel.test.tsx
+++ b/apps/frontend/src/components/conversations/conversation-panel.test.tsx
@@ -21,6 +21,7 @@ import * as userProfileModule from "@/components/user-profile"
 import * as contextsModule from "@/contexts"
 import * as touchCapableModule from "@/hooks/use-touch-capable"
 import * as discussModule from "@/hooks/use-discuss-with-ariadne"
+import * as shareHandoffModule from "@/stores/share-handoff-store"
 import type { BoardViewPost } from "@/hooks/use-stable-board-view"
 
 const WORKSPACE_ID = "ws_1"
@@ -180,6 +181,30 @@ describe("ConversationPanel", () => {
     await user.click(firstRowMenu)
 
     expect(await screen.findByText("Quote reply")).toBeTruthy()
+  })
+
+  it("shares a conversation row with a conversation-aware pointer (conversationId in the handoff)", async () => {
+    const user = userEvent.setup()
+    // Give the share picker one selectable target channel.
+    vi.spyOn(workspaceStoreModule, "useWorkspaceStreams").mockReturnValue([
+      { id: "stream_target", type: "channel", slug: "general", displayName: null, visibility: "public" },
+    ] as never)
+    const queueSpy = vi.spyOn(shareHandoffModule, "queueShareHandoff").mockImplementation(() => {})
+
+    mountPanel({ cached: asCached(makePost()) })
+    await screen.findByText("Opening message body.")
+
+    const [firstRowMenu] = screen.getAllByRole("button", { name: "Message actions" })
+    await user.click(firstRowMenu)
+    await user.click(await screen.findByText("Share message"))
+
+    // The cross-stream picker opens; selecting a target queues the share.
+    await user.click(await screen.findByText("#general"))
+
+    expect(queueSpy).toHaveBeenCalledWith(
+      "stream_target",
+      expect.objectContaining({ messageId: "msg_1", conversationId: CONVERSATION_ID })
+    )
   })
 
   it("reveals the quote affordance when a row is swiped left on touch", async () => {
--- a/apps/frontend/src/components/editor/shared-message-extension.ts
+++ b/apps/frontend/src/components/editor/shared-message-extension.ts
@@ -13,6 +13,14 @@ export interface SharedMessageAttrs {
   authorId: string
   /** The actor type of the source author, cached for the same reason */
   actorType: string
+  /**
+   * The conversation this message was shared from, set only when the share
+   * originated on a conversation surface (board card / conversation panel). When
+   * present the rendered pointer card's back-link reopens the source in that
+   * conversation's side panel instead of its home-stream permalink; in-stream
+   * shares leave it undefined and the card links to the stream permalink.
+   */
+  conversationId?: string
 }
 
 /**
@@ -55,6 +63,13 @@ export const SharedMessageExtension = Node.create({
         default: "user",
         parseHTML: (element) => element.getAttribute("data-actor-type"),
         renderHTML: (attrs) => ({ "data-actor-type": attrs.actorType }),
+      },
+      conversationId: {
+        default: null,
+        parseHTML: (element) => element.getAttribute("data-conversation-id"),
+        // Omit the attribute entirely when there's no conversation origin so an
+        // in-stream share's HTML/markdown stays the legacy two-segment shape.
+        renderHTML: (attrs) => (attrs.conversationId ? { "data-conversation-id": attrs.conversationId } : {}),
       },
     }
   },
--- a/apps/frontend/src/components/message/message-item.tsx
+++ b/apps/frontend/src/components/message/message-item.tsx
@@ -26,6 +26,7 @@ import { DeleteMessageDialog } from "@/components/timeline/delete-message-dialog
 import { EditedIndicator } from "@/components/timeline/edited-indicator"
 import { MessageHistoryDialog } from "@/components/timeline/message-history-dialog"
 import { ReminderPickerSheet } from "@/components/timeline/reminder-picker-sheet"
+import { ShareMessageModal } from "@/components/share/share-message-modal"
 import type { MessageActionContext } from "@/components/timeline/message-actions"
 import { useQuoteReply } from "@/components/timeline/quote-reply-context"
 import { useMessageReactions, stripColons, reactionShortcodes } from "@/hooks/use-message-reactions"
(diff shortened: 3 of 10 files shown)


kristofferremback:
### Code review

No issues found. Checked for bugs, data-flow lifecycle (the `conversationId` round-trip through share attrs → handoff → node attr → markdown serialize/parse → back-link), CLAUDE.md compliance, plan/intent adherence, design quality, security, UX, and mobile.

Highlights the reviewers confirmed:
- **Backward compatibility is real:** `buildSharedMessageHref` emits the legacy two-segment string when `conversationId` is absent; the optional regex group (`[\w-]+`, ≥1 char) matches both two- and three-segment links and never captures an empty string; the pre-existing two-segment round-trip test still passes with no `conversationId` key.
- **`conversationId` is not an access vector:** content hydration/access stays keyed on `(messageId, streamId)` with the existing `shared_messages` grant checks; `conversationId` only feeds `buildConversationPanelPath`, and the conversation panel's own pre-existing, workspace-scoped auth check (INV-62 root-stream boundary) rejects a crafted id — it can't leak content. "No backend change" holds.
- **Field survives the whole pipe and is set only where intended:** the in-stream `MessageEvent` share correctly omits `conversationId` (stays two-segment); the label page omits it (stream-permalink fallback); the composer insertion spreads the full attrs object so nothing drops it; copy/paste round-trips via the paired `parseHTML`/`renderHTML`.
- **Menu/mobile parity:** "Share message" comes from the single shared `messageActions` registry (identical label/icon/position across `MessageItem` and `MessageEvent`), gated the same way on both the desktop menu and the touch `MessageActionDrawer`; the E2E gate hides Share uniformly (E2E sharing stays on the in-stream decrypt-confirm path).

---

PR #1161, opened 2026-07-03
feat(conversations): thread-follow routing for "Reply in conversation"

kristofferremback:
Follow-up to #1146 (Mechanism C). Design basis: `docs/board-view-design.md` "Continuation is recency-biased" + `INV-62` (access inherited through the root).

## Problem

"Reply in conversation" (#1146) arms the current stream's composer and files the send flat into it via `{ intent: "existing" }` (`message-input.tsx`). When the conversation has since **moved into a thread**, that flat send re-interleaves the channel — the exact mess `board-view-design.md` ("Continuation is recency-biased") says to avoid: "posting into the anchor root would re-interleave the channel, the exact mess convert-to-thread avoids."

The board card and conversation panel already route replies recency-biased (`useReplyToBoardPost` + `lastActiveStreamId`). The timeline's inline arm was the one reply surface that didn't follow the conversation.

## Solution

When the armed conversation is live in a thread, hand off to the conversation panel (Mechanism B) instead of filing inline: the panel renders the conversation flattened across its root + threads, and its `BoardReplyComposer` already routes the reply recency-biased into the live thread. Kris's ruling — open the conversation view to reply, so it isn't weird wherever the reply lands — and keep the inline arm for the common same-stream case. The result is **adaptive**:

- Conversation live in the **current stream** → inline strip/send, unchanged from #1146.
- Conversation live in a **thread** → open the conversation panel + land the user in its composer.

### How it works

1. **`conversation-reply-open-store`** — a one-shot request/consume signal keyed by conversation id, mirroring `snippet-request-store` exactly (30s TTL, per-key subscribers). "Reply in conversation" fires from a message row (stream-content React tree) but the reply lands in the conversation panel (a separate panel-slot tree), so the row queues a request and the mounted panel consumes it. Registered for reset in `AccountScope.flushModuleStoreCaches` so it clears on account switch.

2. **`message-input.tsx`** — the armed strip already fetches the conversation's `BoardPost` for its topic label (`useConversationBoardPost`, cached row or one-shot by-id fetch). The same projection carries the most-recently-active stream: `recentMessages.at(-1)?.streamId ?? conversation.streamId`. An effect fires when that resolves to a stream **other than** the current one — `requestConversationReplyOpen(conversationId)`, `openPanel(createConversationPanelId(conversationId))`, then disarm. Same-stream (or unresolved) keeps the inline path. Fires once: disarming nulls `conversationReply`, and the guard holds.

3. **`conversation-panel.tsx`** — consumes the signal on mount (catching a request queued before it existed) and subscribes (catching one that arrives while it's already open for this conversation), raising `autoOpenReply`. Threaded through `ConversationPanelBody` to `BoardReplyComposer`, which expands its resting affordance into the live editor and focuses it (`autoFocus`).

### Key design decisions

**1. Projection-derived last-active stream, not a new wire field.** The redirect reads `BoardPost.recentMessages` (already fetched for the label), not a denormalized `lastActiveStreamId` on the conversation event. This matches #1146's deliberate choice not to denormalize the topic label onto the wire. Tradeoff (below in Out of scope): `recentMessages` can lag a just-landed thread reply, so a single reply can re-interleave before the projection catches up; the steady state (a conversation sitting in a thread) resolves correctly.

**2. Keep the inline arm (Kris's call).** The panel handoff replaces the inline arm only when the conversation is elsewhere. A same-stream reply stays inline and snappy — no panel, no fetch-gated delay. The inline strip renders immediately on arm; the redirect only supersedes it once the projection resolves to a thread.

**3. One-shot store, not a URL param or context.** The message row and the panel live in different React trees, so a shared context can't bridge them; a URL param would leak "open the composer" intent into the panel's shareable link. The transient store is the established pattern here (`snippet-request-store`, `share-handoff-store`).

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/stores/conversation-reply-open-store.ts` | One-shot per-conversation "open the reply composer" signal bridging the message row → conversation panel |
| `apps/frontend/src/stores/conversation-reply-open-store.test.ts` | Unit coverage: request/consume, per-conversation scoping, TTL eviction, subscriber notify/scope |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/message-input.tsx` | Derive last-active stream from the fetched `BoardPost`; redirect to the conversation panel + queue reply-open + disarm when it isn't the current stream |
| `apps/frontend/src/components/conversations/conversation-panel.tsx` | Consume the reply-open signal (mount + subscribe), thread `autoOpenReply` to the composer |
| `apps/frontend/src/components/board/board-reply-composer.tsx` | Optional `autoOpenReply` prop; effect expands + focuses the composer when raised (default false on board cards) |
| `apps/frontend/src/auth/account-scope.tsx` | Reset the new store on account switch |
| `apps/frontend/src/components/timeline/message-input.test.tsx` | `usePanel` spy; board-post mock carries `recentMessages`; new thread-follow redirect test |
| `apps/frontend/src/components/conversations/conversation-panel.test.tsx` | `autoOpenReply` raised on a queued request, collapsed on a plain open |

## Out of scope (deliberate)

- **Exact recency via a wire field.** `recentMessages.at(-1).streamId` is a projection snapshot; a very-recent thread reply not yet in the projection can route a single reply into the channel. A denormalized `lastActiveStreamId` (assigner signature + transaction-order change) would make it exact — same denormalization declined in #1146, deferred here for the same reason.
- **E2E streams.** `BoardReplyComposer` already refuses E2E hosts ("open the note to reply there"); unchanged.
- **`docs/board-view-design.md` open-decision #2** ("lean: content node") is still stale vs the shipped hidden-field design — a #1146 doc cleanup, not this PR.

## Test plan

- [x] `vitest run` — store + message-input + conversation-panel + message-actions: **120 pass**
- [x] `vitest run` — board + event-list + auth suites (regression sweep of touched components): **17 pass**
- [x] `tsc --noEmit -p tsconfig.json` (frontend) — clean
- [x] Pre-commit hook (full monorepo lint / tsc / astro / api-docs) — passed on commit
- [ ] No manual browser pass — verified via the component tests (redirect fires on a thread-live conversation, inline arm on same-stream, panel raises `autoOpenReply`); the live handoff timing (strip → panel) isn't exercised end-to-end

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fn3ieJe79LdC27D5WWixQD)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1161"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785679080&installation_id=129946197&pr_number=1161&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1161&signature=df1c87ea0174f8f0cb5bc65c79797305d6d4cfa36d6333a74633659bc1a077b0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/auth/account-scope.tsx
+++ b/apps/frontend/src/auth/account-scope.tsx
@@ -12,6 +12,7 @@ import { resetStreamStoreCache } from "@/stores/stream-store"
 import { resetDraftStoreCache } from "@/stores/draft-store"
 import { resetShareHandoffStoreCache } from "@/stores/share-handoff-store"
 import { resetSnippetRequestStoreCache } from "@/stores/snippet-request-store"
+import { resetConversationReplyOpenStoreCache } from "@/stores/conversation-reply-open-store"
 import { resetE2eSessionStoreCache } from "@/stores/e2e-session-store"
 import { resetRevealGate } from "@/sync/reveal-gate"
 import { useAuth } from "./hooks"
@@ -70,6 +71,7 @@ function flushModuleStoreCaches(): void {
   resetDraftStoreCache()
   resetShareHandoffStoreCache()
   resetSnippetRequestStoreCache()
+  resetConversationReplyOpenStoreCache()
   resetE2eSessionStoreCache()
   resetRevealGate()
 }
--- a/apps/frontend/src/components/board/board-reply-composer.tsx
+++ b/apps/frontend/src/components/board/board-reply-composer.tsx
@@ -41,13 +41,13 @@ interface BoardReplyComposerProps {
    *  continuation (a reply follows the conversation into its thread). */
   lastActiveStreamId: string | null
   /**
-   * When it turns true, expand the resting affordance into the live composer and
-   * focus it — the conversation panel raises this after being opened via "Reply
-   * in conversation" so the user lands straight in the reply editor. Left false
-   * on board cards (they open on tap). One-shot: a manual collapse afterwards is
-   * respected because the flag doesn't change again.
+   * Monotonic nonce: each increment expands the resting affordance into the live
+   * composer and focuses it — the conversation panel bumps it after being opened
+   * via "Reply in conversation" so the user lands straight in the reply editor. A
+   * counter (not a boolean) so a second request re-opens the composer after a
+   * manual collapse; `0`/absent leaves it collapsed (board cards open on tap).
    */
-  autoOpenReply?: boolean
+  openReplySignal?: number
 }
 
 /**
@@ -72,13 +72,14 @@ export function BoardReplyComposer(props: BoardReplyComposerProps) {
   const buttonRef = useRef<HTMLButtonElement>(null)
 
   // "Reply in conversation" opened the panel and asked it to land in the editor.
-  // Deps on the flag alone, so a later manual collapse isn't reopened.
-  const { autoOpenReply } = props
+  // Keyed on the nonce so each request re-opens the composer even after a manual
+  // collapse; `0`/absent is the resting default.
+  const { openReplySignal } = props
   useEffect(() => {
-    if (!autoOpenReply) return
+    if (!openReplySignal) return
     setResting("idle")
     setOpen(true)
-  }, [autoOpenReply])
+  }, [openReplySignal])
 
   // Quote reply from a message row in this conversation lands here. The resting
   // affordance leaves the form unmounted while collapsed, so this always-mounted
--- a/apps/frontend/src/components/board/board-reply-composer.tsx
+++ b/apps/frontend/src/components/board/board-reply-composer.tsx
@@ -40,6 +40,14 @@ interface BoardReplyComposerProps {
   /** The conversation's most-recently-active stream, for recency-biased
    *  continuation (a reply follows the conversation into its thread). */
   lastActiveStreamId: string | null
+  /**
+   * When it turns true, expand the resting affordance into the live composer and
+   * focus it — the conversation panel raises this after being opened via "Reply
+   * in conversation" so the user lands straight in the reply editor. Left false
+   * on board cards (they open on tap). One-shot: a manual collapse afterwards is
+   * respected because the flag doesn't change again.
+   */
+  autoOpenReply?: boolean
 }
 
 /**
@@ -62,6 +70,15 @@ export function BoardReplyComposer(props: BoardReplyComposerProps) {
   const [open, setOpen] = useState(false)
   const [resting, setResting] = useState<RestingState>("idle")
   const buttonRef = useRef<HTMLButtonElement>(null)
+
+  // "Reply in conversation" opened the panel and asked it to land in the editor.
+  // Deps on the flag alone, so a later manual collapse isn't reopened.
+  const { autoOpenReply } = props
+  useEffect(() => {
+    if (!autoOpenReply) return
+    setResting("idle")
+    setOpen(true)
+  }, [autoOpenReply])
 
   // Quote reply from a message row in this conversation lands here. The resting
   // affordance leaves the form unmounted while collapsed, so this always-mounted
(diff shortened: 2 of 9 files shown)


kristofferremback:
### Code review
**Confidence: 7/7** — Excellent (loop run to convergence)
**Review models:** Orchestrator: claude-opus-4-8 | Reviewers: sonnet ×4 (spec+design, correctness+data-flow, UX, mobile) over two rounds + a final verify pass

Ran the review loop to a clean pass. Findings were surfaced and fixed across the branch's commits; the final pass found nothing.

**Round 1** (vs `47814ad`) — 3 findings, fixed in `c1e29e3`:
1. `message-input.tsx` — the arm handler focused the channel composer before routing was known, popping the mobile keyboard on the wrong composer for a thread-follow → focus deferred to the resolve effect, same-stream only.
2. `message-input.tsx handleSubmit` — the flat `{intent:"existing"}` send fired off armed-state alone; sending before the board-post projection resolved could re-interleave the channel → guard redirects to the panel unless confirmed same-stream.
3. `board-reply-composer.tsx` / `conversation-panel.tsx` — boolean `autoOpenReply` couldn't represent a repeat open request → monotonic `openReplySignal` nonce.

CodeRabbit independently caught #3 and the nonce, plus a navigation race (the routing effect closed over stale `conversationReply` on a stream switch → `streamIdRef` + dropped dep) and asked for `recentMessages` optional-chaining + de-dup → extracted `boardPostLastActiveStreamId`. All ✅ addressed in `c1e29e3`.

**Round 2** (vs `c1e29e3`) — 2 findings, fixed in `ca168cb`:
4. `message-input.tsx` — `conversationReplyLastActiveStreamId` is a live Dexie value; a background reply moving the conversation into a thread could evict a same-stream composer mid-typing → route latched per arm (`routedArmIdRef`).
5. `message-input.tsx` — the send-time redirect left typed content as a silent channel draft (the panel covers the view on mobile) → `toast.info` per INV-63's deferred-action carve-out.

**Final pass** (vs `ca168cb`) — no issues.

---
<details><summary>📐 Plan Adherence</summary>CLEAN — adaptive routing (inline for same-stream, panel handoff for thread-live) matches the stated design; inline arm kept per the user's ruling.</details>
<details><summary>🔍 Bugs / 🔁 Data Flow</summary>CLEAN after fixes — send-before-resolve race, navigation race, live-value re-redirect, and repeat-open all closed; store TTL/listener lifecycle and account-scope reset verified.</details>
<details><summary>📋 CLAUDE.md</summary>CLEAN — INV-38 (no dead code post-rename), INV-63 (`toast.info` deferred carve-out, no happy-path success toast — guard test passes), INV-35/37 (shared `boardPostLastActiveStreamId`; panel keeps its live `displayedReplies` derivation, a superset once the rail loads).</details>
<details><summary>🎛️ UX / 📱 Mobile</summary>CLEAN after fixes — no wrong-composer focus, no mid-composition eviction, redirect is signalled.</details>

Verified: 73 unit/integration tests across the touched + `board` + `use-conversations` suites, `tsc --noEmit` clean, pre-commit hook (full monorepo lint/tsc/astro/api-docs) green on each commit.

---

PR #1166, opened 2026-07-03
feat(e2ee): sealed turns for the Pi and Claude harnesses, end to end

kristofferremback:
**Tracking:** `docs/plans/agent-runtimes-unification-redesign.md` §2.6 (E2EE for external agents) — the client half plus the gate flip. Supersedes #997 (written pre-#1067 against the old HTTP wire; its crypto design carries over, its harness integration does not). No Linear ticket.

## Problem

The sealed wire for external bots existed only server-side: `SealedTurnContext` on the claim, `/sealed-steps`, `/sealed-complete`, BIK wrap-coverage claim gating, actor grants. Nothing consumed it — no harness had an identity key, the SDK (`@threa/bot-runtime-client`) had zero crypto, and `EXTERNAL_SEALED_DELIVERY` was `false`, so no claim ever resolved to `sealed`. Three client-side gaps #997 did not cover made the feature unshippable even with a client:

1. **No sealed interim path.** The Claude channel's whole UX is `send` (interim) + `reply` (final) + relayed permission prompts, all via the plaintext message API — which correctly rejects E2E streams (`E2E_STREAM_PLAINTEXT_UNSUPPORTED`).
2. **No sealed steps off HTTP.** #1067 moved the 150+ step writes per turn onto the `/bot` socket to stop billing the edge Worker; the sealed step endpoints were HTTP-only, so rich sealed traces would have re-inflated exactly that cost.
3. **No way to get a sealed scratchpad served at all.** `pi-local` and `claude-code-channel` dispatch requires a session link (`runtime-kind-config.ts: sessionLinking: "required"`), and links only exist for scratchpads the harness itself creates — which were always plaintext.

## Solution

Sealed support lives once in the SDK and both harnesses consume it; the backend gains the two missing wire pieces; the linked scratchpad a harness creates can be encrypted from birth; and the policy gate flips on.

### How it works

1. **SDK sealed module** (`extensions/bot-runtime-client/src/{crypto,sealed}.ts`): a vendored `@threa/crypto` slice (noble X25519 KEM — Bun's WebCrypto lacks X25519, and the extensions install standalone where the private workspace package can't resolve) with a byte-parity drift guard against the canonical package; `BikKeystore` (per-install X25519 keypair, persisted `0600`, `bik_<ulid>` id, public half on **every** hello/presence — the instance upsert overwrites an omitted key); `openSealedTurnContext` (unwrap SSK per generation with root-stream AAD, decrypt trigger + history); `sealReply`/`sealStep` (enclave-identical `streamId|id|senderId` AAD binding); `mintStreamKeyWraps` (the seal direction, for provisioning); `parseSealedTurnContext` (defensive wire validation); `scrubSealedError`.
2. **Transport**: `recordSealedSteps` on `BotRuntimeTransport` — WS `bot:invocation:sealed-steps` first, per-step HTTP `/sealed-steps` fallback, same best-effort semantics as plaintext steps. Backend side, the frame drives the same `recordSealedSteps` write-op as the HTTP route (extracted `sealed-callbacks.ts`: `authorizeSealedCallback` + `finalizeSealedStep`, shared so the transports can't diverge; `ws-http-schema-parity.test.ts` covers the frame). Sealed steps bump the agent-session heartbeat, mirroring the enclave.
3. **Sealed interims**: `POST …/bot-invocations/:id/sealed-messages` (callback-token auth, client-minted `messageId` binds the AAD and doubles as the idempotency key). `RemoteSession.sendInterim`, `postToStream` (permission relay), and every close path (`reply`, no-response, interrupt notes, idle timeout, intercepted-claim acks) funnel through sealed-aware chokepoints (`completeTurn`), so no close path can leak plaintext.
4. **Harness-created encrypted scratchpads** (two-phase): `POST /bot-runtime/sessions` accepts `e2e:{ownerKeyId}` — the create transaction writes the `e2e_streams` row (INV-E1), the bot's own actor grant, and forces `memoryMode: off` (INV-E2). The harness then mints the generation-0 SSK, wraps it to the owner's UIK (fetched via new `GET /bot-runtime/owner-e2e-key`) + its own BIK, and stores both via `POST /streams/:id/e2e/key-wraps` — bot-actor-only, current generation only, and only while the generation is empty, under a `FOR UPDATE` on the e2e row (INV-20). Two-phase because the wrap AAD binds to the server-minted stream id.
5. **Rich detail when sealed**: pi-remote emits the real command/args/output in sealed trace steps (60K clamp instead of the 10K plaintext cap); plaintext turns keep every existing redaction. `shouldEmitFullTrace` = sealed ∧ `config.sealedFullTrace !== false`; `full` defaults to `false` at each formatter so a missed call site fails safe.
6. **Gate flip**: `EXTERNAL_SEALED_DELIVERY = true`. Without an owner-minted grant the verdict is still `denied: no-key-grant` — flipping alone grants nothing. The consent surface is `InviteBotButton` (next to the Ariadne pill on E2E scratchpads): the owner picks a workspace bot, the invite pins the bot id, and the unlocked client rolls + wraps the stream key to its BIK. The existing revive heal (`computeMissingActorWraps` in the encryption affordance) already covers a bot whose key registers after the invite.

### Key design decisions

**1. Crypto in the SDK, vendored, parity-guarded.** The extensions ship standalone (`install-local.ts` copies them out of the workspace), so `@threa/crypto` can't be a dependency. The slice lives once in `bot-runtime-client` (both harnesses vendor it via their install scripts, which now inherit its `@hpke/*` + `ulid` deps) and `crypto.parity.test.ts` asserts byte-identical AAD layouts, envelope versions, and cross seal/open against the canonical package — the same INV-35 mitigation #997 used, minus the per-harness duplication.

**2. Full trace detail only where the server is blind.** The point of sealed steps is giving the owner more without giving the server anything. The toggle can only opt a sealed turn *back to redacted* — it can never turn full detail on for plaintext (that would reintroduce the "a secret hit the server" liability E2EE removes).

**3. Session-control stays plaintext, acks go silent on E2E.** Session-control invocations never get a sealed context (no agent session row → no callback binding), so a markdown ack on an E2E scratchpad is rightly rejected. The plaintext `/complete` now applies its E2E gate only when a message would be persisted, so `noResponse` closes work — the command still executes, `command:completed` events carry the feedback, and the channel falls back to a silent close on the 400. A sealed session-control ack wire is follow-up work.

**4. Provisioning is all-or-nothing on an empty generation.** Wrap slots are immutable (`ON CONFLICT DO NOTHING`), so a partial or replayed provision could otherwise splice wraps of two different keys. The endpoint refuses any generation that already has wraps (409, which the harness treats as landed) and serializes racers with `FOR UPDATE` — caught in self-review; see Design evolution.

**5. Hydration failure fails the invocation, loudly and scrubbed.** A malformed context, missing BIK, or wrap that won't open fails the claim with a generic reason (`scrubSealedError` — class name only, never a message that could echo decrypted content) instead of TTL-recycling in a loop or leaking plaintext into `/fail`.

## New files

| File | Purpose |
| ---- | ------- |
| `extensions/bot-runtime-client/src/crypto.ts` | Vendored `@threa/crypto` slice (noble KEM, SSK seal/open/wrap, AAD builders, sealed payload) |
| `extensions/bot-runtime-client/src/crypto.parity.test.ts` | Byte-parity drift guard against canonical `@threa/crypto` |
| `extensions/bot-runtime-client/src/sealed.ts` | `BikKeystore`, sealed-claim open, `sealReply`/`sealStep`, `mintStreamKeyWraps`, wire validation |
| `extensions/bot-runtime-client/src/sealed.test.ts` | Keystore persistence/0600/concurrency, open/seal round trips, AAD binding, provisioning round trip |
| `apps/backend/src/features/public-api/sealed-callbacks.ts` | Transport-neutral sealed callback auth + step finalize, shared by HTTP handlers and the WS write-op |
| `extensions/pi-remote/src/threa-remote.sealed.test.ts` | Full-vs-redacted trace policy tests |
| `extensions/remote-session/src/session.sealed.test.ts` | Sealed session behavior with real crypto both directions + the two-phase create |
| `apps/backend/tests/e2e/sealed-harness-session.test.ts` | Full round trip through the real server/DB/outbox/claim machinery (see Test plan) |
| `apps/frontend/src/components/encryption/invite-bot-button.tsx` | Owner consent surface: invite a workspace bot into an E2E scratchpad |

## Modified files

| File | Change |
| ---- | ------ |
| `extensions/bot-runtime-client/src/{transport,types,index}.ts` | `recordSealedSteps` (WS-first + HTTP fallback), BIK fields on hello, exports |
| `apps/backend/src/features/bot-runtimes/{runtime-write-ops,socket-handler,index}.ts` | `recordSealedSteps` contract + `bot:invocation:sealed-steps` frame |
| `apps/backend/src/features/bot-runtimes/service.ts` | `createLinkedScratchpadSession` E2E block (flag row + actor grant + memory off, one txn) |
| `apps/backend/src/features/public-api/{handlers,schemas,routes,runtime-write-ops}.ts` | Sealed-messages endpoint, owner-key endpoint, wrap provisioning, sealed-steps delegation to the shared write-op, `noResponse`-only E2E gate on `/complete`, `e2eEnabled` on session responses |
| `apps/backend/src/routes.ts`, `docs/public-api/openapi.json` | Route mounts + regenerated spec |
| `packages/agent-runtime/src/runtime/negotiate-capabilities.ts` | `EXTERNAL_SEALED_DELIVERY = true` |
| `extensions/pi-remote/src/threa-remote.ts`, `install-local.ts`, `README.md` | BIK on hello/presence, sealed claim/steps/complete/fail, full-trace policy, `e2e` + `sealedFullTrace` config, vendoring |
| `extensions/remote-session/src/{session,client,identity}.ts` | Sealed hydration, sealed-aware chokepoints, two-phase E2E create, `THREA_E2E`/`THREA_BIK_PATH` config |
| `extensions/claude-code-remote/{src/config.ts,install-local.ts,README.md}` | BIK path beside the channel config, vendoring + dep inheritance, docs |
| `apps/frontend/src/pages/stream.tsx`, `components/encryption/index.ts` | Render the invite affordance |
| `.agents/skills/update-pi-remote-plugin/SKILL.md` | Sealed checklist items |

### Rebase integration (#1164)

Rebased onto main after #1164 (per-tool trace steps tailed from the Claude session transcript) landed mid-review — it built the exact hook this PR completes ("`full` mode exists for E2EE-backed links"). `DeliveredTurn` gains `sealed`; `RemoteSession.recordSteps` seals each tailed frame under the stream key and ships the sealed-steps wire (same 50-frame chunking, no statusText); `TranscriptTracer.beginTurn` takes a per-turn `RedactionMode` override that never outlives the turn, driven by `turn.sealed` + a `sealedFullTrace` config (`THREA_SEALED_FULL_TRACE`, default on). So Claude Code sessions on encrypted scratchpads now get the same rich sealed traces as Pi — real commands, patches, outputs — with a test pinning that the next plaintext turn is redacted again.

## Out of scope (deliberate)

- **Sealed attachments** — `THREA_ATTACH:` directives are stripped with a note on sealed turns; uploading plaintext bytes would leave the E2E boundary. The `AttachmentRef` payload machinery exists for a follow-up.
- **Sealed session-control acks** — silent close on E2E scratchpads (decision 3).
- **Enclave continuity for external bots** (interjection pull, rolling summary, digests) — `SealedTurnContext` deliberately omits them (INV-36).
- **`e2e: true` as the harness default** — opt-in config, because an encrypted scratchpad opts out of GAM memory extraction (INV-E2); making it the default is a product call.

- The first provisioning endpoint checked "no wraps for this generation" inside the transaction but without a lock — two concurrent provisions could each pass the check and interleave two different keys into the immutable slots. Fixed with `FOR UPDATE` on the `e2e_streams` row (commit `bf01b145`).
- `sendInterim` on a sealed turn originally sealed a fresh body per attempt, so a retry after a failed POST double-posted. The failed body is now cached per send-seq and retried with the same message id, which the server dedupes (same commit).
- The e2e round-trip test flaked until isolated: a running `dev:test` stack shares `threa_test` and its (older-code) outbox listeners race the suite's. The test file documents running with `TEST_DATABASE_URL` pointed elsewhere; this contention predates the PR.

## Test plan

- [x] SDK `extensions/bot-runtime-client`: 38 pass (parity drift guard + sealed module + transport)
- [x] `extensions/remote-session`: 72 pass (incl. sealed session suite with real crypto both directions + two-phase create)
- [x] `extensions/pi-remote`: 90 pass (incl. full-vs-redacted policy); standalone `install-local.ts` output verified — vendored modules resolve, deps inherited, suite passes in the copy
- [x] `extensions/claude-code-remote`: 47 pass post-rebase (incl. the per-turn full-mode tracer test); standalone install verified pre-rebase
- [x] Backend unit: full root `bun test` — 2294 pass post-rebase; targeted suites (public-api, bot-runtimes, e2e-streams, enclave-runtimes, agent-runtime) — 335 pass
- [x] Backend e2e (isolated DB): **full sealed round trip through the real server** — harness-created encrypted scratchpad, owner seals trigger, sealed dispatch → claim → BIK opens prompt, sealed step + interim + reply, owner opens all output, wire carries only placeholders + ciphertext; plus the whole suite: 321 pass across 24 files
- [x] Typecheck clean: backend (app + tests), frontend, all SDK packages
- [x] OpenAPI spec regenerated (`--check` clean)
- [ ] Frontend vitest: 136 pre-existing failures (`localStorage.clear is not a function`) — identical count on `origin/main` in a clean worktree; environmental, unrelated
- [ ] Live turn through a real running Pi/Claude harness against a browser owner — the e2e test covers the identical wire + crypto; a manual smoke with `THREA_E2E=1` on a real session is the remaining step

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1166"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785685454&installation_id=129946197&pr_number=1166&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1166&signature=b254655c69368fa0e68cd7fdd780899807b60692666dfa927e7d7ae36169bc40"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
> **Superseded by a newer review:** https://github.com/threahq/threa/pull/1166#issuecomment-4883449430

<details><summary>Original review (superseded 2026-07-04)</summary>


### Code review
**Confidence: 5/7** — Good (two non-blocking frontend issues; backend + crypto clean)

Found 2 issues — both on the new bot-invite consent component, neither on the sealed/crypto path (which is clean across all lenses).

1. `apps/frontend/src/components/encryption/invite-bot-button.tsx:66` — The "Invite agent" affordance is hidden entirely when the session is locked (`isUnlocked && candidates.length > 0`), unlike the sibling Ariadne `InviteActorButton` it's meant to mirror, which stays visible while locked and, on click, surfaces a `toast.warning` ("… invited — unlock this scratchpad's encryption to grant it access") via the `invited-unwrapped` path. On a locked encrypted scratchpad, the owner sees the Ariadne invite pill but no "Invite agent" pill and no hint that unlocking restores it. (ux-consistency: divergence from the reference affordance the component is meant to match; a real owner on a locked scratchpad can't tell the capability exists.)
   https://github.com/threahq/threa/blob/c118d052221960bbaa97fa665228dd6fc2685bee/apps/frontend/src/components/encryption/invite-bot-button.tsx#L64-L80

2. `apps/frontend/src/components/encryption/invite-bot-button.tsx:50-63` — Invited bots render as one full pill each (icon + untruncated `bot.name`, no `max-w`/`truncate`, no cap) into the header row at `stream.tsx:605` (`flex items-center gap-2 flex-1 min-w-0`, no `flex-wrap`, no horizontal scroll). Inviting 2+ bots — the feature's own use case (Pi remote + Claude Code channel on one scratchpad) — on a ~375px viewport collapses the title and can push the right-side Search/Context icon buttons off-screen or force page-level horizontal scroll. The same header already solves this: `LabelStack` caps visible chips at 3 + a "+N" overflow. (mobile-layout: materially worsens the pre-existing single-pill crowding rather than repeating the pattern identically; the bounded-chip pattern is right there to reuse.)
   https://github.com/threahq/threa/blob/c118d052221960bbaa97fa665228dd6fc2685bee/apps/frontend/src/components/encryption/invite-bot-button.tsx#L48-L63

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Matches the stated intent: SDK-shared sealed support (one implementation, both harnesses), full-detail-when-sealed with plaintext staying redacted (the toggle cannot unlock plaintext), BIK not derived from the API key, two-phase harness-created encrypted scratchpads, consent UI, gate flipped. No missing corresponding changes found.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No reportable bugs. Correctness reviewer traced and ruled out (with evidence): AAD seal/open symmetry (server echoes the envelope aad verbatim, never recomputes — a relocation fails the GCM tag), callback-token binding identical on the WS frame and HTTP routes (both through `authorizeSealedCallback` → `verifyCallbackToken`'s `timingSafeEqual`), the two-phase provisioning race (SELECT … FOR UPDATE before the empty-generation check), the `/complete` plaintext gate (gated on the same boolean that creates a message row, so `noResponse` can't smuggle content), and the TranscriptTracer mode override (fully replaced/cleared per turn, snapshotted locally in `emitLines`).</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Sealed invocation lifecycle (claim → hydrate → steps/interim/reply/complete/fail) funnels every close through the sealed-aware chokepoint with no plaintext leak. One noted-not-flagged edge: a phase-2 provisioning failure after phase-1 commits leaves an E2E stream with zero wraps — but that "orphaned create" state is already handled by the pre-existing `useOwnerWrapRepair` "Finish setup" affordance (remote-session self-heals via poll-tick retry; pi-remote's one-shot path relies on the repair button). Below the bar.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No introduced invariant violations (checked INV-E1/E2/E7, INV-20, INV-8, INV-4/7, INV-11, INV-51/52, INV-63, INV-40).</details>
<details><summary>🏗️ Design [CLEAN]</summary>No leaky abstractions or parallel implementations flagged — sealed logic lives once in the SDK; the backend WS/HTTP sealed-steps paths share one write-op via `sealed-callbacks.ts`.</details>
<details><summary>🔒 Security [CLEAN]</summary>No high-confidence vulnerabilities. Callback-token binding enforced symmetrically across transports; owner-key validated on session create; wrap provisioning is actor-scoped, current-generation-only, empty-generation-only under a row lock. BIK-on-disk (0600) is intentional, not a finding.</details>
<details><summary>🎛️ UX [1 issue]</summary>Issue 1 above — locked-state affordance divergence from the Ariadne reference.</details>
<details><summary>📱 Mobile [1 issue]</summary>Issue 2 above — unbounded invited-bot pills can overflow the narrow header when 2+ bots are invited.</details>

</details>

kristofferremback:
### Code review

Scope: the delta merged since the prior 5/7 review (`4878252285`) — the session-control-on-E2E work (#1175: claim-gate fix, `sealedAck`, `sealedReply`, harness sealing), the plaintext-step leak gate, the graceful/silent-close fix, and the bot-invite consent-pill follow-ups. **The crypto/backend correctness path is clean** — the data-flow reviewer traced the claim-gate boolean nesting (the `session-control` bypass is `AND`-nested under the mandatory current-generation `EXISTS`; `trigger` is a server-set column, unspoofable), the session-finalize skip (unreachable for session-control), redelivery dedup (`clientMessageId` fast-path), the key-generation check (fails closed → 400, transaction rolls back, no wedge), and the sealed-vs-plaintext auth model (equivalent strength). Zero correctness/security findings.

Found 6 issues, all non-blocking:

1. `apps/frontend/src/components/encryption/invite-bot-button.tsx:58-86` — The truncated bot names and the new "+N" overflow pill disclose their only extra info (which bots can decrypt this scratchpad) through a hover-only Radix `Tooltip`, which no-ops on `pointerType === "touch"`, and the trigger is a bare `<span>` (not focusable). On a phone — the exact ~375px header case this cap was added to solve — tapping "+2" does nothing and the folded names are unreachable for sighted touch users (they survive only in `aria-label`, screen-reader-only). `LabelStack` (`apps/frontend/src/components/labels/label-stack.tsx`) already solves this by branching on `useInputMode() === "touch"` to a tap `Drawer`; that branch wasn't carried over. (ux-accessibility + mobile-touch, both lenses independently)
   https://github.com/threahq/threa/blob/966ddad3316e719937ed554053b41cf964b42c9f/apps/frontend/src/components/encryption/invite-bot-button.tsx#L58-L86

2. `extensions/remote-session/src/session.ts:829` — `completeAck`'s plaintext fallback swallows **any** `ThreaApiError` with `status === 400` as "E2E rejected the plaintext ack, close silently" via `noResponse`. But `/complete` also returns 400 for `CAPABILITY_NOT_DECLARED` (manifest didn't declare `reply`) and other validation errors — those get masked as a successful close instead of surfacing `session-control ack failed: …` (INV-11, fail loud). The sibling pi-remote narrows correctly with `String(error).includes("E2E_STREAM_PLAINTEXT_UNSUPPORTED")`; session.ts can't match the code today because `ThreaClient.request` (client.ts:112-114) throws the body away and `ThreaApiError` carries only `status`. The honest fix threads the server `code` through `ThreaApiError`, then narrows the catch. (bug / INV-11)
   https://github.com/threahq/threa/blob/966ddad3316e719937ed554053b41cf964b42c9f/extensions/remote-session/src/session.ts#L823-L836

3. `apps/frontend/src/components/encryption/invite-bot-button.tsx:63,69,77,83` — Every pill's label/tooltip asserts "{bot} can read this encrypted scratchpad" unconditionally. That was always true when the button rendered only post-unlock; now that the `isUnlocked` gate is dropped, a bot invited **while locked** lands in `useInviteActor`'s `invited-unwrapped` state — recorded but with no wrapped key, so it can't decrypt until the owner unlocks and the repair-wrap runs. The pill renders identically and the only signal of the degraded state is a transient `toast.warning`. Sibling `InviteActorButton` hedges with "{label} is in this scratchpad", which stays true regardless of wrap state. (ux-consistency)
   https://github.com/threahq/threa/blob/966ddad3316e719937ed554053b41cf964b42c9f/apps/frontend/src/components/encryption/invite-bot-button.tsx#L58-L86

4. `apps/backend/src/features/public-api/handlers.ts` (`buildSessionControlSealedAck` ~L554, the claim's `sealedAck` branch ~L1171) — the server-side glue that makes the feature work end to end ships with no unit test (grep for `sealedAck` / `buildSessionControlSealedAck` across `apps/backend/src` is empty). Its sibling `buildSealedClaimContext` — same file, same handler — is thoroughly covered in `sealed-claim.test.ts` including the INV-11 loud-failure cases. The new function's inverse contract (silently return `undefined` instead of throwing) and its wrap-filter (`recipientKind === "bot" && recipientKeyId === instance.publicKeyId && keyGeneration === currentKeyGeneration`) are untested. The `/complete` `sealedReply` side and `openSealedAck` are covered; the claim side isn't. (missing-change / test-coverage)
   https://github.com/threahq/threa/blob/966ddad3316e719937ed554053b41cf964b42c9f/apps/backend/src/features/public-api/handlers.ts#L554-L611

5. `extensions/pi-remote/src/threa-remote.ts:1810-1814` — stale comment: "the ack has no sealed wire yet … Sealed acks are follow-up work" contradicts the `sealSessionControlAck` / `sealedReply` fast path implemented ~100 lines above in this same diff. This branch is now the *fallback* taken only when the bot can't seal (no BIK / wrap race), not "no sealed wire exists". (INV-25, comments describe the current implementation)
   https://github.com/threahq/threa/blob/966ddad3316e719937ed554053b41cf964b42c9f/extensions/pi-remote/src/threa-remote.ts#L1806-L1816

6. `extensions/remote-session/src/session.ts:824-828` — same stale claim, second harness: "a sealed ack wire is follow-up work" contradicts `sealSessionControlAck` implemented ~40 lines above. (INV-25)
   https://github.com/threahq/threa/blob/966ddad3316e719937ed554053b41cf964b42c9f/extensions/remote-session/src/session.ts#L823-L836

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>The diff matches the stated intent (session-control commands work + seal their ack on E2E). OpenAPI regenerated for `sealedReply`; SDK/harness contracts consistent with the backend. No missing corresponding changes.</details>
<details><summary>🔍 Bugs [1 issue]</summary>#2 — over-broad 400 catch in `remote-session` `completeAck` masks non-E2E validation errors (INV-11).</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Redelivery dedup via `clientMessageId: bot-invocation:${claim.id}` (ON CONFLICT DO NOTHING + refetch) covers both the sealed and plaintext ack branches; a redelivered `/complete` discards the second minted ciphertext id. Key-generation checked at complete time (not claim-pinned) but fails closed on a rekey race. Session-finalize block correctly skipped for session-control (no agent_session).</details>
<details><summary>📋 CLAUDE.md Compliance [2 violations]</summary>#5, #6 — stale "sealed acks are follow-up work" comments in both harnesses (INV-25).</details>
<details><summary>🏗️ Design [CLEAN]</summary>Claim-gate fragment stays shared by `claimOne` + `findBootstrapInvocations` (no divergence). `buildSessionControlSealedAck` vs `buildSealedClaimContext` split is justified — session-control has no trigger/history to open, only an ack to seal.</details>
<details><summary>🔒 Security [CLEAN]</summary>No plaintext of a session-control command can reach an E2E stream's storage: the plaintext `/steps` sink is gated on `isE2eStream`, and the ack goes through `sealedReply` (ciphertext) or a `noResponse` close. `sealedReply` auth (instanceId+claimToken) is equivalent to `/sealed-complete`'s callback token (itself a hash of the same claim token). Cross-stream / wrong-generation `sealedReply` rejected.</details>
<details><summary>🎛️ UX [2 issues]</summary>#1 (touch discoverability of the overflow pill), #3 (over-strong "can read" claim for invited-while-locked bots).</details>
<details><summary>📱 Mobile [1 issue]</summary>#1 — the "+N" overflow affordance is functionally inert on touch (hover-only tooltip); mirror `LabelStack`'s `useInputMode`→`Drawer` branch.</details>

PR #1172, opened 2026-07-03
fix(scheduled-messages): signal a scheduled reply whose conversation drifted into a thread

kristofferremback:
Follow-up to #1168 (merged), from the `/code-review` pass on that change — the one surviving finding, split out because #1168 merged before this commit landed.

## Problem

#1168 made a scheduled send carry an armed "Reply in conversation" directive. But arming and *scheduling* now diverges silently from arming and *sending*. The composer's "Replying in <topic>" strip sits above both the Send button and the schedule picker. When the armed conversation has drifted **thread-live** (a background reply moved it into a thread after the same-stream arm latched — `routedArmIdRef` in `message-input.tsx` keeps the strip inline rather than evicting the user), the two paths behave differently:

- **Send** (`handleSubmit`, guard at `message-input.tsx:635`): redirects into the conversation panel (thread-follow) + `toast.info`, keeping the draft.
- **Schedule** (`handleSchedule`): can't follow a live thread — a deferred send has no live thread at fire time, and the panel has no schedule affordance — so it correctly attaches by conversation id and lands flat in the composer's stream. But it did so with **no signal**: same strip, materially different outcome.

## Solution

Signal the divergence; don't change the routing. `handleSchedule` now emits a `toast.info` on the **same guard boundary** `handleSubmit` uses — armed AND the conversation not confirmed same-stream (`conversationReply !== null && conversationReplyLastActiveStreamId !== streamId`, covering both thread-live and not-yet-resolved). The same-stream case stays silent (the Scheduled page is the visible result there). The directive is still forwarded either way; the assigner attaches it cross-stream within one root at fire time.

## Files changed

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/message-input.tsx` | `handleSchedule`: derive `filesFlatIntoConversation` (same boundary as `handleSubmit`); `toast.info(...)` after a successful schedule when true; add `conversationReplyLastActiveStreamId` to the deps |
| `apps/frontend/src/components/timeline/message-input.test.tsx` | same-stream schedule asserts **no** toast; new drift test (arm same-stream → `rerender` with a thread-live `useConversationBoardPost` → schedule) asserts the directive is still forwarded **and** the toast fires |

## Key design decision

**Signal, not reroute.** Mirroring `handleSubmit`'s panel redirect isn't available — a scheduled send fires later (no live thread to follow, and the conversation panel exposes no schedule affordance), so attaching by id + landing in the origin stream is the only sensible deferred semantics (conversation grouping is preserved via the directive; the assigner allows cross-stream-within-one-root). The gap was purely the missing cue, so the fix is purely the cue. `toast.info` is the INV-63 deferred-action carve-out (not a happy-path `toast.success` — success stays silent), so the guard test `no-happy-path-success-toast.test.ts` still passes.

## Test plan

- [x] Frontend `vitest run src/components/timeline/message-input.test.tsx` — 30 pass (adds no-toast-same-stream + toast-on-drift)
- [x] Frontend `vitest run src/lib/no-happy-path-success-toast.test.ts` — 1 pass (the `toast.info` is not a happy-path success toast)
- [x] `tsc --noEmit -p apps/frontend/tsconfig.json` clean
- [x] `/code-review` (3 Sonnet agents) on the parent diff scored 7/7 after this fix; this commit *is* that fix (UX lens finding, 85). Correctness & Data-Flow and Spec & Design lenses were clean.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01BGJNrXBxKiikEuJ1ki4VjF)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1172"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785700742&installation_id=129946197&pr_number=1172&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1172&signature=a0ede5f51eb50a157b2c5d784af61a3e69c86e49941ed6c56248d7fee7df40b2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/message-input.test.tsx
+++ b/apps/frontend/src/components/timeline/message-input.test.tsx
@@ -638,6 +638,51 @@ describe("MessageInput", () => {
         })
       )
       expect(screen.queryByTestId("conversation-reply-strip")).not.toBeInTheDocument()
+      // Same-stream (default mock): the scheduled reply files where the strip
+      // promised, so no divergence signal — success stays silent (INV-63).
+      expect(infoToastSpy).not.toHaveBeenCalled()
+    })
+
+    it("signals the deferred outcome when scheduling an armed reply whose conversation drifted into a thread", async () => {
+      mockComposerState.canSend = true
+      mockComposerState.content = makeDoc("schedule this drifted reply")
+
+      const { rerender } = render(
+        <Wrapper>
+          <MessageInput workspaceId={workspaceId} streamId={streamId} />
+        </Wrapper>
+      )
+      // Arm same-stream: inline strip, route latched, no redirect.
+      act(() => registeredConversationReplyHandler?.({ conversationId: "conv_1" }))
+
+      // A background update moves the conversation into a thread. The latch keeps
+      // the strip inline (no eviction), but the last-active stream is now the thread.
+      vi.spyOn(useConversationsModule, "useConversationBoardPost").mockReturnValue({
+        post: {
+          conversation: { id: "conv_1", streamId, topicSummary: "Pizza plans" },
+          recentMessages: [{ streamId: "thread_789" }],
+        },
+        isLoading: false,
+        notFound: false,
+        refetch: vi.fn(),
+      } as unknown as ReturnType<typeof useConversationsModule.useConversationBoardPost>)
+      rerender(
+        <Wrapper>
+          <MessageInput workspaceId={workspaceId} streamId={streamId} />
+        </Wrapper>
+      )
+
+      await act(async () => {
+        await capturedOnSchedule?.(new Date("2099-01-01T00:00:00.000Z"))
+      })
+
+      // Directive still forwarded (files into the conversation by id)...
+      expect(mockScheduleMutateAsync).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


kristofferremback:
Closing as a duplicate — this exact fix already landed in `main` via #1171 (`64ea392`, "signal when a scheduled reply files into a drifted conversation"): same `toast.info` on the same guard boundary, with a superset of the tests. This PR was opened from a branch that wasn't rebased onto the latest `main`, so it didn't reflect that #1171 had merged. No delta worth keeping.

---

PR #1180, opened 2026-07-05
fix(push): stop silent-push quota burn and harden webpush delivery

kristofferremback:
## Problem

Push subscriptions keep getting silently revoked, and delivery is sometimes delayed or stalls entirely. A multi-agent audit of the push pipeline (21 agents, findings adversarially verified) traced the deregistration to one root cause and the stalls/delays to two more, all in this diff's blast radius:

1. **Silent-push quota burn.** Every stream read emitted `stream:read` → `deliverClearForStreams` → one notification-less `{action:"clear"}` push per stream to **every** subscription the user has (`findByUserId`, no filter — including the reading device itself); mark-all-read sent one per unread stream. Subscriptions are `userVisibleOnly: true`, so a push that shows nothing is a policy violation: Firefox docks a per-subscription quota (16 → 8 → 5 as days-since-visit grow, refreshed only on revisit) and **revokes the subscription at 0** (`_backgroundUnregister`, `UNSUBSCRIBE_QUOTA_EXCEEDED` in `PushService.sys.mjs`); iOS revokes after 3; Chrome shows the junk "This site has been updated in the background" notification. Reading Threa on one device was deregistering the user's other devices. Both prod users run two devices; one is 100% Firefox.
2. **Pipeline wedge.** `webpush.sendNotification` was called with no options — the web-push library only arms a socket timeout when `options.timeout` is numeric (`web-push-lib.js:222`), so a push-service connection that accepts TLS but never responds blocks forever (Node sets no TCP keepalive on these sockets). The outbox handler processes events sequentially under a held cursor lock, so one hung send froze **all** push delivery for every user until redeploy.
3. **Doze-delayed delivery + 28-day stale backlog.** No `Urgency` header means "normal" — FCM/autopush batch those until the device exits Doze, so a real-time message landed minutes-to-hours late on Android. No `TTL` means the 28-day default; no `Topic` means an offline device replayed every queued push individually on reconnect.

Additionally, non-404/410 send failures — including 401/403 VAPID auth rejections, which are a total-outage misconfig, never a per-device problem — were logged at `warn` and indistinguishable from routine noise (INV-11).

## Solution

Stop sending pushes that can't show a notification; make every push that remains carry explicit delivery semantics.

### How it works

1. **Clear-push fan-out removed.** `deliverClearForStream`/`deliverClearForStreams` are deleted; the push outbox handler no longer consumes `stream:read`/`stream:read_all` (those events still flow to the socket broadcast path, untouched). Every event the push listener now handles results in a visible notification.
2. **Cross-device dismissal without pushes.** Two existing/new paths replace the clear push:
   - *Open apps* (unchanged): `workspace-sync.ts` already posts `SW_MSG_CLEAR_NOTIFICATIONS` to the SW on the `stream:read`/`stream:read_messages`/`stream:read_all` socket echoes.
   - *Closed apps* (new): `lib/notification-sweep.ts` + `useNotificationSweep` (mounted as `NotificationSweeper` in `workspace-layout.tsx`). Once a fresh workspace bootstrap lands, it closes every displayed notification whose tag is a stream (`stream_…` / `stream_…:mention`) with nothing unread — keep-set = `unreadCounts > 0` ∪ `unreadActivities[].streamId`. Non-stream tags (`rewrap:…`, `session-expired`, `threa-test`) are never touched. Stale-data safe by construction: the bootstrap query cache is in-memory and only ever filled by a network fetch this page load, and the sweep re-keys on the unread-set signature, not the cache reference.
3. **Explicit delivery options on every send** (`PushDeliveryOptions` in `service.ts`): `timeout: 10s` on all sends (bounds the wedge from indefinite to seconds); `urgency: "high"` + `TTL: 24h` for activity/reminder/rewrap pushes; `TTL: 60s` for the diagnostic test push; `TTL: 7d` + `urgency: "normal"` for session-expired. `Topic` collapses queued pushes per stream on reconnect — the header caps at 32 base64url chars and a prefixed id (`stream_` + 26-char ULID = 33) doesn't fit, so `pushTopic()` uses the bare ULID plus a kind suffix (`m` mentions, `r` rewrap) so different notification groups for the same stream don't collapse into each other.
4. **Failure classification** (`classifySendFailure`, shared by delivery and test paths): 404/410 → evict as before; 401/403 → `logger.error` naming the VAPID misconfig; everything else stays `warn`.
5. **SW tombstone.** The `action === "clear"` branch in `sw.ts` stays for one TTL window — clears queued for offline devices before this deploy carried the 28-day default TTL, and without the guard a straggler would fall through to the display path and render a junk notification. Marked delete-after 2026-08-05.

### Key design decisions

**1. Dismissal moved off web-push entirely, accepting a UX regression.** Alternatives rejected: batching clears into one push (still a silent push — still burns quota, just slower) and sending clears only to recently-active devices (an active device has the app open and gets the socket clear anyway — the only devices a clear push could help are exactly the idle ones it deregisters). The accepted cost: a notification read elsewhere stays on a closed device's lock screen until that device next opens the app. Slack-style instant cross-device dismissal fundamentally requires silent pushes that web push policies punish.

**2. Topic per stream trades badge granularity for reconnect sanity.** With `Topic`, a device offline through 5 messages in one stream gets one push (the newest) instead of five sequential buzzes; the SW's grouped-notification history accumulates only from displayed pushes, so the collapsed case shows the latest message rather than "5 new". The app shows true unread state on open. Preferred over the un-coalesced flood.

**3. No retry queue for transient 5xx/429 send failures (deliberate).** The audit confirmed those are swallowed; a retry queue is real machinery (INV-36 — not justified at current volume where the dominant loss vector was the quota revocation this PR removes). The 401/403 error-level escalation covers the total-outage tail.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/lib/notification-sweep.ts` | Pure tag selector + best-effort sweep that closes stream-tagged notifications with nothing unread |
| `apps/frontend/src/lib/notification-sweep.test.ts` | Selector coverage: read stream + mention group swept, unread kept, system tags never touched |
| `apps/frontend/src/hooks/use-notification-sweep.ts` | Cache-observer on the workspace bootstrap; sweeps when the unread-set signature changes |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/push/service.ts` | Clear methods deleted; `PushDeliveryOptions` + `pushTopic` + `classifySendFailure`; timeout/TTL/urgency/topic on every `sendNotification` |
| `apps/backend/src/features/push/outbox-handler.ts` | `stream:read`/`stream:read_all` branches removed; doc comment states the visible-notification-only rule |
| `apps/backend/src/features/push/service.test.ts` | New `delivery options` suite: message topic, mention topic separation, test-push TTL |
| `apps/frontend/src/sw.ts` | Clear branch kept as tombstone for in-flight 28-day-TTL clears; comment with delete-after date |
| `apps/frontend/src/pages/workspace-layout.tsx` | Mounts `NotificationSweeper` beside the other null-render hook hosts |
| `apps/frontend/src/hooks/use-unread-counts.ts` | Comment updated — cross-device dismissal is socket + sweep, not push |
| `apps/frontend/src/hooks/index.ts` | Barrel export |

## Out of scope (deliberate)

- **Suppression accuracy + subscription liveness** (over-delivery for panels/threads/board views; passive reading not counting as presence; periodic `getSubscription()` liveness check for Firefox's deferred-revocation blind window) — follow-up PR, stacked on this branch.
- **Random per-installation device key** (the UA-hash key can't distinguish same-model devices and drifts on browser updates) — audit finding verified low-impact for the reported symptoms; hygiene follow-up.
- **Outbox graceful drain on SIGTERM** (redeploy mid-batch can re-buzz an already-delivered batch) — verified LOW severity at current volume.

## Test plan

- [x] `bun test apps/backend/src/features/push` — 9 pass (3 new)
- [x] Full backend suite `bun run test` — 2323 unit + 321 e2e pass (one unrelated `claude-code-channel-session` e2e flake on first run, green in isolation and on rerun)
- [x] `apps/frontend` `bun test src/lib/notification-sweep.test.ts` — 3 pass
- [x] Frontend suite failure set diffed against a clean `origin/main` worktree — byte-identical (157 pre-existing environmental `localStorage` failures on this machine, both sides; CI runs these green)
- [x] Monorepo `bun run typecheck` clean; pre-commit lint/typecheck/migration checks green
- [x] `web-push` timeout claim verified against the installed lib source (`web-push-lib.js:222` — timeout armed only when `options.timeout` is numeric)
- [ ] Live end-to-end push against real FCM/Mozilla endpoints — needs a deployed env with VAPID keys; plan is to verify on staging (`staging` label) or after prod deploy via the in-app "Send test" diagnostic and a two-device read test

<details>

**Goal:** Stop web push from destroying its own subscriptions and harden delivery, per the July 2026 push audit (21-agent workflow, findings adversarially verified).

**What was built:**
- Removed the silent clear-push fan-out (audit finding #1, CONFIRMED critical→high): no more notification-less pushes on stream reads / mark-all-read.
- Replacement dismissal: existing socket fast-path for open apps + new bootstrap-time notification sweep for closed apps.
- `timeout: 10s` on every webpush call (finding #3, CONFIRMED high — sequential outbox pipeline under a cursor lock could wedge indefinitely on one hung connection).
- `Urgency: high` + bounded TTL + per-stream `Topic` (finding #4, CONFIRMED — Doze-deferred delivery, 28-day stale backlog, no coalescing).
- 401/403 VAPID failures escalated to error-level (finding #2, CONFIRMED — silent total outage otherwise).

**Design evolution:** none mid-PR; the shape followed the audit's fix sketches with one narrowing — no retry queue for transient send failures (INV-36).

**Known edge (accepted):** if a push is displayed while this device's bootstrap cache hasn't yet learned that stream is unread AND another stream's read lands in that sub-second window, the sweep can close the fresh banner early. Requires the app open with a live socket (user is looking at Threa anyway); the unread badge still shows.

**What's NOT included:** suppression accuracy (panel/thread URL matching), passive-presence signals, periodic subscription liveness — stacked follow-up PR.

**Status:** complete, all suites green.

</details>

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1180"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785824473&installation_id=129946197&pr_number=1180&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1180&signature=a5929feb94a271aecb179ca269c4c926806d3ebd2f64381b523bc78e98749811"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/push/outbox-handler.ts
+++ b/apps/backend/src/features/push/outbox-handler.ts
@@ -2,8 +2,6 @@ import type { Pool } from "pg"
 import {
   type ActivityCreatedOutboxPayload,
   type OutboxEvent,
-  type StreamReadOutboxPayload,
-  type StreamsReadAllOutboxPayload,
   type SavedReminderFiredOutboxPayload,
   type EnclaveRewrapNudgeOutboxPayload,
 } from "../../lib/outbox"
@@ -24,7 +22,11 @@ interface PushNotificationHandlerDeps {
 
 /**
  * Listens for outbox events and delegates push delivery to PushService.
- * Handles activity:created (show notification) and stream:read (clear notifications).
+ * Handles activity:created, saved_reminder:fired, and enclave:rewrap_nudge —
+ * every event here results in a VISIBLE notification. stream:read events are
+ * deliberately not consumed: pushing a notification-less "clear" burns the
+ * browser's silent-push quota and gets the subscription revoked (see
+ * PushService.sendAndEvictStale); dismissal is socket + sweep instead.
  * Infrastructure-only: cursor management, batching, and error handling (INV-34).
  */
 export class PushNotificationHandler extends DebouncedOutboxHandler {
@@ -43,7 +45,7 @@ export class PushNotificationHandler extends DebouncedOutboxHandler {
   // it. Sequential stops at the first failure, ensuring the cursor never
   // advances past un-delivered events.
   //
-  // Push events are low-volume (activity:created, stream:read) so sequential
+  // Push events are low-volume (activity:created, reminders, nudges) so sequential
   // within a batch of 10 has negligible latency impact — the real throughput
   // win is the dedicated realtime pool isolating push from background workers.
   //
@@ -57,26 +59,6 @@ export class PushNotificationHandler extends DebouncedOutboxHandler {
         return
       }
       await this.pushService.deliverPushForActivity(payload)
-      return
-    }
-
-    if (event.eventType === "stream:read") {
-      const payload = event.payload as StreamReadOutboxPayload
-      if (!payload?.workspaceId || !payload?.authorId || !payload?.streamId) {
-        logger.warn({ eventId: event.id }, "Skipping malformed stream:read payload")
-        return
-      }
-      await this.pushService.deliverClearForStream(payload.workspaceId, payload.authorId, payload.streamId)
-      return
-    }
-
-    if (event.eventType === "stream:read_all") {
-      const payload = event.payload as StreamsReadAllOutboxPayload
-      if (!payload?.workspaceId || !payload?.authorId || !payload?.streamIds?.length) {
-        logger.warn({ eventId: event.id }, "Skipping malformed stream:read_all payload")
-        return
-      }
-      await this.pushService.deliverClearForStreams(payload.workspaceId, payload.authorId, payload.streamIds)
       return
     }
 
--- a/apps/backend/src/features/push/service.test.ts
+++ b/apps/backend/src/features/push/service.test.ts
@@ -1,10 +1,10 @@
 import { describe, it, expect, spyOn, beforeEach, afterEach } from "bun:test"
 import webpush from "web-push"
 import type { Pool } from "pg"
-import { ActivityTypes, PrefNotificationLevels } from "@threa/types"
+import { ActivityTypes, PrefNotificationLevels, SavedStatuses } from "@threa/types"
 import { PushService } from "./service"
 import { PushSubscriptionRepository } from "./repository"
-import type { ActivityCreatedOutboxPayload } from "../../lib/outbox"
+import type { ActivityCreatedOutboxPayload, SavedReminderFiredOutboxPayload } from "../../lib/outbox"
 
 function makeActivityPayload(): ActivityCreatedOutboxPayload {
   return {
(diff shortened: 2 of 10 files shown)


kristofferremback:
### Code review

No issues found at the ≥80 reporting threshold. Checked for bugs, data-flow lifecycle (sweep pipeline end-to-end, delivery-option round-trips, session-expired partition), CLAUDE.md compliance, plan adherence, design quality, security, UX, and mobile PWA lifecycle.

Notable verifications performed by the reviewers (all resolved as non-issues):

- `stream:read`/`stream:read_all` outbox rows are still written and still consumed by the broadcast listener for the socket clear fast-path — only the push consumer was removed; no orphaned callers of `deliverClearForStream(s)` anywhere.
- webpush option names/casing (`timeout`, `TTL`, `urgency`, `topic`) verified against `@types/web-push` and `web-push-lib.js`; the library itself enforces the ≤32-char base64url topic, and `pushTopic()`'s 26-char ULID + 1-char suffix stays under it for every call site with no suffix collisions.
- Sweep tag scheme cross-checked against `resolveTag` in `sw-notification-format.ts`; `data.workspaceId` is genuinely stamped by the SW so the per-workspace filter isn't a no-op; thread stream ids exist in the bootstrap key space so thread notifications aren't spuriously swept.
- The "sweep closes a just-displayed notification" race doesn't materialize: with a socket connected, `activity:created` reaches the client ahead of the third-party push relay; disconnected, no JS runs to sweep; on resume, catch-up flushes atomically behind the sync engine's event gate before the signature changes.
- Sub-threshold observations (recorded, not blocking): saved-reminder pushes share the 24h TTL — a device offline >24h misses the lock-screen nudge but the persistent activity-feed row still surfaces it on open (~75); the sweep clears a stream's tag+mention pair on per-stream granularity (~55).

Also addressed during review: both CodeRabbit findings (workspace-scoping the sweep — real cross-workspace bug on shared origins — and delivery-option test coverage for the saved-reminder/rewrap/session-expired paths) fixed in b0dec097; the sweep-freshness comment corrected in 396fc18a.

PR #1181, opened 2026-07-05
fix(push): suppress accurately on panel/thread views and detect dead subscriptions

kristofferremback:
> Stacked on #1180 (`fix/push-quota-hardening`) — merge that first.

## Problem

Two audit-confirmed defects in how the client decides "the user can already see this":

1. **Over-delivery for the stream you're reading** (the reported "banner pops for the stream I'm viewing, then instantly resolves"). Two independent causes:
   - The SW's suppression derives the viewed stream from the URL path (`isViewingStream`, `/^\/w\/([^/]+)\/s\/([^/]+)/` in `sw-notification-format.ts`). But threads and conversations open as `?panel=<streamId|conv:id>` on whatever route is current (`panel-context.tsx`), and board/activity/search conversations live on non-`/s/` routes entirely. A thread reply's push carries the *thread's* stream id while the path names the root (or `/board`) → no match → the push displays over the very messages the user is reading, and the read-advance clear then dismisses it.
   - "Presence" counts only `pointerdown`/`keydown`/`touchstart` (`use-page-interaction.ts`). Desktop wheel-scrolling fires none, so a user reading for >2 min without clicking ages out of `PRESENCE_INTERACTION_WINDOW_MS` — the SW stops suppressing (and the backend's attended-device gate stops routing) for someone actively looking at the app.
2. **Dead subscriptions go unnoticed for up to ~a day.** Firefox revokes a subscription when its silent-push quota exhausts and defers `pushsubscriptionchange` to an idle timer / browser restart. `usePushNotifications` only re-verifies on mount, `visibilitychange`, and `online` — a long-lived focused tab fires none of those, so the settings UI shows "Enabled" while delivery is dead. (#1180 removes the quota burn that *causes* most revocations; this closes the recovery blind window for the ones that still happen.)

## Solution

Let the pages that render a stream say so explicitly, and let the push hook notice when the browser quietly dropped the subscription.

### How it works

1. **Visible-streams registry** (`lib/visible-streams.ts`): a refcounted set of on-screen stream ids, published (coalesced per microtask) into the existing `threa-presence` Cache Storage bucket — the same page↔SW channel `sw-presence.ts` already uses. Two registration points:
   - `VisibleStreamPresence` in `workspace-layout.tsx` registers the URL-derived streams (main stream + `?panel=` values that are bare `stream_…` ids).
   - `ConversationPanel` registers `panelStreamIds` — `[post.conversation.streamId, ...post.streamIds]`, the exact set it already computes for `usePanelStreamSubscriptions` — which the URL cannot express (`?panel=conv:…`).
2. **SW suppression** (`sw.ts` push handler) now suppresses when a focused same-origin client exists AND (the URL matches the pushed stream — unchanged fallback — OR the pushed `streamId` is in the visible set), still gated on `isDevicePresent()`. Multi-tab: the cache entry is last-writer-wins and the registering hook re-publishes on window `focus`, so the focused tab's set is authoritative; a stale entry alone can't suppress because the focused-client and presence gates still apply.
3. **`wheel` counts as interaction** (`use-page-interaction.ts`), feeding both the SW presence timestamp and the backend attended-device heartbeat through the one shared tracker. Deliberately `wheel` (hardware input), not `scroll`: auto-follow and virtualizer adjustments fire `scroll` on an abandoned-but-focused tab, which would fake exactly the presence this gate exists to age out. Mobile touch-scroll is already covered by `touchstart`; keyboard scroll by `keydown`.
4. **Liveness re-verify** (`use-push-notifications.ts`): every 5 minutes (`LIVENESS_CHECK_INTERVAL_MS`), `pushManager.getSubscription()` — a local call — and if the browser holds no subscription while the hook is active, re-run the idempotent `subscribe()`. Bypasses `RESUBSCRIBE_THROTTLE_MS` on purpose: it only acts when the subscription is genuinely gone. Recovery window for a Firefox quota revocation on a long-lived tab drops from up-to-a-day to ≤5 min.
5. **Honest comment** in `sw.ts` `pushsubscriptionchange`: the no-window branch claimed "persist to IndexedDB so the app can sync on next load" — no IDB write ever existed. The comment now states the real recovery path (mount re-subscribe / liveness interval).

### Key design decisions

**1. Registration over URL parsing.** Alternative rejected: teaching the SW to parse `?panel=` and resolve `conv:` ids — the SW has no access to the conversation→stream mapping (it lives in the page's query cache), and every new surface shape would need new URL grammar in the SW. Registration inverts it: the component that renders messages declares its streams; the SW only does set membership. The URL fallback stays so a registration gap degrades to today's behavior, never worse.

**2. Residual over-delivery accepted:** reading without any wheel/key/pointer event for >2 min (e.g. a short stream that needs no scrolling) still ages out of presence and shows the banner. Counting focus alone as presence would revert the deliberate "focused tab left behind keeps eating notifications" fix that the interaction gate exists for.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/lib/visible-streams.ts` | Refcounted registry + Cache Storage publish/read (page + SW shared) |
| `apps/frontend/src/lib/visible-streams.test.ts` | Registry: dedup/sort, refcount release, same-tick swap coalescing, double-release safety |
| `apps/frontend/src/hooks/use-visible-streams.ts` | Per-tab singleton registry wiring + focus re-publish; content-keyed effect |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/sw.ts` | Suppression reads the visible set (URL match kept as fallback); `pushsubscriptionchange` comment corrected |
| `apps/frontend/src/pages/workspace-layout.tsx` | `VisibleStreamPresence` null-host registers URL-derived streams |
| `apps/frontend/src/components/conversations/conversation-panel.tsx` | Registers `panelStreamIds` while mounted |
| `apps/frontend/src/hooks/use-page-interaction.ts` | `wheel` listener added (with the wheel-not-scroll rationale) |
| `apps/frontend/src/hooks/use-push-notifications.ts` | 5-min `getSubscription()` liveness interval → re-subscribe on loss |
| `apps/frontend/src/hooks/index.ts` | Barrel export |

## Out of scope (deliberate)

- Draft panels and any future panel kinds register nothing — they don't render a stream's messages, so there's nothing to suppress.
- Backend attended-device routing changes (escalation/re-fanout) — audit verified the "sole copy surfaces nowhere" claim REFUTED; the narrow OS-DND edge isn't worth routing complexity now.
- Random per-installation device key — separate hygiene follow-up.

## Test plan

- [x] `bun test src/lib/visible-streams.test.ts src/lib/notification-sweep.test.ts` — 6 pass
- [x] Frontend + monorepo typecheck clean; pre-commit lint green
- [x] Frontend suite failure set diffed against clean `origin/main` — byte-identical (the 157 pre-existing environmental `localStorage` failures on this dev machine; CI runs these green)
- [ ] `use-push-notifications` liveness interval has no unit test — that file's tests can't run on this machine (the pre-existing environmental failure above), and shipping an unexecuted test is worse than an honest gap; the logic is a 6-line interval around the existing `subscribe()`
- [ ] Live two-device suppression check (thread/conversation open in a panel while a reply arrives) — needs a deployed env; verify on staging (`staging` label) or post-deploy

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1181"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785825130&installation_id=129946197&pr_number=1181&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1181&signature=4e7385229824825cf5239226fc17bb029348c21edd2ca09d15cdb7434ecd62bc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/board/board-card-auto-read.test.tsx
+++ b/apps/frontend/src/components/board/board-card-auto-read.test.tsx
@@ -158,11 +158,12 @@ describe("BoardCard viewport auto-read (full wiring: real card, real controller,
     await act(async () => {})
     expect(screen.getByText("Opening body.")).toBeTruthy()
 
-    const io = FakeIntersectionObserver.instances.at(-1)
-    expect(io, "an IntersectionObserver should have been constructed").toBeTruthy()
     const rowEl = document.querySelector('[data-message-row][data-message-id="m_open"]')
     expect(rowEl, "the opening row should carry data-message-row").toBeTruthy()
-    expect(io!.observed.has(rowEl!), "the opening row should be observed").toBe(true)
+    // The card constructs several observers (auto-read rows, visible-streams
+    // viewport gate) — pick the one actually watching the opening row.
+    const io = FakeIntersectionObserver.instances.find((instance) => instance.observed.has(rowEl!))
+    expect(io, "an IntersectionObserver should be observing the opening row").toBeTruthy()
 
     act(() => {
       io!.fire([{ target: rowEl!, isIntersecting: true }])
--- a/apps/frontend/src/components/board/board-card.tsx
+++ b/apps/frontend/src/components/board/board-card.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useRef, useState } from "react"
+import { useEffect, useMemo, useRef, useState } from "react"
 import { Link } from "react-router-dom"
 import { useQuery } from "@tanstack/react-query"
 import { Hash, FileEdit, User, MessageSquareText, ChevronDown, PanelRight, type LucideIcon } from "lucide-react"
@@ -11,7 +11,7 @@ import { useConversationAutoRead } from "@/components/message/use-conversation-a
 import { BoardReplyComposer } from "@/components/board/board-reply-composer"
 import { QuoteReplyProvider } from "@/components/timeline/quote-reply-context"
 import { TextSelectionQuote } from "@/components/timeline/text-selection-quote"
-import { useActors } from "@/hooks"
+import { useActors, useVisibleStreams } from "@/hooks"
 import { useWorkspaceUserId } from "@/hooks/use-workspaces"
 import { useConversationService, usePanel, createConversationPanelId } from "@/contexts"
 import { conversationKeys } from "@/hooks/use-conversations"
@@ -146,6 +146,26 @@ export function BoardCard({ workspaceId, post, contextLabel, streamType }: Board
     registerExplicitUnread: setExplicitUnreadListener,
     getReadTruth,
   })
+
+  // The card is a first-class reading surface (live message bodies, viewport
+  // auto-read above), so while any part of it is on screen its streams count
+  // as visible for push suppression — otherwise a push banners the exact
+  // message the user is reading on the board. Viewport-gated (not mount-gated)
+  // so off-screen cards on a long board don't suppress streams the user can't
+  // see.
+  const [cardInViewport, setCardInViewport] = useState(false)
+  useEffect(() => {
+    const el = cardRef.current
+    if (!el || typeof IntersectionObserver === "undefined") return
+    const observer = new IntersectionObserver(([entry]) => setCardInViewport(entry.isIntersecting))
+    observer.observe(el)
+    return () => observer.disconnect()
+  }, [])
+  const cardVisibleStreamIds = useMemo(
+    () => (cardInViewport ? [...new Set([streamId, ...(post.streamIds ?? [])])] : []),
+    [cardInViewport, streamId, post.streamIds]
+  )
+  useVisibleStreams(cardVisibleStreamIds)
 
   const renderMessage = (message: RenderableMessage, continuation: boolean) => (
     <MessageItem
--- a/apps/frontend/src/components/conversations/conversation-panel.tsx
+++ b/apps/frontend/src/components/conversations/conversation-panel.tsx
@@ -22,7 +22,7 @@ import { BoardReplyComposer } from "@/components/board/board-reply-composer"
 import { QuoteReplyProvider } from "@/components/timeline/quote-reply-context"
 import { TextSelectionQuote } from "@/components/timeline/text-selection-quote"
 import { SidebarToggle } from "@/components/layout"
-import { useActors } from "@/hooks"
+import { useActors, useVisibleStreams } from "@/hooks"
 import { useWorkspaceUserId } from "@/hooks/use-workspaces"
 import { useStreamName } from "@/hooks/use-stream-name"
 import { useConversationService, usePanel, parseConversationPanel, useSidebar } from "@/contexts"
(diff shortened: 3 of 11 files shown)


kristofferremback:
### Code review
**Confidence: 7/7** — Excellent (3 issues found, all fixed in 07844d91 before merge)

Found 3 issues, all fixed on this branch:

1. `apps/frontend/src/hooks/use-visible-streams.ts` — **bug (92)**: a background tab's registration published unconditionally to the last-writer-wins cache entry, clobbering the focused tab's set. Concrete sequence: focused Tab A on stream X; stream Y opened in background Tab B (middle-click) registers and publishes `["stream_Y"]`; a push for Y arrives → some focused client exists (A) + Y ∈ set → suppressed, and neither tab shows the user anything. **Fixed:** publishes are gated on `document.hasFocus()`; the focus listener re-asserts the focused tab's set on every focus gain.
2. `apps/frontend/src/lib/visible-streams.ts` — **data-flow (85)**: the "no TTL is safe" doc argument was falsified by an abrupt tab close (no cleanup fires) while another Threa tab stays focused — the closed tab's streams stayed suppressed indefinitely. **Fixed** by the same ownership rule: only the focused tab writes, and any focus transition rewrites the entry, so a dead tab's entry survives only until the next Threa tab gains focus (and while none is focused, the SW's focused-client gate keeps it inert). Both doc comments now state the actual invariant.
3. `apps/frontend/src/components/board/board-card.tsx` — **ux-consistency (85)**: board cards render live message bodies and viewport-auto-read them, but never registered visibility — reproducing the exact reported complaint ("push for the message I'm reading") on `/board`. **Fixed:** cards register `[conversation.streamId, ...post.streamIds]` while intersecting the viewport (off-screen cards register nothing, so a long board doesn't over-suppress).

Verified clean by the other lenses:

- Only `StreamPanel` and `ConversationPanel` (and now board cards) render full stream content — no other missing registration surface; the registry doesn't duplicate `WorkspaceSyncHandler`'s sync-engine `visibleStreamIds` (different concern).
- Mobile PWA lifecycle: closing a mobile panel genuinely unmounts it (key-based reconciliation), so cleanup fires; an OS process kill leaves no focused `WindowClient`, and `isDevicePresent`'s 2-min window ages out — the stale-entry scenario doesn't survive both gates on a single device.
- The `wheel` listener is passive/capture and timestamp-only — no interference with swipe/scroll choreography; touchscreens don't dispatch `wheel`.
- Liveness recovery is invisible-and-correct: `isSubscribed` never flips false mid-recovery, so no settings-toggle flapping; `visibilitychange` → throttled resubscribe already covers mobile resume, the interval covers long-lived static tabs.
- Sub-threshold (recorded): the module-level registry singleton lacks the "INV-9 exception:" comment style its two precedents carry (~45); reading a short stream >2 min with zero wheel/key/pointer events still over-delivers — the documented accepted residual (~60).

PR #1182, opened 2026-07-05
feat(e2ee): sealed attachments for the Pi and Claude harnesses, both directions

kristofferremback:
**Tracking:** the "Sealed attachments" item from #1166's out-of-scope list ("`THREA_ATTACH:` directives are stripped with a note on sealed turns … The `AttachmentRef` payload machinery exists for a follow-up"). No Linear ticket.

## Problem

Sealed turns dropped attachments in both directions. Outbound, `stripSealedAttachmentDirectives` (remote-session) and its pi-remote twin replaced every `THREA_ATTACH:` line with "_attachments aren't supported in encrypted scratchpads yet_" — uploading through the plaintext path would have pushed file bytes outside the E2E boundary. Inbound, `openSealedTurnContext` discarded the `attachmentRefs` already riding inside the sealed payloads it decrypts, and `buildTurnContent` skipped the attachment scan entirely on sealed turns (the plaintext message list only holds ciphertext placeholders). Three server-side gaps kept it that way: the public-API upload handler rejected `e2e=true` outright (`E2E_UPLOAD_UNSUPPORTED` — its comment said the flag returns "in the Pi-remote-with-files slice"; this is that slice), the sealed-messages/sealed-complete schemas had no way to bind an attachment row to the message, and nothing stopped a mixed plaintext/E2E binding from leaking a real filename into a sealed stream's event payload.

## Solution

Reuse the first-party client's attachment scheme wholesale — the harness now does exactly what the browser does on an E2E stream, and the browser already renders the result (`E2eAttachmentList`, untouched).

### How it works

1. **One encrypt primitive** — `encryptAttachmentBytes` moves from `apps/frontend/src/lib/crypto/attachment-crypto.ts` into canonical `packages/crypto/src/attachment.ts` beside its existing `decryptAttachmentBytes` (INV-35); the frontend module re-exports. Fresh single-use 32-byte AES-256-GCM key per file, fixed `threa-attachment-v1` AAD, generation 0.
2. **Vendored into the SDK** — the same pair lands in `extensions/bot-runtime-client/src/crypto.ts` (the extensions install standalone; `@threa/crypto` can't resolve there). `crypto.parity.test.ts` gains cross-module round trips both directions plus AAD/generation constant equality, same drift-guard pattern as #1166.
3. **Outbound (`THREA_ATTACH:` on a sealed turn)** — `uploadSealedReplyAttachments` (remote-session `attachments.ts`) / `uploadSealedAttachment` (pi-remote): encrypt locally, upload only ciphertext as multipart with `e2e=true` under the placeholder name `encrypted`/`application/octet-stream`, build the `AttachmentRef` (key/iv/real filename/mime/size) from local knowledge. The refs seal into the payload via `sealReply(…, { attachmentRefs })` — no `attachment:<id>` markdown links; an E2E viewer renders from refs. The ids ride the new `attachmentIds` wire field so the server can bind the placeholder rows.
4. **Inbound** — `openSealedTurnContext` now surfaces `promptAttachmentRefs` and per-item history `attachmentRefs`. `buildTurnContent`'s sealed branch (and pi-remote's `hydrateSealedClaim`) dedupes trigger-first (`selectSealedInboundRefs`), fetches the opaque S3 ciphertext via the existing presigned-URL endpoint, decrypts locally, and writes the real filenames into `.threa-attachments/<invocationId>/` with the same manifest as plaintext turns. Best-effort per file, like plaintext.
5. **Backend** — `attachmentIds` (max 16) on `sendSealedInvocationMessageSchema` and `completeSealedInvocationSchema.reply`, passed through to `createMessage`/`createMessageInTransaction`, which already validated workspace + sharing-safety and treats E2E ciphertext as shareable. The upload gate lifts for bot API keys only (user keys keep the 400 — they still have no sealed message-write path, so their E2E upload could never be referenced). OpenAPI spec regenerated.
6. **INV-E1 backstop at the write sink** — the attachment sibling of `assertE2eContentMatch` in `event-service.ts`: a sealed message may only bind `e2eOnly` rows (a plaintext row would surface its real filename in the E2E stream's event payload) and a plaintext message never binds an `e2eOnly` row. Covers every caller, not just the new endpoints.

### Key design decisions

**1. Bot-only `e2e=true` on the public-API upload.** The original block's stated reason — "no ciphertext message-write path to bind them to" — is obsolete for bots since #1166 (sealed-messages/sealed-complete), but still true for user API keys. Lifting it for bots only keeps the fail-loud property (INV-11) instead of accepting uploads that could never be referenced.

**2. Explicit `attachmentIds` on the sealed wire, not link parsing.** The plaintext path binds via `attachment:<id>` links parsed out of the markdown; a sealed body is ciphertext the server can't parse, so the ids cross as a first-class field — cleartext metadata the server already owns (it stores the rows), while key/iv/filename ride only inside the sealed payload.

**3. Retries reuse uploads.** The sealed-interim cache (`pendingSealedInterim`) and the reply `prepared` cache both carry the uploaded `attachmentIds`/refs, so a retry after a failed POST re-sends the same sealed body and the same ids — no re-upload, no orphaned ciphertext copy, and the server's idempotency keys dedupe the message row.

**4. Failure notes are sealed too.** A missing/unreadable `THREA_ATTACH:` file becomes an "Attachment upload failed:" note inside the sealed body (local paths included — they're visible only to the owner), mirroring the plaintext behavior without leaking anything.

## New files

None — the work extends existing modules.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/crypto/src/{attachment,index}.ts` | `encryptAttachmentBytes` + `EncryptedAttachment` move in from the frontend (INV-35) |
| `apps/frontend/src/lib/crypto/attachment-crypto.ts` | Becomes a re-export of the canonical pair; keeps the ref-cache bridge |
| `extensions/bot-runtime-client/src/crypto.ts` | Vendored attachment encrypt/decrypt + AAD/generation constants |
| `extensions/bot-runtime-client/src/sealed.ts` | `openSealedTurnContext` surfaces `promptAttachmentRefs` + history `attachmentRefs` |
| `extensions/bot-runtime-client/src/{index.ts,crypto.parity.test.ts,sealed.test.ts}` | Exports, parity round trips, refs-surface tests |
| `extensions/remote-session/src/attachments.ts` | `uploadSealedReplyAttachments`, `selectSealedInboundRefs`, `downloadSealedInboundAttachments` |
| `extensions/remote-session/src/client.ts` | `SealedWireReply` (body + `attachmentIds`), `sealedAttachments` on `ClaimedInvocation` |
| `extensions/remote-session/src/session.ts` | Encrypt-upload-seal replaces strip-with-note in `sendInterim`/`reply`; sealed branch of `buildTurnContent` downloads from refs; `completeTurn` carries refs/ids; `stripSealedAttachmentDirectives` deleted (INV-38) |
| `extensions/pi-remote/src/threa-remote.ts` | Hand-rolled mirror: `uploadSealedAttachment`, `downloadSealedContextAttachments`, sealed prompt note now offers the directive |
| `apps/backend/src/features/public-api/schemas.ts` | `attachmentIds` (max 16) on sealed-messages + sealed-complete reply |
| `apps/backend/src/features/public-api/handlers.ts` | Bot-key `e2e=true` uploads; both sealed endpoints pass `attachmentIds` through |
| `apps/backend/src/features/messaging/event-service.ts` | INV-E1 backstop: sealed message ⇔ `e2eOnly` attachment, both directions |
| `apps/backend/tests/client.ts` | `uploadFile`/`botUploadFile` accept extra form fields (the `e2e` flag) |
| `apps/backend/tests/e2e/sealed-harness-session.test.ts` | Round trip now carries an attachment each way; bot key gains attachment scopes |
| `docs/public-api/openapi.json` | Regenerated (`--check` clean) |
| `extensions/{pi-remote,claude-code-remote}/README.md`, `docs/claude-code-channel.md`, `.agents/skills/update-pi-remote-plugin/SKILL.md` | Stale "not supported on sealed turns" text replaced with the actual behavior |

Three reviewer agents (crypto/E2E-boundary, backend invariants, harness wiring) over the diff; four findings fixed in `31656e24`:

- **Stale interim replay** — the sealed-interim retry cache keyed on `seq` alone, so a runtime that abandoned a failed send and posted *different* text at the same seq silently got the stale body replayed (and its new attachments dropped). The cache now keys on the input text too.
- **Edit-path backstop gap** — the INV-E1 attachment check covered creates only; a plaintext edit could still reference an `e2eOnly` row (undecryptable chip + an unscanned-bytes distribution vector). `_validateEditAttachmentReferences` now rejects it (sealed edits are blocked upstream, so only that direction was open).
- **>16 directives wedged the turn** — the server caps `attachmentIds` at 16 but neither harness clamped, so 17 `THREA_ATTACH` lines uploaded fine and then every completion retry 400'd until the idle timeout ate the reply. Both harnesses now clamp and note the overflow in the sealed body.
- **Attachment-only interims** read "(empty message)" beside their files; they now keep an empty body like a user's attachment-only sealed send.

Dismissed as pre-existing/follow-up: the S3 storage key trusting the multipart filename (placeholder-forcing covers the DB row; all in-repo clients send `"encrypted"` — defense-in-depth for a follow-up), pi-remote re-uploading on a failed sealed complete (no `prepared`-style cache; orphan ciphertext is undecryptable garbage, matching its plaintext shape), no unbound-attachment sweeper (pre-existing for all uploads), the backstop throwing plain `Error` → 500 rather than 400 (matches its sibling checks; converting the family is a separate change), and the `postToStream` plaintext fallback window (#1166 shape; the server rejects it).

## Out of scope (deliberate)

- **Sealed session-control acks** — already shipped separately (#1175); the silent close survives only as the no-BIK/wrap-race fallback.
- **Sealed attachments on message edits** — the public API has no sealed edit path; the backstop covers the create sink.
- **Enclave-produced attachments** — Ariadne still never emits `attachmentRefs` in replies (its `commitMessage` passes sources only); unchanged.
- **`e2e: true` as the harness default** — still a product call (INV-E2 memory tradeoff), per #1166.

## Test plan

- [x] `packages/crypto`: 27 pass (10 pre-existing envelope failures are Bun's missing X25519 WebCrypto, identical on `origin/main`)
- [x] SDK `extensions/bot-runtime-client`: 47 pass (parity both directions, refs surfacing)
- [x] `extensions/remote-session`: 85 pass (sealed upload/refs/failure-note/inbound-manifest, retry-reuses-uploads, stale-replay guard, 16-id clamp)
- [x] `extensions/pi-remote`: 94 pass (sealed outbound + inbound with real crypto over mocked fetch)
- [x] `extensions/claude-code-remote`: 47 pass
- [x] Backend unit: full `bun test src/` — 2327 pass; new coverage for the INV-E1 backstop (create both directions + edit path), the bot-only upload gate, and `attachmentIds` pass-through on both sealed endpoints
- [x] Backend integration: 554 pass
- [x] Backend e2e (isolated DB): full suite 321 pass across 24 files — the sealed round trip now uploads an encrypted attachment owner→harness (opened from `promptAttachmentRefs`, fetched, decrypted to the original bytes) and harness→owner on both the interim and the reply, asserting the wire carries only placeholder summaries and real filenames never appear in any event JSON
- [x] Typecheck: full monorepo + all four extension tsconfigs clean; lint + migrations + OpenAPI `--check` clean (pre-commit hook)
- [ ] Frontend vitest: not run — known pre-existing environmental failures (`localStorage.clear`); the touched module's own suite (`attachment-crypto.test.ts`, 7 pass) runs green under bun
- [ ] Live smoke through a real running harness with `THREA_E2E=1` — the e2e test covers the identical wire + crypto; a manual pass on a real session remains
- [x] Pre-PR review (3 reviewer agents): 4 findings fixed (see Design evolution), 5 dismissed as pre-existing/follow-up

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1182"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785825604&installation_id=129946197&pr_number=1182&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1182&signature=352c2579a6e7de67faa75f253e440c773030ea5e243e9a24fdeabfcf8e8d8d6b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
@coderabbitai review

kristofferremback:
@coderabbitai full review

PR #1226, opened 2026-07-06
feat(board): archived-streams toggle, default hidden

kristofferremback:
## Problem

The workspace board leaked cards under archived streams. `ConversationRepository.findByWorkspaceForViewer` (`apps/backend/src/features/conversations/repository.ts`) filters the cross-stream feed by lens, scope, type, label, hide, and mute — but had no `archived_at` clause, so an archived channel's conversations kept surfacing as board cards. Archiving a channel already seals it everywhere else (the sidebar hides it, sends are rejected — `service.ts:307,320,330`), but the board kept showing its topics.

## Solution

Add a `showArchived` axis, defaulting to hide, mirrored the way every other board filter already is: an SQL predicate on the seed/pagination query and a read-side twin in `use-stable-board-view` so the shared IDB feed re-filters live without a refetch. A card is "archived" when its **effective root** is archived — archiving marks only the root row and a thread inherits its root's lifecycle (INV-62), so the match is by `COALESCE(root_stream_id, id)`, the same rule the mute filter uses.

### How it works

1. **Backend seed filter.** `boardArchivedExcludeSql(showArchived)` returns `AND NOT EXISTS (…)` that joins the anchor stream to its effective root and drops the row when `arch_root.archived_at IS NOT NULL`; it returns an empty fragment when `showArchived` is true. Spliced into `findByWorkspaceForViewer` alongside `hiddenCond`/`mutedCond`. Threaded through `ListWorkspaceConversationsOptions.showArchived` (service) and a `?archived=true` query flag on the handler, Zod-validated as `z.enum(["true","false"]).optional().transform(v => v === "true")` (INV-55) so absence means the default hide.

2. **Frontend seed query.** `?archived` is parsed in `board.tsx` (INV-59 — refresh/share reproduce the view), added to the `useWorkspaceConversations` query key + params, and sent on the wire by `conversationsApi.listByWorkspace` only when true (hide is the server default, so the wire stays clean).

3. **Client-side mirror.** `use-stable-board-view` gains an `archivedRootStreamIds` exclusion (a `ReadonlySet` of archived root ids) plus `showArchived`. `isExcluded` drops a post whose effective root is in that set unless `showArchived` — the read-side twin of `boardArchivedExcludeSql`. The set is resolved live in `board.tsx` from the workspace streams store (`streams.filter(s => s.archivedAt)`), so archiving a channel drops its cards the instant the row updates, without a refetch — exactly the behavior the mute exclusion established.

4. **Filter bar.** An "Archived" toggle (`board-filter-bar.tsx`) opts in; `?archived` is in `BOARD_FILTER_PARAMS`, so Clear-filters and the post-from-filtered-view reset sweep it.

### Key design decisions

**1. `showArchived` broadens, it isn't an archived-only view.** ON includes archived cards *alongside* active ones — the literal meaning of the flag and the inverse of the mute exclusion, not a separate "archive" screen. Because archived channels carry old `last_activity_at`, their cards sort to the bottom of the activity-desc feed; the client's `reconcileStableView` classifies them below the frozen floor and appends them silently (no disruptive "N new" pill), so toggling on adds them where they belong chronologically rather than jumping the view.

**2. Match by effective root, not anchor.** Archiving writes `archived_at` only on the root (`StreamRepository.update` in the archive path; a thread stays `active` and inherits — INV-62). Filtering on the anchor's own `archived_at` would miss every thread-anchored conversation under an archived channel. The SQL join and the client's `post.rootStreamId ?? streamId` lookup both resolve to the effective root, kept identical on purpose.

**3. `showArchived` lives in the exclusion state, not the filter/viewKey.** Like mute (and unlike lens/scope), toggling it doesn't reset the frozen view — the drop is instant and the appearance is graceful via the below-floor append. Keeping it out of `viewKey` avoids a jarring wholesale re-commit on toggle.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/conversations/repository.ts` | `boardArchivedExcludeSql` helper + `showArchived` option, spliced into the feed query |
| `apps/backend/src/features/conversations/service.ts` | `showArchived` on `ListWorkspaceConversationsOptions` (flows through the existing `{...options}` spread) |
| `apps/backend/src/features/conversations/handlers.ts` | `?archived` Zod flag + wire to the service option |
| `apps/frontend/src/api/conversations.ts` | `showArchived` param → `?archived=true` on the wire when true |
| `apps/frontend/src/hooks/use-conversations.ts` | `showArchived` in the query key + `listByWorkspace` params |
| `apps/frontend/src/hooks/use-stable-board-view.ts` | `archivedRootStreamIds` + `showArchived` on `BoardExclusionState`; `isExcluded` drop rule |
| `apps/frontend/src/pages/board.tsx` | parse `?archived`, resolve archived root ids from the streams store, wire query + exclusions + filter bar |
| `apps/frontend/src/components/board/board-filter-params.ts` | `BOARD_ARCHIVED_PARAM`/`BOARD_ARCHIVED_ON` + add to `BOARD_FILTER_PARAMS` |
| `apps/frontend/src/components/board/board-filter-bar.tsx` | "Archived" toggle button + props |
| `apps/frontend/src/hooks/use-stable-board-view.test.tsx` | `exclOf` helper; 3 new archived cases |

## Out of scope (deliberate)

- **Server-side archived filtering relies on the streams-store set only for the archived-*while-viewing* case.** A card whose root was already archived before the toggle is never seeded (the server excludes it by default), so it's only in IDB if it was archived while on screen — and in that case the streams store holds the row (the socket update set `archivedAt`). The client filter isn't a general "is any root archived" oracle; it's the live-drop twin, matching how mute works.
- **No archived-only lens, no per-card "archived" badge.** The toggle includes archived cards in the existing feed; surfacing *which* cards are archived is a separate UI concern.
- **No new preference.** `showArchived` is a transient URL toggle (like the other filter axes), not a saved default — that's the separate per-user-default-lens follow-up.

## Test plan

- [x] `apps/frontend` `use-stable-board-view` — 42 pass (3 new: drop under archived root, keep on opt-in, drop-a-committed-card-immediately)
- [x] `apps/frontend` `board.test` / `board-filter` / `board-saved-views` — 30 pass
- [x] `apps/frontend` `board-filter-params` / `use-conversations` — 27 pass
- [x] Pre-commit gate: lint (all apps), `tsc --noEmit` across every package/app, astro check, migration check, api-docs check — all clean
- [ ] Repo-level `boardArchivedExcludeSql` not integration-tested — no live Postgres in this env; the predicate mirrors the covered mute path exactly and passed all typechecks. Worth an eyeball on the auto-deployed `pr-*-staging.threa.io`.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01TCp9YRC9BC8HrtTub95ojN)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1226"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785964069&installation_id=129946197&pr_number=1226&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1226&signature=4a9679e7cb23dfcdecbf63e3f470e6ad51da36849f9dbdbd817b1e6285bc3c82"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/conversations/handlers.test.ts
+++ b/apps/backend/src/features/conversations/handlers.test.ts
@@ -261,6 +261,17 @@ describe("Conversation Handlers", () => {
       })
     })
 
+    test("threads showArchived: true through when ?archived=true", async () => {
+      await handlers.listByWorkspace(mockReq({ query: { archived: "true" } }), mockRes())
+      expect(mockListByWorkspace).toHaveBeenCalledWith("ws_1", "usr_1", expect.objectContaining({ showArchived: true }))
+    })
+
+    test("rejects a non-boolean ?archived with a 400", async () => {
+      await expect(handlers.listByWorkspace(mockReq({ query: { archived: "yes" } }), mockRes())).rejects.toMatchObject({
+        status: 400,
+      })
+    })
+
     test("passes cursor: undefined when none is supplied", async () => {
       await handlers.listByWorkspace(mockReq({ query: {} }), mockRes())
       expect(mockListByWorkspace).toHaveBeenCalledWith("ws_1", "usr_1", {
--- a/apps/backend/src/features/conversations/handlers.ts
+++ b/apps/backend/src/features/conversations/handlers.ts
@@ -57,6 +57,11 @@ const listWorkspaceConversationsSchema = listConversationsSchema.extend({
   excludeTypes: csvTypeListSchema.optional(),
   labels: csvIdListSchema(MAX_BOARD_SCOPE_LABELS, "labels").optional(),
   excludeLabels: csvIdListSchema(MAX_BOARD_SCOPE_LABELS, "excludeLabels").optional(),
+  // Absent = the board's default (hide archived). `?archived=true` opts in.
+  showArchived: z
+    .enum(["true", "false"])
+    .optional()
+    .transform((value) => value === "true"),
   cursor: z.string().min(1).optional(),
 })
 
@@ -148,6 +153,7 @@ export function createConversationHandlers({
         excludeStreamTypes: query.excludeTypes,
         scopeLabelIds: query.labels,
         excludeLabelIds: query.excludeLabels,
+        showArchived: query.showArchived,
         limit: query.limit,
         cursor: decodeBoardCursor(query.cursor),
       })
--- a/apps/backend/src/features/conversations/handlers.ts
+++ b/apps/backend/src/features/conversations/handlers.ts
@@ -57,11 +57,14 @@ const listWorkspaceConversationsSchema = listConversationsSchema.extend({
   excludeTypes: csvTypeListSchema.optional(),
   labels: csvIdListSchema(MAX_BOARD_SCOPE_LABELS, "labels").optional(),
   excludeLabels: csvIdListSchema(MAX_BOARD_SCOPE_LABELS, "excludeLabels").optional(),
-  // Absent = the board's default (hide archived). `?archived=true` opts in.
-  showArchived: z
+  // `?archived=true` opts into archived cards; absent = the board's default
+  // (hide archived), which the service reads as `?? false`. Kept
+  // undefined-when-absent (not coerced to `false`) so the options object stays
+  // uniform with the other optional axes.
+  archived: z
     .enum(["true", "false"])
     .optional()
-    .transform((value) => value === "true"),
+    .transform((value) => (value === undefined ? undefined : value === "true")),
   cursor: z.string().min(1).optional(),
 })
 
@@ -153,7 +156,7 @@ export function createConversationHandlers({
         excludeStreamTypes: query.excludeTypes,
         scopeLabelIds: query.labels,
         excludeLabelIds: query.excludeLabels,
-        showArchived: query.showArchived,
+        showArchived: query.archived,
         limit: query.limit,
         cursor: decodeBoardCursor(query.cursor),
       })
--- a/apps/backend/src/features/conversations/repository.ts
+++ b/apps/backend/src/features/conversations/repository.ts
@@ -193,9 +193,10 @@ function boardHiddenExcludeSql(userId: string) {
  * marks only the root row — a thread inherits its root's lifecycle (INV-62) — so
  * a conversation is archived exactly when its effective root is. Skipped when
  * `showArchived` is true (the viewer opted into seeing archived cards);
- * otherwise the board's default hides them. Mirrored client-side in
- * `use-stable-board-view` (the `archivedRootStreamIds` exclusion) — keep the
- * effective-root rule identical.
+ * otherwise the board's default hides them. The projection stamps each surviving
+ * card's effective-root archived state onto `BoardPost.rootArchived`, which the
+ * client's `use-stable-board-view` reads to re-hide archived cards the instant
+ * the viewer toggles archived back off — keep the effective-root rule identical.
  */
 function boardArchivedExcludeSql(showArchived: boolean) {
   if (showArchived) return sql``
--- a/apps/backend/src/features/conversations/repository.ts
+++ b/apps/backend/src/features/conversations/repository.ts
@@ -188,6 +188,27 @@ function boardHiddenExcludeSql(userId: string) {
 }
 
 /**
+ * Archived-stream exclusion, matched by effective ROOT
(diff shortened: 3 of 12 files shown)


kristofferremback:
### Code review

Two rounds. Round 1 (against `13b68c5`) surfaced 3 real issues; all fixed in `d42dd18`. Round 2 (against `d42dd18`) is clean across every lens.

**Round 1 findings — all resolved:**

1. `apps/backend/src/features/conversations/handlers.ts` — **wire param mismatch**: the flag was schema-named `showArchived` while the client sends `?archived=true`, so the backend read `query.showArchived` (always undefined) and ignored the toggle. Also broke two exact-object handler tests (CI red). Fixed: schema field renamed to `archived`, mapped to the `showArchived` service option; added handler coverage (opt-in threads through, non-boolean 400s).
2. `apps/frontend/src/hooks/use-stable-board-view.ts` — **stale archived cards after toggle-off**: the client mirror derived archived roots from the workspace streams store, but the bootstrap (`listWithPreviews`) seeds active streams only, so a stream archived before the session was never in the store — cards seeded under `?archived=true` kept rendering after toggling archived off (they persist in IDB; `bulkPut` never deletes). Fixed with a per-card `BoardPost.rootArchived` flag stamped by the projection from the effective root's `archived_at` — authoritative, survives in IDB, joins `hasCapturedMemo`/`isMine` as a per-fetch board field.
3. `apps/frontend/src/pages/board.tsx` — **`showArchived` wasn't a first-class filter axis**: omitted from the scroll-anchor reset key every other dimension feeds, and a silent exclusion the view never reacted to. Fixed by moving it onto `BoardViewFilter` so it rides the stable-view `viewKey` + scroll reset — toggling now re-commits the feed and scroll-resets like lens/scope.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Param `archived` consistent client↔server; `rootArchived` populated on both the feed and single-post (`getBoardPost`) projection paths; `archived` swept by Clear-filters and roundtripped by saved views.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Toggle on→off→on sequences traced; every archived IDB row carries `rootArchived=true` (only ever seeded under `?archived=true`), so `isExcluded` drops them on toggle-off. `rootArchived` preserved through `toCached`/`mergeBoardConversation`/optimistic writes (all spread the existing row).</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>`rootArchived` round-trips projection → wire → IDB → `isExcluded` with no write path dropping it. Mid-session-archive staleness is the documented per-fetch board-field model (same as `hasCapturedMemo`/`isMine`).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-62 (archiving marks root only), INV-55 (Zod), INV-59 (URL-derived view), INV-21/63 (toggle no-shift, no toast) all upheld.</details>
<details><summary>🏗️ Design [CLEAN]</summary>SQL/JS lockstep held (both key on the effective root's archived state). Single-source per-card flag replaced the unreliable streams-store set.</details>
<details><summary>🔒 Security [CLEAN]</summary>`boardArchivedExcludeSql` is parameterized template SQL, workspace-scoped inside the NOT EXISTS.</details>
<details><summary>🎛️ UX [CLEAN]</summary>Toggle feeds both viewKey + scroll reset (parity with other filters); button `outline`→`secondary` + `aria-pressed` gives immediate confirmation; correct ARIA toggle-button.</details>
<details><summary>📱 Mobile [CLEAN]</summary>Toggle matches the sibling filter controls' `h-7 shrink-0 rounded-full` shell; direct-action button, no hover-only affordance, no mobile branches introduced.</details>

---

PR #1229, opened 2026-07-07
fix(memos): judge memo-worthiness by durability + agency, not topic

kristofferremback:
## Problem

Memo generation between two DM participants had been producing mostly junk: of 155 active memos in one production DM, 111 were `context`/`learning`, and a large share were things the prompts *already* claimed to exclude — transient status ("View run är trasig", "mirror fungerar nu") that is false within the hour, and reactions/opinions about AI models ("4.7 känns inte särskilt speciell"). Two distinct issues sat behind the "weird memos" report; this PR addresses the ongoing one (the quality of what gets captured). The one-time July-5 wave of memos sourced from months-old messages was diagnosed separately as a deploy-race that has self-healed (the staleness-sweep guard in `accumulator-outbox-handler.ts:47` is now universally deployed) — out of scope here.

The instinct to fix this by forbidding topics (e.g. "model talk is not knowledge") is wrong: between two people who *build* with a model, "Fable interprets unclear answers as a green light" is a genuine validated learning. Relevance is not a property of the subject.

## Solution

Move the worthiness test from *topic* to *intent*, on two topic-neutral axes, and give the classifier the temporal context it needs to apply one of them.

### How it works

- **Durability + agency, in the prompts** (`memos/config.ts`). The classifier and memorizer now judge on two axes that turn on what the participants *produced*, never on the subject:
  - **Durability** — would the core still be true/useful in six months? A decision, a worked-out procedure, a validated finding lasts; a passing state ("it's broken right now") is stale within the hour.
  - **Agency** — did they produce it (decide, work out, validate) or just react to something outside their control? Reactions fail; the *same subject* passes the moment it becomes a decision or a validated learning. Stated explicitly as topic-neutral in both prompts so "model talk" isn't blanket-banned.
- **Transient status** is named as a non-memo in both prompts, with the counter-example inline (the *fix* for that same bug is worthy — judge what was produced, not the subject).
- **`context` sharpened** — it was undefined to the model and had become the catch-all absorbing transient state. The memorizer prompt now defines it as durable "why it is this way" background, to be reached for **last**.
- **Temporal grounding as a signal, not a cutoff** — the memo message rendering gains a relative `age` attribute (e.g. `age="3 days ago"`) via `formatRelativeDate` (`lib/temporal.ts:211`, already used by the researcher agent). `MemoService.processBatch` passes one `now` per batch so both the classifier and memorizer can weigh whether captured state is still live. The classifier previously got *no* "now" reference at all. This is deliberately not a recency gate — a conversation resumed after a vacation still memoizes.

### Key design decisions

**1. Topic-neutral intent test, not a topic blocklist** — per the reviewing user's ruling: forbidding subjects (model talk, news) is over-indexing on one workspace's noise. A validated learning about a tool you build with is real knowledge; the discriminator is durability + agency. Also avoids the language/keyword-heuristic trap (INV-54).

**2. Temporal as guidance, not a hard cutoff** — production data showed no continuous old-message drift to gate against (every memo outside the July-5 deploy-race was created within ~14 days of its source). A hard recency floor would wrongly drop conversations that legitimately resume after a vacation/long weekend. So the age is an input the model weighs, not a filter.

**3. One renderer for production and evals (no duplication)** — `renderMessagesXml` in `message-formatter.ts` is the single source of truth for the `<messages>` wire format. Production `MessageFormatter.formatMessages` resolves author names from the DB then renders through it; the eval fixtures already have names and render through the same function. Only the DB-bound name resolution is skipped in evals — never the rendering — so the prompt evals exercise can't drift from production (INV-35/37, INV-45).

### Design evolution (reviewer feedback)

The first version had `formatEvalMessages` hand-duplicate the `<message>` wire format, and the new `age` attribute was added to *both* renderers with a "keep in sync" comment. A reviewer flagged this as embedding application code in the test instead of testing it — the two would drift. Fixed by extracting `renderMessagesXml` (decision 3 above); the duplicate renderer and the fixture's separate `escapeXml` are gone, and the eval now escapes content/attributes exactly as production does.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/memos/config.ts` | Classifier + memorizer prompts reframed to durability/agency; transient-status excluded; `context` sharpened; `age` attribute explained |
| `apps/backend/src/lib/ai/message-formatter.ts` | Extract `renderMessagesXml` (shared wire-format renderer) + `relativeTo` → `age` attribute; `formatMessages` renders through it; drop private `formatSingleMessage` |
| `apps/backend/src/features/memos/service.ts` | `processBatch` passes one `relativeTo: now` per batch into the pre-format |
| `apps/backend/src/lib/ai/message-formatter.test.ts` | +2 tests: age omitted by default, rendered when `relativeTo` set |
| `apps/backend/evals/fixtures/memo.ts` | `formatEvalMessages` now calls the shared `renderMessagesXml` (was a hand-rolled copy); dead `escapeXml` removed (INV-45) |
| `apps/backend/evals/suites/memo-classifier/{cases,types}.ts` | `transient-status-not-worthy-001`, `tool-behavior-learning-001` (worthy model-talk contrast); `transient` category |
| `apps/backend/evals/suites/memorizer/{cases,types}.ts` | `transient-status-yields-nothing-001`, `tool-behavior-learning-captured-001`; `transient` category |

## Out of scope (deliberate)

- **The July-5 deploy-race pollution** (39 memos from stale March/April messages, one DM). Self-healed root cause; cleanup is a separate data write the user opted to defer.
- **No pipeline/schema change.** The `age` attribute is additive; the message-count split on `<message` is unaffected. Memo dedup/supersession/queueing untouched.

## Test plan

- [x] `bun test src/lib/ai/message-formatter` — 24 pass (incl. 2 new age tests; all pass unchanged after the `renderMessagesXml` extraction)
- [x] `bun test src/features/memos` — 54 pass
- [x] `bunx tsc --noEmit -p apps/backend/tsconfig.json` (incl. `evals/**`) — clean
- [ ] Eval suites **blocked in CI** — the `/eval` workflow crashes at startup because the `OPENROUTER_API_KEY` Actions secret (slash-commands.yml:145) is unset. Not a code failure. Once the secret is added, run `/eval -s memo-classifier` and `/eval -s memorizer`; success criterion: `garbage-leak-rate` / `over-capture-rate` stay at 0 on the new transient cases, and the knowledge cases (`tool-behavior-learning-*`, `procedure-setup-001`, `decision-measured-001`) pass.

---

### DIFF
--- a/apps/backend/evals/fixtures/memo.ts
+++ b/apps/backend/evals/fixtures/memo.ts
@@ -8,6 +8,7 @@ import { ulid } from "ulid"
 import type { Memo } from "../../src/features/memos"
 import type { Conversation } from "../../src/features/conversations"
 import { MemoStatuses, MemoTypes, ConversationStatuses } from "@threa/types"
+import { formatRelativeDate } from "../../src/lib/temporal"
 import type { EvalClassifierMessage } from "../suites/memo-classifier/types"
 
 function escapeXml(text: string): string {
@@ -15,15 +16,18 @@ function escapeXml(text: string): string {
 }
 
 /**
- * Render messages exactly as MessageFormatter.formatMessages does in
- * production (the classifier counts messages by splitting on "<message"), but
- * from eval fixtures instead of DB rows.
+ * Render messages exactly as MessageFormatter.formatMessages does in production
+ * with `{ includeIds: true, relativeTo: now }` (the classifier counts messages by
+ * splitting on "<message"), but from eval fixtures instead of DB rows. Keep the
+ * attribute shape — including the relative `age` — in step with
+ * MessageFormatter.formatSingleMessage so evals exercise the real prompt (INV-45).
  */
 export function formatEvalMessages(messages: EvalClassifierMessage[], now: Date): string {
   const rendered = messages.map((m, i) => {
     const id = m.id ?? `msg_${ulid()}`
     const createdAt = new Date(now.getTime() - (m.minutesAgo ?? (messages.length - i) * 2) * 60_000)
-    return `<message id="${id}" authorType="${m.authorType}" authorId="${m.authorId}" authorName="${escapeXml(m.authorName)}" createdAt="${createdAt.toISOString()}">${escapeXml(m.contentMarkdown)}</message>`
+    const age = escapeXml(formatRelativeDate(createdAt, now))
+    return `<message id="${id}" authorType="${m.authorType}" authorId="${m.authorId}" authorName="${escapeXml(m.authorName)}" createdAt="${createdAt.toISOString()}" age="${age}">${escapeXml(m.contentMarkdown)}</message>`
   })
   return `<messages>\n${rendered.join("\n")}\n</messages>`
 }
--- a/apps/backend/evals/fixtures/memo.ts
+++ b/apps/backend/evals/fixtures/memo.ts
@@ -8,28 +8,28 @@ import { ulid } from "ulid"
 import type { Memo } from "../../src/features/memos"
 import type { Conversation } from "../../src/features/conversations"
 import { MemoStatuses, MemoTypes, ConversationStatuses } from "@threa/types"
-import { formatRelativeDate } from "../../src/lib/temporal"
+import { renderMessagesXml } from "../../src/lib/ai/message-formatter"
 import type { EvalClassifierMessage } from "../suites/memo-classifier/types"
 
-function escapeXml(text: string): string {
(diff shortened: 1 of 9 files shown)


kristofferremback:
/eval -s memo-classifier

---

kristofferremback:
/eval -s memorizer

---

kristofferremback:
/eval -s memo-classifier

---

kristofferremback:
/eval -s memorizer

---

kristofferremback:
/eval -s memorizer -r 5

---

PR #1190, opened 2026-07-05
feat(agents): Stop/Redirect affordances on the activity card (roadmap 2.2)

kristofferremback:
**Roadmap:** [`docs/plans/ariadne-collaborator-roadmap.md` §2.2](https://github.com/threahq/threa/blob/main/docs/plans/ariadne-collaborator-roadmap.md) — Stop/Redirect affordances on the activity card. Frontend half of #1177 (roadmap 2.1, generalized session abort).

## Problem

Since #1177 the backend gracefully halts **any** running companion session on `agent_session:research:abort` — but the frontend still gates its Stop affordance on the old research-only model. `agent-session-event.tsx` renders `StopResearchButton` only while `sessionCanAbort` says the current step is a research type (`isAbortableStepType` over `workspace_search`/`research`), so the button pops in and out mid-run as steps change, is labelled "Stop research" even though it now stops the whole session, and never appears during `thinking`/`web_search`/`message_sent` steps despite Stop working there. The same stale gate sits on the trace dialog's in-flight step header. And there is no affordance for the other half of "jump in and steer": the runtime already folds mid-run user messages into a running session (`NewMessageAwareness` → `reconsidering`), but nothing in the UI tells the user that typing while the agent works is a supported, useful thing to do.

## Solution

Gate the card's actions on the one thing that matters — "is this session running" — and make steering legible with a Redirect button that is pure frontend.

### How it works

1. **Gating collapses onto the card's own derived status.** `AgentSessionEvent` already derives `running | completed | failed | deleted` from its lifecycle events; the Redirect/Stop pair renders whenever that status is `running`. The per-session `sessionCanAbort` map (built from `isAbortableStepType(activity.currentStepType)` in both `event-list.tsx` and `stream-content.tsx`, threaded through `TimelineItemRenderContext` and the row memo comparator) is deleted end-to-end. The buttons are stable for the entire run — nothing pops in or out as steps change (INV-21).
2. **Stop** calls `onStopSession(sessionId)`, threaded from `EventList`/`StreamContent` where `useAbortSession` (renamed from `useAbortResearch`, INV-49) emits the unchanged wire event `agent_session:research:abort`. The backend routes it to the in-process abort registry or the enclave's `requestAbort` flag (#1177); any stream member with access can stop.
[...]

PR #1231, opened 2026-07-07
fix(board): thread Stop + in-place Redirect onto the running-session card

kristofferremback:
## Problem

The board's running-session trace card was inert. `board-row-item.tsx` rendered `<AgentSessionEvent events={row.events} />` with no `onStopSession` and no redirect wiring, so on the board (and the conversation panel, which shares the same row component):

- **Stop was hidden.** `AgentSessionEvent` gates the Stop button on `onStopSession && …` — with the prop absent, it never rendered.
- **Redirect was a no-op.** Its handler walks `closest("[data-editor-zone]")` for a composer to focus; the board card has no `data-editor-zone`, so `findVisibleZoneEditor` returned null and the click did nothing.

The timeline card gets both because `event-list.tsx` / `stream-content.tsx` build `handleStopSession` from `useAbortSession(useSocket())` and thread it in, and the main/thread views wrap their composer in `data-editor-zone`. The board never did — so "jump in and steer" (roadmap 2.2, #1190) worked in the timeline but not on the board.

## Solution

Thread the same Stop path the timeline uses into the board's shared row component, and give `AgentSessionEvent` a redirect override for surfaces whose composer is lazily mounted (the board card's reply composer is collapsed to a resting affordance until opened, so there is no editor in the DOM to walk to).

### How it works

1. **Stop** — `BoardEventRowItem` (`board-row-item.tsx`) now sources `useAbortSession(useSocket())` and passes `onStopSession={(sessionId) => abortSession({ sessionId, workspaceId })}` into `AgentSessionEvent`. `useSocket()` reads a default-`null` context, so no provider is required in tests; `useAbortSession(null)` no-ops. This fixes the board **and** the conversation panel — both route session rows through `BoardEventRowItem` and both previously lacked Stop.
2. **Redirect override** — `AgentSessionEvent` gains an optional `onRedirect?: () => void`. When present, `handleRedirect` calls it (the surface owns opening + focusing its composer) instead of the DOM walk; when absent, the existing `[data-editor-zone]` walk runs unchanged (timeline, thread panel). Either branch shows the same "fold your message into the current work" hint for `REDIRECT_HINT_MS` (5s).
3. **Board card opens in place** — `board-card.tsx` holds an `openReplySignal` nonce and passes `onRedirectSession={() => setOpenReplySignal(n => n + 1)}` through `BranchedBoardRows` → `BoardEventRowItem` → `AgentSessionEvent.onRedirect`. The bump drives `BoardReplyComposer`'s existing `openReplySignal` prop, which expands the resting affordance into the live composer and focuses it (`InlineComposerForm` defaults `autoFocus`). The running session then absorbs the next message via the runtime's existing `NewMessageAwareness` fold-in — no backend call.

## Key design decisions

**1. Redirect opens the card composer in place, not navigate-away.** Chosen by the user (a-vs-b call). The board card already has a `BoardReplyComposer`; the fix is to open+focus it via the `openReplySignal` primitive already used by the conversation panel, so "fold into current work" stays literally true without yanking the user out of the board overview. Navigating to the stream/thread was the alternative — simpler and always-works, but breaks the in-place steer.

**2. `onRedirect` override rather than forcing a `data-editor-zone` onto the board card.** The card's composer is collapsed until tapped, so even with a `data-editor-zone` wrapper `findVisibleZoneEditor` would find no mounted editor. The override lets the surface open its own composer; the DOM-walk default is kept for timeline/thread where the composer is always mounted.

**3. Stop wired in the shared `BoardEventRowItem`, so the conversation panel gets it too.** One wiring site fixes both board surfaces (INV-35). The panel keeps its existing `data-editor-zone="panel"` DOM-walk Redirect (works when its composer is open) — no `onRedirectSession` threaded there, no regression.

Both actions stay buttons with `stopPropagation` inside the card's trace `<Link>` (INV-40); the running card's right slot already reserves their footprint (INV-21); no `toast.success` anywhere (INV-63).

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/agent-session-event.tsx` | Add optional `onRedirect`; `handleRedirect` uses it when present, else the `[data-editor-zone]` walk |
| `apps/frontend/src/components/board/board-row-item.tsx` | `BoardEventRowItem` sources `useAbortSession(useSocket())`, threads `onStopSession` + `onRedirect` into the session row; accepts `onRedirectSession` |
| `apps/frontend/src/components/board/branch-rows.tsx` | Thread `onRedirectSession` through `BranchedBoardRowsProps` to the event row |
| `apps/frontend/src/components/board/board-card.tsx` | `openReplySignal` nonce + `openReplyComposer`; wire to both `BranchedBoardRows` and `BoardReplyComposer` |
| `apps/frontend/src/components/timeline/agent-session-event.test.tsx` | New: Redirect calls `onRedirect` (board) instead of walking the DOM, still shows the hint |
| `apps/frontend/src/components/board/board-card.test.tsx` | New: running-session card shows Stop + Redirect; Redirect opens the card composer in place (stubs `InlineComposerForm`) |

## Out of scope (deliberate)

- **Conversation panel Redirect is unchanged** — it keeps the DOM-walk against its `data-editor-zone="panel"` (works when its composer is open). Only Stop was newly threaded there. Making the panel also open-on-Redirect would mean threading its existing `openReplySignal` setter down; not needed for the board ask.
- **Live counts / substep on board session cards** — the board still renders `AgentSessionEvent` without `liveCounts`/`liveSubstep` (unchanged); this PR is Stop/Redirect only.

## Test plan

- [x] `bun run --cwd apps/frontend typecheck` — clean (`tsc --noEmit`)
- [x] `agent-session-event.test.tsx` — 9 pass (1 new: `onRedirect` path)
- [x] `board-card.test.tsx` — 13 pass (1 new: running-session Stop/Redirect + open-in-place)
- [x] `src/components/board/`, `conversation-panel.test.tsx`, `src/components/trace/`, `event-list.test.tsx` — 116 pass together, no regressions
- [x] `eslint` on all six changed files — clean
- [x] Full pre-commit monorepo typecheck (14 workspaces) + lint + api-docs `--check` — passed; no `openapi.json` change (frontend-only)
- [ ] `/code-review` loop to 7/7 — running now (per request); findings will be pushed to this branch and noted here

---

---
_Generated by [Claude Code](https://claude.ai/code/session_0132xdweS3K2DQt8PkcMP1Uo)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1231"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785998221&installation_id=129946197&pr_number=1231&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1231&signature=ffc8e123e52ed74c1b1b6b31b43ac3e5c4fafcefd1ac9b447a4c386a7536c720"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/board/board-card.test.tsx
+++ b/apps/frontend/src/components/board/board-card.test.tsx
@@ -22,6 +22,8 @@ import * as userProfileModule from "@/components/user-profile"
 import * as syncEngineModule from "@/sync/sync-engine"
 import * as contextsModule from "@/contexts"
 import * as queueDraftModule from "@/hooks/use-queue-draft-message"
+import * as inlineComposerModule from "@/components/board/board-inline-composer"
+import { spyOnExport } from "@/test/spy"
 
 const WS = "ws_1"
 const STREAM = "stream_1"
@@ -282,6 +284,43 @@ describe("BoardCard agent activity", () => {
     expect(await screen.findByText("Interrupted, retrying…")).toBeTruthy()
     expect(screen.queryByText("Session failed")).toBeNull()
     expect(await screen.findByText(/2 steps/)).toBeTruthy()
+  })
+
+  it("shows Stop + Redirect on a running session, and Redirect opens the card composer in place", async () => {
+    const user = userEvent.setup()
+    // A running session's row mounts the live progress rail (useAgentActivity),
+    // which reads the socket — give it a fake one.
+    const { socket } = fakeSocket()
+    vi.spyOn(contextsModule, "useSocket").mockReturnValue(socket)
+    // Prevent the real editor mounting when Redirect opens the composer — the
+    // wiring under test is Redirect → open card composer, not editor mechanics.
+    spyOnExport(inlineComposerModule, "InlineComposerForm").mockReturnValue(((props: { placeholder: string }) => (
+      <div data-testid="reply-composer-open">{props.placeholder}</div>
+    )) as never)
+
+    // A running session (started, no terminal event) triggered by the card's
+    // opening message: the board card must expose the same Stop/Redirect pair the
+    // timeline card does.
+    await db.events.bulkPut([
+      sessionEvent("agent_session:started", 30, {
+        sessionId: "sess_run",
+        triggerMessageId: "m_open",
+        personaName: "Ariadne",
+      }),
+    ])
+    mountCard()
+
+    expect(await screen.findByRole("button", { name: "Stop" })).toBeTruthy()
+    expect(screen.getByRole("button", { name: "Redirect" })).toBeTruthy()
+    // Composer starts collapsed to its resting affordance.
+    expect(screen.queryByTestId("reply-composer-open")).toBeNull()
+
+    await user.click(screen.getByRole("button", { name: "Redirect" }))
+
+    // Redirect opened the card's own composer in place (no navigation) and the
+    // fold-in hint confirms the running session will absorb the next message.
+    expect(await screen.findByTestId("reply-composer-open")).toBeTruthy()
+    expect(screen.getByText("Ariadne will fold your message into the current work")).toBeTruthy()
   })
 
   it("does NOT render a session whose invoking message is not a conversation member", async () => {
--- a/apps/frontend/src/components/board/board-card.tsx
+++ b/apps/frontend/src/components/board/board-card.tsx
@@ -77,6 +77,12 @@ export function BoardCard({ workspaceId, post, contextLabel, streamType }: Board
   const conversationService = useConversationService()
   const { openPanel, getPanelUrl } = usePanel()
   const [expanded, setExpanded] = useState(false)
+  // A running session's Redirect bumps this nonce to expand + focus the card's
+  // reply composer in place (its editor is collapsed until opened, so there's no
+  // DOM zone to walk to). A counter, not a boolean, so a repeat Redirect re-opens
+  // it after a manual collapse.
+  const [openReplySignal, setOpenReplySignal] = useState(0)
+  const openReplyComposer = useCallback(() => setOpenReplySignal((n) => n + 1), [])
   // Scopes text-selection quoting to this card so a selection routes into this
   // card's composer, not a sibling card's provider.
   const cardRef = useRef<HTMLDivElement>(null)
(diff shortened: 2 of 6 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: Opus | Reviewers: sonnet x4 (spec+design, correctness+data-flow, UX, mobile)

No blocking issues found. Checked CLAUDE.md compliance, plan/intent adherence, design quality, correctness & data-flow lifecycle, security, UX, and mobile/responsive.

The change reuses the existing `openReplySignal` nonce on `BoardReplyComposer` (the same primitive `conversation-panel.tsx` already drives) rather than a parallel path (INV-35); both `BranchedBoardRows` call sites and the composer are wired; `BoardEventRowItem` sourcing `useAbortSession(useSocket())` fixes Stop on the board **and** the conversation panel; the `onRedirect` override branches cleanly and the conversation panel is left on the DOM-walk path as intended. No `toast.success` (INV-63), no navigation-as-button (INV-40), no layout-shift change (INV-21).

<details><summary>🎛️ UX — 2 sub-threshold observations (below the ≥80 bar, not blocking)</summary>

Both are properties of the reused `openReplySignal` primitive / board layout, not regressions this diff introduces:

1. `apps/frontend/src/components/board/board-reply-composer.tsx` — a repeat Redirect while the card composer is **already open** doesn't re-focus it: `setOpen(true)` is a no-op when already `true`, so `InlineComposerForm` never remounts and its mount-time `autoFocus` doesn't re-fire. The first Redirect (collapsed→open) focuses correctly. Same limitation the conversation panel's "Reply in conversation" already ships with.
2. `apps/frontend/src/components/board/board-card.tsx` — on a very long card the composer can sit below the fold, so focusing it may scroll the 5s "fold your message into the current work" hint out of view. Mitigated in practice: a running-session row sorts to the card **tail**, adjacent to the composer, so both are usually visible together.

</details>

---

PR #1230, opened 2026-07-07
feat(messages): collapse long individual messages, user-tunable threshold

kristofferremback:
## Problem

A single very long message dominates the feed. The stream timeline and board cards already fold their *sub-blocks* — code blocks, blockquotes, quote replies, stream descriptions (each with its own measure→clamp→fade→toggle) — and the board folds a whole *card* by pixel height. But an individual message body of prose has no fold: a wall of text pushes everything else off-screen in the timeline, in the conversation panel, on the label page, and inside an expanded board card.

## Solution

Fold a long message body behind a **Show more/less** toggle past a rendered line count, reusing the exact machinery the stream-description fold already uses. Threshold is a new user preference, `messageCollapseThreshold` (default **16** lines), sitting alongside the existing code/blockquote/board-card thresholds.

### How it works

1. **Reusable `CollapsibleBody`** (`apps/frontend/src/lib/markdown/collapsible-body.tsx`) — extracted verbatim from `DescriptionSetEvent`'s private `DescriptionBody`: `useMeasuredLineCount` (rendered height ÷ line-height, `null` until real layout) → `collapsible = lineCount > threshold + 0.5` → `useBlockCollapse` for the persisted per-message toggle → CSS `max-height` clamp + bottom fade + chevron toggle. It marks its subtree via `InsideCollapsibleBlockProvider active={canToggle}` so a *folded* message hides its inner code/quote chrome and folds as one unit; a short (non-collapsible) message leaves inner blocks folding individually, exactly as today. `DescriptionBody` now delegates to it — no behavior change to descriptions.
2. **Both message render sites wrap their body** in `<MarkdownBlockProvider messageId={…}><CollapsibleBody kind="message" …>`:
   - `message-event.tsx` `MessageLayout` — the **stream timeline** row.
   - `message-item.tsx` `richBody` — the shared **board card / conversation panel / label page** row.
   The `MarkdownBlockProvider` (previously created implicitly by passing `messageId` to `MarkdownContent`) is now explicit at the fold boundary so `useBlockCollapse` reads the same per-message scope; the `messageId` prop is dropped from the inner `MarkdownContent` to avoid a redundant nested provider. Attachments, link previews, memo/giphy embeds, and reactions stay **outside** the fold — only the text body clamps.
3. **`useMessageCollapseThreshold`** reads the preference via `usePreferencesOptional` (safe in tests/no-provider), falling back to the default.
4. **New preference plumbing** — key-value override store, so **no migration**:
[...]

PR #1234, opened 2026-07-07
perf(board): virtualize the feed with virtua

kristofferremback:
## Problem

The production board is unusable-laggy. The workspace board renders every
conversation card at once — the floor measurement put a realistic Active board at
**~430 cards** (`docs/board-view-design.md`) — and each `BoardCard` is heavy: it
mounts several ref-counted Dexie `liveQuery` rails (`useBoardCardMessages` →
`useMergedStreamRail`), two `IntersectionObserver`s (viewport + sticky-header
sentinel), a `ResizeObserver` (tall-measure), and renders full message bodies.
With no windowing, all ~430 mount and re-render on every IDB tick. `virtua`
adoption for the board was the deferred follow-up the design doc flagged
("§ Stable view … Virtualization … must be designed with the virtualizer, not
retrofitted").

## Solution

Virtualize the feed with `virtua`, mounting only the on-screen cards. The board
becomes an owned overflow scroller driving a single `<Virtualizer>` over one flat
row list (recency headers, cards, load-more), exactly the pattern the timeline
already uses (`stream-content.tsx` / `use-timeline-scroll.ts`).

### How it works

1. **Owned scroller + one flat list.** `board.tsx` replaces the Radix
   `ScrollArea` and the per-section non-virtualized render with a plain
   `overflow-y-auto` div (`overflowAnchor: none`) whose ref is handed to
   `<Virtualizer scrollRef=…>`. The recency sections are flattened into one
   `feedRows` array — `{ kind: "header" | "card" | "load-more" }` — so the whole
   board is a single virtualizer, not one per section (a section can hold hundreds
   of cards). Inter-card spacing (`pb-3`) and section separation (`pt-6` on
   non-first headers) live on the rows, since a flat list has no per-section
   flex-gap.
2. **Composer via `startMargin`.** The composer stays a non-virtualized element
   above the rows in the same scroller (so its open/draft state survives a scroll
   away and back). A `ResizeObserver` on its wrapper feeds its height to
   `<Virtualizer startMargin=…>` so item offsets stay aligned as it opens/grows —
   the same device the timeline uses for its top spacer.
3. **`BoardFeedList` seam.** The `<Virtualizer>` is wrapped in a new plain
   function component so tests can swap it for a passthrough via namespace `spyOn`
   (INV-48) — `virtua` renders **zero** items under jsdom's zero-height, no-op-
   `ResizeObserver` layout, so the full-page integration tests need the seam.
4. **Scroll-anchor handled by `virtua`.** `use-board-scroll-anchor.ts` is deleted
   (INV-38): `virtua` maintains scroll position across item measurement and
   above-fold async reflow (avatars/images loading above the viewport), which is
   exactly what the hand-rolled `ResizeObserver`-based anchor did. Off-screen cards
   are unmounted, so they can't reflow while off-screen in the first place.
5. **Reset + reveal.** Changing lens/scope scrolls to the top pre-paint (a layout
   effect keyed on the filter string) so a shorter subset doesn't land past its
   end; the "N new" pill's `revealNew` sets `scrollTop = 0` then `commit()`.

### Key design decisions

**1. No `shift`.** `virtua`'s `shift` maintains position from the *end* on
prepend (the timeline's reverse-infinite-scroll case). The board never prepends
into the rendered list: new cards wait behind the "N new" pill and only land via
`revealNew` (scroll-to-top → commit), and "Load more" appends older pages below
the committed floor (`reconcileStableView`). So `shift` is omitted; content growth
of an existing card is handled by `virtua`'s per-item resize anchoring.

**2. Pill → viewport-top overlay (behavior change).** The "N new" pill moves from
an in-flow `sticky top-2` element to an `absolute` overlay at the top of the
scroll viewport. In-flow sticky can't persist across a virtualized feed — a
sticky element only sticks within its parent's box, and the only parent that spans
the feed is the one whose child is the virtualizer container, which fights clean
`startMargin` measurement. The overlay is the standard "new content" pill pattern
(X/Slack) and keeps it reachable at any scroll depth. Trade-off: at `scrollTop 0`
the pill floats over the collapsed composer button.

**3. `virtua` isolated behind `BoardFeedList`, not mocked with `vi.mock`.** INV-48
forbids `vi.mock`; `virtua`'s `Virtualizer` is a `forwardRef` object, not a
spyable function. A plain-function wrapper is spyable via the sanctioned namespace-
`spyOn` pattern the board tests already use for `useBoardPosts`, and it doubles as
the module boundary for the virtualization detail.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/board/board-feed-list.tsx` | The virtualization boundary — a plain function wrapping `<Virtualizer>`; the single seam tests replace with a passthrough |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/pages/board.tsx` | Radix `ScrollArea` + per-section render → owned scroller + `<BoardFeedList>` over a flat `feedRows` list; composer `startMargin`; filter-keyed scroll-to-top; pill → absolute overlay |
| `apps/frontend/src/components/board/board-card.tsx` | Sticky-header `IntersectionObserver` root now matches `[data-board-scroll-viewport]` (the owned scroller) or a Radix viewport |
| `apps/frontend/src/pages/board.test.tsx` | `spyOn(BoardFeedList)` passthrough (jsdom renders no `virtua` items); recency-sections test asserts document order instead of `.closest("section")` |

## Deleted files

| File | Reason |
| ---- | ------ |
| `apps/frontend/src/hooks/use-board-scroll-anchor.ts` | `virtua` maintains scroll position across measurement + above-fold reflow; the hand-rolled anchor (and its only consumer, board.tsx) is now dead (INV-38) |

## Out of scope (deliberate)

- **Load-more bugs 11/12** (per-card "N more" middle-gap read as a prepend; the
  load-more control's stacking vs the sticky header) — same scroll path, now
  easier to fix on the flat list, but a separate change.
- **`board-card.tsx` internals** — only the observer-root selector changed. The
  card's own layout/collapse/sticky-header markup is untouched, so this stays clear
  of the in-flight #1230 (`MessageItem` fade sites) and #1231 (`openReplySignal`).
- **No backend stable-paging key** — still the deferred follow-up
  (`reconcileStableView`'s per-page client cursor is unchanged).

## Test plan

- [x] `bunx vitest run src/pages/board.test.tsx` — 15 pass (passthrough seam; recency test rewritten to document-order)
- [x] `bunx vitest run src/components/board src/hooks/use-stable-board-view.test.tsx src/hooks/use-board-card-messages.test.tsx src/hooks/use-board-stream-subscriptions.test.tsx src/lib/board` — 162 pass
- [x] Full pre-commit gate on commit: prettier + all-app eslint + `tsc --noEmit` across all 14 workspaces + dockerfile/migration checks + OpenAPI `--check` — clean
- [x] **Browser-verified in Chromium** (the two flagged risks): windowing unmounts off-screen cards (top: cards 0–1 mounted; after scrolling: 10–12, card 0 gone); `position: sticky` header pins to the scroller top inside `virtua`'s `contain: layout` items (`cardTop: -149` while `headerTop: 1`)
- [ ] Not driven against the live prod board (no auth/data harness in-session) — the windowing + sticky mechanics are verified against real `virtua` in a standalone Chromium harness instead

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ed75QksFcXxLsm6KLNVTFh)_

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1234"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786018654&installation_id=129946197&pr_number=1234&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1234&signature=da29f1c0244ad0678cbff1f9ece1b791f821b0f6e21aeb1a5b66de6f5c714877"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: Sonnet ×7 across 3 rounds (spec+design, correctness+data-flow, UX, mobile)

No issues remain. The review iterated with the branch — every finding below was found, fixed, and re-reviewed clean at `6de99ce`.

<details><summary>Resolved during review (5 findings)</summary>

1. **Pill overlapped the composer** (ux + correctness + mobile, flagged ×3) — the fixed-`top-2` "N new" overlay sat on top of the collapsed compose box at scroll-top (the common case). Fixed via 4→5.
2. **Typing re-rendered the whole feed** (reactivity) — the composer's `ResizeObserver` bumps `startMargin` on every line-wrap; the rows were mapped inline, so each bump rebuilt all ~430 elements (`labelsFor` + a fresh `BoardCard`) and re-rendered every mounted card, on the page this PR de-janks. Fixed by memoizing `renderedRows` (deps exclude `startMargin`) + a stable `labelsFor` `useCallback`. (`8a246f8`)
3. **Composer measurement went stale on mobile** (mobile-regression) — the `ResizeObserver` was a mount-once effect on a plain ref; a mobile conversation-panel open/close remounts the composer and the observer stayed bound to the dead node, freezing `startMargin`. Fixed with a callback-ref-into-state. (`8a246f8`)
4. **`top: startMargin` clipped the pill** (correctness + ux, flagged ×2 in round 2) — the first pill fix anchored it to the composer's measured height, but an expanded mobile composer (75dvh) pushes that offset past the `overflow-hidden` edge, making "N new" unreachable while composing.
5. **Pill → non-scrolling row above the scroller** (`6de99ce`) — final fix for 1 + 4. The pill now lives outside the scroller in the board column, so it never overlaps the composer, never clips, and stays reachable at any scroll position or composer height. Appearing shifts the feed down once, matching the pre-virtua board.

</details>

Also verified clean: the `virtua` seam (`BoardFeedList`) is `spyOn`-swappable (INV-48, not `vi.mock`); `use-board-scroll-anchor` removal left no dangling refs (INV-38); no E2E spec depends on the removed `BOARD_CARD_ATTR` / Radix-viewport selector; sticky-header-in-`virtua` and windowing were browser-verified.

Note: the `evaluate-results` (Playwright) CI failure on an earlier commit is unrelated — no E2E spec navigates to the flag-gated board.

---

kristofferremback:
### Code review — pre-merge pass
**Confidence: 7/7** — Excellent (supersedes the earlier review comment)
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: Sonnet ×3 (correctness+data-flow, spec+design, UX+mobile)

Re-reviewed the full diff at merge readiness, weighted toward the reveal-latch added since the last pass. One real finding, fixed; one sub-threshold note, addressed.

<details><summary>Fixed this pass</summary>

- **Reveal latch mutated a ref during render** (state-lifecycle, 82) — `useBoardRevealLatch` set `hasRevealedRef.current = true` in the render body, the render-time-ref-mutation anti-pattern the codebase explicitly guards against (`use-timeline-scroll.ts:245`, `use-board-card-messages.ts:611`): a StrictMode/concurrent render that observes readiness true but is discarded before commit would latch a reveal that never painted. Moved the latch write into a `useLayoutEffect` (commits only on a real paint); the workspace-switch reset stays in render (idempotent, keyed on id — the sibling carve-out). The current render still reveals via the `revealReadyNow` term the instant it's ready. (`024d6d5`)

</details>

<details><summary>Noted, sub-threshold</summary>

- **Pill nudges the composer** (ux-consistency, 50) — the "N new" banner is a non-scrolling row above the scroller, so when it appears (`newCount` 0→1) it shrinks the scroller and shifts the composer + feed down once. The pre-virtua pill lived inside the scroll content below the composer, so it only shifted the feed. Kept the banner (it's what avoids the composer-overlap and the mobile-expanded-compose clip that the in-scroll and overlay variants hit); corrected the code comment, which had wrongly claimed parity with the old board. If the composer nudge reads as janky in dogfooding, the clean removal is pinning the composer above the scroller — a deliberate composer-UX change, left as a follow-up.

</details>

Verified clean: reveal-latch workspace-reset + mobile-panel-toggle-non-reset behavior, no stale-content case; virtualization + memoized rows + callback-ref composer measurement interaction; INV-38/18/9/63; no dead code left after the accumulated edits.

Not in this PR (pre-existing main behavior, follow-ups): optimistic post should slot in exactly as the composer clears (stream-view model) rather than a beat later; new-scratchpad materialization latency.

---

PR #1236, opened 2026-07-07
fix(agents): "Interrupted, retrying…" instead of a terminal-fail flash on retryable turns

kristofferremback:
## Problem

A persona turn that throws marks its session `FAILED` and emits a terminal `agent_session:failed` stream event on **every** attempt (`withCompanionSession` catch, `apps/backend/src/features/agents/companion/session.ts`). The persona worker then re-throws (`persona-agent-worker.ts`), the queue retries the same session — resuming it `FAILED`→`RUNNING` via `findByTriggerMessage` — up to `maxRetries` (default 5, `manager.ts`), and a successful retry emits `agent_session:completed`.

The board/timeline session card derives its status from the presence of those events (`deriveStatus`, `agent-session-event.tsx`). `completed` wins over `failed` — but only once *both* exist. The `failed` event lands first, so at that instant the card is terminal red "Session failed", then the later `completed` flips it to "Session complete". A session that ultimately succeeds flashes red mid-run.

The signal needed to avoid this — "are retries exhausted?" — lived only in the queue manager (`newFailedCount >= maxRetries`), two layers above the emit site. The handler's `Job` was `{ id, name, data }` with no attempt count.

## Solution

Emit the terminal `agent_session:failed` **only when retries are actually exhausted**; a retryable attempt emits a new non-terminal `agent_session:interrupted` ("Interrupted, retrying…") instead. Thread the queue's retry accounting down to the emit site so it can tell the two apart.

### How it works

1. **Thread the attempt count.** `Job` gains optional `attempt` (the message's `failedCount`) and `maxAttempts`; `QueueManager.processMessage` populates them on dispatch (hoisting the existing `maxRetries` lookup above the `try`). `persona-agent-worker` reads them off the job and forwards through `PersonaAgentInput` → `withCompanionSession` params.
2. **Branch the failure emit.** The catch computes `willRetry = attempt + 1 < maxAttempts` — exactly the manager's DLQ predicate (`newFailedCount >= maxRetries`), so the terminal `failed` fires on precisely the attempt that gets dead-lettered. Retryable → `agent_session:interrupted`; last attempt (or no accounting, e.g. evals/tests) → `agent_session:failed`. The row is still marked `FAILED` either way: orphan cleanup only scans `RUNNING`, so it skips a between-retries session, and the retry resumes it via the existing `FAILED`→`RUNNING` path.
3. **Suppress the trace-dialog false-fail.** `withCompanionSession` returns `willRetry` on its `failed` result; `persona-agent` calls `trace.notifyFailed()` (session-room socket, drives the open trace dialog) only when `!willRetry`. `notifyActivityEnded()` still fires to stop the live ticker; the retry re-emits `notifyActivityStarted()`.
4. **`agent_session:interrupted` is a broadcast lifecycle event** — registered in `EVENT_TYPES`, `AGENT_SESSION_EVENT_TYPES`, `TIMELINE_BROADCAST_EVENT_TYPES`, and `STREAM_ROW_SPEC` (the shared `AGENT_SESSION` spec). It's appended and delivered to every member via the outbox (`OutboxRepository` type unions + payload map + `STREAM_SCOPED_EVENTS`), rides `handleAppendEvent` in `stream-sync`, groups into the session card via `getSessionId`, and takes a dense broadcast slot — so it moves with its session and survives a page refresh, exactly like its siblings.
5. **Frontend `retrying` state.** `deriveStatus` gains an `interruptedPayload` and a `"retrying"` status with precedence deleted › completed › failed › **retrying** › running — a terminal event always supersedes it. `buildStatusConfig` renders it amber (not the red terminal-fail) with a spinner, a live step count (snapshot as the backoff floor — see decision 3), and the live substep. Board and timeline share `AgentSessionEvent`, so both surfaces get the fix; the board keeps `interrupted` out of `SESSION_TERMINAL_EVENT_TYPES`, so a retrying session still mounts the live progress rail. `interrupted` is added to `event-item`'s null-render group and `event-list`'s `ZERO_HEIGHT_EVENT_TYPES` so the raw event never renders standalone alongside the grouped card.

### Key design decisions

**1. `interrupted` is broadcast, not a non-broadcast patch.** First modeled as non-broadcast (analogous to `agent:follow_up_cancelled`, a correlation-patch that takes no slot). Caught in self-review: the message-move flow (`event-service.ts`) fetches a session's events via `findAgentSessionEventsBySessionIdsForUpdate` and assigns each a fresh `(sequence, broadcastSequence)` pair, asserting every one is a broadcast type — a non-broadcast `interrupted` there would take a broadcast slot it shouldn't, and *excluding* it from the move would orphan it on the source stream (a phantom "retrying" card with no `started`). `interrupted` is genuinely an appended event delivered to every member (it flows through `handleAppendEvent` like `started`/`failed`), not a patch-by-correlation, so broadcast is the correct model — and it makes the move flow and refresh-survival work with no special-casing. `event-repository`'s move query switched from a hardcoded 4-type literal to `AGENT_SESSION_EVENT_TYPES` so it stays in sync (INV-33).

**2. Retry state lives in the event log, not a column (INV-57).** No new session column; the transient "retrying" is an append-only lifecycle event committed in the same transaction as the `FAILED` mark (INV-7). The DB status stays `FAILED` between retries so orphan cleanup and the resume path are unchanged.

**3. Retrying follows the live step count, snapshot as the backoff floor.** A retry restarts its step counter from 1 (`SessionTrace` is fresh per attempt), so the card shows `liveSteps > 0 ? liveSteps : interruptedPayload.stepCount` — during backoff the live rail reports 0 and the card holds the interrupted snapshot; once the retry is actively progressing the count follows the live rail (and the live substep renders) so it moves every step instead of freezing and reading as a hang. (First cut used the snapshot only, then `max(live, snapshot)`; both froze the count during the retry's re-walk of the prior steps — caught in the /code-review loop, see the two follow-up commits.)

**4. In scope: the in-process persona path only.** The enclave `/fail` endpoint and orphan cleanup emit `agent_session:failed` on genuinely terminal failures ("a loop error is not retried", `session-handlers.ts`; stale heartbeat = dead). Those are correct and untouched — the flash was unique to the queue-retry path.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/types/src/agent-trace.ts` | `AgentSessionInterruptedPayload` (sessionId, stepCount, attempt, maxAttempts, error, interruptedAt) |
| `packages/types/src/constants.ts` | `agent_session:interrupted` added to `EVENT_TYPES`, `TIMELINE_BROADCAST_EVENT_TYPES`, `AGENT_SESSION_EVENT_TYPES` |
| `packages/types/src/stream-rows.ts` | Map `interrupted` to the broadcast `AGENT_SESSION` spec |
| `packages/types/src/stream-rows.test.ts` | Add `interrupted` to the `BOARD_EVENT_ROW_TYPES` assertion |
| `packages/types/src/index.ts` | Export the new payload type |
| `apps/backend/src/lib/queue/job-queue.ts` | `Job.attempt` / `Job.maxAttempts` (optional) |
| `apps/backend/src/lib/queue/manager.ts` | Hoist `maxRetries`; pass `attempt`/`maxAttempts` into the handler call |
| `apps/backend/src/features/agents/persona-agent-worker.ts` | Forward `job.attempt`/`job.maxAttempts` into `agent.run` |
| `apps/backend/src/features/agents/persona-agent.ts` | `PersonaAgentInput.attempt`/`maxAttempts`; pass to `withCompanionSession`; gate `notifyFailed` on `!willRetry` |
| `apps/backend/src/features/agents/companion/session.ts` | Params + `willRetry`; catch emits `interrupted` (retryable) vs `failed` (terminal); `WithSessionResult.failed` carries `willRetry` |
| `apps/backend/src/lib/outbox/repository.ts` | `agent_session:interrupted` in event-type unions, payload map, `STREAM_SCOPED_EVENTS` |
| `apps/backend/src/features/streams/event-repository.ts` | Move query uses `AGENT_SESSION_EVENT_TYPES` instead of a 4-type literal |
| `apps/frontend/src/components/timeline/agent-session-event.tsx` | `retrying` status: `deriveStatus` precedence, amber `buildStatusConfig` case, live-count-with-snapshot-floor |
| `apps/frontend/src/sync/stream-sync.ts` | `on`/`off` `agent_session:interrupted` → `handleAppendEvent` |
| `apps/frontend/src/components/timeline/event-item.tsx` | `interrupted` in the grouped-session null-render case |
| `apps/frontend/src/components/timeline/event-list.tsx` | `interrupted` in `ZERO_HEIGHT_EVENT_TYPES` |
| `apps/backend/src/features/agents/companion/session.test.ts` | Interrupted vs failed vs no-accounting branches |
| `apps/frontend/src/components/board/board-card.test.tsx` | Persisted-interrupted card renders "Interrupted, retrying…" (refresh survival) |
| `apps/frontend/src/components/timeline/agent-session-event.test.tsx` | Retrying render + precedence (completed/failed supersede) + no Stop/Redirect + live-count/backoff-floor |

## Out of scope (deliberate)

- **Enclave & orphan-cleanup failure paths** — genuinely terminal, correctly keep emitting `agent_session:failed`.
- **Trace-dialog "retrying" state** — the open trace dialog (`use-agent-trace`, session room) no longer flashes a false `failed` (the `willRetry` guard), and stays "running" through the brief backoff until the retry's steps resume or it terminates. A dedicated retrying banner in the trace dialog wasn't built — the reported flash was the board/timeline card.

## Test plan

- [x] Backend `bun test src/features/agents/companion/session.test.ts` — 8 pass (interrupted-on-retryable, failed-on-last-attempt, terminal-when-no-accounting)
- [x] Backend `bun test src/lib/queue/ src/features/agents/` — 442 pass
- [x] Types `bun test` — 109 pass (incl. `stream-rows` broadcast-slot mirror invariant, INV-61)
- [x] Frontend `bunx vitest run src/components/timeline/ src/components/board/` — 488 pass
- [x] `tsc --noEmit` across all 14 workspaces clean (via the pre-commit gate: lint, typecheck, astro, migrations, API-docs)
- [x] Self-review over the full diff — 1 finding fixed pre-commit: the move-flow broadcast-slot issue (decision 1), which flipped `interrupted` from non-broadcast to broadcast
- [x] `/code-review` (4 Sonnet lenses: spec+design, correctness+data-flow, UX, mobile) → 7/7. One UX finding (frozen retry step count) fixed over two commits (see decision 3); all other lenses clean
- [ ] No live-DB harness, so `findAgentSessionEventsBySessionIdsForUpdate` (the `FOR UPDATE` move query) is exercised only through its existing stubbed-repo tests — `interrupted` now shares the identical broadcast path as its siblings, so there is no interrupted-specific move logic to integration-test

---


kristofferremback:
### Code review
**Review models:** Orchestrator: opus-4-8 | Reviewers: sonnet x4 (spec+design, correctness+data-flow, UX, mobile)

No blocking issues found. Checked for CLAUDE.md/invariant compliance, plan adherence, design quality, correctness, data-flow lifecycle, security, UX, and mobile/responsive across the 19-file diff.

The review loop surfaced and resolved one real UX issue before reaching 7/7:

- **Frozen retry progress** (`agent-session-event.tsx`) — the "retrying" card first showed the interrupted **snapshot** step count. Since a retry restarts its step counter from 1 (`SessionTrace` is fresh per attempt), the count sat frozen while the retry re-walked the prior attempt's steps — a spinner over a static number that reads as a hang for a session that failed deep. Fixed in two steps: (1) show the live substep while retrying and floor the count with the snapshot; (2) follow the live count once the retry is actively progressing (`live > 0`), falling back to the snapshot only during backoff — so the count moves every step and never freezes. The one-time backward dip at retry start (snapshot → live) is honest "restarting" behavior and scored below the bar.

Verified in passing (context, not findings):

- **INV-61** — `agent_session:interrupted` is correctly a dense broadcast slot (added to `TIMELINE_BROADCAST_EVENT_TYPES`, `STREAM_ROW_SPEC` via the shared `AGENT_SESSION` spec), delivered to every member, and moves with its session in the message-move flow. No per-viewer gaps.
- **INV-57 / INV-7** — retry state lives in the append-only event log, not a new column; the `FAILED` mark + event + outbox stay in one transaction. Orphan cleanup (`onlyIfStatus: RUNNING`) correctly skips a between-retries `FAILED` session.
- **`willRetry` = `attempt + 1 < maxAttempts`** exactly mirrors the manager's DLQ predicate (`newFailedCount >= maxRetries`), fed from the same `maxRetries` source — the terminal `failed` fires on precisely the attempt that dead-letters. The `!willRetry` guard also suppresses the trace-dialog `notifyFailed`, so both the persisted-event and live-socket paths are fixed.

<details><summary>📐 Plan Adherence [CLEAN]</summary>New event type wired consistently across all 19 surfaces (EVENT_TYPES, AGENT_SESSION_EVENT_TYPES, broadcast set, spec, outbox unions/payload-map/scoped-events, sync handlers, grouping, ZERO_HEIGHT, event-item, deriveStatus/render). No missing corresponding change.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>willRetry/DLQ boundary math verified (incl. maxAttempts=1 and last-attempt cases); interruptedPayload non-null when status="retrying"; each retry is a fresh run() with its own notify pair (no shared-closure/listener leak).</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>interrupted event round-trips (persist → outbox → handleAppendEvent → group → refresh). FAILED-between-retries interacts correctly with orphan cleanup and the FAILED→RUNNING resume path. Live substep plumbing confirmed effective on retry (activity_started re-emitted per attempt).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-61, INV-57, INV-7, INV-4, INV-33 (event-repository move query switched from a 4-type literal to AGENT_SESSION_EVENT_TYPES), INV-21 (single-line subtitle preserved) all verified.</details>
<details><summary>🏗️ Design [CLEAN]</summary>No new list sprawl; interrupted flows through the shared constants/spec. Broadcast (not non-broadcast patch) is the correct model — makes the move flow and refresh-survival work with no special-casing.</details>
<details><summary>🔒 Security [CLEAN]</summary>No user-input trust boundary changed; no injection/auth/exposure surface.</details>
<details><summary>🎛️ UX [CLEAN]</summary>Retry state resolved (see above): amber non-alarming spinner, live count + substep, no Stop/Redirect while retrying, survives refresh, no INV-21 layout shift.</details>
<details><summary>📱 Mobile [CLEAN]</summary>Pure status-config addition reusing the existing responsive card layout; title `min-w-0 truncate`, single-line subtitle — no phone-width overflow or height growth.</details>

---

PR #1254, opened 2026-07-09
fix(timeline): resolve thread read watermark against the rendered window (ghost unread threads)

kristofferremback:
## Problem

Prod report (DM conversation `conv_01KX47HPDDXPA5P1BC3QPCFMFH`): a user has a `(1)` in the tab title that survives a hard refresh, and the thread "Sinnet 4.6 och kil på 5" keeps its blue unread strip in the sidebar even after opening the thread. Opening it also shows **no red unread divider** — the thread *looks* fully read while the server keeps counting 1 unread.

The server is right: `stream_members.last_read_event_id` for the reporter still pointed at seq 1 (`member_added`) with the reply at seq 2. The frontend never fired `POST /streams/:id/read` while the user was viewing the thread. Reproduced end-to-end against a local `dev:test` stack with Playwright: a second thread member opening the thread (focused, visible, rows on screen) never issues the read call — via `?m=` deep link *and* via plain open.

Two stacked frontend bugs, one shared root: **a fresh thread member's read watermark is seeded on their `member_added` event (backend `addToStream`), but threads suppress membership events from the rendered window** (`THREAD_HIDDEN_EVENT_TYPES` in `stream-content.tsx`).

1. **Auto-read never fires** — `useLastSeenEvent` resolves the watermark against the rendered window (`displayEvents`). `member_added` isn't in it → `readSeq` resolves to `null` ("pointer outside the loaded window") → the emit guard suppresses `lastSeenEventId` permanently → `useAutoMarkAsRead` has no target, so the read POST never happens. **Every non-creator thread member was unable to auto-read any thread without replying in it.**
2. **The frontier scan was vetoed anyway** — the thread view renders its parent message via `ThreadParentMessage` → `EventItem`, emitting `data-event-id` with the *parent stream's* event id. `recompute()` bailed entirely when the visible range's boundary row wasn't mappable (`topIdx === undefined`). In a short thread the parent banner never leaves the viewport, so every scan was vetoed even where bug 1 didn't apply.
3. **No red unread strip** (same root as 1) — `useUnreadDivider` returns `undefined` when `lastReadEventId` isn't found in its `events` (`use-unread-divider.ts:97-100`), so no divider latched: the thread renders with no unread marking at all while genuinely unread.

## Solution

Resolve the watermark into rendered-window space once, in `StreamContent`, and make the frontier scan tolerate foreign rows.

### How it works

- **`remapSuppressedWatermark(lastReadEventId, events, displayEvents)`** (new pure helper in `stream-content.tsx`): a watermark present in the loaded `events` but hidden from `displayEvents` is remapped to the nearest **preceding** rendered event, or `null` when none precedes it — equivalent read positions, since suppressed events aren't readable content. A watermark outside the loaded window is returned unchanged so the existing mark-unread-below-window suppression semantics keep applying. Threads only (`isThread` guard); channels/DMs render membership events and are unaffected.
- The remapped `frontierLastReadEventId` feeds `useLastSeenEvent`, `useUnreadDivider`, and the divider's `readStateResolved` gate. `isDividerReadPast` and `useNewMessageIndicator` keep the raw watermark — they resolve against unfiltered `events`, where `member_added` exists.
- **`useLastSeenEvent.recompute`**: rows whose `data-event-id` isn't in the loaded window map are now skipped when collecting visible rows, instead of poisoning the range lookup and vetoing the scan. Only `EventItem` emits `data-event-id`, and gap placeholders don't, so the only foreign row today is the thread parent banner — but any future out-of-window row gets the same treatment.

### Key design decisions

**1. Remap in `StreamContent`, not inside the hooks** — both consumers (`useLastSeenEvent`, `useUnreadDivider`) resolve against the same filtered window, and the filtering (`THREAD_HIDDEN_EVENT_TYPES`) lives in `StreamContent`. One remap where the filter is defined beats teaching each hook about thread display rules. The hooks' null-pointer suppression stays intact for its real purpose (mark-as-unread below the loaded window).

**2. Nearest *preceding* rendered event, not nearest following** — the watermark means "read through here"; mapping forward would silently mark the next message read. Mapping backward (or to `null`) can only surface *more* as unread, matching the server's count (messages with sequence > watermark).

**3. Skip foreign rows rather than excluding the parent banner's `data-event-id`** — the attribute is load-bearing for scroll targeting and the divider `querySelector`; removing it from `ThreadParentMessage` would break deep-linking to the parent. Filtering at the consumer keeps one emit path.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/stream-content.tsx` | New `remapSuppressedWatermark` helper + `frontierLastReadEventId` memo; feeds the read frontier, unread divider, and its `readStateResolved` gate |
| `apps/frontend/src/components/timeline/stream-content.test.ts` | 5 remap tests: no-predecessor → null, mid-thread → nearest preceding, rendered/out-of-window/null passthrough |
| `apps/frontend/src/hooks/use-last-seen-event.ts` | `recompute` skips rows not in the loaded window instead of bailing on unmappable range boundaries |
| `apps/frontend/src/hooks/use-last-seen-event.test.ts` | Regression test: foreign parent-banner row visible at top, reply still marks seen |

## Out of scope (deliberate)

- **Sidebar highlight ambiguity** — the reporter's second complaint (active-stream highlight and unread highlight look alike) is a UX/design change, untouched here.
- **Fresh threads not appearing in the sidebar without a refresh** — both users needed a reload to see the new thread at all; a separate live-delivery gap to investigate.
- Existing stuck-unread rows in prod self-heal on next open (the read now fires); no backfill needed.

## Test plan

- [x] Live repro before/after on `dev:test` + Playwright (two users, thread on B's message, A replies, B opens): before — no `POST /read`, watermark stuck at `member_added`; after — `POST /read` fires with the reply's event id, `stream_members.last_read_event_id` advances to seq 2 (verified in Postgres)
- [x] `vitest run` on the four touched suites + `use-unread-divider` + `use-auto-mark-as-read` — 97 pass
- [x] `vitest run src/components/timeline src/hooks src/sync` — 116 files pass; 3 files fail identically on a clean tree (known Node-25 localStorage reds, verified via `git stash` A/B)
- [x] Full monorepo lint + typecheck clean (pre-commit hook runs both across all workspaces)
- [ ] `bun run test:e2e` not run — the affected flow is covered by the scripted Playwright repro above; no existing e2e spec exercises thread read state

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1254"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786221897&installation_id=129946197&pr_number=1254&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1254&signature=ea66a740d61dec57936280f2f8be1f788748096cedc87187068f05ad7bfe7b7c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/stream-content.test.ts
+++ b/apps/frontend/src/components/timeline/stream-content.test.ts
@@ -9,8 +9,10 @@ import {
   shouldRunEdgePagination,
   shouldShowOlderSkeletons,
   resolveDateJumpAnchor,
+  remapSuppressedWatermark,
 } from "./stream-content"
 import { localStartOfDayMs } from "@/lib/dates"
+import type { StreamEvent } from "@threa/types"
 
 const DEADLINE = 4000
 
@@ -466,5 +468,43 @@ describe("resolveDateJumpAnchor", () => {
       { eventType: "message_created" as const, createdAt: REF.toISOString(), payload: {} },
     ]
     expect(resolveDateJumpAnchor({ events, targetDayMs, hasOlderEvents: true })).toBeNull()
+  })
+})
+
+describe("remapSuppressedWatermark", () => {
+  const evt = (id: string, eventType: string) => ({ id, eventType }) as unknown as StreamEvent
+  // A fresh two-event thread: the member_added seeding event is suppressed
+  // from the rendered window; only the reply renders.
+  const events = [evt("e_member", "member_added"), evt("e_reply", "message_created")]
+  const displayEvents = [evt("e_reply", "message_created")]
+
+  it("remaps a suppressed watermark with no rendered predecessor to null (nothing read)", () => {
+    // The ghost-unread-thread bug: the watermark on member_added was
+    // unresolvable in the rendered window, so no divider rendered and
+    // auto-read never fired — the thread stayed unread forever.
+    expect(remapSuppressedWatermark("e_member", events, displayEvents)).toBeNull()
+  })
+
+  it("remaps a suppressed watermark to the nearest preceding rendered event", () => {
+    const longer = [
+      evt("e_msg1", "message_created"),
+      evt("e_member2", "member_added"),
+      evt("e_msg2", "message_created"),
+    ]
+    const rendered = [evt("e_msg1", "message_created"), evt("e_msg2", "message_created")]
+    expect(remapSuppressedWatermark("e_member2", longer, rendered)).toBe("e_msg1")
+  })
+
+  it("passes a rendered watermark through unchanged", () => {
+    expect(remapSuppressedWatermark("e_reply", events, displayEvents)).toBe("e_reply")
+  })
+
+  it("passes a watermark outside the loaded window through unchanged (suppression semantics apply)", () => {
+    expect(remapSuppressedWatermark("e_below_window", events, displayEvents)).toBe("e_below_window")
+  })
+
+  it("passes null/undefined (nothing read / still hydrating) through unchanged", () => {
+    expect(remapSuppressedWatermark(null, events, displayEvents)).toBeNull()
+    expect(remapSuppressedWatermark(undefined, events, displayEvents)).toBeUndefined()
   })
 })
--- a/apps/frontend/src/components/timeline/stream-content.tsx
+++ b/apps/frontend/src/components/timeline/stream-content.tsx
@@ -242,6 +242,35 @@ export function canActOnDeepLinkNavigation<T extends string>(args: {
   return args.hasEvents
 }
 
+/**
+ * Remap a thread watermark that points at a suppressed membership event to the
+ * nearest PRECEDING rendered event (null when none precedes it — "nothing read
+ * yet", an equivalent read position since suppressed events aren't readable
+ * content). A fresh thread member's watermark is seeded on their member_added
+ * event by the backend, but threads hide membership events from the rendered
+ * window — so consumers that resolve the watermark against `displayEvents`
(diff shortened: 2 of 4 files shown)


kristofferremback:
### Code review

No issues found. Checked for bugs, data-flow lifecycle (watermark round-trip across remap → frontier → divider → mark-read echo, out-of-window passthrough, mid-thread re-add, gap-placeholder contiguity), CLAUDE.md compliance, plan adherence (all `lastReadEventId` consumers audited: the three filtered-window consumers get the remap, the three raw-window consumers correctly keep the raw watermark), design quality, security, UX (fresh-member red divider, no over-counting vs server, no cold-load flash-then-stick, red→gray dimming, main-view/panel consistency), and mobile (composer-band exclusion untouched, no gesture/layout surface).

PR #1288, opened 2026-07-11
feat: persona config revision history — view and restore previous versions

kristofferremback:
Persona config **revision history** — view and restore previous versions of a persona's config. Kris follow-up to the persona builder (#1285).

**Stacked on #1287** (`feat/ai-agents-framing`) — set that as the base / merge it first; the diff here is only the revisions delta.

## Problem

Committing a persona override was one-way: no record of what the config was before, and no way back. `stream_brief_revisions` established an append-only revision pattern but shipped write-only — no list or restore surface anywhere. This adds the full loop for persona config.

### Backend

- **`persona_config_revisions`** (append-only, migration `20260711221827…`): `acrev_` ULID, `workspace_id`, `agent_id`, monotonic `version INT`, `patch JSONB` (the sparse override at that version), `created_by_kind/id`, `created_at`; index `(workspace_id, agent_id, version DESC)`. Mirrors `stream_brief_revisions`, patch is JSONB not TEXT.
- **Capture**: a revision is inserted inside `PersonaConfigService.setOverride`'s existing transaction, right after `upsertActive` succeeds — same txn as the override write and the `agent_config:updated` outbox (INV-7), so the audit trail can never miss an accepted write. `version` is `COALESCE(MAX(version),0)+1` derived in-txn; race-safe because the override upsert already takes a `FOR UPDATE` row lock on the active `(workspace, agent)` row (verified), so at most one writer appends a revision per persona at a time (INV-20) — no new column on `agent_config_overrides` needed.
- **Read**: `GET /api/workspaces/:workspaceId/personas/:personaId/revisions` (admin, cap 50, logs on truncation), newest-first.
- **Restore**: `POST …/revisions/:revisionId/restore` (admin, body `{expectedUpdatedAt}`) loads the revision and re-commits its patch through `setOverride` **whole** — so restore is never destructive: it appends a new revision equal to the restored one (restoring v2 while on v5 creates v6 == v2). Same optimistic-concurrency 409 as a normal save; a foreign revision 404s; a schema-drifted revision (e.g. an `enabledTools` entry since retired from the enum) returns 422 `PERSONA_REVISION_INCOMPATIBLE` instead of a raw 500.

### Frontend

- `personasApi.listRevisions` / `restoreRevision` + `personaKeys.revisions` + `useRevisions` / `useRestoreRevision` (cache-write-on-success so the editor re-seeds to restored values; 409/422 surfaced to the caller).
- **`PersonaHistoryPanel`** — a `ResponsiveDialog` from a "History" button in the editor footer. Lists revisions newest-first: version, a **"Current"** badge on the newest (no Restore on it), who (`getActorName`, neutral fallback), local relative+full timestamp (INV-42), and a concise **"Changed: …"** field summary computed against the previous revision's patch. Each older row has Restore → AlertDialog confirm → re-commit with the live `overrideUpdatedAt`; 409 reuses the save's amber conflict banner, 422 toasts "out of date", success closes the dialog (no success toast, INV-63).

## Key design decisions

**1. Restore reuses `setOverride` whole.** Restore is "re-commit an old patch," which is exactly what `setOverride` does — including the optimistic-concurrency guard, the draft delete, the outbox broadcast, and (now) appending a revision. No override-write logic is duplicated, and restore is append-only by construction.

**2. Version derived MAX+1 in-txn, no schema change.** The alternative — a `version` column on `agent_config_overrides` — was avoided after verifying the upsert's `FOR UPDATE` lock serializes writers (so MAX+1 can't collide). One fewer column, one fewer backfill.

**3. Shared `applyCommittedOverride` cache writer.** Save and Restore return the identical `{persona, updatedAt}` and reconcile the config + list caches identically; the verify pass caught the duplication and collapsed both `onSuccess` bodies into one writer (which also invalidates the revisions list, so History reflects a fresh Save).

## New files

| File | Purpose |
| ---- | ------- |
| `apps/backend/src/db/migrations/20260711221827_persona_config_revisions.sql` | Append-only revisions table |
| `apps/backend/src/features/agents/persona-config-revision-repository.ts` | insert / list / findById (workspace-scoped) |
| `apps/frontend/src/components/persona-editor/persona-history-panel.tsx` | History dialog + Restore |
| Colocated `*.test.ts(x)` for the repo, service revision paths, handlers, and the panel |

## Modified files (highlights)

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/agents/persona-config-service.ts` | Capture revision in setOverride txn; `listRevisions`; `restoreRevision` |
| `apps/backend/src/features/agents/persona-config-handlers.ts` + `routes.ts` | GET revisions / POST restore (admin) |
| `packages/types/src/persona-config.ts` | `PersonaConfigRevision`, `RestorePersonaRevisionInput`, single-sourced author-kind union |
| `apps/frontend/src/hooks/use-personas.ts` | `personaKeys.revisions`, hooks, shared `applyCommittedOverride` |
| `apps/frontend/src/components/persona-editor/persona-form.ts` | `changedPatchFields` + `PERSONA_FIELD_LABELS`; `patchesEqual` derives from them |
| `apps/frontend/src/components/persona-editor/persona-editor-form.tsx` | History button in footer; shared `applyOverrideConflict` |

## Out of scope (deliberate)

- No E2E spec (component + backend suites cover it; the write/read/restore round-trip could extend `persona-customization.spec.ts` later).
- Restoring the current version (a no-op) — hidden behind the "Current" badge.

## Test plan

- [x] Backend `bun test src/features/agents` — 532 pass / 0 fail (revision repo, service capture/list/restore, handlers)
- [x] Frontend `vitest run components/persona-editor` — panel lists newest-first + Current, change-summary correct, Restore round-trips with expectedUpdatedAt, 409 inline
- [x] Monorepo typecheck — 0 errors
- [x] Visual QA (screenshot): the History dialog with a "Current" v2, a restorable v1, who/when, and "Changed: …" summaries
- [x] Per-step adversarial review (2 Opus verifiers each): caught + fixed a duplicated onSuccess writer, stale revisions cache after Save, restore-of-drifted-revision → clean 422, cap false-positive at exactly 50

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1288"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786396040&installation_id=129946197&pr_number=1288&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1288&signature=ab820f2fa2c5a5f1f8ef36ab0d23d489c9cccc81f43ad806920b7a5201f99f3d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
Rebased onto main (post-#1287 merge) and added **restore-to-default (v0)** per review: an empty override patch now means "identical to defaults", so `setOverride` removes the override via a new race-safe `AgentConfigOverrideRepository.deleteActive` (same FOR UPDATE + optimistic-concurrency guard as the upsert) rather than storing an empty row. The History panel gains a synthetic "Built-in default" floor below the committed revisions — Current when no override exists, and its Restore commits an empty patch through the Save path whole (reusing the 409 conflict handling). 538 backend agents tests + 27 frontend persona tests green.

PR #1283, opened 2026-07-11
feat(prefs): configurable unread open position — open at marker or latest

kristofferremback:
Supersedes #1265 (rebased onto current `main`, conflicts resolved). Close #1265 in favor of this.

## Summary
Adds a user preference `unreadOpenPosition` (`"latest" | "marker"`, default `"latest"`) controlling where a stream lands on open:
- **latest** (default, unchanged): open at the live bottom; keeps the "N new messages" affordance.
- **marker** (Discord-style): land on the first unread via the `scrollToMessage` refine loop, with the existing "Jump to latest" affordance back down.

## Design
- **Decision helper `resolveUnreadMarkerOpen`** — pure function returning `"scroll" | "wait" | "skip"`, decided once per stream open (`unreadMarkerOpenRef` latch) after load/settle/read-state resolve. Deep-links, jump mode, and user gestures veto it; a null (unhydrated) preference holds at `"wait"` so a marker user's cold boot still lands right. Scroll runs in a `useLayoutEffect` so the marker landing is pre-paint (no bottom-flash).
- **Backend** — Zod enum derived from `UNREAD_OPEN_POSITION_OPTIONS` + `simpleKeys` entry; picks up `flattenUpdates`/`mergeOverrides`/`getDefaultValue` generically (mirrors the `boardDefaultLens` / `labelRemoveOnMove` precedent). Default `"latest"` everywhere, so existing behavior is untouched when the pref is absent.
- **Settings** — radio group under Appearance; labels/descriptions live frontend-only (INV-46).

## Files
- `packages/types/src/{preferences.ts,index.ts}` — enum + type, exported once
- `apps/backend/src/features/user-preferences/{handlers.ts,service.ts}` — Zod enum + simpleKeys
- `apps/frontend/src/components/settings/appearance-settings.tsx` — radio group
- `apps/frontend/src/components/timeline/stream-content.tsx` — `resolveUnreadMarkerOpen` + the open-decision effect
- tests: `stream-content.test.ts` (helper unit tests), `user-preferences/handlers.test.ts`, hook fixtures

## Verification
- types / frontend / backend typecheck clean
- `stream-content.test.ts` 59 pass; backend `user-preferences` tests pass
- Rebased onto latest `main`, merges cleanly

## Review
[...]

PR #1291, opened 2026-07-12
fix: open-at-unread-marker lands at the top of the first unread, for any row type

kristofferremback:
## Problem

Opening a stream with `unreadOpenPosition: "marker"` (PR #1283) often landed at the bottom showing the "N new messages" pill instead of at the divider — and even when it landed, a first unread message taller than the viewport showed its middle, forcing a scroll up to find the start. Kris's video of the failure: first unread was a "Session complete" card, stream opened at the tail.

Three compounding defects in `stream-content.tsx`:

1. **Non-message divider anchors took a fragile one-shot path.** The marker-open effect extracted `payload.messageId` from the divider event and only engaged the `scrollToMessage` refine loop when one existed. A session card, `member_added`, or `description_set` anchor has none, so it fell to `scrollToFirstUnread()`'s one-shot `scrollToIndex` — on a freshly mounted virtua list with unmeasured row heights, which lands pages off target or throws (swallowed by its `catch`). The refine loop exists precisely because a one-shot scroll on a fresh window lands ~14 messages off (PR #1265 finding).
2. **Group rows were unaddressable in the DOM.** `session_group`/`command_group` wrappers rendered no `data-event-id`, so even the refine loop's DOM-precise pass (and the plain thread scroller's `querySelector`) could not find them.
3. **The refine loop centered its target and required `fullyVisible` to converge** — unsatisfiable for a row taller than the viewport, so it churned its full 1.2s bound and settled centered, top off-screen.

## Solution

Marker opens always go through the refine loop, targeting the divider row by **event id** with a new `align: "start"` mode that pins the row near the viewport top — the same landing as the jump-to-first-unread pill.

### How it works

- `findTimelineTargetIndex` (`event-list.tsx`) resolves a scroll target as message id, plain event id, or a group's first-event id — the union of `findMessageItemIndex` and `findEventItemIndex`, mirroring `isFirstUnread`'s anchor rule. The refine loop uses it for both the early in-window bail and the per-tick live index.
- The loop's DOM query is now `[data-message-id="…"], [data-event-id="…"]`, and `session_group`/`command_group` wrappers carry `data-event-id={item.events[0]?.id}`.
- `scrollToMessage(targetId, opts?)` gains `align: "center" | "start"`. Start mode scrolls the target's top to `scroller.top + UNREAD_MARKER_TOP_GAP_PX` (56px — shared constant with the pill's `scrollToIndex` offset, clears the sticky date header and leaves a sliver of read context) and converges on top-alignment instead of `fullyVisible` — a target too close to the tail settles at the scroll clamp (`atMaxScroll` + visible). Center mode is byte-for-byte the old behavior; deep links and search jumps are untouched.
- The marker-open effect drops the `payload.messageId` extraction and calls `scrollToMessage(dividerEventId, { align: "start" })`, falling back to `scrollToFirstUnread()` only for the non-virtualized thread scroller.

### Key design decisions

**1. Event-id targeting instead of teaching the one-shot path to retry** — the refine loop already owns every hard case (unmeasured rows, window shifts mid-loop, user-gesture abort, the 1.2s bound); duplicating that in `scrollToFirstUnread` would be a second convergence loop (INV-35). The pill handler keeps its one-shot: by the time it's clickable the list is measured, and it now shares `UNREAD_MARKER_TOP_GAP_PX` so both land identically.

**2. Top-align only for marker opens; deep links keep centering.** A deep-linked message is the point of interest and reads best centered; the marker is a reading *cursor* — everything below it is unread, so it belongs at the top (matches the pill, and Discord's behavior the pref was modeled on).

**3. Side effect: group cards now participate in the auto-read frontier scan.** `useLastSeenEvent` collects `[data-event-id]` rows; group first-event ids map into the loaded window, so the existing `map.has(id)` guard (the PR #1254 parent-banner fix) admits them. The frontier already absorbs row-less events (reactions) between visible rows, so this is consistent — and it closes a ghost-unread gap where a session card as the only unread row gave the scan nothing to advance on.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/timeline/stream-content.tsx` | `scrollToMessage` → generic target + `align` option; marker-open effect targets the divider event id with `align: "start"`; `UNREAD_MARKER_TOP_GAP_PX` shared with the pill path |
| `apps/frontend/src/components/timeline/event-list.tsx` | `findTimelineTargetIndex`; `data-event-id` on session/command group wrappers |
| `apps/frontend/src/components/timeline/event-list.test.ts` | `findTimelineTargetIndex` coverage incl. the group-first-event case `findMessageItemIndex` misses |

## Out of scope (deliberate)

- The `hideSessionCards` context can still leave `dividerEventId` pointing at a filtered-out row (marker open then falls back to latest, and no divider renders at all) — pre-existing, orthogonal to this fix.
- `resolveUnreadMarkerOpen`'s decision gates (deep-link/jump/gesture vetoes, pref-hydration wait) are unchanged.

## Test plan

- [x] `vitest` — `event-list.test.ts` (64), `stream-content.test.ts`, `use-unread-divider.test.ts`: 142 pass
- [x] Frontend `tsc --noEmit` clean (app + test configs); full lint/typecheck pre-commit hook green
- [x] Playwright browser specs guarding the shared refine loop: `edit-last-message` (incl. "scrolls off-screen message into view"), `infinite-scroll`, `composer-bottom-spacing` — 12 pass
- [x] Live verify against `dev:test` (scripted Playwright, desktop 1280×800 **and** mobile 390×844 emulation), fresh browser context per pref value:
  - marker + single taller-than-viewport unread message → divider lands ~36px below scroller top, `LONG-MESSAGE-START` visible
  - marker + non-message first unread (`member_added`/`description_set` anchor, unread block > viewport) → divider at top; with a short unread run it settles at the scroll clamp with the divider visible (expected)
  - `latest` pref → opens at bottom, unchanged
  - `?m=` deep link with marker pref → linked message centered, marker scroll vetoed
- [ ] Real agent-session card as divider anchor not reproduced live (dev:test has no `OPENROUTER_API_KEY`); covered by the group-first-event unit test + the `member_added` non-message anchor live scenario, which exercises the same event-id path

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1291"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786425090&installation_id=129946197&pr_number=1291&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1291&signature=0b5e10fc9b9a2ffe1bd4d71df06205ac5c21f2d8147717eccd6837d89e80d1bf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/timeline/event-list.test.ts
+++ b/apps/frontend/src/components/timeline/event-list.test.ts
@@ -9,6 +9,7 @@ import {
   findFirstMessageId,
   findMessageItemIndex,
   findEventItemIndex,
+  findTimelineTargetIndex,
   getTimelineItemKey,
   groupTimelineItems,
   injectGapItems,
@@ -565,6 +566,50 @@ describe("findEventItemIndex", () => {
 
   it("returns -1 when the event is not in the window so the cold-load scroll falls back to the bottom", () => {
     expect(findEventItemIndex([eventItem("evt_1")], "evt_gone")).toBe(-1)
+  })
+})
+
+describe("findTimelineTargetIndex", () => {
+  function messageItem(id: string, messageId: string): TimelineItem {
+    return {
+      type: "event",
+      event: createEvent({ id, sequence: id, eventType: "message_created", payload: { messageId } }),
+    }
+  }
+
+  it("resolves a message id like the deep-link lookup", () => {
+    const items: TimelineItem[] = [messageItem("evt_1", "msg_a"), messageItem("evt_2", "msg_b")]
+    expect(findTimelineTargetIndex(items, "msg_b")).toBe(1)
+  })
+
+  it("resolves a plain event id (a non-message divider anchor like a description row)", () => {
+    const items: TimelineItem[] = [
+      messageItem("evt_1", "msg_a"),
+      {
+        type: "event",
+        event: createEvent({ id: "evt_desc", sequence: "2", eventType: "description_set", payload: {} }),
+      },
+    ]
+    expect(findTimelineTargetIndex(items, "evt_desc")).toBe(1)
+  })
+
+  it("resolves a session group by its first event id — the marker-open case the message lookup misses", () => {
+    const items: TimelineItem[] = [
+      messageItem("evt_1", "msg_a"),
+      {
+        type: "session_group",
+        sessionId: "sess_1",
+        sessionVersion: 1,
+        events: [createEvent({ id: "evt_sess", sequence: "2", eventType: "message_created", payload: {} })],
+      },
+      messageItem("evt_3", "msg_c"),
+    ]
+    expect(findMessageItemIndex(items, "evt_sess")).toBe(-1)
+    expect(findTimelineTargetIndex(items, "evt_sess")).toBe(1)
+  })
+
+  it("returns -1 when the target is not in the window (never feed virtua a stale index)", () => {
+    expect(findTimelineTargetIndex([messageItem("evt_1", "msg_a")], "evt_gone")).toBe(-1)
   })
 })
 
--- a/apps/frontend/src/components/timeline/event-list.tsx
+++ b/apps/frontend/src/components/timeline/event-list.tsx
@@ -426,6 +426,18 @@ export function findEventItemIndex(items: TimelineItem[], eventId: string): numb
   })
 }
 
+/**
+ * Index of the timeline item addressed by `targetId` — a message id, a plain
+ * event id, or the first-event id of a command/session group. The union of
+ * `findMessageItemIndex` and `findEventItemIndex`, for scroll targets that can
+ * be any row the unread divider anchors on (a session card's first unread is
+ * an `event_…` id inside a group, which the message lookup alone misses).
+ */
+export function findTimelineTargetIndex(items: TimelineItem[], targetId: string): number {
+  const idx = findMessageItemIndex(items, targetId)
+  return idx >= 0 ? idx : findEventItemIndex(items, targetId)
+}
+
 /** Returns a stable key string for a timeline item */
 export function getTimelineItemKey(item: TimelineItem): string {
   switch (item.type) {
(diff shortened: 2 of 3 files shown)


kristofferremback:
### Code review

No issues found. Checked for plan/intent adherence, CLAUDE.md compliance, design quality, bugs, data-flow lifecycle, security, UX consistency, and mobile/touch behavior.

Highlights from the verification passes:

- **Correctness:** the dual `[data-message-id], [data-event-id]` query cannot resolve the wrong row (prefixed ULIDs, INV-2 — a `msg_` string never appears in `data-event-id`). All five start-align sequences converge: far-above target, near-tail clamp (`atMaxScroll` settle), taller-than-viewport row (top-alignment is height-independent — the reason centering was insufficient), unrendered-target `scrollToIndex` → DOM handoff, and user-gesture abort. The refine loop re-resolves by id every tick, so prepends and jump-window swaps that shift indexes are tolerated; dropping `events` from the marker effect deps is safe (the body no longer reads it).
- **Auto-read side effect verified safe:** group rows expose their *first* event id, and `advanceFrontier`'s strict contiguity guard (`topIdx <= frontier + 1`) means a visible group card cannot advance the frontier past unseen message rows above it — conservative by construction, and it closes the session-card-only ghost-unread gap.
- **Mobile:** the near-tail clamp settle cannot hide the divider behind the floating composer — `ComposerFooterSpacer` (last scroll item, `height: var(--composer-height)`) pushes real content above the band at max scroll.
- **Consistency:** marker open and the jump-pill resolve the same divider row and share `UNREAD_MARKER_TOP_GAP_PX` (56px), so both land identically; deep-link callers pass no `align` and keep the old centering byte-for-byte.

PR #1297, opened 2026-07-12
fix: mobile image gallery goes full-bleed and hides its chrome on tap

kristofferremback:
## Problem

Three iterations of mobile-gallery feedback (Kris, scratchpad hand-off):

1. **Browser chrome covered content.** The mobile gallery was a centered floating card sized `max-sm:h-[90vh] max-sm:max-w-[95vw]` (`apps/frontend/src/components/image-gallery.tsx`). `vh` measures the **large** viewport (browser chrome collapsed), while the card is centered inside the **dynamic** viewport — with the URL bar expanded the card overflowed and chrome covered the top of the image and the action bar. The override dated to the original gallery PR (#266).
2. **The gallery's own chrome sat on top of the image.** The action bar (copy/download/close pill) and the filename bar were always visible. Kris: "I want to hide it on tap … really get this all on the outermost zoomed-out layer."
3. **Zoom must not interact with the toggle.** Kris: "double tapping to zoom it stays, same as pinch … if I tap once the chrome should disappear, tap again, or zoom all the way out, the chrome should reappear."

## Solution

**Full-bleed mobile layout.** Drop the centered-card override and let the base `DialogContent` mobile layout apply: `max-sm:inset-0` on a `position: fixed` element tracks the dynamic viewport, so the gallery is always exactly the visible area. Desktop and ≥640px keep the centered dialog, now `90dvh` instead of `90vh` (identical on desktop, correct on landscape phones). `max-sm:gap-0` / `max-sm:overflow-hidden` cancel the base's mobile defaults (twMerge doesn't merge across breakpoint prefixes — noted in `dialog.tsx`). The video-embed width calc `85vh` → `85dvh` for the same reason.

**Single tap toggles the gallery chrome (touch layout).** A single tap on the media — anywhere that isn't a nav zone, the text panel, or a `<video>` — hides the action bar and filename bar with a 200ms fade; tap again restores them. Zoomed or not: while zoomed the nav zones are disabled, so the whole surface is toggle. Edge-zone tap navigation (unzoomed) and swipe are unchanged; the toggle resets to visible on every open.

**Zoom gestures never hide the chrome.** The toggle is deferred past the double-tap window (`DOUBLE_TAP_MS = 300`, now exported from `use-zoom-pan.ts` instead of being an inline magic number, INV-33) and cancelled the moment a zoom commits — so the first tap of a double-tap zoom never flips the chrome, and pinch (two touches) never schedules one. Zooming all the way back out (`onZoomChange(false)`, fired synchronously by `commit` in `use-zoom-pan`) always re-shows it.

### Key design decisions

**1. Deferred toggle instead of suppression-only.** `use-zoom-pan` `preventDefault()`s only the *second* tap's click — the first tap's click always lands. Acting on it immediately would flip the chrome at the start of every double-tap zoom. The 300ms deferral is the same trade every photo viewer makes (single-tap latency for double-tap disambiguation); two quick taps that *don't* zoom (text-slide margin, video poster) cancel each other, so a double-tap is never a toggle anywhere.

**2. Hidden chrome is `inert`, and focus is parked first.** `opacity-0` alone would leave invisible buttons tab-focusable and visible to AT, so the pill gets `inert`. But Radix focuses the first pill button on open, and making a focused subtree inert ejects focus to `<body>` — which Radix's dismissable layer reads as focus-outside and closes the dialog (found live; jsdom can't reproduce inert focus ejection). The toggle parks focus on the strip container (`tabIndex={-1}`) before flipping.

**3. Safe-area insets on the bars.** `index.html` sets `viewport-fit=cover`, so in standalone/PWA mode the notch and home indicator would overlap the action bar / filename bar. Both use `max(...)` with the env insets; on desktop the insets are 0 and the values collapse to the previous `0.75rem`/`1rem`.

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/image-gallery.chrome-toggle.test.tsx` | Mounts the real gallery (touch mocks): deferred single-tap hide/show, reset-on-reopen, double-tap never toggles, edge taps navigate, zoom keeps chrome + tap-while-zoomed toggles + zoom-out re-shows |

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/components/image-gallery.tsx` | Mobile full-bleed layout (dynamic viewport); `90vh` → `90dvh`; deferred tap-to-toggle chrome (zoom-aware) with `inert` + focus parking; safe-area insets; `85vh` → `85dvh` in video-embed sizing |
| `apps/frontend/src/hooks/use-zoom-pan.ts` | Export `DOUBLE_TAP_MS` (replaces inline `300`) so the gallery toggle and the double-tap detector share one window |
| `apps/frontend/src/components/gallery/pdf-viewer.tsx` | Move the tap-exclusion marker off the full-bleed wrapper so PDF slides keep the chrome toggle reachable (opus review finding) |

## Out of scope (deliberate)

- Desktop keeps its always-visible chrome (hover already governs the arrows); the toggle is touch-layout only.
- Pre-existing hybrid-device quirk observed while verifying: opening the gallery with a mouse and then touching it flips `useInputMode` mid-dialog and Radix registers the layout swap as a pointer-down-outside, closing the dialog. Reproduces without this diff; not addressed here.

## Test plan

- [x] `bun run test -- image-gallery.chrome-toggle image-gallery.reopen media-gallery-context` — 9 pass
- [x] Full pre-commit lint/typecheck suite clean
- [x] Live-verified on dev:test (scripted Playwright touch session): open → chrome visible; double-tap → zoomed (screenshot) with chrome **still visible**; single tap while zoomed → hidden; double-tap zoom-out → chrome **reappears**; single tap at 1× → hidden. Layout: mobile 390×844 full-bleed `(0,0)`; desktop `1152×720` = 90vw×90dvh; landscape 90dvh centered.
- [ ] Real-device check of dynamic browser chrome collapse/expand and real pinch gestures — desktop Chromium can't emulate either; pinch never yields a click, so it cannot reach the toggle path by construction.

---

### DIFF
--- a/apps/frontend/src/components/gallery/pdf-viewer.tsx
+++ b/apps/frontend/src/components/gallery/pdf-viewer.tsx
@@ -71,9 +71,11 @@ export function PdfViewer({ url, filename, variant = "gallery" }: PdfViewerProps
     return isInline ? (
       <div className="w-full">{renderer}</div>
     ) : (
-      <div data-gallery-text-viewer="true" className="absolute inset-0 pb-16 pt-14 sm:py-16">
-        {renderer}
-      </div>
+      // No data-gallery-text-viewer here: the marker lives on the renderer's
+      // inner scroll region (mirroring markdown/html/text) so the reserved
+      // pt-14/pb-16 strips stay valid chrome-toggle tap zones — marking this
+      // full-bleed wrapper made PDF slides untappable, stranding hidden chrome.
+      <div className="absolute inset-0 pb-16 pt-14 sm:py-16">{renderer}</div>
     )
   }
 
--- a/apps/frontend/src/components/image-gallery.chrome-toggle.test.tsx
+++ b/apps/frontend/src/components/image-gallery.chrome-toggle.test.tsx
@@ -1,0 +1,125 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
+import { render, screen, waitFor, within, fireEvent } from "@testing-library/react"
+import userEvent from "@testing-library/user-event"
+import { createMemoryRouter, RouterProvider } from "react-router-dom"
+import { MediaGalleryProvider } from "@/contexts"
+import { attachmentsApi } from "@/api"
+import { AttachmentList } from "@/components/timeline/attachment-list"
+import * as useMobileModule from "@/hooks/use-mobile"
+import * as inputModeModule from "@/hooks/use-input-mode"
+import * as touchCapableModule from "@/hooks/use-touch-capable"
+import type { AttachmentSummary } from "@threa/types"
+
+const WIDTH = 400
+let offsetWidthSpy: PropertyDescriptor | undefined
+
+beforeEach(() => {
+  vi.restoreAllMocks()
+  vi.spyOn(attachmentsApi, "getDownloadUrl").mockResolvedValue("https://example.com/img")
+  vi.spyOn(inputModeModule, "useInputMode").mockReturnValue("touch")
+  vi.spyOn(touchCapableModule, "useTouchCapable").mockReturnValue(true)
+  vi.spyOn(useMobileModule, "useIsMobile").mockReturnValue(true)
+  offsetWidthSpy = Object.getOwnPropertyDescriptor(HTMLElement.prototype, "offsetWidth")
+  Object.defineProperty(HTMLElement.prototype, "offsetWidth", { configurable: true, get: () => WIDTH })
+})
+
+afterEach(() => {
+  if (offsetWidthSpy) Object.defineProperty(HTMLElement.prototype, "offsetWidth", offsetWidthSpy)
+})
+
+const img = (id: string, filename: string): AttachmentSummary => ({
+  id,
+  filename,
+  mimeType: "image/png",
+  sizeBytes: 1024,
+})
+
+function renderGallery() {
+  const attachments = [img("img_1", "photo1.png"), img("img_2", "photo2.png"), img("img_3", "photo3.png")]
+  const router = createMemoryRouter(
+    [
+      {
+        path: "/s",
+        element: (
+          <MediaGalleryProvider>
+            <AttachmentList attachments={attachments} workspaceId="ws_1" />
+          </MediaGalleryProvider>
+        ),
+      },
+    ],
+    { initialEntries: ["/s"] }
+  )
+  render(<RouterProvider router={router} />)
+}
+
+/** The touch tap surface: the strip's clipping container, which owns onClick. */
+function tapContainer(): HTMLElement {
+  const dialog = screen.getByRole("dialog")
+  const strip = dialog.querySelector<HTMLElement>('div[style*="will-change"]')
+  const container = strip?.parentElement
+  if (!container) throw new Error("tap container not found")
+  // jsdom rects are all zeros; the tap handler derives its left/right/middle
+  // zone from the container rect, so give it a real one.
+  container.getBoundingClientRect = () =>
+    ({
+      left: 0,
+      top: 0,
+      right: WIDTH,
+      bottom: 800,
+      width: WIDTH,
+      height: 800,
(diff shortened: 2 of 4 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good

Found 1 issue:

1. `apps/frontend/src/components/gallery/pdf-viewer.tsx:74` — On PDF slides the chrome toggle is unreachable: the gallery PDF path puts `data-gallery-text-viewer` on the full `absolute inset-0` wrapper (both the pdf.js path at `pdf-viewer.tsx:74` and the iframe fallback marks content filling the same wrapper), so every tap on the slide is excluded from the toggle. Markdown/html/text put the marker on the inner panel, leaving the reserved `pt-14`/`pb-16` strips tappable. Repro: hide chrome on an image slide, swipe to a PDF — no tap re-shows the chrome (close button is inert; swipe-down dismiss and swiping back still work). Pre-existing marker placement, but this PR amplifies it from "no tap-nav on PDFs" into "chrome can be stuck hidden on PDFs" (ux-consistency: violates the stated "tap again → the chrome should reappear" on one slide type)
   https://github.com/threahq/threa/blob/2b2a2765c140828d6b9611c6344bb391afb5b58d/apps/frontend/src/components/gallery/pdf-viewer.tsx#L71-L77

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>Full-bleed <sm via base DialogContent inset-0; ≥sm centered at 90dvh; safe-area insets backed by viewport-fit=cover in index.html; single-tap toggle deferred past the shared DOUBLE_TAP_MS; zoom transitions cancel pending toggles and zoom-out re-shows chrome; reset-on-open. Every stated requirement is implemented.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Traced: single tap (schedule→fire), double-tap zoom (tap-1's click schedules, tap-2's touchend commits zoom → onZoomChange cancels the pending toggle within the 300ms window; tap-2's click is preventDefault'd), double-tap on non-zoom regions (second click hits the cancel branch), tap while zoomed (nav disabled → toggle), close-with-pending-timer (null-safe callback + reset on next open), goTo→zoomableRef.reset() at scale 1 does not fire onZoomChange (no spurious chrome show). No stale closures; live reads go through refs.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Invariants verified: chrome visible on every open (layout effect cancels pending + sets visible before paint); a zoom transition never hides chrome (handleZoomChange only cancels or shows); at most one pending toggle (schedule only when the ref is null; second tap cancels).</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>Comments are load-bearing why (INV-25); opacity+inert avoids layout shift (INV-21); no toasts (INV-63 n/a); DOUBLE_TAP_MS centralized instead of a duplicated magic 300 (INV-33).</details>
<details><summary>🏗️ Design [CLEAN]</summary>handleZoomChange stable via useCallback (no re-render regression vs the previous bare setIsZoomed); toggle state machine is three refs + one state, no parallel implementation.</details>
<details><summary>🔒 Security [CLEAN]</summary>Frontend gesture/layout change; no data flow from user input to any sink.</details>
<details><summary>🎛️ UX [1 issue]</summary>The PDF-slide toggle gap above. Also verified: user can always exit with chrome hidden (tap-to-reshow on image/video/text slides, swipe-down dismiss everywhere, Radix Escape unaffected by inert); the 300ms single-tap deferral applies only to the toggle — edge-zone navigation stays immediate.</details>
<details><summary>📱 Mobile [CLEAN]</summary>tailwind-merge (v3.5) confirmed to drop the base max-sm:p-4/gap-4/overflow-y-auto against the caller's same-prefix cancellations; touch-laptop (isMobile false, touch input) still populates containerRef so focus parking works; sub-threshold note: at ≥sm the centered card measures safe-area insets from the viewport edge, not the card edge — cosmetic over-padding on notched landscape phones only.</details>

PR #1298, opened 2026-07-12
feat(board): gold-thread flash on the viewer's own just-posted card

kristofferremback:
## Problem

When you post from the board's overlay composer, your own card just appears at the top of the feed — indistinguishable from any other card, and easy to lose track of in a busy feed. The reveal is correct but silent: no signal that *this* is the thing you just wrote. The board deserves a small moment of delight on its primary action, in Threa's own visual language (the Ariadne golden thread), without being flashy.

## Solution

A brief warm pulse in the primary/golden-thread color rings the viewer's own just-posted card, then fades. It rides the existing post→reveal path — the id of the new conversation is threaded from the mutation through the compose-overlay store to the board's reveal handler, which flashes the matching card for ~1.6s.

### How it works

1. `useCreateBoardPost` now **returns the conversation id** on both branches — the client-minted id for a new-scratchpad target, the server `conversationId` for a stream target (previously both returned `undefined`). This is the id the flash keys on, and it matches the optimistic card's `conversation.id`.
2. `compose-overlay-store` threads it through: `notifyComposePosted(conversationId)` → the registered `onPosted(conversationId)`. The store's post-success handler ref is now `(conversationId?: string) => void`.
3. `BoardOverlayComposerBody` captures the mutation's return and passes it to `onPosted(postedConversationId)` after closing the overlay.
4. `board.tsx` holds `flashConversationId` in state for **1600ms** (a hair past the 1400ms pulse so it completes), clears the prior timer if you post again mid-flash, and clears the timer on unmount. It passes `flash={row.post.conversation.id === flashConversationId}` to each `BoardCard`, and adds `flashConversationId` to the `renderedRows` memo deps.
5. `BoardCard` gains a `flash?` prop that toggles the `.board-post-flash` class → a 1.4s `outline` keyframe that ripples the ring outward (`outline-offset` 0→7px) and fades the color (primary/0.8→primary/0).

### Key design decisions

**1. `outline`, not `box-shadow` or `ring`** — the card carries a resting drop shadow. Animating a `box-shadow` would fight it and animating a border/ring would shift layout. `outline` sits outside the box model, follows the card's `border-radius` in modern browsers, and disturbs neither the shadow nor the layout (INV-21).

**2. Flash keys on the conversation id, not "the top card"** — the viewer's post doesn't always land at the top (lens/filter ordering), and re-renders shouldn't replay the pulse. Keying the animation on an id that must equal exactly one card's `conversation.id`, held in state and then cleared, makes it fire once for the right card and never re-fire.

**3. Reuses the post-reveal channel already in place** — no new store, no new event. The overlay already tells the board "a post landed so it can surface"; this only adds *which* post to that same signal, so a post from a non-board surface (where nothing is registered) still simply lands with no flash.

**4. Respects `prefers-reduced-motion`** — the keyframe is disabled under reduced-motion; the card just appears, as before.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/frontend/src/hooks/use-conversations.ts` | `useCreateBoardPost` returns the conversation id on both post branches |
| `apps/frontend/src/stores/compose-overlay-store.ts` | `onPosted` handler + `notifyComposePosted` carry the conversation id |
| `apps/frontend/src/components/board/board-overlay-composer.tsx` | Capture the mutation's return, pass it to `onPosted(id)` |
| `apps/frontend/src/pages/board.tsx` | `flashConversationId` state + timer, `flash` prop on cards, memo dep |
| `apps/frontend/src/components/board/board-card.tsx` | `flash?` prop → `.board-post-flash` class on the card root |
| `apps/frontend/src/index.css` | `.board-post-flash` + `@keyframes board-post-flash` + reduced-motion guard |
| `apps/frontend/src/hooks/use-conversations.test.tsx` | Assert the mutation resolves with the conversation id (was `undefined`) |
| `apps/frontend/src/stores/compose-overlay-store.test.ts` | Assert `onPosted` receives the id |
| `apps/frontend/src/components/board/board-overlay-composer.test.tsx` | `mutateAsync` resolves an id; assert `onPosted("conv_1")` |

## Out of scope (deliberate)

- **Timeline / fullscreen editor** — the flash is a board-feed affordance; the timeline shows your sent message inline already. Not wired there.
- **Secondary board-oriented sidebar** — discussed and shelved (the current sidebar is timeline-based; a board-first nav is a separate, larger idea).

## Test plan

- [x] `apps/frontend` — `use-conversations`, `compose-overlay-store`, `board-overlay-composer`, `board` suites — 54 pass
- [x] `tsc --noEmit` across all workspaces (pre-commit hook) — clean
- [x] Visual QA on the dev:test stack (Playwright): posted to a #design channel, captured the gold ring rippling around the new card at ~t480ms and faded by ~t980ms — pulse fires on the correct card, subtle, on-brand
- [x] `prefers-reduced-motion`: keyframe disabled, card appears without the pulse (CSS guard)

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1298"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786445713&installation_id=129946197&pr_number=1298&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1298&signature=518c17f8d4c834ac4b10c6cf877bb12183daab3ecf50b77a1de8de82f1687678"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/auth/account-scope.tsx
+++ b/apps/frontend/src/auth/account-scope.tsx
@@ -12,6 +12,7 @@ import { resetStreamStoreCache } from "@/stores/stream-store"
 import { resetDraftStoreCache } from "@/stores/draft-store"
 import { resetShareHandoffStoreCache } from "@/stores/share-handoff-store"
 import { resetComposeOverlayStoreCache } from "@/stores/compose-overlay-store"
+import { resetBoardFlashStoreCache } from "@/stores/board-flash-store"
 import { resetSnippetRequestStoreCache } from "@/stores/snippet-request-store"
 import { resetConversationReplyOpenStoreCache } from "@/stores/conversation-reply-open-store"
 import { resetE2eSessionStoreCache } from "@/stores/e2e-session-store"
@@ -75,6 +76,7 @@ function flushModuleStoreCaches(): void {
   resetConversationReplyOpenStoreCache()
   resetE2eSessionStoreCache()
   resetComposeOverlayStoreCache()
+  resetBoardFlashStoreCache()
   resetRevealGate()
 }
 
--- a/apps/frontend/src/components/board/board-card.test.tsx
+++ b/apps/frontend/src/components/board/board-card.test.tsx
@@ -24,6 +24,7 @@ import * as contextsModule from "@/contexts"
 import * as queueDraftModule from "@/hooks/use-queue-draft-message"
 import * as inlineComposerModule from "@/components/board/board-inline-composer"
 import * as inputModeModule from "@/hooks/use-input-mode"
+import { setBoardFlash, resetBoardFlashStoreCache } from "@/stores/board-flash-store"
 import { spyOnExport } from "@/test/spy"
 import { MESSAGE_ROW_HEAD_PADDING } from "@/components/message/message-row-layout"
 
@@ -191,6 +192,38 @@ describe("BoardCard unread dot", () => {
     mountCard()
     await screen.findByText("Opening body.")
     expect(screen.queryByLabelText("Unread")).toBeNull()
+  })
+})
+
+describe("BoardCard post flash", () => {
+  beforeEach(() => {
+    vi.spyOn(conversationReadModule, "useConversationReadController").mockReturnValue({
+      value: readValue,
+      hasUnread: () => false,
+      markReadSilently: () => Promise.resolve(),
+      setExplicitUnreadListener: () => {},
+      getReadTruth: () => ({ lastReadSequence: null, readMessageIds: [] }),
+    })
+  })
+  afterEach(() => resetBoardFlashStoreCache())
+
+  it("rings the card while its own conversation is flashing, then clears", async () => {
+    mountCard()
+    const body = await screen.findByText("Opening body.")
+    expect(body.closest(".board-post-flash")).toBeNull()
+
+    act(() => setBoardFlash("conv_1"))
+    expect(body.closest(".board-post-flash")).not.toBeNull()
+
+    act(() => setBoardFlash(null))
+    expect(body.closest(".board-post-flash")).toBeNull()
+  })
+
+  it("does not ring a card whose conversation isn't the flashed one", async () => {
+    mountCard()
+    const body = await screen.findByText("Opening body.")
+    act(() => setBoardFlash("conv_other"))
+    expect(body.closest(".board-post-flash")).toBeNull()
   })
 })
 
--- a/apps/frontend/src/components/board/board-card.tsx
+++ b/apps/frontend/src/components/board/board-card.tsx
@@ -56,6 +56,8 @@ interface BoardCardProps {
   scrollerRef?: RefObject<HTMLDivElement | null>
   /** virtua's imperative handle for the board feed, paired with `scrollerRef`. */
   listRef?: RefObject<VirtualizerHandle | null>
+  /** Play the one-shot gold-thread pulse — the viewer's own just-posted card. */
+  flash?: boolean
 }
 
 /** How many trailing messages a nested branch previews on a collapsed card; the
@@ -85,6 +87,7 @@ export function BoardCard({
   dmPeerUserId,
   scrollerRef,
   listRef,
+  flash,
 }: BoardCardProps) {
   const { conversation } = post
   const { getActorName, getActorAvatar } = useActors(workspaceId)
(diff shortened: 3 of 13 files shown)


kristofferremback:
### Code review
**Confidence: 5/7** — Good (three non-blocking issues found; all fixed in `b533f56b`)

Reviewed at `f60cc7a7`. Found 3 issues — all addressed in the follow-up commit `b533f56b`:

1. `apps/frontend/src/pages/board.tsx:707-716` — Adding `flashConversationId` to the `renderedRows` memo deps reintroduced the exact regression that memo's own comment says it exists to prevent: it rebuilt all ~430 card elements **twice per post** (set + 1.6s clear), and since `BoardCard` isn't `React.memo`'d, every mounted card re-rendered both times. (performance — flagged by both the spec/design and mobile lenses.)
   https://github.com/threahq/threa/blob/f60cc7a7fc3f05326d2d0f1b5935581c7385e13f/apps/frontend/src/pages/board.tsx#L707-L716
   **Fixed:** the flashed id now lives in a module store (`board-flash-store`) that each `BoardCard` subscribes to by its own id via `useBoardFlash`, so only the one card whose flashed-ness flips re-renders. `flashConversationId` is gone from the memo.

2. `apps/frontend/src/pages/board.tsx:574-592` — On the board-home lens (the common case, `currentViewSurfacesOwnPost === true`), `handlePosted` doesn't navigate, so nothing scrolls the virtualized feed to top. A viewer who'd scrolled down before posting would get the new card prepended off-screen — and the golden-thread flash, the whole point of the change, would play invisibly. (ux-feedback.)
   https://github.com/threahq/threa/blob/f60cc7a7fc3f05326d2d0f1b5935581c7385e13f/apps/frontend/src/pages/board.tsx#L574-L592
   **Fixed:** the stay-put branch now jumps `scrollerRef.current.scrollTop = 0` (mirroring `revealNew` and the `resetKey` layout effect), so the reveal + flash are on-screen regardless of prior scroll.

3. `apps/frontend/src/components/board/board-card.tsx` — No automated test exercised the visible flash itself (the plumbing was covered, the render behavior wasn't). (missing-change / test-coverage.)
   **Fixed:** added `board-flash-store.test.ts` (per-id flash, move-to-newer, account-switch reset) and a `BoardCard` test asserting `.board-post-flash` tracks the store for the card's own conversation id.

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>The gold-flash is wired end-to-end as intended; the mutation return, store threading, and card prop all line up.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>Correctness lens traced the full path: timer lifecycle (ref survives re-renders, cleared on unmount + on re-post), id-match on both the new-scratchpad (client-minted) and stream-target (server) branches, optimistic-write-throws path, mount timing (optimistic row awaited before the handler fires), and route-remount (single `board/:lens?` route, no `key`, so `navigate(boardHome)` doesn't remount) — all sound.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Conversation id round-trips mutation → `notifyComposePosted` → `onPosted` → flash target → the one matching card. `reconcileOptimisticBoardPost` patches fields on the same keyed row, never swaps the id, so echo/refetch can't invalidate the flash target.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>INV-21 (outline/outline-offset animate without reflow), INV-63 (in-place signal, not a toast), INV-9 (new module store's reset registered in `flushModuleStoreCaches`) all respected.</details>
<details><summary>🏗️ Design [1 concern, fixed]</summary>The renderedRows-memo re-render regression (issue 1).</details>
<details><summary>🔒 Security [CLEAN]</summary>No user-input flow, no injection surface.</details>
<details><summary>🎛️ UX [1 issue, fixed]</summary>The off-screen-flash-when-scrolled-down gap (issue 2). Otherwise: no annoying replay on unrelated re-renders, 1.6s clear debounced + cleaned up, reduced-motion leaves a fully transparent (literal `transparent`) outline, consistent across board-FAB and quick-switcher entry points.</details>
<details><summary>📱 Mobile [CLEAN]</summary>outline-offset 7px ripple stays within the `px-2` mobile gutter and fades to near-transparent before its widest extent — no clipping or horizontal overflow; drawer-close eats only a few hundred ms of the 1.6s hold; not hover-gated; reduced-motion correct.</details>

PR #1342, opened 2026-07-14
feat(frontend): conversation panel docks the real composer; branch replies arm it with a dismissible chip

kristofferremback:
## Problem

Kris (with video): "The composer on the conversations view is still not correct. It shouldn't load the button as it does right now, it should just load the composer as usual. … When replying to a sub conversation show the chip we have saying we're replying to a sub conversation, though that chip needs a cancel cross, which when tapped moves the draft (if any) to the root conversation, at which point it just shows the standard composer again."

Three defects behind that:

1. On mobile the conversation panel's footer rendered a resting "Write a reply…" **button** (and a "✎ draft" pill) instead of the composer; tapping it mounted an editor mid-flow.
2. **Two disagreeing "mobile" definitions**: the panel goes full-screen on `useSidebar().isMobile` (viewport < 640px OR coarse pointer) while `InlineComposerForm`'s floating gate used bare `useIsMobile()` (viewport only) — on a coarse-pointer device ≥ 640px the panel was full-screen but the reply editor rendered **mid-screen in the message flow**, the exact weirdness in the video.
3. Replying to a sub-conversation mounted a separate inline editor under the branch instead of retargeting the panel's composer.

## Solution

The panel footer now permanently mounts the real composer on every device — desktop always-open (unchanged, round-5 semantics), mobile shows `MessageComposer`'s own collapsed⇄focused bar (timeline parity). Sub-conversation replies **arm** that docked composer with a dismissible "Replying in {branch}" strip; the × relocates the draft to the root conversation.

### How it works

1. **`docked` prop on `InlineComposerForm`** — suppresses the floating-portal path (the footer IS the dock) and passes `initialMobileChromeOpen={false}` so mobile rests as the collapsed bar. `BoardReplyComposer` gained `alwaysDocked` (replacing `desktopAlwaysOpen`); the panel passes it, board cards keep their resting-button model.
2. **One mobile predicate** — new `useIsMobileOrCoarse()` in `hooks/use-pointer.tsx` is now consumed by BOTH `useSidebar().isMobile` and the composer's floating gate, so the two surfaces provably can't drift. The duplicate `useIsCoarsePointer` in `use-mobile.tsx` was deleted (INV-35/38); `use-auto-mark-as-read.ts` repointed to the canonical hook.
3. **The chip is the timeline's chip** — `ConversationReplyStrip` extracted from `message-input.tsx` into `components/composer/` and reused verbatim by the timeline and the panel (`Layers` + "Replying in **{title}**" + ×, `aria-label="Cancel reply in conversation"`).
4. **Arming** — a branch tail's Reply (and branch-scoped `?stash=` deep links) call `onArmBranchReply` instead of mounting an inline form; `ConversationPanelBody` holds the armed state and retargets the footer form's `draftKey` (`board:branch-reply:<id>`), send routing, and `scheduleTarget`. Pending branches keep the inline form (no real conversation id to arm against — brief-sanctioned).
5. **Cancel × = `relocateLoadedDraft`** (new, `use-draft-message.ts`): capture the live editor content → **tombstone-delete** the branch draft → mint a **fresh** loaded draft under the root scope (stashing any existing root draft first — no-loss). Content visibly survives into the root composer.

### Design evolution (self-review — a real data-integrity bug found and fixed live)

The workflow's original cancel was flush → `rescopeScopeDrafts` → disarm. Live verification against a real backend showed the draft **resurrecting under the branch scope after reload** — twice, for two independent reasons:

- `rescopeScopeDrafts` moved the IDB rows but not the **localStorage staging buffer**; the startup reconcile "recovered" the orphaned staged entry under the old scope. Fixed in `rescopeScopeDrafts` (the buffer now follows the move) — this also covered the pre-existing subtopic→branch migration.
- Deeper: the queue worker claims a push op and snapshots the row in separate transactions, so an in-flight pre-move push whose claim isn't yet visible gets coalesced onto — and its completion deletes the only pending marker, losing the move server-side **forever** (probe showed the server row permanently branch-scoped while local said root). No enqueue-layer ordering wins that race — the in-flight HTTP can always land last. `enqueueDraftUpsert` gained `forceNewOp` (lineage-carrying supersede) for `rescopeScopeDrafts`, and the cancel flow switched to `relocateLoadedDraft`, which sidesteps the class entirely: the delete tombstone (`supersededWriteIds`) exists precisely to suppress late in-flight upserts, and the fresh target row has no version history to fight.

Each step was verified with a live probe polling the server drafts API through the cancel: final state is exactly one row, root-scoped, stable across reload.

## Modified files

| File | Change |
| ---- | ------ |
| `components/board/board-inline-composer.tsx` | `docked`, `replyTarget` + cancel-relocate, floating gate on shared predicate, fullscreen-expand gate on `floating` |
| `components/board/board-reply-composer.tsx` | `alwaysDocked` (both platforms), `ArmedReply` retargeting (draftKey/streamId/scheduleTarget/send) |
| `components/board/use-inline-branch-composer.tsx` | `onArmBranchReply`; non-pending branch Reply arms the footer; `queueExistingBranchReply` extracted |
| `components/conversations/conversation-panel.tsx` | armed state + `focusSeq`, `alwaysDocked` footer, stale `?stash=` comment rewritten |
| `components/composer/conversation-reply-strip.tsx` | New: the shared dismissible reply-target strip |
| `components/timeline/message-input.tsx` | Consumes the extracted strip (pure extraction) |
| `hooks/use-pointer.tsx` | `useIsMobileOrCoarse()`; `use-mobile.tsx` duplicate hook deleted; `sidebar-context.tsx` + `use-auto-mark-as-read.ts` repointed |
| `hooks/use-draft-message.ts` | `relocateLoadedDraft`; `rescopeScopeDrafts` moves the staging buffer + forces a lineage-carrying op |
| `sync/draft-sync.ts` | `enqueueDraftUpsert({ forceNewOp })` |
| 8 test files | New coverage: docked-never-floats, coarse-tablet floats, armed strip + cancel-relocate, retargeting, relocate semantics (fresh row / tombstone / stash-existing), staging-buffer move, lineage supersede |

## Out of scope (deliberate)

- Board **cards** keep the resting-button + inline branch composers (round-5 ruling: "board cards and inline sub-conversations" are the exceptions). Their visual parity pass is the next stacked PR.
- After a successful branch send the footer **stays armed** to that branch (the timeline disarms) — you're still "in" that sub-topic; × is the deliberate way back. Flagged for Kris.
- New-sub-topic gestures keep the inline form (no target exists until send).

## Test plan

- [x] From `apps/frontend`: board + conversations + composer + timeline + hooks + sync — 1832 pass, 0 fail; `tsc --noEmit` clean
- [x] Opus 2-lens verify (2 rounds): predicate-duplication finding fixed by consolidating to `useIsMobileOrCoarse` everywhere; stale-comment + scratch-file minors fixed by coordinator
- [x] Live-verified on dev:test (scripted Playwright, seeded branch via `newSubtopic` directive), 12/12: panel has NO resting button and a mounted editor (desktop + 390px); branch Reply arms the chip; × disarms, content survives visibly; **server-truth probe**: post-cancel exactly one draft row under `board:reply:<root>`, stable across reload (the two resurrection bugs above each reproduced before their fix and gone after)
- [x] Mobile 390px: collapsed bar at rest, focused chrome shows send, armed chip renders (the implementer's flagged focus/collapse race did not manifest)
- [ ] `src/lib/dates.test.ts` — 4 pre-existing TZ failures on main, untouched

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1342"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786579644&installation_id=129946197&pr_number=1342&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1342&signature=76f9994cef3a8aada5bd655312ac563e2fb855432319e86d8e1ca4dcb2bfcc6f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


kristofferremback:
Collapsed and merged into main via #1343 (squash of #1342 + #1343 into one commit, per Kris's request to collapse the stack). Closing without a separate merge — this PR's changes are already in main.

PR #1358, opened 2026-07-16
fix(scheduled-messages): Send Now fires future rows regardless of clock skew + editor gains Send now/Delete

kristofferremback:
## Problem

Send Now on a scheduled message silently did nothing. On a future-scheduled row, tapping Send Now (composer popover, mobile action drawer, or `/scheduled` page) appeared to no-op — the only reliable way to send early was to edit the scheduled time into the past.

Root cause is in the backend, `ScheduledMessagesService.sendNow` (`apps/backend/src/features/scheduled-messages/service.ts`). It forced a future row by rewriting `scheduled_for = new Date()` (the **app server** clock), then re-ran the worker CAS `tryStartSend`, whose `WHERE` requires `scheduled_for <= NOW()` (the **DB** clock). When the app clock runs even a few hundred ms ahead of Postgres — normal on Railway, where app and DB are separate containers — the rewritten instant lands just past the DB's `NOW()`, the CAS matches zero rows, and the method throws `409 SCHEDULED_MESSAGE_NOT_PENDING`. The composer popover/drawer call `sendNowMutation.mutate(id)` with no `onError`, so the 409 was swallowed → "nothing happens." Editing the time to the past works because a past `scheduled_for` clears the guard regardless of skew.

Secondary UX issue (mobile): the composer's scheduled popover rendered the desktop Send/Edit/Cancel triplet **persistently** on touch (`reveal-actions`, `opacity:1` for touch), cluttering the row, while the edit dialog offered no Send-now or Delete.

## Solution

Claim the row for sending in a single CAS anchored on the DB clock, and stop rewriting `scheduled_for` at all.

### How it works

**Backend.** `ScheduledMessagesRepository.tryStartSend` gains a `force?: boolean`. When true, the SQL guard becomes `AND (${force}::boolean OR scheduled_for <= NOW())` — the schedule check short-circuits. `sendNow` now does one `tryStartSend({ bypassFence: true, force: true })`: `force` ignores the (future) scheduled time, `bypassFence` ignores the user's own edit-dialog lock, and the transition is evaluated entirely on the DB clock. A `null` return now unambiguously means the row left `pending` (worker won / cancelled / already sent) → `409 NOT_PENDING`. The old rewrite-and-retry block (update `scheduled_for`, re-CAS, second `expectedVersion` failure path) is deleted. The worker `fire` path and the past-time `update` path pass no `force`, so their `scheduled_for <= NOW()` guard is unchanged.

**Frontend — composer popover** (`scheduled-messages-picker.tsx`). The inline action cluster switches from `reveal-actions` to `reveal-actions-hover-only`: hidden (and `pointer-events:none`) on touch, hover/focus-revealed for a mouse. On mobile the row now relies on tap → edit dialog and long-press → action drawer, matching the row's own doc comment.

**Frontend — edit dialog** (`scheduled-edit-dialog.tsx`). Adds Delete + Send now beside Save/Send on both desktop and mobile via a shared, exported `ScheduledEditActions` cluster (module-level, honors INV-18). Send now persists the current editor content first (`updateMutation`), then force-sends (`sendNowMutation`); if the save itself already sent the row (past-time Save=Send path), the second call is skipped. Delete routes through `useCancelScheduled` (handles local placeholders). The overloaded "Send original" secondary button and the footer "Cancel" are dropped — desktop `DialogContent` has its own close X, mobile `Drawer` dismisses by swipe/outside-tap, both routing through `handleClose` (which releases the edit fence).

### Key design decisions

**1. Force the CAS instead of rewriting `scheduled_for`.** The rewrite coupled the send to app↔DB clock agreement; the send has no reason to depend on wall-clock at all once the user explicitly asked for it. Forcing the guard off removes the skew surface *and* a full round-trip and the fragile double-`expectedVersion` failure path.

**2. Only `sendNow` forces.** The worker `fire` and past-time `update` still require `scheduled_for <= NOW()` — a forced worker tick would defeat the whole scheduling mechanism. `force` is opt-in per call, mirroring the existing `bypassFence` shape.

**3. Dialog "Send now" saves-then-sends.** The dialog is where content is edited, so Send now must ship what the user sees, not the last-saved copy. Save-then-force-send reuses the reliable past-time atomic send when the time is already past and otherwise force-sends.

## Modified files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/scheduled-messages/repository.ts` | `tryStartSend` gains `force` — drops the `scheduled_for <= NOW()` guard via an OR branch |
| `apps/backend/src/features/scheduled-messages/service.ts` | `sendNow` is one forced CAS; deletes the `scheduled_for` rewrite/re-CAS block |
| `apps/frontend/src/components/composer/scheduled-messages-picker.tsx` | Popover row inline actions `reveal-actions` → `reveal-actions-hover-only` (hidden on touch) |
| `apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx` | Adds Delete + Send now via shared `ScheduledEditActions`; drops "Send original"/Cancel; `handleSendNow` save-then-send |

## New files

| File | Purpose |
| ---- | ------- |
| `apps/frontend/src/components/scheduled/scheduled-edit-actions.test.tsx` | Gating of the dialog action cluster (pending → Delete/Send now/Save; failed → Remove only; `canSubmit` disables writes) |

## Out of scope (deliberate)

- The `/scheduled` full-page row (`ScheduledItem`) keeps its persistent touch icons + long-press drawer — the report was about the composer popover, and that page's list rows don't open an editor on tap the same way. Can follow up if the same touch-declutter is wanted there.
- No new integration test for the clock-skew path itself: it only manifests when app and DB clocks diverge, which a single-host test harness can't reproduce. The behavior is pinned at the repository layer (force drops the guard) and the service layer (single forced CAS, no rewrite).

## Test plan

- [x] Backend `bun test src/features/scheduled-messages` — 41 pass (adds: repo `force` guard discriminator; service `sendNow` single-forced-CAS + no-rewrite + NOT_PENDING-on-lost-race)
- [x] Frontend `vitest run scheduled-edit-actions.test.tsx` — 3 pass
- [x] Frontend `vitest run scheduled-messages-picker.test.tsx use-scheduled.test.ts` — 5 pass (unaffected, confirms no regression)
- [x] Frontend `vitest run no-happy-path-success-toast.test.ts` — 1 pass (no `toast.success` added)
- [x] `tsc --noEmit` clean — backend (0 errors) and frontend
- [ ] Live verification deferred to **staging** — the fix is only observable where app and DB run on separate clocks; localhost shares one, so the bug can't reproduce here. Add the `staging` label and test Send Now on a future-scheduled message from mobile.

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1358"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786776319&installation_id=129946197&pr_number=1358&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1358&signature=7c2793c873e197dd5537f6e7639e98ec52e49ceba9bd01f77d33d061bb050357"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/backend/src/features/scheduled-messages/repository.test.ts
+++ b/apps/backend/src/features/scheduled-messages/repository.test.ts
@@ -256,6 +256,28 @@ describe("ScheduledMessagesRepository.tryStartSend (worker CAS)", () => {
 
     expect(captured.values).toContain(true)
   })
+
+  it("drops the scheduled_for guard when force=true so Send Now fires a still-future row regardless of clock skew", async () => {
+    // Send Now must not depend on scheduled_for <= NOW(): the row is
+    // future-scheduled and the user asked to send it now. `force` short-
+    // circuits the schedule check in SQL (the OR branch) rather than the
+    // service rewriting scheduled_for to a JS `new Date()` and racing DB
+    // clock skew.
+    const captured: Captured = { text: null, values: null }
+    const db = createQuerier(captured)
+
+    await ScheduledMessagesRepository.tryStartSend(db, {
+      workspaceId: "ws_1",
+      id: "sched_01",
+      ttlSeconds: 10,
+      force: true,
+    })
+
+    // The guard survives only as an OR branch behind the force flag.
+    expect(captured.text).toContain("OR scheduled_for <= NOW()")
+    // force=true is passed (bypassFence defaults false), so a `true` appears.
+    expect(captured.values).toContain(true)
+  })
 })
 
 describe("ScheduledMessagesRepository.releaseEditFence", () => {
--- a/apps/backend/src/features/scheduled-messages/repository.ts
+++ b/apps/backend/src/features/scheduled-messages/repository.ts
@@ -357,18 +357,28 @@ export const ScheduledMessagesRepository = {
    * worker defers because of the fence, and the user's PATCH 409s with
    * NOT_PENDING because of the same fence.
    *
+   * `force: true` additionally drops the `scheduled_for <= NOW()` guard so a
+   * user-initiated Send Now fires a still-future row immediately. This anchors
+   * the send on the DB clock (the CAS) rather than a JS `new Date()`: forcing
+   * a future row by rewriting `scheduled_for = new Date()` and re-running the
+   * guard raced app↔DB clock skew — an app clock a few hundred ms ahead of
+   * Postgres put the rewritten time just past the DB's NOW() and the send
+   * silently 409'd. Only the user-initiated `sendNow` path forces; the worker
+   * and the past-time save keep the guard.
+   *
    * Returns null on:
    *  - status not pending (cancelled, already sending/sent, failed)
-   *  - scheduled_for in the future (stale leased queue row that survived a
-   *    reschedule cancel — see service.fire's pre-check comment)
+   *  - scheduled_for in the future AND `force` is false (stale leased queue
+   *    row that survived a reschedule cancel — see service.fire's pre-check)
    *  - edit_active_until in the future AND `bypassFence` is false (worker
    *    path only; an editor is active, defer)
    */
   async tryStartSend(
     db: Querier,
-    params: { workspaceId: string; id: string; ttlSeconds: number; bypassFence?: boolean }
+    params: { workspaceId: string; id: string; ttlSeconds: number; bypassFence?: boolean; force?: boolean }
   ): Promise<ScheduledMessage | null> {
     const bypassFence = params.bypassFence ?? false
+    const force = params.force ?? false
     const result = await db.query<ScheduledMessageRow>(sql`
       UPDATE scheduled_messages SET
         status = ${ScheduledMessageStatuses.SENDING},
@@ -379,7 +389,7 @@ export const ScheduledMessagesRepository = {
       WHERE id = ${params.id}
         AND workspace_id = ${params.workspaceId}
         AND status = ${ScheduledMessageStatuses.PENDING}
-        AND scheduled_for <= NOW()
+        AND (${force}::boolean OR scheduled_for <= NOW())
         AND (
           ${bypassFence}::boolean
           OR edit_active_until IS NULL
--- a/apps/backend/src/features/scheduled-messages/service.test.ts
+++ b/apps/backend/src/features/scheduled-messages/service.test.ts
@@ -443,6 +443,60 @@ describe("ScheduledMessagesService.update (optimistic CAS)", () => {
   })
 })
 
+describe("ScheduledMessagesService.sendNow", () => {
+  afterEach(() => mock.restore())
+
+  it("forces a still-future row to send now in one CAS, without rewriting scheduled_for", async () => {
+    // Regression: sendNow used to force a future row by rewriting
+    // scheduled_for = new Date() and re-running the worker CAS (which
+    // requires scheduled_for <= NOW()). An app clock a few hundred ms ahead
+    // of Postgres put the rewritten instant just past the DB's NOW(), the
+    // reclaim CAS matched nothing, and the send silently 409'd. The fix
+    // forces the guard off in a single CAS anchored on the DB clock — no
+    // scheduled_for rewrite, no second round-trip.
+    const service = setupService()
+    const row = fakeScheduled({ version: 2, scheduledFor: FUTURE, queueMessageId: "q_1" })
+
+    spyOn(ScheduledMessagesRepository, "findById").mockResolvedValue(row)
+    const update = spyOn(ScheduledMessagesRepository, "update").mockResolvedValue(row)
+    const tryStartSend = spyOn(ScheduledMessagesRepository, "tryStartSend").mockResolvedValue({
+      ...row,
+      status: ScheduledMessageStatuses.SENDING,
+    })
+    spyOn(ScheduledMessagesRepository, "markSent").mockResolvedValue({
+      ...row,
+      status: ScheduledMessageStatuses.SENT,
+      sentMessageId: "msg_42",
+    })
+    const cancelById = spyOn(QueueRepository, "cancelById").mockResolvedValue(undefined as any)
+    spyOn(OutboxRepository, "insert").mockResolvedValue({} as any)
+
+    const result = await service.sendNow({ workspaceId: WORKSPACE_ID, userId: USER_ID, id: SCHEDULED_ID })
+
+    expect(result.status).toBe(ScheduledMessageStatuses.SENT)
+    // One CAS, forced past both the schedule guard and the user's own fence.
+    expect(tryStartSend).toHaveBeenCalledTimes(1)
+    expect(tryStartSend).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({ force: true, bypassFence: true })
+    )
+    // No scheduled_for rewrite — the clock-skew source is gone.
+    expect(update).not.toHaveBeenCalled()
+    // The pending queue job is cancelled so the worker can't double-fire.
+    expect(cancelById).toHaveBeenCalledWith(expect.anything(), "q_1")
+  })
+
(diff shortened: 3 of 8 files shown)


kristofferremback:
### Code review — superseded

This review has been superseded by an updated pass after fixes: https://github.com/threahq/threa/pull/1358#issuecomment-4990017627

<details><summary>Original round-1 review</summary>


### Code review
**Confidence: 4/7** — Acceptable (core fix is clean; one real mobile-layout breakage + three UX-consistency nits)
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: sonnet x4 (spec+design, correctness+data-flow, UX, mobile)

The backend clock-skew fix (forced CAS), the dialog save-then-send flow, and the double-send guard were traced end-to-end and are **correct** — no bug/data-flow/security findings. Findings are mobile layout + cross-surface UX consistency.

Found 4 issues:

1. `apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx` (mobile Drawer footer) — **mobile-layout**: the footer `EditorActionBar` is a single non-wrapping `flex` row (no `flex-wrap`/`overflow-x-auto`). On the mobile drawer it carries 5 left icons (expand, format, emoji, mention, attach ≈ 28px each) **plus** the new `ScheduledEditActions` cluster (Delete 32px + "Send now" ~90px + Save ~60–90px). Total ≈340–360px vs ~328px usable on a 360px phone — the primary **Save/Send** button can be clipped or pushed off-screen. Adding Delete as a third trailing button is what tipped it over. This breaks the core send interaction on small phones.
   https://github.com/threahq/threa/blob/e4fb02b8478a94c807d0569eed84b0b4740183ca/apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx#L370-L405

2. `apps/frontend/src/components/scheduled/scheduled-item.tsx#L158` — **ux-consistency**: the composer popover row now hides its action triplet on touch (`reveal-actions-hover-only`), but the `/scheduled` full-page row still uses `reveal-actions` (persistent on touch). Same `ScheduledActions` component, now divergent touch behavior between the two lists a mobile user moves between. (Called out as deliberate "out of scope" in the PR body — surfacing so it's a conscious choice.)
   https://github.com/threahq/threa/blob/e4fb02b8478a94c807d0569eed84b0b4740183ca/apps/frontend/src/components/scheduled/scheduled-item.tsx#L155-L166

3. `apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx` (isPast footer) — **ux-discoverability**: when the send time is already past, the primary button reads "Send" (`saveLabel = isPast ? "Send" : "Save"`) and "Send now" sits right beside it — both persist edits and send immediately. Two adjacent buttons that do effectively the same thing, with no copy distinguishing them. Consider hiding "Send now" when `isPast` (the primary already means send-now).
   https://github.com/threahq/threa/blob/e4fb02b8478a94c807d0569eed84b0b4740183ca/apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx#L540-L575

4. `apps/frontend/src/components/scheduled/scheduled-edit-dialog.tsx` vs `scheduled-actions.tsx#L85` — **ux-consistency**: the dialog labels the destructive action "Delete" (`isFailed ? "Remove" : "Delete"`), while the shared `ScheduledActions` cluster (popover, drawer, /scheduled page) labels the same `useCancelScheduled` mutation "Cancel" for pending rows. Same effect, different verb — "Delete" reads more permanent. Align on one verb.
   https://github.com/threahq/threa/blob/e4fb02b8478a94c807d0569eed84b0b4740183ca/apps/frontend/src/components/scheduled/scheduled-actions.tsx#L81-L89

---
<details><summary>📐 Plan Adherence [CLEAN]</summary>The diff matches the stated intent: force-CAS backend fix, touch-hide on the popover row, Delete + Send now in the editor. No missing corresponding changes.</details>
<details><summary>🔍 Bugs [CLEAN]</summary>No logic/edge/race findings. sendNow's single forced CAS, the worker-fire/edit-fence interaction, and the dialog's save-then-send (with the `saved.status !== "sent"` double-send guard) were traced and hold.</details>
<details><summary>🔁 Data Flow [CLEAN]</summary>Scheduled-row write paths (schedule → confirm → update → sendNow → cancel) preserve fields; no dropped version/optimistic-lock. Dropping the `scheduled_for` rewrite leaves no invariant broken.</details>
<details><summary>📋 CLAUDE.md Compliance [CLEAN]</summary>No INV violations introduced. No happy-path `toast.success` added (INV-63 ok). Shared `ScheduledEditActions` is module-level (INV-18 ok). Minor: stale JSDoc above `ScheduledRow` (picker) still says "stays persistent for a finger" — now false (INV-25), scored 70, below threshold; worth a one-line fix.</details>
<details><summary>🏗️ Design [CLEAN]</summary>Shared cluster factored cleanly; force flag mirrors the existing bypassFence shape.</details>
<details><summary>🎛️ UX [3 issues]</summary>Findings 2, 3, 4 above.</details>
<details><summary>📱 Mobile [1 issue]</summary>Finding 1 above — footer overflow risk on small phones.</details>

</details>

kristofferremback:
### Code review (round 2 — after fixes)
**Review models:** Orchestrator: Claude Opus 4.8 | Reviewers: sonnet x3 (correctness+design, UX, mobile)

All four round-1 findings are resolved and no new issues surfaced. Backend was unchanged since round 1 (already clean); this pass focused on the frontend fixes.

**Verified resolved:**
1. **Mobile footer overflow** → "Send now" collapses to an icon on mobile (`compact`). Mobile reviewer's width math on a 360px phone: left cluster 112px (4 icons — `showExpand={false}`) + trailing 120px (Cancel 32 + Send-now-icon 32 + Save ~56) + gaps ≈ 260px vs ~326px usable — **~66px margin**, up from being right at/over the ceiling before.
2. **Cross-surface consistency** → `/scheduled` rows now use `reveal-actions-hover-only`, matching the popover; touch = tap→editor, long-press→drawer on both. Nothing became unreachable.
3. **Duplicate buttons when past-due** → "Send now" hidden when `isPast` (`showSendNow = isPending && !isPast`); the past-due dialog shows one clear submit ("Send"). Pending rows never lose their submit affordance.
4. **Verb mismatch** → dialog now uses "Cancel"/"Remove", matching the shared `ScheduledActions` cluster.

Also fixed: the stale `ScheduledRow` JSDoc (INV-25). Icon-only mobile "Send now" keeps `aria-label`/`title` (operable + discoverable). No happy-path `toast.success` (INV-63), no INV-18 nesting.

---
<details><summary>🔍 Bugs / 🔁 Data Flow / 🏗️ Design [CLEAN]</summary>`isPast`/`compact` branching is a single top-level component, props passed directly, no stale-closure or duplicated logic. No submit-affordance dead state.</details>
<details><summary>🎛️ UX [CLEAN]</summary>All three round-1 UX findings verified resolved; the new "Cancel" verb reads clearly (the dialog has no separate dismiss button, so it's unambiguous). Icon-only mobile Send now stays discoverable.</details>
<details><summary>📱 Mobile [CLEAN]</summary>Overflow resolved with real margin; 32px touch targets consistent with the component's existing Delete button; Trash vs paper-plane icons visually distinct.</details>

PR #1359, opened 2026-07-16
fix: surface [#text](url) links and in-app references in "In this stream"

kristofferremback:
## Problem

A Claude Channel message ending "PR open: [#1358](https://github.com/threahq/threa/pull/1358)" never showed the PR link in the "In this stream" panel.

Root cause is in `parseInlineMarkdown` (`packages/prosemirror/src/markdown.ts`): the generic-link branch re-parses the link *label* with mention/channel triggers enabled, so `#1358` matched the bare-channel pattern and became a `channelLink` node (unresolved id `"1358"`) with the real href relegated to a `link` mark. `collectLinkUrls` only read link marks off `text` nodes, so the URL vanished from link extraction — no panel row, and no link-preview candidate at ingestion. `[@handle](https://…)` had the same failure via the mention pattern.

Separately, in-app link previews (`message_link`, `stream_link`, `memo_link`, `conversation_link`, `delegation_link`) were filtered out of the panel's link list by `isInAppLinkContentType` with a comment saying they "belong to 'related'" — but no related bucket exists, so a shared message reference was dropped entirely.

## Solution

Three layers, each independently correct:

1. **Parser** — link labels are display-only. The generic-link branch now parses `linkText` with `enableMentions: false, enableChannels: false`, so `[#1358](https://…)` / `[@handle](https://…)` parse as plain text carrying a link mark. Pointer forms (`[#slug](channel:stream_x)`, `[@slug](user:usr_x)`) are unaffected — they're handled by `parseActorPointer` before this branch.
2. **Extractor** — `collectLinkUrls` reads link marks off *every* node type, not just text nodes. This matters because contentJson at rest is immutable: documents with the malformed channelLink shape already exist in prod (including the reporting message), and this makes their URLs surface without any backfill.
3. **Panel** — `deriveStreamContext` includes in-app previews as badged link items ("Message" / "Channel" / "Memo" / "Conversation" / "Task", `previewKind: "in-app"`) instead of dropping them. The row already routed own-origin URLs through react-router (`resolveInternalAppPath`), so navigation lands in-app; the external-link glyph is now shown only for URLs that actually open externally.

Verified end-to-end against the real prod message content: both the re-parsed markdown and the stored malformed contentJson shape now yield `https://github.com/threahq/threa/pull/1358` from `collectLinkUrls`.

## Modified files

| File | Change |
| ---- | ------ |
| `packages/prosemirror/src/markdown.ts` | Generic-link branch parses the label with mentions/channels disabled |
| `packages/prosemirror/src/extractors.ts` | `collectLinkUrls` reads link marks off any node type; text-node bare-URL scan unchanged |
| `apps/frontend/src/lib/stream-context/derive.ts` | `linkBadge` takes the full preview; in-app contentTypes map to badges instead of being skipped |
| `apps/frontend/src/lib/stream-context/types.ts` | `previewKind` gains `"in-app"` |
| `apps/frontend/src/components/stream-context/stream-context-row.tsx` | In-app badge tone (`bg-primary/10 text-primary`); external-link glyph gated on the URL actually being external |
| `packages/prosemirror/src/{markdown,extractors}.test.ts`, `apps/frontend/src/lib/stream-context/derive.test.ts` | Regression tests for the parse quirk, the stored malformed shape, and in-app badging (the old "excludes in-app" test now asserts inclusion) |

## Out of scope (deliberate)

- `quoteReply` nodes and `memoEmbed` references aren't surfaced as panel items — quote-replies are usually intra-stream noise; memo references only appear when a `memo_link` preview row exists. Follow-up if wanted.
- Old malformed messages keep their `channelLink` node in contentJson (rendered body unchanged); only extraction is healed. Re-serializing such a message on edit still produces the odd `[#1358](channel:1358)` form — pre-existing, untouched.
- No link previews are retro-generated for old messages (previews are created at ingestion); their panel rows show the bare URL without title/favicon.

## Test plan

- [x] `packages/prosemirror` `bun test` — 144 pass (incl. 4 new: `[#…](url)` / `[@…](url)` parse, link marks on non-text nodes, non-http marks excluded)
- [x] Frontend `vitest run src/lib/stream-context src/components/stream-context` — 33 pass (incl. new in-app badge + stored-malformed-shape tests)
- [x] Frontend `vitest run src/lib/markdown src/components/editor` — 503 pass
- [x] Backend `bun test src/features/{messaging,mentions,link-previews,public-api}` — 490 pass
- [x] `bun run typecheck` — monorepo clean
- [x] End-to-end sanity: real prod message markdown and its stored malformed contentJson both yield the PR URL from `collectLinkUrls`
- [ ] Live UI check of the prod message's panel row — needs this deployed; the stored-shape unit test reproduces the exact contentJson

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1359"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786778021&installation_id=129946197&pr_number=1359&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1359&signature=6b8f28dcbe510ced7c0cfc74b4c69f394010492924feca79e863b51cd0433650"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/frontend/src/components/stream-context/stream-context-row.tsx
+++ b/apps/frontend/src/components/stream-context/stream-context-row.tsx
@@ -103,14 +103,14 @@ function LinkLeading({ item }: { item: LinkContextItem }) {
   )
 }
 
-function BadgePill({ label, tone }: { label: string; tone: "github" | "linear" }) {
+function BadgePill({ label, tone }: { label: string; tone: "github" | "linear" | "in-app" }) {
   return (
     <span
       className={cn(
         "shrink-0 rounded px-1 py-px text-[10px] font-semibold uppercase tracking-wide",
-        tone === "github"
-          ? "bg-emerald-500/15 text-emerald-600 dark:text-emerald-400"
-          : "bg-indigo-500/15 text-indigo-600 dark:text-indigo-400"
+        tone === "github" && "bg-emerald-500/15 text-emerald-600 dark:text-emerald-400",
+        tone === "linear" && "bg-indigo-500/15 text-indigo-600 dark:text-indigo-400",
+        tone === "in-app" && "bg-primary/10 text-primary"
       )}
     >
       {label}
@@ -139,6 +139,7 @@ export function StreamContextRow({
   // Set when the row's primary action opens the in-stream gallery; the value is
   // the gallery key (`?smedia=`) — an attachment id or a `giphy:` sentinel.
   let galleryKey: string | null = null
+  let opensExternally = false
 
   switch (item.category) {
     case "link": {
@@ -146,13 +147,16 @@ export function StreamContextRow({
       primaryText = item.title ?? prettyHost(item.url)
       secondaryText = item.title ? prettyHost(item.url) : item.snippet || prettyHost(item.url)
       if (item.refCount > 1) secondaryText = `${secondaryText} · ${item.refCount}×`
-      if (item.badge)
-        badge = <BadgePill label={item.badge} tone={item.previewKind === "linear" ? "linear" : "github"} />
+      if (item.badge) {
+        const tones = { linear: "linear", "in-app": "in-app", github: "github", generic: "github" } as const
+        badge = <BadgePill label={item.badge} tone={tones[item.previewKind]} />
+      }
       // A link to our own origin routes in-app (react-router), matching how the
       // message body renders links (lib/markdown/components.tsx). Modifier- and
       // middle-clicks fall through to the native <a> so "open in new tab" still
       // works. External links keep opening in a new browsing context.
       const internalPath = resolveInternalAppPath(item.url)
+      opensExternally = internalPath == null
       primaryAction = internalPath ? (
         <a
           href={internalPath}
@@ -344,10 +348,11 @@ export function StreamContextRow({
             <div className="flex items-center gap-1.5">
               {badge}
               <span className="truncate text-sm font-medium leading-snug">{primaryText}</span>
-              {item.category === "link" && (
+              {item.category === "link" && opensExternally && (
                 // Persistent (not hover-gated) so a link row reads as "opens
                 // externally" at a glance, distinct from the rows that jump to
                 // the message — and visible on touch, where there is no hover.
+                // In-app links navigate inside Threa, so they don't get it.
                 <ExternalLink className="size-3 shrink-0 text-muted-foreground/70" aria-hidden />
               )}
             </div>
--- a/apps/frontend/src/lib/stream-context/derive.test.ts
+++ b/apps/frontend/src/lib/stream-context/derive.test.ts
@@ -209,7 +209,7 @@ describe("deriveStreamContext", () => {
     expect(links[0].sourceMessageId).toBe("msg_2")
   })
 
-  it("excludes in-app message_link previews from the links list", () => {
+  it("surfaces in-app message_link previews as badged link items", () => {
     const events = [
       messageEvent("2026-06-23T10:00:00.000Z", {
         linkPreviews: [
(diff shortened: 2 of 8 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: Claude Fable 5 | Reviewers: opus x4 (spec+design, correctness+data-flow, UX, mobile)

Found 1 issue (fixed in 6633c0e3):

1. `apps/frontend/src/lib/stream-context/derive.ts:129-135` — `stream_link` was badged with a hardcoded "Channel", but a stream link can point at a scratchpad or DM; the timeline reference card branches Channel/Scratchpad/Conversation by `streamType`, which is per-viewer (`inAppData`) and absent on broadcast events, so the panel can't do the same. (ux-consistency: mislabel a real user hits in a scratchpad-primary product.) **Fixed:** type-agnostic labels — "Stream", plus "Memory"/"Delegation" to match this panel's own filter chips.
   https://github.com/threahq/threa/blob/739557e0bd88edc670c79916a4aab5e66a7ef689/apps/frontend/src/lib/stream-context/derive.ts#L129-L135

---

Both other `collectLinkUrls` consumers verified safe: `in-app-links.ts` self-gates hrefs with `/^https?:\/\//`; backend `url-utils.ts` routes through `filterAndDeduplicateUrls` (SSRF blocklist). The `previewKind` union widening is fully covered — the `tones` map is exhaustive and `LinkLeading` falls through to the `Link2` icon for in-app. Parser scope is correct: only the label recursion disables mentions/channels; pointer forms are consumed by `parseActorPointer` before the fallback.
</details>

Round-trip verified: `#1358`+link mark serializes to `[#1358](url)` and re-parses to the identical text node. `collectLinkUrls` branches are mutually exclusive (no mark/bare-scan double-read); link marks sit on leaf nodes, and `seenInMessage` dedup prevents refCount inflation regardless.
</details>

Backend ingestion parses fresh markdown, so preview generation never sees the malformed node shape going forward. Editing an old malformed doc still serializes `[#1358](channel:1358)` — pre-existing serializer behavior, not amplified by this PR (which stops creating such docs and heals the read path); noted as out of scope in the PR description.
</details>

New comments are load-bearing "why" (INV-25); BadgePill uses a `cn` chain, not nested ternaries (INV-47); INV-64 wire-format round-trip for real mentions/channels intact.
</details>

`generic: "github"` in the tones map is dead-but-safe (generic previews never carry a badge).
</details>

`https?:` gates preserved on both the mark-reading path and the preview path; no new scheme admission. No SSRF bypass from broadening extraction (backend consumer keeps its blocklist).
</details>
<details><summary>🎛️ UX [1 issue — fixed]</summary>

The stream_link badge mislabel above (fixed). Two sub-threshold observations recorded for posterity: badge vocabulary previously drifted from the panel's filter chips (folded into the same fix); an in-app link with no preview title falls back to "app.threa.io" as the row title — the timeline card resolves a per-viewer name via `useStreamName`, the panel row doesn't. Candidate follow-up if it bothers in practice.
</details>

Badge is `shrink-0` + title truncates — no overflow at ~360px. Full-row `inset-0` tap target unregressed; dropping the ExternalLink glyph on in-app rows loses no touch affordance (the whole row remains tappable and navigates in-app).
</details>

PR #1899, opened 2026-08-16
fix: stream bootstrap counters require a known fetch start

kristofferremback:
## Why

`applyStreamBootstrap`'s counter publication guards against restoring a pre-message snapshot with a `fetchStartedAt`-vs-`counterTouchedAt` comparison (#1895 added this and tested it), but bypassed the guard entirely when `fetchStartedAt` was `undefined`. Every current call site passes the timestamp, so the bypass is latent — but a future call site omitting it silently reopens the race: a bootstrap snapshotted before a live message and resolved after it restores the stale count, and because `latestOrdinals` max-merges, the absorbed ordinal means the re-delivered `stream:activity` can never raise unread again — the auto-read report is then suppressed entirely (`use-auto-mark-as-read` sees 0/0 on a partial mark).

## Changes

Counter apply and post-transaction publication now require a known `fetchStartedAt`; without one the counters are skipped (read-state application is unchanged — it has its own freshness path). Test pins the closed bypass: no timestamp → no counter write, IDB and query cache both untouched.

## Verification

`stream-sync.test.ts` 148/148 green (includes #1895's existing in-flight-ordinal races plus the new bypass pin); monorepo lint+typecheck green via pre-commit.

## Residual risk

None beyond the intended behavior change: a hypothetical caller relying on timestamp-less counter publication would now not publish counters — no such caller exists.

### DIFF
--- a/apps/frontend/src/sync/stream-sync.test.ts
+++ b/apps/frontend/src/sync/stream-sync.test.ts
@@ -992,6 +992,47 @@ describe("applyStreamBootstrap — per-stream unread count", () => {
     })
   })
 
+  it("skips counter publication entirely when the caller provides no fetchStartedAt", async () => {
+    // Without a fetch-start timestamp the touched-at guard cannot order the
+    // snapshot against a live ordinal, so the counters must not apply at all —
+    // a caller that wants counter publication passes `fetchStartedAt`.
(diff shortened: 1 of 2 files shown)


kristofferremback:
@coderabbitai review

PR #1907, opened 2026-08-19
fix: merge the published-track registry by kind instead of overwriting

kristofferremback:
## Problem

Every camera-on video call is silent in both directions. `CallService.publishTracks` overwrites `call_endpoints.published_tracks` with the declared set, but the client transport (`CloudflareSfuTransport.publish`) declares one track per call — mic first, then camera — so the camera publish erases the mic entry. Peers pull only what the registry advertises, and `CallManager.applyRoster`'s prune then tears down the transient mic pull that briefly existed. Prod evidence from the 2026-08-19 08:28/08:29 UTC calls in `ws_01KJDZZV5H57W3H2SJ11PA554B`: all four endpoints ended with camera-only or empty registries, no mic entry anywhere. Audio-only calls never hit this (single publish), which is why earlier desktop tests sounded fine.

## Solution

Registry writes become read-merge-write under the call lock, shared by publish and unpublish.

- `publishTracks` merges the declared tracks into the current registry by kind (same kind → replaced, other kinds → kept).
- `closeTracks` prunes from the same in-tx re-read instead of the `fenceEndpoint` read taken before the CF round-trip — the old shape could lose a concurrent publish that committed between fence and write.
- Both go through a new private `mutateRegistry`: lock call row → re-read endpoint → mutate → fenced `setPublishedTracks` → roster bump + snapshot, all in one transaction (INV-20). Entries from another `media_incarnation` never carry into the write (rebind resets the registry at cf/session create; the fenced write already 409s stale writers).
- Backend-only over client-declares-full-set: CF `addLocalTracks` adds tracks, so re-declaring already-published tracks to CF on every publish would churn SDP for nothing; the registry is the backend's row, so the merge belongs where the write is.

## Files

| File | Change |
| ---- | ------ |
| `apps/backend/src/features/calls/service.ts` | `mutateRegistry` (read-merge-write under call lock); `publishTracks` merges by kind; `closeTracks` prunes in-tx |
| `apps/backend/src/features/calls/service.test.ts` | Regression: camera publish keeps mic; same-kind republish replaces; cross-incarnation entries dropped; closeTracks prune |

## Test plan

- [x] `bun test src/features/calls/` — 153 pass (81 in service.test.ts)
- [x] monorepo lint + typecheck + migration check (pre-commit hook, observed green)
- [ ] live two-device call — needs a deploy; the prod rows above are the failing baseline to compare against

---

---
<a href="https://app.blacksmith.sh/threahq/codesmith/threa/pr/1907"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789763501&installation_model_id=20758&pr_number=1907&repository=threahq%2Fthrea&return_to=https%3A%2F%2Fgithub.com%2Fthreahq%2Fthrea%2Fpull%2F1907&signature=b9607dcf4220e017f0365814299a5318613ccc7d689e80d8f54d97a8b34f4402"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/apps/backend/src/features/calls/service.test.ts
+++ b/apps/backend/src/features/calls/service.test.ts
@@ -1699,6 +1699,126 @@ describe("CallService.publishTracks", () => {
     expect(result.snapshot.rosterVersion).toBe(4)
   })
 
+  it("should keep the mic entry when a camera publish declares only the camera", async () => {
+    stubWithClient()
+    stubTransaction()
+    stubFence(
+      fakeEndpoint({
+        id: "callep_1",
+        cfSessionId: "sess_1",
+        mediaIncarnation: "inc_1",
+        publishedTracks: [{ kind: "mic", trackName: "mic0" }],
+      })
+    )
+    const setTracks = spyOn(CallEndpointRepository, "setPublishedTracks").mockResolvedValue(fakeEndpoint())
+    spyOn(CallRepository, "bumpRosterVersion").mockResolvedValue(5)
+    spyOn(CallParticipantRepository, "listRoster").mockResolvedValue([])
+
+    await makeServiceWithCf(fakeCloudflare()).publishTracks({
+      workspaceId: "ws_1",
+      callId: "call_1",
+      userId: "usr_1",
+      endpointId: "callep_1",
+      mediaIncarnation: "inc_1",
+      sdp: { type: "offer", sdp: "o" },
+      tracks: [{ kind: "camera", mid: "1", trackName: "cam1" }],
+    })
+
+    expect(setTracks).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({
+        publishedTracks: [
+          { kind: "mic", trackName: "mic0" },
+          { kind: "camera", trackName: "cam1" },
+        ],
+      })
+    )
+  })
+
+  it("should replace the entry when the same kind republishes", async () => {
+    stubWithClient()
+    stubTransaction()
+    stubFence(
+      fakeEndpoint({
+        id: "callep_1",
+        cfSessionId: "sess_1",
+        mediaIncarnation: "inc_1",
+        publishedTracks: [
+          { kind: "mic", trackName: "mic-old" },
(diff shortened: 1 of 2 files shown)


kristofferremback:
### Code review
**Review models:** Orchestrator: claude-fable-5 | Reviewers: sonnet x2 (spec+design, correctness+data-flow)

No issues found. Checked for bugs, data-flow lifecycle (full `published_tracks` walk: insert → rebind reset → publish merge → close prune → endpoint close, plus concurrent publish/close and takeover-mid-publish interleavings), AGENTS.md compliance (INV-20 lock ordering, INV-66 incarnation fencing), plan adherence, design quality, and security on this backend diff.

Notes from the pass, for the record:
- Every `published_tracks` writer takes the call-row lock first, so the in-tx re-read in `mutateRegistry` cannot interleave with another registry writer.
- The `current.mediaIncarnation !== params.mediaIncarnation → empty base` guard is belt-and-suspenders on top of the fenced `WHERE media_incarnation = …` write; both hold.
- Client `diffPulls` keys by `(cfSessionId, trackName)` and tolerates the merged registry unchanged.
- Pre-existing, not amplified: `publishSchema` doesn't dedupe kinds within one call's declared set; the client only ever declares one track per publish.

